PR #130, opened 2020-05-25
Change data preparation from class to function

pplantinga:
A bit nicer interface for calling the preparation. Instead of:

```python
preparer = TIMIT_Preparer(...)
preparer()
```
Just do:
```python
prepare_timit(...)
```
- [x] timit
- [x] librispeech

### DIFF
--- a/recipes/LibriSpeech/librispeech_prepare.py
+++ b/recipes/LibriSpeech/librispeech_prepare.py
@@ -19,9 +19,17 @@ from speechbrain.data_io.data_io import (
 )
 
 logger = logging.getLogger(__name__)
+OPT_FILE = "opt_librispeech_prepare.pkl"
+SAMPLERATE = 16000
 
 
-class LibriSpeechPreparer:
+def prepare_librispeech(
+    data_folder,
+    splits,
+    save_folder,
+    select_n_sentences=None,
+    use_lexicon=False,
+):
     """
     This class prepares the csv files for the LibriSpeech dataset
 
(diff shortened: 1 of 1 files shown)


recipes/LibriSpeech/librispeech_prepare.py
```diff
@@ -42,298 +50,317 @@ class LibriSpeechPreparer:
...
-    >>> LibriSpeechPreparer(data_folder,splits,save_folder)
+    >>> data_folder = 'datasets/LibriSpeech'
+    >>> splits = ['train-clean-100', 'dev-clean', 'test-clean']
+    >>> save_folder = 'librispeech_prepared'
+    >>> prepare_librispeech(data_folder, splits, save_folder)
     """
+    data_folder = data_folder
+    splits = splits
+    save_folder = save_folder
+    select_n_sentences = select_n_sentences
```
csukuangfj, line 61: This line does not make sense.

recipes/LibriSpeech/librispeech_prepare.py
```diff
@@ -42,298 +50,317 @@ class LibriSpeechPreparer:
...
+        )
 
-            split = self.splits[split_index]
+        text_dict = text_to_dict(text_lst)
 
-            wav_lst = get_all_files(
-                self.data_folder + "/" + split, match_and=[".flac"]
-            )
+        if select_n_sentences is not None:
+            select_n_sentences = select_n_sentences[split_index]
```
csukuangfj, line 104: The documentation of this function says `select_n_sentences` is an `int`,
how can it support the index operator `[]`?

csukuangfj:
The above two issues still exist in the latest master.

recipes/LibriSpeech/librispeech_prepare.py
pplantinga: You're right they should be removed.

recipes/LibriSpeech/librispeech_prepare.py
pplantinga: You're right, the `select_n_sentences` variable is not used and should be removed. Feel free to open a PR for that!

PR #442, opened 2021-01-07
New data loading: Save a mid epoch checkpoint every x minutes.

Gastron:
Implements the functionality. Currently not tested at all.

### DIFF
--- a/recipes/TIMIT/ASR/CTC/experiment.py
+++ b/recipes/TIMIT/ASR/CTC/experiment.py
@@ -40,7 +40,6 @@ class ASR_Brain(sb.Brain):
     def compute_objectives(self, predictions, batch, stage):
         "Given the network predictions and targets computed the CTC loss."
         pout, pout_lens = predictions
-        ids = batch.id
         phns, phn_lens = batch.phn_encoded
 
         if stage == sb.Stage.TRAIN and hasattr(self.hparams, "env_corrupt"):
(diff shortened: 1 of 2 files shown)


speechbrain/utils/distributed.py
```diff
@@ -51,25 +55,134 @@ def run_on_main(
...
+    Arguments
+    ---------
+    run_opts: list
+        a list of arguments to parse, most often from `sys.argv[1:]`
+    """
+    if run_opts["distributed_launch"]:
+        if "local_rank" not in run_opts:
+            raise ValueError(
+                "To use DDP backend, start your script with:\n\t"
+                "python -m torch.distributed.lunch [args]\n\t"
```
csukuangfj, line 125: It is `launch`, not `lunch`. The same typo exists for the following comments.

speechbrain/utils/distributed.py
```diff
@@ -51,25 +55,134 @@ def run_on_main(
...
+        else:
+            if run_opts["local_rank"] + 1 > torch.cuda.device_count():
+                raise ValueError(
+                    "Killing process " + str() + "\n"
+                    "To use DDP backend, start your script with:\n\t"
+                    "python -m torch.distributed.lunch [args]\n\t"
+                    "experiment.py hyperparams.yaml --distributed_launch=True "
+                    "--distributed_backend=nccl"
+                )
+        if "RANK" in os.environ is None or os.environ["RANK"] == "":
```
csukuangfj, line 138: The return value of `"RANK" in os.environ`  is a boolean, which can never be `None`.

So
```python
"RANK" in os.environ is None
```
does not make sense.

speechbrain/utils/distributed.py
TParcollet: Solved, thanks.

speechbrain/utils/distributed.py
TParcollet: Where did you find that it is a boolean ? From what I see from the doc, it is a string.

csukuangfj: It is from https://docs.python.org/3/reference/datamodel.html#object.__contains__

---

You can write a test
```python
>>> isinstance('' in {}, bool)
True
```

`os.environ` is a dict. 

`"RANK" in os.environ` will invoke the `__contains__` method of the dict, which returns `True` or `False`.
It never returns a string, I believe.

---

> From what I see from the doc, it is a string.

Could you point me to the doc you are reading?

TParcollet: https://docs.python.org/3/library/os.html

@aheba

csukuangfj: https://docs.python.org/3/library/os.html#os.environ
reads
> A mapping object representing the string environment.

which shows `os.environ` is indeed a dict, i.e., a mapping object. Cannot find where it says `os.environ` is a string or the result of
`"RANK" in os.environ`  is a string.

TParcollet: Right, could you do a PR to fix this ? So we can test it :D

PR #472, opened 2021-01-31
Dynamic Batching (Tests)

mravanelli:
The goal of this PR is to test dynamic batching in different recipes and find the best hyperparameters for that.


mravanelli:
I'm now resuming the PR on dynamic batching and I'm doing some debugging tests using TIMIT. Actually, it looks very promising.  With the parameters, I pushed I can train a TIMIT seq2seq system with the same performance (i.e., PER=13.8%-14.3%) but with a major speed-up factor (e.g., 66%) without exceeding 12 GB of memory.

However, there is a speed issue when using DDP. When I use 2 gpus, it looks like both GPUs are processing exactly the same data (i.e, the sentences aren't split across gpus).  The amount of iteration, in fact, is the same with and without DDP (259).  For more details, you can see the time-per epoch reported below of different gpus:

_On RTX8000_
```
No Dynamic Batch, no DPP: 1 min 55 sec
No Dynamic Batch, yes DPP: 1 min 28 sec
Yes Dynamic Batch, no DPP: 1 min 25 sec
Yes Dynamic Batch, yes DPP: 1 min 52 sec
```

_On K80_
```
No Dynamic Batch, no DPP: 6 min 50 sec
No Dynamic Batch, yes DPP: 4 min 00 sec
Yes Dynamic Batch, no DPP: 5 min 51 sec
Yes Dynamic Batch, yes DPP: 6 min 36 sec

```
Is there something to change in the dynamic batching part to support multiple GPUs? @popcornell @aheba

TParcollet:
One day, we will have to get back to this PR. This day, I'll need to take vacations for a month XD

PR #487, opened 2021-02-11
AISHELL-1 recipe

lorenlugosch:


Hap-Zhang:
Hi, @mravanelli 
I test the recipe of AISHELL, and meet a error when training in epoch 35, can you give me some instruction, and thank you very much! The error is as follows:

image

mravanelli:
This error happens when we meet some numerical issue in the loss. I noticed
in the past one some challenging recipes like speech recognition on small
datasets with noise. I don't know if Loren has seen this issue in this
recipe before.  Do you have the same error if you resume training?

lorenlugosch:
One thing @Hap-Zhang could test is whether there is a batch with (encoder output sequence length < label sequence length), which would have infinite loss with CTC. I never encountered this issue when training on AISHELL, but maybe I just got lucky.

PR #514, opened 2021-02-16
WFST-based ASR decoding

aheba:
Hey @mravanelli, @TParcollet,

This PR will add WFST-based decoding to our E2E models. 

- [ ] add common functions for dealing with fsts. we will integrate [K2](https://github.com/k2-fsa/k2) create token/lexicon/grammar fst, composing decoding graph...
- [ ] add Viterbi decoder on graph
- [ ] use pre-trained model
- [ ] add entity-based example like: what's the weather in #city. my number is #digits. Where entities are defined by "#" in the Grammar file and are converted to all possible entries.


csukuangfj:
@aheba

trangtv57:
hi @aheba 
WFST of k2 just support output of CTC, does it support model with rnn-transducer?

mravanelli:
I think we can close this PR, right? Soon, there might be a join effort with the k2 team to create such an interface.

PR #535, opened 2021-02-23
added MHA with relative pos encodings as in Conformer and TransformerXL

popcornell:
Need to do docs. 
I will be busy with other stuff before SB release so if one of you @JianyuanZhong  or @ycemsubakan can do the docs + integrate it in Conformer it will be cool.


recipes/LibriSpeech/ASR/transformer/hparams/conformer.yaml
```diff
@@ -29,23 +32,27 @@ data_folder_rirs: !ref <data_folder>
...
 ctc_weight: 0.3
-gradient_accumulation: 4
+gradient_accumulation: 2
 gradient_clipping: 5.0
 loss_reduction: 'batchmean'
 sorting: random
 
+# stages related parameters
+stage_one_epochs: 90
+lr_adam: 1.0
```
TParcollet, line 53: adam and 1.0 ??

tests/unittests/attention.py
```diff
@@ -0,0 +1,24 @@
...
+    from speechbrain.nnet.attention import RelPosMultiHeadAttention
+
+    bsz = 2
+    emb_dim = 4
+    k_len = [12, 10]
+    q_len = [10, 12]
+    bias = [True, False]
+    head_dim = [3, None]
+
+    for kl in k_len:
```
TParcollet, line 15: let's loop!

recipes/LibriSpeech/ASR/transformer/hparams/conformer.yaml
popcornell: I did not change that maybe @JianyuanZhong knows more

popcornell: I think it is because this recipe uses Warmup LR schedule see NoamScheduler

mravanelli:
@popcornell what's the status of this PR? It looks like there are some conflicts to solve.

popcornell:
The status is that we get very good accuracy more than Transformer with Conformer yet WER is still worse now. 
I can resolve the conflicts in few minutes, but I'd merge this only when we have better results with Conformer

kalpitdixit:
hi, thanks for working on this. Seeing that the last update here was ~4 weeks back, is the Conformer implementation still being actively worked on ?

I have a use-case of long-sequences (upto 5 mins), so was hoping to use the Conformer (vs. Transformer) for sub-sampling.

popcornell:
If you are brave enough you can use this branch already. @TParcollet is going to merge it in a couple of days BTW.

PR #554, opened 2021-03-14
Fix: Remove separate linter messages which failed from forks

Gastron:
Separate nice linter flags seemed to fail from forks. Propose removing those for now.


pplantinga (approved):
Linters pass now, great!

PR #555, opened 2021-03-14
Change length to lengths everywhere

pplantinga:
Best to have a consistent vocabulary when talking about the same thing, for our relative lengths tensors we sometimes use "length" and sometimes "lengths". This PR changes a number of the former to the latter.


mravanelli:
I'm not sure about the effect of this in tutorials. IMO it is to risky to this this now. I propose keeping that for the next version

PR #560, opened 2021-03-15
Adding feature to use corresponding backend for torchaudio

sharockys:
Adding feature allowing torchaudio to use soundfile on windows and sox_io on mac/linux.
Otherwise the import of speechbrain will fail directly on windows.

### DIFF
--- a/recipes/DNS/enhance/spectral_map/train.py
+++ b/recipes/DNS/enhance/spectral_map/train.py
@@ -24,7 +24,7 @@ from speechbrain.utils.torch_audio_backend import get_torchaudio_backend
 torchaudio_backend = get_torchaudio_backend()
 torchaudio.set_audio_backend(
     torchaudio_backend
-) 
+)
 logger = logging.getLogger(__name__)
 
 try:
--- a/recipes/DNS/enhance/spectral_map/train.py
+++ b/recipes/DNS/enhance/spectral_map/train.py
@@ -22,9 +22,7 @@ from speechbrain.utils.distributed import run_on_main
 from speechbrain.utils.torch_audio_backend import get_torchaudio_backend
 
 torchaudio_backend = get_torchaudio_backend()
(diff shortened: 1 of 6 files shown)


mravanelli:
That's a nice add. In fact, windows users cannot use the sox-io backend, but the soundfile is working.

sharockys:
Could you please help me with the checks problems? 
I tested those two files with black and pytest and they passed the tests.

aelimame (approved):
Tested on my Windows machine and it works. Thanks

UrosOgrizovic:
Should SoundFile be added to requirements.txt?

sharockys:
> Should SoundFile be added to requirements.txt?

It seems that soundfile is installed with torchaudio on windows.  https://github.com/pytorch/audio/blob/master/requirements.txt#L4

UrosOgrizovic:
@sharockys After creating a venv, installing both requirements (which installs torchaudio==0.8.0) and lint-requirements, `pytest tests` gives me: `RuntimeError: Backend "soundfile" is not one of available backends: []`. If I run `pip install SoundFile` after that, the tests do pass.

sharockys:
> @sharockys After creating a venv, installing both requirements (which installs torchaudio==0.8.0) and lint-requirements, `pytest tests` gives me: `RuntimeError: Backend "soundfile" is not one of available backends: []`. If I run `pip install SoundFile` after that, the tests do pass.

Thank you for testing. I added soundfile in requirements as torchaudio did.

mravanelli:
everything seems to work! We can merge it. Than you @sharockys, @aelimame, @UrosOgrizovic. @TParcollet, after that I think window support is completed (we should change the documentation on that part).

mravanelli (approved):

aelimame:
I can confirm that torchaudio on Windows does not automatically install soundfile backend, so good call on adding it to the requirements.txt

PR #576, opened 2021-03-16
Fairseq wav2vec integration 

TParcollet:
This is a PR to integrate pretrained wav2vec models from fairseq. This PR reopens PR #545


TParcollet:
@pplantinga How can we disable the testing for this file ? We don't want to install fairseq from scratch, but we need to keep the example.

pplantinga:
Probably we want to do something similar with transducer, where the test is only run if numba is installed. I think you can look at `conftest.py` for an example?

mravanelli:
@TParcollet, do you think it makes sense to fetch the wav2vect model from HuggingFace?

TParcollet:
I have to think about that. They do not release all models in HF but we could add this feature

flozi00:
> I have to think about that. They do not release all models in HF but we could add this feature

https://huggingface.co/models?filter=wav2vec2
175 pretrained models in 60 languages
WER of around 15% for languages like spanish, german and so on.
Making them accessible in speechbrain would be a very big step

TParcollet:
Right, let's do this after the Interspeech deadline then. It should be quite easy honestly.

TParcollet:
@flozi00 We now support HuggingFace based wav2vec2 models ;)

flozi00:
great, still studying the code to learn how to contribute to.
I would like to train multiple models on commonvoice, did you speak with the HF team about if we are allowed to upload community models on their hub ?
Will directly start an training with transducer.
Is there an ready to use notebook or something similar ? Otherwise I will check out the recipes

TParcollet:
I'm working on a TIMIT recipe that I will push here soon. You can push anything you want to HF (or even the official SpeechBrain HF depending on what you would like to push).

flozi00:
using the cachedir argument while loading the pretrained model makes it stocking, if it is already cached in the default cache dir

TParcollet:
Not sure to understand.

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -0,0 +1,95 @@
...
+    >>> outputs = model(inputs)
+    >>> outputs.shape
+    torch.Size([10, 1,  768])
+    """
+
+    def __init__(self, source, save_path, output_norm=True, freeze=True):
+        super().__init__()
+
+        # Download the model from HuggingFace and load it.
+        self.model = Wav2Vec2Model.from_pretrained(source, cache_dir=save_path)
```
flozi00, line 62: If the save_path/cache_dir changes and the model was already cached this line takes endless time on my machine
After removing the cache_dir argument it loaded within some seconds

TParcollet: Really. ? interesting, I don't see any reason why this would happen. Hum.

TParcollet: For an unknown reason, also models obtained from HF seem to not converge when fine-tuned. While Fairseq ones work really well. I need to investigate this.

TParcollet: Ok, it was just a normalization issue eheh

flozi00: Did a quick run with commonvoice recipe, transducer
Just replaced the model with wave2vec2, getting mismatch error, I think I have to learn more about your design how to configure it correctly. 
Looking forward to your recipe :)

TParcollet: I can share with you my TIMIT recipe if you wish.

mravanelli:
Titouan, why don't you push the TIMIT recipes that achieve PER=9% here?

TParcollet:
I will do it, but I'm trying many different things right now to lower the score. I will push the final one. (tmr I think)

speechbrain/lobes/models/huggingface_wav2vec.py
flozi00: Would be cool 😎

TParcollet:
@mravanelli @flozi00 I pushed the fairseq example on TIMIT that obtains 9% of PER. I'll try to switch it to HuggingFace tmr, but you can also simply do it by switching the lobes that you call in the yaml :D

TParcollet:
I swapped to the HuggingFace approach for the standard recipe. PER is now 8.84% in solely 16 epochs.

flozi00:
> I swapped to the HuggingFace approach for the standard recipe. PER is now 8.84% in solely 16 epochs.

Sounds good
Which steps do I have to do to use commonvoice dataset ?
Is it possible to calculate multiple scores while evaluation by default ? Like WER, CER and PER ?

TParcollet:
I will provide a tutorial soon (hopefully ...). To use it with CommonVoice, it is fairly simple. Just take the wav2vec2 block on the YAML (and the concept of having two optimizers for fine-tuning) and put it on the standard CV recipe. IMPORTANT: I think that the CRDNN encoder should be removed and replaced with something way simpler (like the VanilaNN on the TIMIT recipe). However, I have no empirical validation of this claim. It is just based on the fact that wav2vec2 already is a big encoder, so I suspect that nothing more is needed ... For the multi logs, I believe that the CV recipe reports CER and WER already by default, you can check the code of the recipe to see how this is added.

flozi00:
> I will provide a tutorial soon (hopefully ...). To use it with CommonVoice, it is fairly simple. Just take the wav2vec2 block on the YAML (and the concept of having two optimizers for fine-tuning) and put it on the standard CV recipe. IMPORTANT: I think that the CRDNN encoder should be removed and replaced with something way simpler (like the VanilaNN on the TIMIT recipe). However, I have no empirical validation of this claim. It is just based on the fact that wav2vec2 already is a big encoder, so I suspect that nothing more is needed ... For the multi logs, I believe that the CV recipe reports CER and WER already by default, you can check the code of the recipe to see how this is added.
> 
> Note: be prepared to use a LOT of VRAM.

hmm
I did something wrong, the memory and training time are not changing, will wait for your recipe to compare and learn from.

TParcollet:
Tutorial: https://colab.research.google.com/drive/17Hu1pxqhfMisjkSgmM2CnZxfqDyn2hSY?usp=sharing

TParcollet:
@flozi00 There is already a tutorial (see above) and the recipe is on the TIMIT recipe folder.

mravanelli:
I tested it and everything seems to work. I got a PER=8.04% on TIMIT (I'm uploading this model on the google shared folder). I did some minor changes in the code and tutorial. As for the code, I added a pretrain flag that when it is False, it instantiates the randomly initialized model (this is useful when we will integrate the training part as well in SpeechBrain).

One issue is that with the current solution we have to create a novel experiment.py file every time we wanna use wav2vect. This is mainly due to the two optimizers used for the wav2vect system and the other parts. 
This makes the current solution not that scalable. As mentioned privately, I think we should find a solution that binds the model and the corresponding optimizers in the yaml file directly. This might require some changes in the brain class though. This change might take time and I propose to merge this as it is for now.

PR #587, opened 2021-03-19
Recipe for classification on GSC V2

DavidRP1:
Add a recipe to train X Vectors and Ecapa TDNN for classification on the Google Speech Commands V2 dataset.


TParcollet:
Hey, do not hesitate to flag your work with work in progress or ready to review ;) Thanks for the PR !

mravanelli:
I tested an I'm finally able to replicate your best numbers (which are actually the state-of-the-art for the google speech command task).  I think we can merge it. Thank you @DavidRP1!

TParcollet:
@mravanelli should we think about uploading the model to HuggingFace ?

mravanelli:
Yes, we can (even though this isn't a really big model and very limited in
terms of possible practical applications).

PR #589, opened 2021-03-19
initial implementation of dynamic whamr reverberation

ycemsubakan:
I started working on dynamically adding reverberation for the whamr dataset. This is especially needed when we do speedaugment. Not ready to merge yet.


ycemsubakan:
- [ ] Wham / Whamr dynamic mixing

popcornell:
@ycemsubakan added same procedure as used in Librimix it requires however pyloudnorm. 
This guarantees it is equal 100%. We might even avoid pyloudnorm altogether however by using SB normalize. It is not the same but probably won't affect performance too much

recipes/WSJ0Mix/separation/dynamic_mixing.py
```diff
@@ -45,12 +46,55 @@ def build_spk_hashtable(hparams):
     return spk_hashtable, spk_weights
 
 
+def build_spk_hashtable_librimix(hparams):
```
popcornell, line 51: are you sure we should put everything here  on WSJ2mix recipe ?

ycemsubakan: I think we can attack all of the datasets with the same code?

mravanelli:
@ycemsubakan, any update on that? I think we should finalize this as soon as possible (ideally before ICASSP, where many people will likely take a look at the code)

ycemsubakan:
Right, I will try to finalize this this week..

recipes/WSJ0Mix/meta/rir_constants.py
```diff
@@ -0,0 +1,8 @@
+NUM_BANDS = 4
```
mravanelli, line 1: Please, add docstring with a description of the parameters + which other file is using it

recipes/WSJ0Mix/meta/wham_room.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
```
mravanelli, line 1: Please add a general docstring + a function specific one. If this is taken from somewhere else, it should be mentioned a credits should be given.

recipes/WSJ0Mix/separation/README.md
```diff
@@ -61,10 +62,39 @@ Here are the SI - SNRi results (in dB) on the test set of WSJ0 - 2/3 Mix dataset
 | | SepFormer. WHAMR! |
 | --- | --- |
 |NoAugment | 11.4 |
+|SpeedAugment | 13.7|
```
mravanelli, line 65: We should add here a link to the output folders stored in google drive.  I also think we should mention where are the pre-trained models and easy inference function in HuggingFace

mravanelli:
Thank you @ycemsubakan for this important PR! I left some minor comments. 
The major comment I have is the following:
It looks to me that we are trying to do too many things in this code. This is good because we save code, but bad because it might be less easy to understand. Moreover, everything is under the folder WSJ0Mix, but we actually support also Wham, whamr, and LibriSpeech. As a result, the code is quite long (e.g, data preparation) and often contains statements like: "`if "wham" in Path(hparams["data_folder"]).stem:`"
I'm wondering if it would be better to separate the code into three/four folders like WSJMix, WHAM, WHAMR, LibriMix. This way users know that we also support other datasets. Of course, there might be some overlap in the code, but maybe the result is cleaner.  What do you think? As far as I remember, @popcornell  has a similar argument.

ycemsubakan:
In my opinion separating the code into multiple folders would not be ideal, as in that case we would need to update all of the `train.py` scripts each time we want to make a change. I think the differences are not that much, so I vote for keeping one folder. 

This being said, we could potentially compress the code to make things easier to understand maybe.

mravanelli:
I'm not convinced about this solution. People only see the WSJ0 folder and they cannot figure out that there are many other datasets inside that. As you mention, having a separate train.py for each dataset  might create redundancy. However, this is what we do in other recipes, including for instance speech recognition. There is significant overlap in the code of librispeech, TIMIT, and CommonVoice but we have a separate train.py for all the cases. Any though from @popcornell and from the others?

popcornell:
IMO right now it is confusing as the recipe is called WSJ0mix right ? 
So maybe it must be called  Separation as it handles many datasets for separation ? 

But then this brings a fundamental question:
recipes should be organized only by datasets or also by tasks ? IMO you should just go for one clear way of organizing recipes for all SB, and right now all recipes are organized by datasets. 

Of course splitting recipes by datasets will lead to significant boilerplate code. 
However maybe much of the boilerplate code can be moved directly to Speechbrain source. Maybe a sub class of Brain called SeparationBrain can be defined in SB and then used by all recipes.

ycemsubakan:
Right, I think if we can put all the shared code in something called `SeparationBrain`, then we can have separate recipes. 

But yes, we should think about what we should call the recipe. One option is calling it `Separation`, and then having a folder for each dataset. The other option is to have a separate recipe for each dataset. Currently, in speechbrain we are doing the second one. The advantage with the current option is that we can have multiple tasks for a given dataset. In our case this might be advantageous as we might want to add different tasks for a given dataset.

TParcollet:
Hey thanks for the PR. My point of view on this question: We decided to go for the datasets > code for an excellent reason, peoples are used to it :D I vote for doing this. It is also easier to read imho.

ycemsubakan:
Right, for the time being we will go with the dataset--> code organization. But I think it makes a lot of sense to use a parent class structure that would be inherited by several recipes for datasets. We can perhaps think about how to move towards that structure in the future, because the current structure requires `train.py` files that repeat a lot of things..

mravanelli:
I did several tests on WHAM, WHAMR, WSJ0 and everything seems to work properly. 
I only have the following minor comments on wham/whamr:
1. `WHAMandWHAMR`  => `WHAM&WHAMR` (maybe this is shorter?)
2. `README.md` => It looks like some parts still mention WSJMIX (e.g., `# Speech separation with WSJ0MIX`). 
3. `README.md` python train.py hyperparams/sepformer-wham.yaml --data_folder => `python train.py hparams/sepformer-wham.yaml --data_folder=your_data_folder`
Moreover, "hyperparams" should be replaced with "hparams" in the running example.

4. `/recipes/WHAMandWHAMR/meta/README.md`  => add an example about how to run the rirs preparation script (even though it is pretty clear from the code)

5. `resample_folder in preprocess_dynamic_mixing.py` looks quite general. I think it could be better to put it in `speechbrain/utils/data_utils.py ` and remove `preprocess_dynamic_mixing.py` from here and all the other places. What do you think?

6. `hparams/sepformer-whamr.yaml`, `hparams/sepformer-whamr.yaml` => replace WSJ with WHAM/WHAMR (e.g. # Dataset : WSJ0-2mix and WSJ0-3mix)

Comments on WSJ recipe:
7. `recipes/WSJ0Mix/separation/README.md` => remove WHAM/WHAMR and Librimix from here (e.g.,results and run examples). Note that the WSJ0-3Mix training is wrong because still contains 2 speakers.

8. As for the WSJMIX recipe, the system uses `pyloudnorm`, which is not added in the extra-dependency and not needed for wham/whamr.

mravanelli:
Few more comments on the LibriSpeech recipe:

1. `recipes/LibriMix/meta/README.md` => still refers to WSJ.
2. `recipes/LibriMix/prepare_data.py` => the docstring still refers to wsj.
3. In the readme files, we should report the link where we made available the output folder (link to our google drive) and maybe even where people can find the pre-trained model with easy inference functions (link to HF). This is valid especially for wham and whamr (but I would suggest uploading on drive the stable results with librimix as well).
4. `recipes/LibriMix/separation/README.md`  => the call example with 3mix actually refers to 2 mixtures (yourpath/Libri2Mix).
5. ` recipes/LibriMix/separation/dynamic_mixing.py` => it looks here that there are things that are not needed (e.g., get_wham_noise_filenames).
6. add  pyloudnorm in `extra-dependencies.txt`
7. it looks like` oct2py==5.2.0`, `octave-kernel==0.32.0`, `pyroomacoustics==0.3.1` aren't needed for this recipe, right?
8. `LibriMix/separation/hparams/sepformer-libri2mix.yaml` => still contains wsj in header and comments, and variables (e.g, base_folder_dm: `/yourpath/wsj0-processed/si_tr_s/`). The same for the 3 mix.
9. We should run an experiment without wsjmix pre-training as well.

danpovey:
> Right, for the time being we will go with the dataset--> code organization. But I think it makes a lot of sense to use a parent class structure that would be inherited by several recipes for datasets. We can perhaps think about how to move towards that structure in the future, because the current structure requires `train.py` files that repeat a lot of things..

I think Mirco is right that it's better to have a little repetition than to have things like inheritance, in cases like this.  Having a more complex structure puts a burden on the users (to understand it), and on the maintainers (testing becomes more complex, and there can be unanticipated and un-tested side effects from changes).  Yes, sometimes if there is repetition things can get out of date; but I think the total cost is less than the cost from unnecessary sharing.

ycemsubakan (commented):
> > Right, for the time being we will go with the dataset--> code organization. But I think it makes a lot of sense to use a parent class structure that would be inherited by several recipes for datasets. We can perhaps think about how to move towards that structure in the future, because the current structure requires `train.py` files that repeat a lot of things..
> 
> I think Mirco is right that it's better to have a little repetition than to have things like inheritance, in cases like this. Having a more complex structure puts a burden on the users (to understand it), and on the maintainers (testing becomes more complex, and there can be unanticipated and un-tested side effects from changes). Yes, sometimes if there is repetition things can get out of date; but I think the total cost is less than the cost from unnecessary sharing.

Right, from what I see, this re-organization will be nice at the end. The brain class for separation is repeating few methods, but for now it's okay I think.

mravanelli:
Thank you, @ycemsubakan! On my side, everything runs. I only have the following few comments:
1- I removed the swp files that were pushed accidentally.
2-` meta/wham_room.py` => add docstring
3-` separation/dynamic_mixing.py` =>  add docstring `build_spk_hashtabl`, `get_wham_noise_filenames`, `audio_pipeline`
4- `recipes/WSJ0Mix/meta/wham_room.py`, `create_whamr_rirs.py`, `rir_constants.py` => are these files needed in the WSJMix folder?

mravanelli:
Thank you @ycemsubakan!

PR #596, opened 2021-03-21
Updates to sequential class to clarify usage

pplantinga:
Add a few error messages and update CRDNN to add `input_size` option.

Still need to update a bunch of recipes to use `input_size` for CRDNN instead of `input_shape`


mravanelli:
LGTM. I did some tests and everything seems to work. Thank you @pplantinga.

Issue #594, opened 2021-03-21
interfaces.py separate_file() not working properly?


UrosOgrizovic:
Specs: Windows Version	10.0.19042 Build 19042, Python 3.8.3

I'm following the Source Separation tutorial that can be accessed from [this page](https://speechbrain.github.io/tutorial_processing.html).

Since the audio I have is already mixed, I tried to use `model.separate_file()`, based off the HuggingFace speechbrain/sepformer-wsj02mix [code](https://huggingface.co/speechbrain/sepformer-wsj02mix). There are two issues with that:

1. CPU, RAM and disk usage on my PC shoot up into the high nineties, causing the computer to freeze.
2. After the computer unfreezes (presumably, after the calculations are finished), an error is thrown:
```
Traceback (most recent call last):
  ...
    est_sources = model.separate_file(path='data/test/audio/speech.wav')
  File "C:\...\venv\lib\site-packages\speechbrain\pretrained\interfaces.py", line 710, in separate_file
    est_sources = self.separate_batch(batch)
  File "C:\...\venv\lib\site-packages\speechbrain\pretrained\interfaces.py", line 669, in separate_batch
    est_mask = self.modules.masknet(mix_w)
  File "C:\...\venv\lib\site-packages\torch\nn\modules\module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\...\venv\lib\site-packages\speechbrain\lobes\models\dual_path.py", line 1124, in forward
    x = self.dual_mdl[i](x)
  File "C:\...\venv\lib\site-packages\torch\nn\modules\module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\...\venv\lib\site-packages\speechbrain\lobes\models\dual_path.py", line 975, in forward
    inter = self.inter_mdl(inter)
  File "C:\...\venv\lib\site-packages\torch\nn\modules\module.py", line 889, in _call_impl
    result = self.forward(*input, **kwargs)
  File "C:\...\venv\lib\site-packages\speechbrain\lobes\models\dual_path.py", line 597, in forward
    return self.mdl(x + pos_enc)[0]
RuntimeError: The size of tensor a (2830) must match the size of tensor b (2500) at non-singleton dimension 1
```

My guess is that the data that `separate_file()` passes to `separate_batch()` is incorrect:
```
source, fl = split_path(path)
path = fetch(fl, source=source, savedir=savedir)
[...]

PR #597, opened 2021-03-22
Separation pretrained fix

ycemsubakan:
As requested in issue #594 , integrated resampling in pretrained separation models.

### DIFF
--- a/speechbrain/nnet/CNN.py
+++ b/speechbrain/nnet/CNN.py
@@ -3,6 +3,7 @@
 Authors
  * Mirco Ravanelli 2020
  * Jianyuan Zhong 2020
+ * Cem Subakan 2021
 """
 
 import math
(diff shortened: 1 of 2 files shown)


speechbrain/pretrained/interfaces.py
```diff
@@ -705,8 +705,16 @@ def separate_file(self, path, savedir="."):
         """
         source, fl = split_path(path)
         path = fetch(fl, source=source, savedir=savedir)
+        batch, fs = torchaudio.load(path)
+
+        # resample the data if needed
+        data_folder = self.hparams.data_folder
+        if ("wsj" in data_folder or "wham" in data_folder) and fs != 8000:
```
mravanelli, line 712: This looks weird to be. I think we should only check if the actual sampling frequency is the one specified in the yaml file, right? This class should be dataset-independent...

speechbrain/nnet/CNN.py
```diff
@@ -654,6 +657,179 @@ def _check_input(self, shape):
         return in_channels
 
 
+class ConvTranspose1d(nn.Module):
```
mravanelli, line 660: Ok. this part should be removed and we should use the one just merged in the development branch

mravanelli (changes requested):

PR #603, opened 2021-03-24
Allows TorchScript tracing of pretrained ASR models

Tpt:
Fixes the `torch.jit.trace` warnings and errors on pretrained ASR models `compute_features`, `normalize` and `asr_encoder` models.

Tracing of `beeam_search` still fails but it is going to be significantly more work to make it work.

### DIFF
--- a/speechbrain/nnet/CNN.py
+++ b/speechbrain/nnet/CNN.py
@@ -845,8 +845,7 @@ def get_padding_elem(L_in: int, stride: int, kernel_size: int, dilation: int):
         padding = [kernel_size // 2, kernel_size // 2]
 
     else:
-        L_out = (L_in - dilation * (kernel_size - 1) - 1) / stride + 1
-        L_out = int(L_out)
+        L_out = (L_in - dilation * (kernel_size - 1) - 1) // stride + 1
 
         padding = [(L_in - L_out) // 2, (L_in - L_out) // 2]
     return padding
--- a/speechbrain/processing/NMF.py
+++ b/speechbrain/processing/NMF.py
@@ -175,8 +175,10 @@ def reconstruct_results(
                 dim=-1,
             )
         )
-        shat1 = ISTFT(X1hat_stft.unsqueeze(0).permute(0, 2, 1, 3))
-        shat2 = ISTFT(X2hat_stft.unsqueeze(0).permute(0, 2, 1, 3))
+        X1hat_stft = X1hat_stft.unsqueeze(0).permute(0, 2, 1, 3)
+        X2hat_stft = X2hat_stft.unsqueeze(0).permute(0, 2, 1, 3)
+        shat1 = ISTFT(X1hat_stft)
+        shat2 = ISTFT(X2hat_stft)
 
         div_factor = 10
         x1 = shat1 / (div_factor * shat1.std())
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -696,9 +696,7 @@ class Filterbank(torch.nn.Module):
         x_db -= self.multiplier * self.db_multiplier
 
         # Setting up dB max
-        new_x_db_max = torch.tensor(
-            float(x_db.max()) - self.top_db, dtype=x_db.dtype, device=x.device,
-        )
+        new_x_db_max = x_db.max() - self.top_db
         # Clipping to dB max
         x_db = torch.max(x_db, new_x_db_max)
 
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -193,7 +193,7 @@ class ISTFT(torch.nn.Module):
 
     This class computes the Inverse Short-Term Fourier Transform of
     an audio signal. It supports multi-channel audio inputs
-    (batch, time_step, n_fft, 2, n_channels [optional]).
+    (batch, time_step, n_fft, n_channels [optional], 2).
 
     Arguments
     ---------
@@ -301,6 +301,9 @@ class ISTFT(torch.nn.Module):
             x = x.reshape(-1, x.shape[2], x.shape[3], x.shape[4])
         elif len(or_shape) == 4:
             x = x.permute(0, 2, 1, 3)
+
+        # isft ask complex input
+        x = torch.complex(x[...,0], x[...,1])
 
         istft = torch.istft(
             input=x,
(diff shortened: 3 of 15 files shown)


pplantinga (commented):
These changes look good, avoiding python types is going to be more efficient and allows for TorchScript tracing. If someone can verify that this does indeed allow TorchScript tracing, we can go ahead and merge it.

TParcollet:
Isn't removing the device a potential issue ?

Tpt:
> Isn't removing the device a potential issue ?

`new_x_db_max` is computed from `x_db` and float constants and `x_db` is itself computed from `x` and float constants. So, I believe it is safe to assume that `new_x_db_max` will end up on the same device as `x`.

Tpt:
> If someone can verify that this does indeed allow TorchScript tracing, we can go ahead and merge it.

I can add unittests for that if you want.

mravanelli:
Thank you @Tpt! I can do a test to make sure we have exactly the same performance as before with a pretrained model. As you mentioned, it could be great to take this opportunity to make the beamsearch part jitable with  `torch.jit.trace`. What are the issues that you are encountering here?

mravanelli:
I did some tests:
1. I confirm that now the pipeline from feature extraction to the ASR posterior probability is  jitable. To double check:
```
 import torch
asr_model = EncoderDecoderASR.from_hparams(source="speechbrain/asr-crdnn-rnnlm-librispeech", savedir="pretrained_models/asr-crdnn-rnnlm-librispeech")
traced_foo = torch.jit.trace(asr_model.modules.normalize, (torch.rand([4,40]),torch.ones(4)))
 traced_foo = torch.jit.trace(asr_model.modules.compute_features, (torch.rand([4,1600])))
traced_foo = torch.jit.trace(asr_model.modules.asr_encoder, (torch.rand([4,100,40])))
```
2. I did some tests with a pre-trained model and,  as expected, the performance is the same as before
3. I confirm that the current version of the beaformer is not currently jitable:
```
>>> traced_foo = torch.jit.trace(asr_model.modules.beam_searcher, (torch.rand([4,50,512]),torch.ones(4)))
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/dataio/dataio.py:706: TracerWarning: torch.as_tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  mask = torch.as_tensor(mask, dtype=dtype, device=device)
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:1204: TracerWarning: torch.Tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  cond, tensor, torch.Tensor([fill_value]).to(tensor.device)
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:485: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if eos_indices.shape[0] > 0:
/home/mila/r/ravanelm/anaconda3/envs/py18/lib/python3.9/site-packages/torch/tensor.py:587: RuntimeWarning: Iterating over a tensor might cause the trace to be incorrect. Passing a tensor of different shape won't change the number of iterations executed (and might lead to errors or silently give incorrect results).
  warnings.warn('Iterating over a tensor might cause the trace to be incorrect. '
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:488: TracerWarning: Converting a tensor to a Python number might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  index = index.item()
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:535: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  top_lengths = torch.tensor(
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:548: TracerWarning: Converting a tensor to a Python number might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  topk_log_probs = [top_lengths[index.item()] for index in indices]
/network/tmp1/ravanelm/speechbrain_github/speechbrain_jit2/speechbrain/decoders/seq2seq.py:1311: TracerWarning: Converting a tensor to a Python list might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  res = filter_seq2seq_output(p.tolist(), eos_id=eos_id)
  File "<stdin>", line 1, in <module>
  File "/home/mila/r/ravanelm/anaconda3/envs/py18/lib/python3.9/site-packages/torch/jit/_trace.py", line 733, in trace
    return trace_module(
  File "/home/mila/r/ravanelm/anaconda3/envs/py18/lib/python3.9/site-packages/torch/jit/_trace.py", line 934, in trace_module
    module._c._create_method_from_trace(
RuntimeError: Only tensors, lists, tuples of tensors, or dictionary of tensors can be output from traced functions
```
4. I think it is a good idea adding some unittest to make sure the most important modules of speechbrain are jitable.

Tpt:
@mravanelli Thank you for your investigation and tests. I have added `ŧorch.jit.trace` calls to most of the Torch Modules provided by SpeechBrain. `ŧorch.jit.trace` automatically test that the traced module output is close to the source module output on the provided input.

mravanelli:
Thank you very much @Tpt! I added few more tests. There are few things that are missing to have full jitability. One is the beamformer. Beyond that, I would like to make sure that all the models that we proposed in lobes.models are jitable (I tried some of them but there is something to fix). For now, I think we can merge this PR and eventually ask another one for the remaining issues.

Tpt:
Thank you!

PR #608, opened 2021-03-28
implemented automatic resampling for the separation pretrained class

ycemsubakan:
Moved the earlier pull request to here. We automatically resample in the pretrained class for separation.


mravanelli:
Thank you @ycemsubakan. I tested it and everything works as expected!

PR #615, opened 2021-03-31
Amp4 gradient clipping

vielzeuf:
As stated in issue#612, the change is adding an unscaling step for amp before gradient clipping (as recommended by Pytorch documentation).


mravanelli:
Thank you @vielzeuf. I tested and everything seems to work. I would like to open soon an new PR to expose the `auto_mix_prec` params in all the recipes (and set it to True where we see some seed improvements). Keep me updated on the model you are testing!

PR #616, opened 2021-03-31
Add single value (per Epoch) logging to TensorboardLogger

aelimame:
I noticed that when logging to Tensorboard using TensorboardLogger a full list of per batch metrics is logged at each epoch (a lot of values). This makes visualization of training progress basically unreadable and defats the purpose of using Tensorboard.

I did a minor change to make it possible to log a single value at the end of the epoch. And it is now much more readable, and we can clearly see the progress of the training at each epoch.

I also kept the list (Per batch) logging for now (Will auto detect if single value or a list is submitted to log_stats() ). This is to keep backward compatibility with existing recipes.
But I think we need to remove it all together since it will not be very useful nor readable (lots of values) and because the steps won't represent the real training epoch number in this case.

### DIFF
--- a/recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml
+++ b/recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml
@@ -24,7 +24,7 @@ train_log: !ref <output_folder>/train_log.txt
 enhanced_folder: !ref <output_folder>/enhanced_wavs
 
 # Basic parameters
-use_tensorboard: False
+use_tensorboard: True
 tensorboard_logs: !ref <output_folder>/logs/
 
 # FFT paremeters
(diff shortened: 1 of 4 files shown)


mravanelli:
@pplantinga are you ok with this change in the train_logger? It would be great if someone regularly using the tensorboard double check this PR.

pplantinga:
I think that having the option to do either per-batch or per-epoch statistics is a good solution. I've seen charts made both ways and there are tasks/datasets for which one or the other makes more sense. Maybe the `TODO` can be removed, since I think it makes sense to support both approaches.

aelimame:
> I think that having the option to do either per-batch or per-epoch statistics is a good solution. I've seen charts made both ways and there are tasks/datasets for which one or the other makes more sense. Maybe the `TODO` can be removed, since I think it makes sense to support both approaches.

Well I think it depends on the specific usage when logging per batch values (Basically the size of the batches and dataset). We can keep both if you think it is ok.
I'll remove the TODO and just keep the comment that it supports both.

mravanelli:
Thank you @aelimame! I tested it and everything seems to work. I think we can merge it!

PR #622, opened 2021-04-03
Include ckpt_interval_minutes: 15

mschonwe:
Looks like this recipe didn't get the checkpointing interval set.  Given an epoch can take a while, and prone to memory overflow on smaller GPUs, it is nice to have more intra-epoch checkpoints.


mravanelli:
Right. I remember we had it but @JianyuanZhong remove it. @JianyuanZhong is there any reason for not having it?

TParcollet:
@pplantinga any idea why tests are failing so much on the different yaml files while they are not changed w.r.t master branch ?

pplantinga:
There's a bunch of warnings (which are always there) but only a single error:
```
./recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml
  41:56     error    trailing spaces  (trailing-spaces)
```
Try removing the trailing spaces.

TParcollet:
LGTM merging.

PR #623, opened 2021-04-03
Added the message for num_workers for MacOS users

ycemsubakan:


mravanelli:
thank you for the fix!

PR #624, opened 2021-04-03
Change boolean run options to flags

pplantinga:
Several boolean run options are set up to pass values, but record `True` regardless of value, e.g.

```bash
> python train.py train.yaml --auto_mix_prec=True
> python train.py train.yaml --auto_mix_prec=False
```
both set `auto_mix_prec` option to `True`.

More canonical approach is to use flags instead of taking values, this PR changes it to:
```bash
> python train.py train.yaml --auto_mix_prec
```

### DIFF
--- a/docs/experiment.md
+++ b/docs/experiment.md
@@ -43,10 +43,11 @@ For more details on YAML and our extensions, please see our dedicated [tutorial]
 ## Running arguments
 SpeechBrain defines a set of running arguments that can be set from the command line args (or within the YAML file).
 - `device`: set the device to be used for computation.
-- `data_parallel_backend`: default False, if True, use `data_parallel` for multigpu training on a single machine.
-- `data_parallel_count`: default "-1" (use all gpus), if > 0, use a subset of gpus available [0, 1, ..., data_parallel_count].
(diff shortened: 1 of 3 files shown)


mravanelli:
I tested and it seems to work. One annoying thing is that some changes in the documentation (https://speechbrain.readthedocs.io/en/latest/multigpu.html) and tutorials (https://colab.research.google.com/drive/13pBUacPiotw1IvyffvGZ-HrtBr9T6l15?usp=sharing#scrollTo=ck1qe3KvpZPV) are needed as well. This brings me to a question. Do we want to have documentation and tutorials to be aligned with the development or the main branch? Note that sometimes in tutorials we do "pip install speechbrain" (main branch) and in other cases "git clone ..." that points (by default) to the development branch.

mravanelli:
I think one possible solution, would be to create in the next few days a version '0.5.5' with all the latest changes and bug fixes. This way, we can merge the develop branch into the main branch and updates the doc as well.

pplantinga:
Every version of the documentation will have listed the version that it refers to. Maybe we should list at the top of each tutorial the version that it was written with, along with `pip install speechbrain==0.5.4` as the install or `git clone speechbrain; git checkout [tagged version]`.

For people who clone the repository, I think we can best help them by maintaining a "tight" release schedule, where we release quite frequently (once a week or so). This way `develop` never gets too far out of sync with the most recent documentation.

PR #625, opened 2021-04-05
Modify Fbank to expose f_max, win_length and hop_length arguments.

aelimame:
It would be very useful to have the `f_max`, `win_length` and `hop_length `arguments exposed. Otherwise it is impossible to set the desired number of output time steps when using `Fbank`.

The default values have been set based on `STFT()` and `Filterbank()` default values to keep the same behavior for existing recipes.

### DIFF
--- a/speechbrain/lobes/features.py
+++ b/speechbrain/lobes/features.py
@@ -245,6 +245,7 @@ class MFCC(torch.nn.Module):
         )
 
         self.compute_fbanks = Filterbank(
+            sample_rate=sample_rate,
             n_fft=n_fft,
             n_mels=n_mels,
             f_min=f_min,
--- a/speechbrain/lobes/features.py
+++ b/speechbrain/lobes/features.py
@@ -32,8 +32,15 @@ class Fbank(torch.nn.Module):
     sample_rate : int (default: 160000)
         Sampling rate for the input waveforms.
     f_min : int (default: 0)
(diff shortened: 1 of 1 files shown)


TParcollet:
Hey thank you for the update ! This looks like a nice idea, but then could you add it also to MFCC and add the documentation to the docstring ?

Thanks !

mravanelli:
Right @TParcollet. At this point, I would expose the `f_min` parameter as well.

aelimame:
Yes @TParcollet, I totally forgot about the docstring.
@mravanelli  `f_min` is already there.
I'll also do the same for MFCC.

aelimame:
I also noticed that default values are not specified in the docstring for MFCC, so I added them.

TParcollet:
@mravanelli LGTM. Did not try it, but test are passing and the changes aren't changing anything drastically.

aelimame:
@mravanelli Looks like when using a sample rate other than 16KHz the expected range we get (fmax) does not seem to be right.
There is a discussion with Cem in the Slack channel with some screenshots comparing the output @8KHz.

aelimame:
OK update on the situation. Looks like there is a bug and it is because we don't pass the `sample_rate` parameter to `FilterBank()`. It is only passed to `STFT()` so using a different sample rate than the default 16KHz breaks things...
I'll fix this and I think we should be fine.

mravanelli:
thank you @aelimame for this fix. I think we can merge it!

PR #626, opened 2021-04-05
Fix merging and splitting not using the "space" arg

pplantinga:
Bug with merging and splitting tokens, used by error rate metrics. Also add option to use different character in the merging and splitting, and use this new functionality to fix the mimic loss recipe.


TParcollet:
@pplantinga what is the status of this PR ?

PR #632, opened 2021-04-07
 catch and fix file encoding

flozi00:
This PR catches an error I recognized while running commonvoice

### DIFF
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -210,7 +210,7 @@ def create_csv(
         raise FileNotFoundError(msg)
 
     # We load and skip the header
-    loaded_csv = open(orig_tsv_file, "r").readlines()[1:]
+    loaded_csv = open(orig_tsv_file, "r", encoding="utf-8").readlines()[1:]
     nb_samples = str(len(loaded_csv))
 
     msg = "Preparing CSV files for %s samples ..." % (str(nb_samples))
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -125,7 +125,7 @@ def load_data_csv(csv_path, replacements={}):
     '/home/utt1.wav'
     """
 
-    with open(csv_path, newline="") as csvfile:
+    with open(csv_path, newline="", encoding="utf-8") as csvfile:
         result = {}
         reader = csv.DictReader(csvfile, skipinitialspace=True)
         variable_finder = re.compile(r"\$([\w.]+)")
--- a/speechbrain/tokenizers/SentencePiece.py
+++ b/speechbrain/tokenizers/SentencePiece.py
@@ -224,7 +224,7 @@ class SentencePiece:
                 )
                 break
             row_idx += 1
-            if(len(row) > index_label):
+            if len(row) > index_label:
                 sent = row[index_label]
                 if self.char_format_input:
                     (sent,) = merge_char([sent.split()])
--- a/speechbrain/tokenizers/SentencePiece.py
+++ b/speechbrain/tokenizers/SentencePiece.py
@@ -206,7 +206,7 @@ class SentencePiece:
             + " sequences from:"
             + self.annotation_train
         )
-        annotation_file = open(self.annotation_train, "r")
+        annotation_file = open(self.annotation_train, "r", encoding="utf-8")
         reader = csv.reader(annotation_file)
         headers = next(reader, None)
         if self.annotation_read not in headers:
@@ -214,7 +214,7 @@ class SentencePiece:
                 self.annotation_read + " must exist in:" + self.annotation_train
             )
         index_label = headers.index(self.annotation_read)
-        text_file = open(self.text_file, "w+")
+        text_file = open(self.text_file, "w+", encoding="utf-8")
         row_idx = 0
         for row in reader:
             if self.num_sequences is not None and row_idx > self.num_sequences:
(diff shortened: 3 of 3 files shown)


TParcollet:
@aheba could you have a look at the change made on SentencePiece on this PR ? The one on CV LGTM.

Gastron:
Hey, thanks for the PR!

Hardcoding UTF-8 might be problematic for core library code (dataio, tokenizer): what if someone needs to use a different encoding? Though I am not sure if we have any defined stance on encodings in SpeechBrain. [This  post](http://python-notes.curiousefficiency.org/en/latest/python3/text_file_processing.html) lists different strategies for dealing with text files. Before this PR, the library code doesn't specify an encoding explicitly, so Python looks at the OS defaults and the PYTHONIOENCODING environment variable (this is the `Files in a typical platform specific encoding` strategy, new strategy would be `Files in a consistent, known encoding`). Seems like the PR is really about the Common Voice recipe, though, am I right, @flozi00 ? Any thoughts on this? I feel like this is a bigger question.

Hardcoding an encoding for a particular recipe seems natural, when that dataset has a consistent encoding.

TParcollet:
I don't really know if CV is consistent on his encoding scheme though .....

flozi00:
Could we pass the encoding scheme from the hyperparameters ?
The german commonvoice is not 100% clean and contains characters from france and greek too, that'swhy the default encoding crashed.
Retrieving the encoding from system isn't clean solution, if I want to train other languages too, then I would have to change system settings for each training.

TParcollet:
Sure, simply create a yaml variable and get it when calling commonvoice_prepare with hparams["your_variable"].

flozi00:
Can we set this variable global ?
So we don't need to pass at every function call, just import it from somewhere. Like this:

file_utils.py
`file_encoding = "utf-8"`

Recipe:
```
from speechbrain.utils.file_utils import file_encoding
file_encoding = "my encoding" #override default value
```

In core of the library:
```
from speechbrain.utils.file_utils import file_encoding

open(filepath, "r", encoding=file_encoding)
```

Gastron:
Without explicitly specifying the encoding, Python3 looks at a few environment sources to determine which encoding to expect. Doesn't that achieve the same globally specified encoding, but keep the ability to change the global default?

flozi00:
The way with Environment variables has some more limitations I think. If you are okay with the global config variable I would start at weekend with doing it. Just tell me which way you want to go, Python environment variables or global variable in code :)

mravanelli:
Thank you @flozi00. What do you think @Gastron about this change?

Gastron:
I like the environment variable approach (i.e. keeping the core library as is.), because it uses the Python standard approach. I'm happy to hear more elaboration on the possible downsides of that approach, I can surely be missing something. But if nothing else, I'd like to go with the current environment variable approach and just change the recipe in this PR.

speechbrain/dataio/dataio.py
```diff
@@ -125,7 +125,7 @@ def load_data_csv(csv_path, replacements={}):
     '/home/utt1.wav'
     """
 
-    with open(csv_path, newline="") as csvfile:
+    with open(csv_path, newline="", encoding="utf-8") as csvfile:
```
Gastron, line 128: Let's just discard this change

speechbrain/tokenizers/SentencePiece.py
```diff
@@ -206,15 +206,15 @@ def _csv2text(self):
             + " sequences from:"
             + self.annotation_train
         )
-        annotation_file = open(self.annotation_train, "r")
+        annotation_file = open(self.annotation_train, "r", encoding="utf-8")
```
Gastron, line 209: Again, just discard these.

speechbrain/tokenizers/SentencePiece.py
```diff
@@ -206,15 +206,15 @@ def _csv2text(self):
...
         reader = csv.reader(annotation_file)
         headers = next(reader, None)
         if self.annotation_read not in headers:
             raise ValueError(
                 self.annotation_read + " must exist in:" + self.annotation_train
             )
         index_label = headers.index(self.annotation_read)
-        text_file = open(self.text_file, "w+")
+        text_file = open(self.text_file, "w+", encoding="utf-8")
```
Gastron, line 217: And this

speechbrain/tokenizers/SentencePiece.py
```diff
@@ -224,11 +224,12 @@ def _csv2text(self):
                 )
                 break
             row_idx += 1
-            sent = row[index_label]
-            if self.char_format_input:
-                (sent,) = merge_char([sent.split()])
-                sent = " ".join(sent)
-            text_file.write(sent + "\n")
+            if len(row) > index_label:
```
Gastron, line 227: This seems to be a silent skip for empty CSV lines or lines which have missing columns. Feels like this should rather be fixed at the CSV creation stage, right? I feel like a silent skip should maybe only happen for empty lines. But this is a minor comment.

recipes/CommonVoice/common_voice_prepare.py
```diff
@@ -210,7 +210,7 @@ def create_csv(
         raise FileNotFoundError(msg)
 
     # We load and skip the header
-    loaded_csv = open(orig_tsv_file, "r").readlines()[1:]
+    loaded_csv = open(orig_tsv_file, "r", encoding="utf-8").readlines()[1:]
```
Gastron, line 213: Keep this if Common Voice is reliably utf-8.

Gastron (changes requested):
Added formal review, but just basically reiterates my earlier comment.

PR #679, opened 2021-04-23
gradio web demo

AK391:
added easy to use web demo for sepformer on gradio hub


mravanelli:
Hi,
thank you very much for this PR. After an internal discussion, we don't really want to add this in the main repo. This would force users to have gradio as a mandatory external dependency (which is something we are trying to minimize). It however good to keep the SpeechBrain interfaces on your website. We just ask you to add a clear link to the original repo.

PR #685, opened 2021-04-23
Implicitly discover doctests so we follow rules in conftest.py

pplantinga:
Doctests are failing locally due to failure to skip tests that require `fairseq`. I'm not sure why this is working on GitHub CI, its possible `fairseq` is getting installed. This fix just uses automatic docstring discovery to run the tests, but I'm not sure why the old approach was used, so maybe we need a different approach to fixing this.

We also need to decide if we want these tests to be run on GitHub (if they aren't already).

### DIFF
--- a/tests/.run-doctests.sh
+++ b/tests/.run-doctests.sh
@@ -1,7 +1,4 @@
 #!/bin/bash
 set -e -u -o pipefail
 
-# It is ugly but currently we have to specifically filter out transducer loss here.
-# Otherwise pytest will stubbornly fail on import if you don't have numba.
-# (In contrast, implicitly discovered files which produce ImportErrors are ignored.)
-git ls-files speechbrain | grep -e "\.py$" | grep -v "transducer_loss.py" | xargs pytest --doctest-modules
+pytest --doctest-modules speechbrain/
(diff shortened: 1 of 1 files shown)


Gastron (commented):
Responded on Slack - let's move discussion here at this point. Just reposting the gist of my response:

The comment in `.run-doctests.sh` is missing the main point for the convoluted approach. The point was that locally you can easily have some additional (untracked) files in the speechbrain directory - something you’re just working on but haven’t committed yet. Those files would get tested in the pre-push hook as well, even though they’re in no way related to the commit. I remember this was a practical annoyance too.

The `git ls-files` approach is still imperfect - won't deal with pushing different branches, won't deal with uncommitted changes in tracked files.

I'm ok with this change, but it will reintroduce the old annoyance - or we can also decide to just update the comment on it, and just add another custom `grep -v`.

Or if you’re up to it, you can even make the local tests scripts more sophisticated (checkout the correct branch, stash changes, undo stashing checkout the previous branch, all this with proper dealing for errors/ctrl-c, I think).

mravanelli:
Ok, I think for now the best approach is to restore the grep approach and update the list of files that we don't have to test.

pplantinga:
We still need to add a comment to this code explaining the not-straightforward approach that's taken.

mravanelli:
Right, feel free to add it and merge!

PR #706, opened 2021-04-29
Less alarming logging in dataloader

Gastron:
The logging in the SaveableDataLoader was too alarming for something that happens in a normal experiment.


mravanelli:
I agree on that. Also my students thought that something went wrong after reading the original message. Can we merge it @TParcollet ?

TParcollet:
lgtm

PR #718, opened 2021-04-30
fairseq wav2vec 2.0 fix the normalize argument

TParcollet:
This PR fixes a problem where the normalize argument isn't found in the cfg file causing a wrong normalization (or an error).

### DIFF
--- a/speechbrain/lobes/models/fairseq_wav2vec.py
+++ b/speechbrain/lobes/models/fairseq_wav2vec.py
@@ -88,8 +88,10 @@ class FairseqWav2Vec2(nn.Module):
         # Hence, we check if the model has be trained with or without it.
         # If the information isn't contained in the checkpoint it is set to False.
         if input_norm is None:
-            if not hasattr(cfg["task"], "normalize"):
+            if hasattr(cfg["task"], "normalize"):
                 self.normalize = cfg["task"].normalize
+            elif hasattr(cfg, "normalize"):
+                self.normalize = cfg.normalize
             else:
                 self.normalize = False
         else:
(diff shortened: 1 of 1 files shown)


speechbrain/lobes/models/fairseq_wav2vec.py
```diff
@@ -86,19 +86,12 @@ def __init__(
 
         # wav2vec pretrained models may need the input waveform to be normalized
         # Hence, we check if the model has be trained with or without it.
-        # If the information isn't contained in the checkpoint IT HAS TO BE GIVEN
-        # BY THE USER.
+        # If the information isn't contained in the checkpoint it is set to False.
         if input_norm is None:
-            if hasattr(cfg, "normalize"):
-                self.normalize = cfg.normalize
+            if not hasattr(cfg["task"], "normalize"):
```
mhn226, line 91: Shouldn't have "not" presented here

mhn226: With some older versions of fairseq wav2vec model, for instance (facebook/wav2vec2-large-960h), "normalize" seems to be outside of cfg["task"]. One might need to check this case also.

`if input_norm is None:
     if hasattr(cfg["task"], "normalize"):
         self.normalize = cfg["task"].normalize
     elif hasattr(cfg, "normalize"):
          self.normalize = cfg.normalize
      else:
          self.normalize = False
else:
    self.normalize = input_norm`

TParcollet: SHould be better now

PR #725, opened 2021-05-02
Fixes typos

calvinballing:
Each typo is corrected in its own commit, as listed below.


mravanelli:
That is super useful! Thank you!

Issue #580, opened 2021-03-17
Support for sharded datasets (e.g WebDataset API) for efficiently working on network-attached filesystems


nikvaessen:
First of all: thanks for making such an amazing tool public!

I was interested in using the VoxCeleb speaker recognition recipe as a baseline/starting point for a research project. However, a major issue is how the data is stored on disk. If I understand correctly SpeechBrain follows the pattern that you have a CSV/JSON manifest, wherein each training sample has a path to a (e.g audio) file on a filesystem mounted on the machine.

In the case of VoxCeleb1 and 2, this amounts to ~1.200.000 files in a single folder. On a local NVMe SSD this is fine, but our cluster uses a network-attached filesystem for providing most storage space. This has severe performance implications - loading ~100.000 audio files (a single epoch of VoxCeleb1) with a network-attached filesystem takes an hour. There is a fairly straightforward solution to this problem: storing your data in so-called shards (e.g TFRecords in Tensorflow). In PyTorch there is a current proposal [1] to include the WebDataset [2, 3] library as an API for sharded datasets. The basic premise of this library is storing your data in multiple `.tar` or `.tar.gz` files.  In my case, this reduced the loading time of a single VoxCeleb1 epoch from an hour to 3 minutes. 

I wanted to discuss a potential integration of WebDatasets in SpeechBrain. Currenty,  `speechbrain.dataio.datasets.DynamicItemDataset` inherits from `torch.util.data.Dataset`, while the `WebDataset` class uses the `torch.util.data.IterableDataset` API. I am not yet familiar enough with the code so I had the following question(s):

* Is there an existing solution for efficiently using speechbrain on network-attached file systems?
* If not, would there be interested in merging a PR offering support for `WebDataset`?
* Any suggestions for the best way to integrate `WebDataset`?

links:
[1] https://github.com/pytorch/pytorch/issues/38419
[2] https://github.com/tmbdev/webdataset
[3] https://pytorch.org/blog/efficient-pytorch-io-library-for-large-datasets-many-files-many-gpus/

TParcollet:
@Gastron @popcornell

Gastron:
# Efficient data IO on NFS

[...]

PR #734, opened 2021-05-05
On-the-fly dynamic batching and LoopedLoader (for webdataset-style IterableDataset support)

Gastron:
After working with [webdataset](https://github.com/tmbdev/webdataset) for a while, and informed by the discussion in issue #580, this setup seemed the best for now:

- Don't care about exact epochs (seeing every sample exactly once). Instead set a nominal epoch length (number of updates per epoch). This way we don't need to care about how the shards are split and how many workers we have loading the data.
- Create batches with on-the-fly dynamic batching. I also tried sorting the data in the shards, so that adjacent samples would have the similar lengths. That would still require somehow dynamically reducing the batch size as we get to longer utterances, and it required a lot of effort before-hand.


Gastron:
I want to add some example next, but sharing this here already for discussion.

Gastron:
@mravanelli I added a version of the TIMIT recipe, which uses webdataset and dynamic batching. On my side it seems to run just a little slower than the normal version (7% slower :/). I tried to set the dynamic batching parameters so that the average batch size would be close to 8. I got 13.74 test PER :) Do you want to test this on your system? Should be as simple as `pip install webdataset` and then running training with `train_webdataset.py`

speechbrain/core.py
```diff
@@ -965,11 +969,17 @@ def fit(
             Whether to display the progress of each epoch in a progressbar.
         """
 
-        if not isinstance(train_set, DataLoader):
+        if not (
+            isinstance(train_set, DataLoader)
+            or isinstance(train_set, LoopedLoader)
+        ):
```
pplantinga, line 975: Is there no way to make the `LoopedLoader` inherit from `DataLoader`? It seems like you should only need one check for these...

speechbrain/dataio/iterators.py
```diff
@@ -0,0 +1,218 @@
...
+def padding_ratio(lengths):
+    # How much of batch is padding:
+    return 1.0 - sum(lengths) / total_length_with_padding(lengths)
+
+
+@dataclass(order=True)
+class RatioIndex:
+    ratio: float
+    index: int
+
```
pplantinga, line 34: I'm not super familiar with `@dataclass`... what's the rationale for using that here rather than e.g. named tuple or something?

pplantinga (commented):
I admit that I don't totally understand everything going on in this PR, but I think the looped loader is a good addition so I'm thinking this is ready to be added to the repo.

speechbrain/dataio/iterators.py
pplantinga: Ah, I guess its for the automatic generation of `__lt__` `__gt__` etc.

Gastron: This particular example could have been a namedtuple as well.  The other dataclass, LengthItem, makes it so that only the length value is ever used in comparisons. Since that one was a dataclass, I made this like that too. I think the main difference in usage is that tuples are immutable.

speechbrain/core.py
Gastron: Not really, DataLoader is very complex and this really does a different thing.

TParcollet:
@Gastron THis is a good PR. How do you propose that we proceed with the review process?

TParcollet:
@Gastron, could you provide a tutorial with this PR? I think that such an addition should be properly explained. The "large" scale aspect is important.

TParcollet:
Ok, currently testing this PR. @Gastron a tutorial DEFINITELY is needed before merging. By the way, could you tell me how your dynamic batching relates to the one introduced in the other PR? Will we have 2 different dynamic batching strategies ?

TParcollet:
__main__ - Average batchsize: 7.9523809523809526 -> wouldn't this info better be on the log.log only ?

TParcollet:
Got a test PER of 13.69. I will review the code more carefully tmr I think. @Gastron see my above question, and once we have a tutorial, we can merge I guess.

TParcollet:
2min2 on a 1080 ti.

Gastron:
The strategies are only different because the whole problem is also different. It might be possible to bring them closer in parametrisation, but I don't think having different parametrisations is such a big problem since the approaches are not interchangeable at all.

TParcollet (changes requested):
Huge work @Gastron. In general, I find that this add an extra layer of complexity to the recipe and code. And not a small one. As you will see in my comments, I think it is crucial to a acknowledge this and to say it: "It's more complex, but it provides X,Y,Z benefit, and we have a tutorial here". Why ? Imagine a newcomer comes to SpeechBrain and start with such a recipe. From a first view, he will say: This look as hard to understand as ESPnet :p.

Side note: Would you agree in helping us reviewing the other dynamic batch PR ? I think you are definitely the one that can "unify" both approaches in term of integration to the core library.

mravanelli:
Hi @Gastron ,
thank you very much for this PR.
Here are my comments:
1. I tried it and it is running on my side. I didn't finish a full experiment, but the results look reasonable in the first few epochs. In terms of speed, it looks very fast (I'm pretty impressed by that). Each epoch takes  2m 03 sec vs 2m 00 with the standard recipe.
2. I agree with Titouan that this adds an extra level of complexity that might be difficult to digest by the users. Adding a tutorial will help.
3. I think we should use this approach only for the recipes that use very big datasets (i.e., datasets likely larger than the SSD). I don't see a recipe with such a big dataset. Do you have in mind to create one?
4. Having this recipe on TIMIT was super good to debug it and compare its performance with the other recipes. However, I think the final destination should be in templates.(https://github.com/speechbrain/speechbrain/tree/develop/templates/speech_recognition/ASR). Templates are running recipes that users can use as reference. This way, we can take the opportunity to add there more comments on what is going on. I would also clarify in the docstring why and when we need this approach. What do you think?

Gastron:
Thanks for the detailed comments @TParcollet and @mravanelli! I think a lot of the confusing parts here are particularly confusing because I used TIMIT. The benefits are really realised on large datasets, and here I took the pre-existing TIMIT dataprep as a starting point, but normally for WebDataset the JSON data manifests would not be needed (just make TAR files right away). So I think my two main takeaways are then:

1. Make the tutorial, which will explain _why_ we do certain things as well.
2. I like Mirco's proposal that in the end, we would not include this modified TIMIT recipe, it was just a good debug tool. A template could be the way to go.

recipes/TIMIT/ASR/seq2seq/hparams/train_webdataset.yaml
```diff
@@ -0,0 +1,206 @@
...
+manifests:
+    train: !ref <output_folder>/train.json
+    valid: !ref <output_folder>/dev.json
+    test: !ref <output_folder>/test.json
+shards_root: !PLACEHOLDER  # Where to put the shards (top-dir), e.g. /tmp/timit-shards
+examples_per_shard: 1024
+skip_prep: False # Skip data preparation including shards
+
+
+# Training parameters
```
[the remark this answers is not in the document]
Gastron: More comments can be added for sure. But here I'd also like to hear your opinion, so I'll explain briefly. So the dynamic batching adds new utterances to the batch until the batch total number of elements (like `.numel()`) exceeds `target_batch_numel`, expect `max_batch_numel` should not be exceeded. Since the elements are in samples here, the numbers are large and unintuitive. However, they can be converted to seconds (like 23.75s here). 

Personally find it kind a nice to think of this dynamic batchsize as target number of seconds in a batch, do you feel the same? That's why I think this could be decent way to parametrize a dynamic batch size.

recipes/TIMIT/ASR/seq2seq/train_webdataset.py
```diff
@@ -0,0 +1,366 @@
...
+        .then(
+            sb.dataio.iterators.dynamic_bucketed_batch,
+            **hparams["dynamic_batch_kwargs"],
+        )
+        .repeat()
+    )
+    # For the valid and test data, we need to specify lengths:
+    valid_data = (
+        wds.WebDataset(
+            valid_shards, length=math.ceil(400 / hparams["eval_batchsize"])
```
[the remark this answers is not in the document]
Gastron: Yeah that's not explained... But it's just the number of samples in the TIMIT validation set.

recipes/TIMIT/ASR/seq2seq/hparams/train_webdataset.yaml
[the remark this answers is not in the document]
TParcollet: I think it totally makes sense. This would ease the relation to VRAM. I think ESPnet does the "same" even though they use seq_len which is quite disconnected from time if you don't know the parameterization.

TParcollet:
@Gastron is the tutorial ready? If so, I can have a look at it. Maybe the template can follow in a new PR.

mravanelli:
As discussed privately with @Gastron, we can merge this PR now and create another PR with the template (+ tutorial) in the near future.

mravanelli:
Also, this could be the way to go for very big datasets like [GigaSpeech](https://github.com/SpeechColab/GigaSpeech) that might not fit in SSDs or local disks

PR #738, opened 2021-05-07
Adding new CoopNet recipe + logger for wandb

nicoduchene:
CoopNet recipe written over the course of coursework.

Please let me know if anything should be changed.

Thanks for all the hard work!


speechbrain/utils/train_logger.py
```diff
@@ -144,15 +146,49 @@ def log_stats(
                 if stat not in self.global_step[dataset]:
                     self.global_step[dataset][stat] = 0
                 tag = f"{stat}/{dataset}"
-
```
nicoduchene, line 147: I think my code was not up to date here. I will edit asap.

nicoduchene:
@pplantinga  I'm not sure who should be reviewing this, perhaps you can suggest other reviewers if need be?

recipes/Voicebank/MTL/CoopNet/CoopNet.py
```diff
@@ -0,0 +1,224 @@
...
+    def save(self, path):
+        key = path.stem.split("optimizer_")[1]
+        torch_save(self.optimizers[key], path)
+
+    @checkpoints.mark_as_loader
+    def load(self, path, end_of_epoch, device):
+        del end_of_epoch
+        del device
+        key = path.stem.split("optimizer_")[1]
+        torch_recovery(self.optimizers[key], path)
```
Gastron, line 96: Just a quick question: This looks like the multioptimizer needs to be added multiple times to the checkpointer. Couldn't this class just save all its optimizers in one file?

nicoduchene: Hi Gastron, 
Sorry for the late reply.
From what I remember I set this up in this way because I thought that it could be useful to retrieve an optimizer directly from a file instead of loading all the optimizers and selecting one. 
I did not end up using it, so I could change it if a change is desirable.

mravanelli:
Hi @nicoduchene,
thank you very much for this PR! As mentioned privately, I started the review of this PR.  Apparently, everything is working and giving the expected results (I did some minor changes as you can see from my last commit). I only have issues when running this code:

`python train.py hparams/pretrain_asr_and_se.yaml --data_folder /localscratch/noisy-vctk-16k/`

```
  File "/network/tmp1/ravanelm/speechbrain_github/speechbrain_coopnet/recipes/Voicebank/MTL/CoopNet/train.py", line 115, in enhance_speech
    enhanced_mag = torch.expm1(predictions[f"feats_{layer}"])
KeyError: 'feats_0'
```

It looks like the part that defines "feats_" is missing when `enhance_type=noisy`. I would take this opportunity to also the other options (i.e., null, mapping, masking, noisy, clean).

TParcollet:
Hi @nicoduchene, where are we w.r.t the requested changes? Thanks!

PR #744, opened 2021-05-09
add bleu script

mhn226:
This PR introduces a BLEU measuring script, which exploits sacrebleu (https://github.com/mjpost/sacrebleu) to measure BLEU scores for Speech Translation tasks. It is therefore necessary to install sacrebleu for the script to work:

`pip install sacrebleu`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ joblib>=0.14.1
 numpy>=1.17.0
 packaging
 pre-commit>=2.3.0
-sacrebleu>=1.5.1
 scipy>=1.4.1
 sentencepiece>=0.1.91
 SoundFile; sys_platform == 'win32'
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,14 @@
 -r lint-requirements.txt
+huggingface_hub>=0.0.6
 hyperpyyaml>=0.0.1
 joblib>=0.14.1
 numpy>=1.17.0
 packaging
 pre-commit>=2.3.0
+sacrebleu>=1.5.1
 scipy>=1.4.1
 sentencepiece>=0.1.91
+SoundFile; sys_platform == 'win32'
 torch>=1.7.1
 torchaudio>=0.7.2
 tqdm>=4.42.0
-huggingface_hub>=0.0.6
-SoundFile; sys_platform == 'win32'
--- a/speechbrain/utils/bleu.py
+++ b/speechbrain/utils/bleu.py
@@ -3,12 +3,7 @@ from speechbrain.utils.metric_stats import MetricStats
 try:
     import sacrebleu
 except ImportError:
-    import pip
-
-    pip.main(["install", "sacrebleu"])
-    import sacrebleu
-
-    # raise ImportError("Please install sacrebleu (https://github.com/mjpost/sacreble) in order to use the BLEU metric")
+    print("Please install sacrebleu (https://github.com/mjpost/sacreble) in order to use the BLEU metric")
 
 
 def merge_words(sequences):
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,13 +1,14 @@
 -r lint-requirements.txt
+huggingface_hub>=0.0.6
 hyperpyyaml>=0.0.1
 joblib>=0.14.1
 numpy>=1.17.0
 packaging
 pre-commit>=2.3.0
+sacrebleu>=1.5.1
```
TParcollet, line 8: Please remove it from the general requirements.txt and instead try to see if it's installed when importing. If not, then print a message to install it ?

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
+import sacrebleu
+from speechbrain.dataio.dataio import merge_char, split_word
+from speechbrain.utils.metric_stats import MetricStats
+
+
+def merge_words(sequences):
```
TParcollet, line 6: Could you add a small description of what this function does ? And why we need it ?

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
...
+    results = []
+    for seq in sequences:
+        words = " ".join(seq)
+        results.append(words)
+    return results
+
+
+def detokenize_batch(detokenizer, sentences):
+    """
+    detokenizer: object
```
TParcollet, line 16: Short description of what the function does (even if it's clear from the name, it's better to explain it quickly for the users")

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
...
+    detok_sentences = []
+    for sentence in sentences:
+        sentence = detokenizer.detokenize(sentence)
+        detok_sentences.append(sentence)
+
+    return detok_sentences
+
+
+class BLEUStats(MetricStats):
+    """A class for tracking BLEU.
```
TParcollet, line 30: Docstring ok. Maybe a reference to a paper on BLEU ?

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
...
+        this token in between, and then the whole sequence is split.
+    Example
+    -------
+    >>> bleu = BLEUStats()
+    >>> i2l = {0: 'a', 1: 'b'}
+    >>> bleu.append(
+    ...     ids=['utterance1'],
+    ...     predict=[[0, 1, 1]],
+    ...     target=[[0, 1, 0]],
+    ...     ind2lab=lambda batch: [[i2l[int(x)] for x in seq] for seq in batch],
```
TParcollet, line 57: This line might be confusing. Consider doing a line before and explaining it briefly.

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
...
+        self.space_token = space_token
+
+        self.predicts = []
+        self.targets = []
+
+    def append(
+        self, ids, predict, target, ind2lab=None,
+    ):
+        """Add stats to the relevant containers.
+        * See MetricStats.append()
```
TParcollet, line 85: Link to this doc.

speechbrain/utils/bleu.py
```diff
@@ -0,0 +1,151 @@
...
+        Arguments
+        ---------
+        ids : list
+            List of ids corresponding to utterances.
+        predict : torch.tensor
+            A predicted output, for comparison with the target output
+        target : torch.tensor
+            The correct reference output, for comparison with the prediction.
```
Gastron, line 93: This doc and the implementation seem to assume one reference - but BLEU was specifically designed to handle multiple references, so we should support that as well.

TParcollet:
__main__ - Average batchsize: 7.9523809523809526 -> Wouldn't this information better be in the log.log only ?

TParcollet:
LGTM Thanks @mhn226 !

PR #746, opened 2021-05-10
SLU Multistage - Add Wav2vec2 CTC + Inference

aheba:
Hi, thanks @Moumeneb1  for this PR,

- [ ] add Wav2vec2CTC
- [ ] Load wav2vec2CTC Module from HF and predict text from wav form

### DIFF
--- a/recipes/timers-and-such/multistage/hparams/train_with_wav2vec2_asr.yaml
+++ b/recipes/timers-and-such/multistage/hparams/train_with_wav2vec2_asr.yaml
@@ -1,0 +1,189 @@
+# ############################################################################
+# Model: Multistage SLU
+# Encoder: Pre-trained ASR (Timers and Such LM) -> transcript -> LSTM
+# Decoder: GRU + beamsearch
+# Tokens: BPE with unigram
+# losses: NLL
+# Training: Timers and Such
+# Authors:  Loren Lugosch, Mirco Ravanelli 2020
+# ############################################################################
+
+# Seed needs to be set at top of yaml, before objects with parameters are made
+seed: 1986
+__set_seed: !apply:torch.manual_seed [!ref <seed>]
+experiment: train-real-and-synth
+output_folder: !ref results/TAS_LM/<experiment>/<seed>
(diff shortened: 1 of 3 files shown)


aheba:
This work is related to SLU multi-stage: new perf using Wav2vec2CTC instead of CRDNN. @Moumeneb1 , could you share your numbers

TParcollet:
Hey, can't we decouple their CTC head? I don't like the idea of using their CTC head within SpeechBrain because we already have our.

aheba:
Hey @TParcollet , We use a pre-trained wav2vec2 on Librispeech from HF,  do we have a Librispeech fine-tuned wav2vec within SpeechBrain ?

Or maybe I can convert the output layer ?!

TParcollet:
Let me send you my CV finetuned models (EN).

Moumeneb1:
Hi, 

In the PR we tried to experiment how well can the Wav2vec ASR (trained on CTC) improve the results. 

| Model  | Accuracy|
| ------------- | ------------- |
| CRDNN  LS_LM|  55.55% |
| CRDNN  TAS_LM|  64% |
| Wav2vec as ASR (facebook/wav2vec-base-960h) | 81.02% |
| Wav2vec as ASR (facebook/wav2vec-base-960h) + Bert | 85.53% |

The wav2vec gives better results, it can also be due to the fact that the model is trained with CTC, and doesn't have the looping behaviour as discussed with @lorenlugosch

lorenlugosch:
Cool! If you just use the CRDNN with CTC is it better too? Or is it only if you use wav2vec

lorenlugosch:
I do not. It would probably be easy to swap the original CRDNN encoder into your CTC finetuning script instead of wav2vec, though, if you wanted to try that.

TParcollet:
@aheba https://drive.google.com/file/d/1xOVwifiQNyyUuDT4lHDO9ni2GZSG1ZAz/view?usp=sharing Here is the w2v-en finetuned on commonvoice en (CTC only)

mravanelli:
@aheba, what is the status of this PR?

PR #751, opened 2021-05-11
SpeechBrain 0.6.0

30stomercury:
The goal of this PR is to support pure ctc training and decoding (beam search). Users can set `ctc_weight: 1`  and `ctc_weight_decode: 1` to perform pure ctc training and beamsearch.

Here are the results I got (CTC with transformerlm):
```
WER 5.22 [ 2742 / 52576, 440 ins, 343 del, 1959 sub ] on test-clean
WER 12.41 [ 6494 / 52343, 1041 ins, 762 del, 4691 sub ] on test-other
```

To-dos:
+ Integrate N-gram LM interface in arpa format.
+ Run ctc, joint ctc/att decoding (with and without LM) after modification.


speechbrain/decoders/seq2seq.py
```diff
@@ -644,15 +650,6 @@ def forward(self, enc_states, wav_len):  # noqa: C901
             if t < min_decode_steps:
                 log_probs[:, self.eos_index] = self.minus_inf
 
-            # Set the eos prob to minus_inf when it doesn't exceed threshold.
-            if self.using_eos_threshold:
```
30stomercury, line 648: I modify this part. We should check the eos threshold after all scorers (attn scorer, ctc scorer and lm scorer), or the emission of eos will only be based on the output of attn scorer.

TParcollet:
Thanks @30stomercury, I guess that integrating arpa will be harder as it operates at word level and not bpe ?

30stomercury:
Thanks @TParcollet , I'm investigating kenlm and see if we can have BPE in ARPA format.

30stomercury:
I see, I will take a look.

mravanelli:
@30stomercury, what's the status of this PR? I think the output of this work should be compared with #773, right?

30stomercury:
Hi @mravanelli and @TParcollet, this is the current status if this pr. 
- [ ] Support pure ctc decoding using `CTCPrefixScorer`. 
- [ ] Support NgramScorer with kenlm.
  - [ ] full scorer mode
  - [ ] partial scorer mode
- [ ] Compare ctcscorer + ngramscorer with `ctcdecode`.
- [ ] Run decoding experiments on: seq2seq + ngram lm, ctc + ngram lm. 
  - [ ] AIShell
  - [ ] LibriSpeech
- [ ] A Global Scorer for beam searcher.

Gastron:
Hey, should we implement the top-k hypothesis output (as discussed in PR #761) in this PR instead?

30stomercury:
@Gastron Sure, we can do it here. I will add it to the to-dos.

mravanelli:
Hi @30stomercury,
thank you very much for this PR! I like the idea of making beamsearch more modular by separating the "search" part from the "rescoring" one.
Let me share here some comments (already shared privately):
1. kenlm shouldn't be mandatory.  I suggest importing it one where needed and raise an error is not installed.
2. add the docstrings in scorer.py (make sure to include runnable examples)
3. It looks like `recipes/AISHELL-1/ASR/seq2seq/train.py` contains weird things like ` loss, loss_seq, loss_ctc = 0.0, 0.0, 0.0`,  `# np.save('token_list.npy', token_list)`. I guess this is still a work in progress, right?
4. ScorerBuilder => it is not clear to me how we manage the scoring with RNNLM or Transofrmer LM. I would have expected a type of scorer similar to NGramScorer (maybe called NeuralLMScorer) in scorer.py. Is that part missing?

mravanelli:
Let me also put on the table one idea for the `ScorerBuilder`. At this moment it is a bit hardcoded and not that scalable to other scorers possibly defined by the users.  My suggestion to make everything even more modular is the following:
1. Define all the scorers in the yaml file (e.g, `CTCPrefixScorer,` `NGramScorer`, `CoveragePenalty`).
2. Instantiate the builder in the yaml file as wel such that it takes a list of scorer objects:

```
scorer: !new:speechbrain.decoders.scorer.ScorerBuilder
[
CTCPrefixScorer,
NGramScorer,
CoveragePenalty
]
```

This is something quite similar to the way we manage the augmentation pipeline in some recipes (e.g, https://github.com/speechbrain/speechbrain/blob/develop/recipes/VoxCeleb/SpeakerRec/hparams/train_ecapa_tdnn.yaml#L122). The advantage is that everything looks more transparent and modular. Users, for instance, can define a new scorer without changing the ScorerBuilder.

What do you think?

30stomercury:
Hi @mravanelli , thank you for your suggestions.
1, 2. Currently I don't have a better way to manage importing part, I will finish the docstring first.
3. Yes, that part is still in progress, I would like to make librispeech converge first.
4. I think we can move neural LM part to the scorer. The reason why I didn't do it is because moving RNNLM or Transofrmer LM to scorer part will involve a lot of changes, I'm not sure if other developers are okay with that.

mravanelli:
As for point 4, this PR already requires us to change all the ASR recipes. So we can really take this opportunity to do the change that adds the NeuralLMScorer as well.

30stomercury:
Hi @mravanelli ,
I propose two ways to manage scorers in yaml. The second approach is similar to the way you suggested.
```
# Method1

scorer: !new:speechbrain.decoders.scorer.ScorerBuilder
   eos_index: !ref <eos_index>
   vocab_size: !ref <output_neurons>
   ctc_weight: !ref <ctc_decode_weight>
   #ngram_weight: !ref <ngram_weight>
   transformerlm_weight: !ref <lm_weight>
   coverage_weight: !ref <coverage_penalty>
   ctc_score_mode: !ref <ctc_score_mode>
   ctc_linear: !ref <ctc_lin>
   #lm_path: !ref <ngram_model>
   #tokenizer: !ref <save_folder>/tokenizer.ckpt
   transformerlm: !ref <lm_model>


# Method2

ctc_scorer: !new:speechbrain.decoders.scorer.CTCScorer
   eos_index: !ref <eos_index>
   ctc_fc: !ref <ctc_lin>

coverage_scorer: !new:speechbrain.decoders.scorer.CoverageScorer
   vocab_size: !ref <output_neurons>

transformerlm_scorer: !new:speechbrain.decoders.scorer.TransformerLMScorer
   language_model: !ref <lm_model>

scorer2: !new:speechbrain.decoders.scorer.ScorerBuilder2
   ctc_weight: !ref <ctc_decode_weight>
   transformerlm_weight: !ref <lm_weight>
   coverage_weight: !ref <coverage_penalty>
   full_scorers: [
        !ref <transformerlm_scorer>,
        !ref <coverage_scorer> ]
   partial_scorers: [ !ref <ctc_scorer> ]
```

30stomercury:
I will add a method to validate if users define scorers correctly. E.g., `coverage_scorer` should always by put into `full_scorers`. Or `ctc_weight` should be 1.0 if `ctc_scorer` is put into `full_scorers` list.

mravanelli:
Hi @30stomercury. I like method2 more because looks more modular and users can modify the scoring pipeline directly in yaml file. To make it even more modular, I would suggest using a list of tuples where the first element is the scorer object and the second one is the weight of the scorer:

```
scorer2: !new:speechbrain.decoders.scorer.ScorerBuilder2
   full_scorers: [
        !ref (<transformerlm_scorer>,  <lm_weight>),
        !ref (<coverage_weight>,  <coverage_penalty>), ]
   partial_scorers: [ !ref (<ctc_scorer>, ctc_weight ) ]
```

As an alternative, we can link the scorer object with the corresponding weight in a way similar to what done with the [pretrainer](https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml#L279):

```
scorer2: !new:speechbrain.decoders.scorer.ScorerBuilder2
   full_scorers:
      transformerlm_scorer: !ref <transformerlm_scorer>
      coverage_weight: !ref <coverage_weight>
   partial_scorers:
      ctc_scorer: !ref <ctc_scorer>
   weights:
       transformerlm_scorer: !ref<lm_weight>
       coverage_weight: !ref <coverage_penalty>
       ctc_scorer: !ref <ctc_weight>
```
The second one creates the link using dictionary. It is less compact, but probably more elegant and more in line with what done for the pretrainer. Maybe @Gastron and the others have some suggestions here. What do you think @30stomercury?

Also, what is the difference between full-scorers and partial one?

30stomercury:
Hi @mravanelli , the partial scorers score the topk tokens based on the logprobs after full scorers. Scoring all tokens in vocabulary is too expensive for some scorers, e.g. ngram scorer, therefore they only score on pruned tokens.

See candidates (pruned tokens) in the score() method. We score the logprob with full scorers first, then partial scorers. I will add some descriptions for that.
```
    def score(self, inp_tokens, memory, attn, log_probs, beam_size):
        new_memory = dict()
        # score full candidates
        for k, impl in self.full_scorers.items():
            score, new_memory[k] = impl.score(inp_tokens, memory[k], None, attn)
            log_probs += score * self.weights[k]

        # select candidates for partial scorers
        _, candidates = log_probs.topk(int(beam_size * self.scorer_beam_scale), dim=-1)

        # score patial candidates
        for k, impl in self.partial_scorers.items():
            score, new_memory[k] = impl.score(
                inp_tokens, memory[k], candidates, attn
            )
            log_probs += score * self.weights[k]

        return log_probs, new_memory

```

mravanelli:
Makes sense. What is your preference for the yaml part?

30stomercury:
I prefer using a list of tuples.

mravanelli:
I'm fine with both solutions. Let's hear a bit the others (e.g., @Gastron, @TParcollet ,..)

Gastron:
I'm ok with using `(scorer_func, weight)` -tuples too, but maybe I vote for separate arguments (Mirco's second suggestion). Tuples would force specifying weights for everything, but I think it could be nice to have a default 1.0 weight for all scorers that don't get a weight specified. Having one weights argument but two scorer lists (full/partial) could look a little weird, but referring by key would allow very clear error messages. Specifying weights separately would be more in line with many PyTorch weighted things (like weighted sampling or weighted losses), I think.

Full / partial scorer makes me think of rescoring full utterance outputs vs. scoring a partial output. Perhaps the vocab could be something like fulldist_scorer / topk_scorer.

30stomercury:
I have a default 0.0 weight if scorers are not specified. I can go for solution 2, with an error messages to check if weights and full/partial scorer lists are matched. 
`topk_scorer` can be confused, it looks like scorers are rescoring on topk hyps but our partial scorers are rescoring on topk tokens of all hyps.

30stomercury:
@mravanelli, @Gastron 

I have adapted those changes to yaml files that involve beamsearch part.

For the top-k hypothesis output:

I suggest we modify the train.py to obtain best hyps from topk hyps.
An example of outputting topk hyps,
```
topk_tokens, topk_lens, scores = self.hparams.test_search(x, wav_lens)
# The best hyps of each batch
p_tokens = [
        hyps[:length-1] for hyps, length
        in zip(topk_tokens[:, 0, :].tolist(),
               topk_lens[:, 0].tolist())
]

```
where the shape of `topk_tokens` is `[batch_size, topk, max_hyp_lengths]`, and `topk_length` is `[batch_size, topk]`.

mravanelli:
Hi @30stomercury, 
I'm fine with that. This is very transparent because the beasearcher always returns the topk hypothesis and then the user can select the best one. As for the code snippet, we are doing best hypothesis + conversion to a list jointly. Maybe we can disentangle it like this:

```
topk_tokens, topk_lens, scores = self.hparams.test_search(x, wav_lens)

# Select the best hypothesis
best_hyps = topk_tokens[:, 0, :]
best_len =  topk_lens[:, 0]

# Convert best hypothesis to list
best_hyps_lst = to_list(best_hyps, best_len)
```
As far as I remember we should have somewhere an auxiliary function that do the list conversion. If not, it might make sense to create it, right?

Any insights from @Gastron?

Gastron:
I also like the idea of always returning with the top-k dimension:
> topk_tokens, topk_lens, scores = self.hparams.test_search(x, wav_lens)
but this does changes in a _lot_ of files, including files in users' control (their recipes). It is not backwards compatible, so any users would need to change their code, as well. Adding a new interface (new method to call or similar) would remain backwards compatible, so that is worth considering.

mravanelli:
Thank you Aku,
the refactor is already backward incompatible because we have to change all
the yaml files using beamsearch. This is an opportunity to change
everything once and then release a new version of SpeechBrain.

Gastron:
Alright, yeah this does make sense. We need to just communicate this clearly.

TParcollet:
yes, I am a bit concerned about user-level recipes as @Gastron mentioned ...

30stomercury:
We can make use of `undo_padding`:

```
from speechbrain.utils.data_utils import undo_padding

# Select the best hypothesis
best_hyps, best_lens = topk_tokens[:, 0, :], topk_lens[:, 0]

# Convert best hypothesis to list
p_tokens = undo_padding(best_hyps, best_lens)
```

requirements.txt
```diff
@@ -1,4 +1,5 @@
 -r lint-requirements.txt
+https://github.com/kpu/kenlm/archive/master.zip
```
johntremblay, line 2: @30stomercury I tried to run by installing with pip install kenlm and got some issues with `cython` and `gcc`. I installed the package via a git clone and pip install -e . and didn't get any issue. Have you installed it via pypi?

recipes/AISHELL-1/ASR/seq2seq/hparams/train.yaml
```diff
@@ -203,22 +203,33 @@ pretrainer: !new:speechbrain.utils.parameter_transfer.Pretrainer
...
+   blank_index: !ref <blank_index>
+   ctc_fc: !ref <ctc_lin>
+
+scorer: !new:speechbrain.decoders.scorer.ScorerBuilder
+   full_scorers: [!ref <coverage_scorer>, !ref <ctc_scorer>]
+   weights:
+      coverage: !ref <coverage_penalty>
+      ctc: !ref <ctc_decode_weight>
+
 beam_search: !new:speechbrain.decoders.S2SRNNBeamSearcher
```
johntremblay, line 218: these will need to be refactored given they are now in the seq2seq.py

requirements.txt
30stomercury: Hi @johntremblay, please try 
```
pip install https://github.com/kpu/kenlm/archive/master.zip
```

recipes/AISHELL-1/ASR/seq2seq/hparams/train.yaml
30stomercury: Can you specify which part should be refactored? 
We use `__init__.py` to automatically import modules in `seq2seq.py`. Maybe we should add `score.py` as well.

johntremblay: I should've been more clear sorry about that:
```suggestion
beam_search: !new:speechbrain.decoders.seq2seq.S2SRNNBeamSearcher
```

requirements.txt
johntremblay: It worked thanks!

recipes/AISHELL-1/ASR/seq2seq/hparams/train.yaml
30stomercury: Hi, as mentioned, please refer to `__init__.py` in decoders. Thanks.

KacperKubara:
Hi, what's the status of this PR?

TParcollet:
Should be re-worked and merge before the summer. Big changes incoming :-)

speechbrain/pretrained/interfaces.py
```diff
@@ -256,14 +256,14 @@ class EndToEndSLU(Pretrained):
...
-    >>> tmpdir = getfixture("tmpdir")
-    >>> slu_model = EndToEndSLU.from_hparams(
-    ...     source="speechbrain/slu-timers-and-such-direct-librispeech-asr",
-    ...     savedir=tmpdir,
-    ... )
-    >>> slu_model.decode_file("samples/audio_samples/example6.wav")
-    "{'intent': 'SimpleMath', 'slots': {'number1': 37.67, 'number2': 75.7, 'op': ' minus '}}"
+    #>>> from speechbrain.pretrained import EndToEndSLU
+    #>>> tmpdir = getfixture("tmpdir")
+    #>>> slu_model = EndToEndSLU.from_hparams(
```
anautsch, line 261: Throws now a `TypeError: ('Invalid argument to class speechbrain.decoders.S2SRNNBeamSearcher', "__init__() got an unexpected keyword argument 'coverage_penalty'")`

speechbrain/pretrained/interfaces.py
```diff
@@ -374,14 +374,14 @@ class EncoderDecoderASR(Pretrained):
...
-    >>> tmpdir = getfixture("tmpdir")
-    >>> asr_model = EncoderDecoderASR.from_hparams(
-    ...     source="speechbrain/asr-crdnn-rnnlm-librispeech",
-    ...     savedir=tmpdir,
-    ... )
-    >>> asr_model.transcribe_file("samples/audio_samples/example2.flac")
-    "MY FATHER HAS REVEALED THE CULPRIT'S NAME"
+    #>>> from speechbrain.pretrained import EncoderDecoderASR
+    #>>> tmpdir = getfixture("tmpdir")
+    #>>> asr_model = EncoderDecoderASR.from_hparams(
```
anautsch, line 379: Throws `ImportError: There is no such class as speechbrain.decoders.S2SRNNBeamSearchLM`

anautsch: edits to the pretained interface & the yaml file get it back on track (other edits help to figure out what needs to be done). A revision of the pretrained interface is necessary as well as pretrained yaml files.

anautsch:
lgtm

Getting back to this PR. What needs to be done is to update HF yaml files. It's really just the HF yaml files that need to be updated, the scorer is not affecting trained parameters. This is not done yet, so as long as this PR sits on unstable, HF interfaces will not work with the `unstable` branch.

This PR's tests are working since docstring examples in the pretrained interface were commented out.
@mravanelli @TParcollet  They need to be attended to when merging unstable to develop and working closer towards v0.6 release. For now, I'd like to move on here and wrap this up, hence `unstable` branch after all ;-)
This part needs attention when wrapping up v0.6:
```python
class EndToEndSLU(Pretrained):
    """A end-to-end SLU model.
    The class can be used either to run only the encoder (encode()) to extract
    features or to run the entire model (decode()) to map the speech to its semantics.
    Example
    -------
    #>>> from speechbrain.pretrained import EndToEndSLU
    #>>> tmpdir = getfixture("tmpdir")
    #>>> slu_model = EndToEndSLU.from_hparams(
    #...     source="speechbrain/slu-timers-and-such-direct-librispeech-asr",
    #...     savedir=tmpdir,
    #... )
    #>>> slu_model.decode_file("samples/audio_samples/example6.wav")
    #"{'intent': 'SimpleMath', 'slots': {'number1': 37.67, 'number2': 75.7, 'op': ' minus '}}"
    """
```

To keep things clear, I prefer to merge this PR now and then work directly on the `unstable` branch, so this PR serves as a reference of what works while the other parts can become playground for a few months. During this time, packages to interact with HF will also change (to start somewhere, let's start here).

anautsch:
As mentioned in #1560, some tests need to be skipped for these tests here "to pass", so we can merge from latest `develop` branch into this one.

**CRITICAL**

DO NOT MERGE this PR until the skipped tests are put back in and tested accordingly.
* speechbrain/alignment/ctc_segmentation.py::speechbrain.alignment.ctc_segmentation.CTCSegmentation
* speechbrain/pretrained/interfaces.py::speechbrain.pretrained.interfaces.EncoderDecoderASR
* speechbrain/pretrained/interfaces.py::speechbrain.pretrained.interfaces.EndToEndSLU
* tests/unittests/test_ctc_segmentation.py

Rationale: YAML files on HuggingFace need to be refactored in accordance with this PR's refactorings.
This will break legacy; thus, v0.6. With v0.6, however, further legacy breaks are intended—to be resolved on `unstable-v0.6` branch.

speechbrain/pretrained/interfaces.py
```diff
@@ -592,12 +676,133 @@ def transcribe_batch(self, wavs, wav_lens):
         with torch.no_grad():
             wav_lens = wav_lens.to(self.device)
             encoder_out = self.encode_batch(wavs, wav_lens)
-            predicted_tokens, scores = self.mods.decoder(encoder_out, wav_lens)
-            predicted_words = [
-                self.tokenizer.decode_ids(token_seq)
-                for token_seq in predicted_tokens
-            ]
-        return predicted_words, predicted_tokens
+            if detail_level == InferenceDetailLevel.TOP1_HYP:
```
TParcollet, line 682: I like the idea, but this would be something to apply to all interfaces :/ And maybe we should discuss the format of it as well, because it's making the code of the interface MUCH longer.

anautsch: yeah, well, testing here ^^"
interface needs refactoring...

anautsch: put it in an extra file - available as Easter Egg - if one wants & desires to extend (it's in another "domain")

Adel-Moumen:
Update of the current refactoring. 

I think I have nearly finished working on improving the readability of the seq2seq.py file. 

The idea was to break the forward function of the beam search into smaller functions, changing variables/function names, dropping unused args etc... I also moved some utils function into an util.py file. 

This is the current version of the forward function:
Refactored forward

I choose to not use a lot of data handler classes for multiple reasons. The first one is that PyTorch JIT doesn't support `dataclass` object (or at least in a lot of versions that we support). Furthermore, I found that it was not adding clarity. Even though it reduces the number of parameters it does not indicates just by reading the forward function which args are necessary and which are updated during the beam search. So, I decided to only implement one class that will be used later for adding new features (e.g, scoring on char/words while the model is training on BPE). 

During the refactoring process, I made sure to not break anything. I did retrain a CRDNN model on LibriSpeech to empirically validate the beam search and got the following results:
> speechbrain.utils.train_logger - Epoch loaded: 15 - test loss: 1.17, test CER: 1.85, test WER: 3.77
> speechbrain.utils.train_logger - Epoch loaded: 15 - test loss: 1.43, test CER: 5.67, test WER: 10.57

which seems to demonstrate that everything works. 

If you do think that the beam search is still hard to read, please let me know. :smiley:

Adel-Moumen:
Fix yamls with Beam Search refactoring:

- [x] AISHELL
- [x] CommonVoice
- [x] LibriSpeech
- [ ] SwirtchBoard
- [ ] TIMIT
- [ ] VoiceBank

lucadellalib (commented):
Looks great, thank you! I left a couple of minor comments directly in the corresponding files. Some general comments:

- it's not completely clear to me the difference between full and partial scorer. Maybe we could improve the docstring to explain the meaning of "pruned tokens"?
```
candidates : torch.Tensor
            (batch_size x beam_size, scorer_beam_size). The pruned tokens for
            scoring. If None, scorers will score on full vocabulary set.
```
- If I understand correctly every scorer class can be used both as a full scorer and a partial scorer, right? Otherwise we should add some documentation or change the name such that it is clear which can be used as full and/or partial scorers and which cannot.

- I think the `ScorerBuilder` abstraction is slightly misleading. It is an independent class that does not inherit from `BaseScorerInterface`, but in the end it is a scorer that combines several sub-scorers. What about having a `CombinedScorer` or `MultiScorer` wrapper that inherits from `BaseScorerInterface` and implements the same functionality of `ScorerBuilder`? This way we could have beam search working directly with an instance of `BaseScorerInterface` and let the user choose if using a simple scorer or a combined scorer or even a custom scorer.

Adel-Moumen:
> If I understand correctly every scorer class can be used both as a full scorer and a partial scorer, right? Otherwise we should add some documentation or change the name such that it is clear which can be used as full and/or partial scorers and which cannot.

Yes, every scorers can be used both as a full scorer or as a partial scorer.


> I think the ScorerBuilder abstraction is slightly misleading. It is an independent class that does not inherit from BaseScorerInterface, but in the end it is a scorer that combines several sub-scorers. What about having a CombinedScorer or MultiScorer wrapper that inherits from BaseScorerInterface and implements the same functionality of ScorerBuilder? This way we could have beam search working directly with an instance of BaseScorerInterface and let the user choose if using a simple scorer or a combined scorer or even a custom scorer.

I think I prefer to have the `ScorerBuilder` abstraction that does not inherit from the `BaseScorerInterface`, as we did in this PR. I believe it is easier for users to understand that they need to use this interface to work with scorers, especially if they want to use them as partial or full scorers. Furthermore, I am planning to integrate new scoring ways (see: #1986) which involves using `ScorerBuilder`. Overall, I am not in favor of having multiple possibilities to accomplish the same thing. People could get confused and/or miss informations. 

What are your thoughts on this, @mravanelli and @lucadellalib?

speechbrain/pretrained/fetching.py
```diff
@@ -99,11 +97,8 @@ def fetch(
         The model revision corresponding to the HuggingFace Hub model revision.
         This is particularly useful if you wish to pin your code to a particular
         version of a model hosted at HuggingFace.
-    cache_dir: str or Path (default: None)
-        Location of HuggingFace cache for storing pre-trained models, to which symlinks are created.
-    silent_local_fetch: bool (default: False)
-        Surpress logging messages (quiet mode).
```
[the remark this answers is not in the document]
Adel-Moumen: This line is removed

speechbrain/decoders/utils.py
```diff
@@ -0,0 +1,147 @@
...
+    Example
+    -------
+    >>> tensor = torch.Tensor([[1,2,3], [4,5,6]])
+    >>> cond = torch.BoolTensor([[True, True, False], [True, False, False]])
+    >>> mask_by_condition(tensor, cond, 0)
+    tensor([[1., 2., 0.],
+            [4., 0., 0.]])
+    """
+    tensor = torch.where(
+        cond, tensor, torch.Tensor([fill_value]).to(tensor.device)
```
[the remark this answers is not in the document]
Adel-Moumen: Indeed, and your version is faster!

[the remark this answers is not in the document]
Adel-Moumen: 27.1 µs ± 2.68 µs per loop vs 34.9 µs ± 631 ns per loop

lucadellalib:
I mean a design pattern similar to https://pytorch.org/docs/stable/distributions.html#torch.distributions.transforms.ComposeTransform, i.e. a scorer that allows to combine multiple subscorers according to some criteria (in this case full scorers first followed by partial scorers). Also, if the class is called `ScorerBuilder`, I would expect an object with a `__call__` or `build` method that returns a scorer instance given some configuration options.

Adel-Moumen:
I agree that the naming could be misleading. What about `ScorerInterface` then ? Similar to `TransformerInterface` (see: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/transformer/Transformer.py#L21) and it follows the naming of `BaseScorerInterface`.

30stomercury:
> What about having a CombinedScorer or MultiScorer wrapper that inherits from BaseScorerInterface and implements the same functionality of ScorerBuilder? This way we could have beam search working directly with an instance of BaseScorerInterface and let the user choose if using a simple scorer or a combined scorer or even a custom scorer.

I prefer to keep it simpler. As @Adel-Moumen mentioned, the `BaseScorerInterface` is an interface for users to define their own scorers. The `ScorerBuilder`  is more like an instance that collects a chain of scorers. 

> What about ScorerInterface then ? 

`ScorerInterface` seems like another interface, but we are not using it for inheritance. Could be misleading.

> I mean a design pattern similar to https://pytorch.org/docs/stable/distributions.html#torch.distributions.transforms.ComposeTransform.

Agree. Or similar to `torch.nn.Sequential`, although the order does not matter if we follow the criteria.

recipes/AISHELL-1/ASR/transformer/train.py
```diff
@@ -56,17 +60,26 @@ def compute_forward(self, batch, stage):
...
+
+            # Decide searcher for inference: valid or test search
+            search = getattr(self.hparams, f"{stage.name}_search".lower())
+            topk_tokens, topk_lens, _, _ = search(enc_out.detach(), wav_lens)
+
+            # Select the best hypothesis
+            best_hyps, best_lens = topk_tokens[:, 0, :], topk_lens[:, 0]
+
+            # Convert best hypothesis to list
+            hyps = undo_padding(best_hyps, best_lens)
```
TParcollet, line 82: I don't really understand this change. It makes the reading so much more complicated. Now we have indices and stuffs in the middle, a  getattr function etc.

recipes/CommonVoice/ASR/seq2seq/train_with_wav2vec.py
```diff
@@ -63,30 +63,30 @@ def compute_forward(self, batch, stage):
...
-                return p_seq, wav_lens
         else:
-            p_tokens, scores = self.hparams.beam_searcher(x, wav_lens)
-            return p_seq, wav_lens, p_tokens
+            topk_tokens, topk_lens, _, _ = self.hparams.beam_searcher(
+                x, wav_lens
+            )
+
+            # Select the best hypothesis
+            best_hyps, best_lens = topk_tokens[:, 0, :], topk_lens[:, 0]
```
TParcollet, line 79: Why do we keep returning the best hypotheses like this as a standard? It makes the code harder to read for an edge use case.

recipes/CommonVoice/ASR/transformer/train.py
```diff
@@ -73,15 +73,34 @@ def compute_forward(self, batch, stage):
...
+        elif is_test_search:
+            topk_tokens, topk_lens, _, _ = self.hparams.test_search(
+                enc_out.detach(), wav_lens
+            )
+
+            # Select the best hypothesis
+            best_hyps, best_lens = topk_tokens[:, 0, :], topk_lens[:, 0]
+
+            # Convert best hypothesis to list
+            hyps = undo_padding(best_hyps, best_lens)
```
TParcollet, line 103: I mean, we go from 8 lines to 3 times this number. This looks really sub-optimal to me.

recipes/CommonVoice/ASR/transformer/train_with_whisper.py
```diff
@@ -48,10 +48,16 @@ def compute_forward(self, batch, stage):
         enc_out, logits, _ = self.modules.whisper(wavs, bos_tokens)
 
         hyps = None
-        if stage == sb.Stage.VALID:
-            hyps, _ = self.hparams.valid_greedy_searcher(enc_out, wav_lens)
-        elif stage == sb.Stage.TEST:
-            hyps, _ = self.hparams.valid_greedy_searcher(enc_out, wav_lens)
+        if stage == sb.Stage.VALID or stage == sb.Stage.TEST:
+            # Decide searcher for inference: valid or test search
+            search = getattr(self.hparams, f"{stage.name}_search".lower())
```
TParcollet, line 53: this getattrb with string parsing is AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAH in a recipe.

recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml
```diff
@@ -44,7 +44,7 @@ test_csv:
 # Training parameters
 number_of_epochs: 15
 number_of_ctc_epochs: 5
-batch_size: 8
+batch_size: 4
```
TParcollet, line 47: Where the heck do these changes come from. Changing this will make the corresponding drive model change. Someone tested those params?

recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml
```diff
@@ -105,13 +105,12 @@ max_decode_ratio: 1.0
 valid_beam_size: 80
 test_beam_size: 80
 eos_threshold: 1.5
-using_max_attn_shift: True
-max_attn_shift: 240
-lm_weight: 0.50
-ctc_weight_decode: 0.0
-coverage_penalty: 1.5
+using_max_attn_shift: False
```
TParcollet, line 108: Same story here.

recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_5000.yaml
```diff
@@ -103,16 +103,17 @@ eos_index: 2
...
+test_beam_size: 30
+using_eos_threshold: True
 eos_threshold: 1.5
-using_max_attn_shift: True
-max_attn_shift: 300
-lm_weight: 0.80
-ctc_weight_decode: 0.40
-ctc_window_size: 200
-coverage_penalty: 1.5
+using_max_attn_shift: False
```
TParcollet, line 109: Same.

recipes/SLURP/direct/train_with_wav2vec2.py
```diff
@@ -55,10 +56,17 @@ def compute_forward(self, batch, stage):
...
+            topk_tokens, topk_lens, _, _ = self.hparams.beam_searcher(
+                wav2vec2_out.detach(), wav_lens
             )
-            return p_seq, wav_lens, p_tokens
+
+            # Select the best hypothesis
+            best_hyps, best_lens = topk_tokens[:, 0, :], topk_lens[:, 0]
+
+            # Convert best hypothesis to list
+            hyps = undo_padding(best_hyps, best_lens)
```
TParcollet, line 67: this undo_padding also is mysterious. I see it everywhere, but I don't like that we change something that most users already are very used to .

recipes/TIMIT/ASR/seq2seq/train.py
```diff
@@ -62,22 +62,24 @@ def compute_forward(self, batch, stage):
         logits = self.modules.seq_lin(h)
         p_seq = self.hparams.log_softmax(logits)
 
-        if stage == sb.Stage.VALID:
-            hyps, scores = self.hparams.greedy_searcher(x, wav_lens)
-            return p_ctc, p_seq, wav_lens, hyps
+        hyps = None
+        if stage != sb.Stage.TRAIN:
+            # Decide searcher for inference: valid or test search
+            searcher = getattr(self.hparams, f"{stage.name}_searcher".lower())
```
TParcollet, line 68: :'(

recipes/TIMIT/ASR/seq2seq/train_with_wav2vec2.py
```diff
@@ -51,22 +52,24 @@ def compute_forward(self, batch, stage):
         logits = self.modules.seq_lin(h)
         p_seq = self.hparams.log_softmax(logits)
 
-        if stage == sb.Stage.VALID:
-            hyps, scores = self.hparams.greedy_searcher(x, wav_lens)
-            return p_ctc, p_seq, wav_lens, hyps
+        hyps = None
+        if stage != sb.Stage.TRAIN:
+            # Decide searcher for inference: valid or test search
+            searcher = getattr(self.hparams, f"{stage.name}_searcher".lower())
```
TParcollet, line 58: :'(

speechbrain/decoders/ctc.py
```diff
@@ -113,89 +103,82 @@ def forward_step(self, g, state, candidates=None, attn=None):
...
             ).view(-1)
             x_inflate = torch.index_select(
                 self.x.view(2, -1, self.batch_size * self.vocab_size),
                 2,
                 scoring_index,
-            ).view(2, -1, self.batch_size * self.beam_size, self.num_candidates)
+            ).view(2, -1, n_bh, self.num_candidates)
         # for full search
         else:
             scoring_table = None
```
TParcollet, line 140: We really need to add some comments here, this is some serious black magic.

speechbrain/decoders/scorer.py
```diff
@@ -0,0 +1,850 @@
...
+"""
+
+import torch
+import numpy as np
+import speechbrain as sb
+from speechbrain.decoders.ctc import CTCPrefixScore
+
+
+class BaseScorerInterface:
+    """A scorer abstraction to be inherited by other
```
TParcollet, line 17: What is it usefull for? Add a description of what is the role of such a "scorer" what is the definition of SpeechBrain of a "scorer"?

speechbrain/decoders/scorer.py
```diff
@@ -0,0 +1,850 @@
...
+        - speechbrain.decoders.scorer.CTCPrefixScorer
+        - speechbrain.decoders.scorer.RNNLMScorer
+        - speechbrain.decoders.scorer.TransformerLMScorer
+        - speechbrain.decoders.scorer.KenLMScorer
+        - speechbrain.decoders.scorer.CoverageScorer
+        - speechbrain.decoders.scorer.LengthScorer
+    """
+
+    def score(self, inp_tokens, memory, candidates, attn):
+        """This method scores tokens in vocabulary.
```
TParcollet, line 30: scores w.r.t what? What is a score?

speechbrain/decoders/scorer.py
```diff
@@ -0,0 +1,850 @@
...
+        return (
+            torch.tensor(
+                [1.0], device=inp_tokens.device, dtype=inp_tokens.dtype
+            ).expand(inp_tokens.size(0), self.vocab_size),
+            None,
+        )
+
+
+class ScorerBuilder:
+    """ Builds scorer instance for beamsearch.
```
TParcollet, line 621: More docstring, what is this usefull for?

speechbrain/pretrained/interfaces_topk.py
```diff
@@ -0,0 +1,533 @@
+"""Example extension of interfaces; it's a show-casing Easter Egg.
```
TParcollet, line 1: This has to be put in another PR and we have to check if  we want it or not. Especially since it's impacting most recipes. It should be a edge case, not a standard. Or if it's standard, we should make the use of it in other recipe look better.

TParcollet (changes requested):
I really don't like the new formalism due to topk. I think this should be put into another PR and discussed. Also, some recipes are actually changing (hyperparams like attention shift). This cannot be pushed as is as we don't know what it means in terms of training results.

Adel-Moumen:
Thanks @TParcollet. I will address all of your requests tomorrow! :-)

lucadellalib:
Some results on LibriSpeech test-clean using recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml and https://www.dropbox.com/sh/vdyw6uuiy4z63mp/AABdFqExB9llDGIQRCzsd1U7a?dl=0:

**ctc-prefix-beamsearch branch**

batch_size=1, beam_size=2, LM + CTC
[**1:33:09**<00:00,  2.13s/it]
speechbrain.utils.train_logger - Epoch loaded: 100 - test loss: 1.74, test ACC: 9.75e-01, **test WER: 2.58**

batch_size=1, beam_size=2, LM
[**31:11**<00:00,  1.40it/s]
speechbrain.utils.train_logger - Epoch loaded: 100 - test loss: 1.32, test ACC: 9.75e-01, **test WER: 3.37**

batch_size=32, beam_size=2, LM
[**06:25**<00:00,  4.70s/it]
speechbrain.utils.train_logger - Epoch loaded: 100 - test loss: 3.35e-02, test ACC: 9.75e-01, **test WER: 3.50**


**develop branch**

batch_size=1, beam_size=2, LM + CTC
[**1:32:55**<00:00,  2.13s/it]
speechbrain.utils.train_logger - Epoch loaded: 100 - test loss: 1.32, test ACC: 9.75e-01, **test WER: 2.58**

recipes/CommonVoice/ASR/transformer/train_with_whisper.py
Adel-Moumen: fixed

recipes/CommonVoice/ASR/transformer/train.py
Adel-Moumen: fixed

recipes/CommonVoice/ASR/seq2seq/train_with_wav2vec.py
Adel-Moumen: fixed

recipes/AISHELL-1/ASR/transformer/train.py
Adel-Moumen: fixed

recipes/TIMIT/ASR/seq2seq/train.py
Adel-Moumen: :')

recipes/TIMIT/ASR/seq2seq/train_with_wav2vec2.py
Adel-Moumen: :')

recipes/SLURP/direct/train_with_wav2vec2.py
Adel-Moumen: fixed

recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml
Adel-Moumen: solved

recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_5000.yaml
Adel-Moumen: solved

speechbrain/pretrained/interfaces_topk.py
Adel-Moumen: I dropped the file. Will open a PR after the big release.

30stomercury:
> Hi @30stomercury, I'm fine with that. This is very transparent because the beasearcher always returns the topk hypothesis and then the user can select the best one. As for the code snippet, we are doing best hypothesis + conversion to a list jointly. Maybe we can disentangle it like this:
> 
> ```
> topk_tokens, topk_lens, scores = self.hparams.test_search(x, wav_lens)
> 
> # Select the best hypothesis
> best_hyps = topk_tokens[:, 0, :]
> best_len =  topk_lens[:, 0]
> 
> # Convert best hypothesis to list
> best_hyps_lst = to_list(best_hyps, best_len)
> ```
> 
> As far as I remember we should have somewhere an auxiliary function that do the list conversion. If not, it might make sense to create it, right?
> 
> Any insights from @Gastron?

Hi @TParcollet and @Adel-Moumen, we actually have discussed whether we should return topk before. If we only return the best hyp, the topk argument is useless:
https://github.com/speechbrain/speechbrain/blob/a9be78c0f646e86cf189707cd810536c338407aa/speechbrain/decoders/seq2seq.py#L217

The topk is never used in the original searcher. By returning with the top-k dimension users have more flexibility.
I would suggest changing everything once as the pr is already backward incompatible. Otherwise we may want to remove the topk option, which is less flexible.

Adel-Moumen:
> > Hi @30stomercury, I'm fine with that. This is very transparent because the beasearcher always returns the topk hypothesis and then the user can select the best one. As for the code snippet, we are doing best hypothesis + conversion to a list jointly. Maybe we can disentangle it like this:
> > ```
> > topk_tokens, topk_lens, scores = self.hparams.test_search(x, wav_lens)
> > 
> > # Select the best hypothesis
> > best_hyps = topk_tokens[:, 0, :]
> > best_len =  topk_lens[:, 0]
> > 
> > # Convert best hypothesis to list
> > best_hyps_lst = to_list(best_hyps, best_len)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As far as I remember we should have somewhere an auxiliary function that do the list conversion. If not, it might make sense to create it, right?
> > Any insights from @Gastron?
> 
> Hi @TParcollet and @Adel-Moumen, we actually have discussed whether we should return topk before. If we only return the best hyp, the topk argument is useless:
> 
> https://github.com/speechbrain/speechbrain/blob/a9be78c0f646e86cf189707cd810536c338407aa/speechbrain/decoders/seq2seq.py#L217
> 
> The topk was never used in the original searcher. By returning with the top-k dimension users have more flexibility. I would suggest changing everything once as the pr is already backward incompatible. Otherwise we may want to remove the topk option, which is less flexible.

I tend to agree with @TParcollet on the fact that it is harder to read when you need to deal with the top-k output (lots of code). What I propose if we really wants to support top-k  would be to return by default the one-best, and only if the user specify it (through a param e.g. `return_topk`), return the top-k. What do you think @TParcollet ?

30stomercury:
That's another part we have refactored. There was a `return_topk` flag before. Not really neat for me as the returns of the searcher, w/ or w/o topk are in different shapes.

Adel-Moumen:
> That's another part we have refactored. There was a `return_topk` flag before. Not really neat for me as the returns of the searcher, w/ or w/o topk are in different shapes.

I don't think it is an issue as the user is expecting to get more hyps than the one-best. Furthermore, I think that there are not many users that needs the top-k output. In 99% of the time, people are just using the one-best. Why should we add ~8 lines per recipe for 1%?

30stomercury:
Alright, at user-level it makes sense, hyps other than the best one are usually not useful.

speechbrain/decoders/scorer.py
Adel-Moumen: done

speechbrain/decoders/scorer.py
Adel-Moumen: done

speechbrain/decoders/scorer.py
Adel-Moumen: done

speechbrain/decoders/ctc.py
Adel-Moumen: :s

Adel-Moumen: welcome to beam search with batching :-)

mravanelli:
Thank you all for contributing to this huge PR. As for the top-k, I suggest to proceed as @Adel-Moumen proposed. By default, we can have the best transcription and, if user request it, we can have the top-k.

mravanelli:
Also, @Adel-Moumen, do the numbers shared by @lucadellalib in a previous message makes sense and match your ones?

mravanelli:
Hi @Adel-Moumen,
thank you again for leading this crucial PR needed for the new major version of speechbrain. I did a first code inspection. The code looks clean and well structured. I have the following comments:

**decoders/scorer.py** 

1. In the docstring of  `BaseScorerInterface`,  I suggest better explaining the difference between full and partial rescorer (maybe adding a simple example and a use case for the partial one). I suggest adding a working docstring example here as well.

2. In `CTCScorer`, `RNNLMScorer`, `TransformerLMScorer`, `KenLMScorer`, `CoverageScorer`, and `LengthScorer` I think we need a better docstring for each method. We need to describe better the functionality implemented and explain the expected input/output types and shapes. This is not immediate to understand for users and might improve clarity. For the scoring method, I would add a working example.
In the `LengthScorer,` please add the docstring example as for the other classes.

**decoders/seq2seq.py**

4. I have similar comments regarding the docstring of methods. In some cases, the methods have a description of the input/outputs, in some cases, it is missing. I would suggest improving this aspect and adding docstring examples where possible. This will further improve clarity.

5. It looks like the attention shift is still included in the beamsearch part. I'm wondering if this part can be an external scorer instead. Any thoughts @Adel-Moumen @30stomercury ?

**General**

6. As mentioned by @TParcollet , I noticed that some search/scoring hyperparameters are changing wrt to the previous code. For instance, take a look at recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml. Now we have different `max_attn_shift`, `lm_weight`, `coverage_penalty`. Is there a reason for that?

7. My suggestion is to expose the most important search/scoring hyperparameters in the yaml file. At least for some popular recipes (e.g., LibriSpeech, CommonVoice, Switchboard, TIMIT), we can optimize these hparams with a hyperparameter optimization tool (e.g., Orion).  This can be done quite efficiently as it does not require retraining the model but just running decoding.

Adel-Moumen:
> 8\. It looks like the attention shift is still included in the beamsearch part. I'm wondering if this part can be an external scorer instead. Any thoughts @Adel-Moumen @30stomercury ?

Hello Mirco, 

Thanks for the great feedbacks. Regarding your comments:
1. Will do.
2. Will do.
3. Will do.
4. Will do.
5. I don't think we should add the attention shift as a Scorer. In my opinion, it can't be consider as a Scorer. To me, it doesn't make sense to have an assigned weight for the potential attention shift scorer as your are masking to -inf the frames that are not superior to tmax, which is not a score but rather a masking step. It also requires to have access to log_probs which is not the case currently. 
6. I will change everything. Sorry it was a mistake. 
7. "expose the most important" wydm ?

mravanelli:
Thank you @Adel-Moumen  for addressing my previous comments! After running the LibriSpeech recipes (in debug modality), I encountered some additional issues. Here is a list of the problems I encountered:

1. In `recipes/LibriSpeech/ASR/transformer`:
- Files `hparams/branchformer_large.yaml` and `hparams/conformer_large.yaml` are not converted to the new version.

- When running `python train_with_whisper.py hparams/train_hf_whisper.yaml --data_folder=$SLURM_TMPDIR/librispeech/LibriSpeech --debug`, I got the following error: "RuntimeError: Expected size for the first two dimensions of the batch2 tensor to be: [12, 80], but got: [12, 201]." I suspect this issue might be related to the whisper fix we merged in the development branch. Do you think this problem will be resolved after merging the dev branch?

2. In `recipes/LibriSpeech/ASR/CTC`:
- When executing `python train_with_wav2vec.py hparams/train_sb_wav2vec.yaml`, I encountered the error: "ValueError: 'wav2vec2_hub' is a !PLACEHOLDER and must be replaced." I believe we need to specify the path of the HF hub.
- Running `python train_with_wav2vec.py hparams/downsampled/*.yaml --data_folder=$SLURM_TMPDIR/librispeech/LibriSpeech  --debug` resulted in the error: "ValueError: 'ngram_lm_path' is a !PLACEHOLDER and must be replaced." I believe we need to provide this path to this ln and ensure it is automatically downloaded from Dropbox. Do you have it? @salah-zaiem  might have it.
- The same whisper problem as before occurred when running `python train_with_whisper.py hparams/train_hf_whisper_encoder.yaml --data_folder=$SLURM_TMPDIR/librispeech/LibriSpeech --debug`.

3. In `recipes/LibriSpeech/ASR/seq2seq`:
- Executing `python train.py hparams/train_*.yaml --data_folder=$SLURM_TMPDIR/librispeech/LibriSpeech --debug` produced the error: "ValueError: too many values to unpack (expected 2)" at the line "p_tokens, _ = self.hparams.valid_search(x, wav_lens)".

4. It's possible that other recipes may have similar issues. To minimize such problems, we have created recipe tests located in `tests/recipes` (see the README). Please run the recipe tests for all the modified recipes to ensure everything is running correctly.

5. Additionally, I noticed that the docstring tests are failing, but I'm unsure why. The only modification I made was to the `weight_decay` in some YAML files, which should not cause any docstring tests to fail.

Adel-Moumen:
We run the recipe tests and  fix recipes whose tests are failing.
To check:

- [ ] AISHELL-1
- [ ] Aishell1Mix
- [ ] AMI
- [ ] BinauralWSJ0Mix
- [ ] CommonLanguage
- [x] CommonVoice
- [ ] DVoice
- [ ] ESC50
- [ ] Fisher-Callhome-Spanish
- [ ] fluent-speech-commands
- [ ] Google-speech-commands
- [ ] IEMOCAP
- [ ] IWSLT22_lowresource
- [ ] KsponSpeech
- [ ] LibriMix
- [ ] LibriParty
- [x] LibriSpeech
- [ ] LibriTTS
- [ ] LJSpeech
- [ ] MEDIA
- [ ] REAL-M
- [ ] RescueSpeech
- [ ] SLURP
- [ ] Switchboard
- [ ] timers-and-such
- [ ] TIMIT
- [ ] UrbanSound8k
- [ ] Voicebank
- [ ] VoxCeleb
- [ ] VoxLingua107
- [ ] WHAMandWHAMR
- [ ] WSJ0Mix
- [ ] ZaionEmotionDataset

mravanelli:
Hi @Adel-Moumen, I merged the latest development branch and ran again the recipe tests (including the recently added full-inference tests).
Beyond the known issues due to HF (which we will be fixed after updating the HF repos to the new version of speechbrain), I found the following issues:

1. **PIQ recipe**:
```
(153/186) Running test for ESC50_row_7...
        ... 9.57s
        ERROR: Error in ESC50_row_7 (recipes/ESC50/interpret/hparams/piq.yaml). Check tests/tmp/ESC50_row_7/stderr.txt and tests/tmp/ESC50_row_7/stdout.txt for more info.
```
        
```
  File "/workspace/speechbrain_adelPR/recipes/ESC50/interpret/train_piq.py", line 751, in <module>
    Interpreter_brain.checkpointer.recover_if_possible(
TypeError: recover_if_possible() got an unexpected keyword argument 'device'

```
2.**Conformer transducer recipes**

```
(159/186)
        ERROR: Error in LibriSpeech_row_7 (recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml). Check tests/tmp/LibriSpeech_row_7/stderr.txt and tests/tmp/LibriSpeech_row_7/stdout.txt for more info.
        ...checking files & performance...
        ERROR: The recipe LibriSpeech_row_7 does not contain the expected file tests/tmp/LibriSpeech_row_7/wer_ASR_train.txt
(160/186) Running test for LibriSpeech_row_8...
        ... 30.45s
        ERROR: Error in LibriSpeech_row_8 (recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml). Check tests/tmp/LibriSpeech_row_8/stderr.txt and tests/tmp/LibriSpeech_row_8/stdout.txt for more info.
        ...checking files & performance...
        ERROR: The recipe LibriSpeech_row_8 does not contain the expected file tests/tmp/LibriSpeech_row_8/wer_ASR_train.txt
```
```
  File "/workspace/speechbrain_adelPR/recipes/LibriSpeech/ASR/transducer/train.py", line 520, in <module>
    asr_brain.evaluate(
  File "/workspace/speechbrain_adelPR/speechbrain/core.py", line 1430, in evaluate
    self.on_evaluate_start(max_key=max_key, min_key=min_key)
  File "/workspace/speechbrain_adelPR/recipes/LibriSpeech/ASR/transducer/train.py", line 300, in on_evaluate_start
    ckpt = sb.utils.checkpoints.average_checkpoints(
TypeError: average_checkpoints() got an unexpected keyword argument 'device'
```


3.**ASR Template**

```
(168/186) Running test for LibriSpeech_row_16...
        ... 12.81s
        ERROR: Error in LibriSpeech_row_16 (templates/speech_recognition/ASR/train.yaml). Check tests/tmp/LibriSpeech_row_16/stderr.txt and tests/tmp/LibriSpeech_row_16/stdout.txt for more info.
```
        
```
  File "/workspace/speechbrain_adelPR/templates/speech_recognition/ASR/train.py", line 454, in <module>
    asr_brain.fit(
  File "/workspace/speechbrain_adelPR/speechbrain/core.py", line 1317, in fit
  File "/workspace/speechbrain_adelPR/speechbrain/core.py", line 1218, in _fit_valid
    loss = self.evaluate_batch(batch, stage=Stage.VALID)
  File "/workspace/speechbrain_adelPR/speechbrain/core.py", line 1122, in evaluate_batch
    out = self.compute_forward(batch, stage=stage)
  File "/workspace/speechbrain_adelPR/templates/speech_recognition/ASR/train.py", line 103, in compute_forward
    if stage == sb.Stage.Valid:
  File "/opt/conda/envs/myenv/lib/python3.9/enum.py", line 429, in __getattr__
    raise AttributeError(name) from None
AttributeError: Valid
```

This is something related to this PR because the same tests on the dev branch run smoothly. Any idea?

Adel-Moumen:
Hey @mravanelli, I fixed everything. It was mainly related to the PR in unstable that is changing how we are loading ckpts.

mravanelli:
All tests are passing now! Thank you @Adel-Moumen and @30stomercury for this PR. It took a while, but this is an important step toward SpeechBrain 1.0!

PR #754, opened 2021-05-13
Adding w2v2 for CommonVoice IT / EN / RW / FR recipes

TParcollet:
Here is the English latest results with wav2vec2. For some reasons CTC with greedy decoding works better. We will certainly be able to reduce even further with beam search and LM / Arpa LM.

### DIFF
--- a/recipes/CommonVoice/ASR/CTC/README.md
+++ b/recipes/CommonVoice/ASR/CTC/README.md
@@ -11,14 +11,15 @@ It is important to note that CommonVoice initially offers mp3 audio files at 42H
 Here is a list of the different languages that we tested within the CommonVoice dataset and CTC:
 - French
 - Kinyarwanda
-- Italian
 - English
 
 # Results
 
-| Language | Release | hyperparams file | LM | Val. CER | Val. WER | Test CER | Test WER | HuggingFace link | Model link | GPUs |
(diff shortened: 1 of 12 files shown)


TParcollet:
@mravanelli Here, the PR is now up to date with the NeurIPS submission. The only missing part is for Italian. @lorenlugosch would you be interested in helping me with the English recipe? Results are bad, even with w2v2. Looking at the ctcatt model the output obtained has a lot of noise, and I don't really understand why ... As you mentioned, we do not really have an "English" ASR model, as LibriSpeech cannot be considered as a real-world scenario. @aheba Could you please try to run FR recipes (just half an epoch) for w2v2 ctc+att AND ctc only ?

TParcollet:
I will push the models to HF later, as I need to create all the necessary files ...

mravanelli:
Ok, I added Italian. We can merge it!

lorenlugosch:
I'd love to help with English, but probably won't have time as I'm doing my internship now. I think our WER results might actually be what is expected; it's a very diverse and noisy dataset. I don't know what the SOTA is though

Re: "not having an English model", you're right about LibriSpeech not being realistic, but what I meant in that context was we don't have a pure CTC English model---that would be easy to add for LibriSpeech though (just get rid of the autoregressive decoder and do beam search on the CTC head). It would require retraining the model.

PR #757, opened 2021-05-14
add apply_spec_augment

jjery2243542:
Add a flag for apply_spec_augment


speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         freeze=True,
         freeze_feature_extractor=False,
         pretrain=True,
+        apply_spec_augment=False,
```
TParcollet, line 72: It needs to be added as a parameter explaining why we might wanna remove it

TParcollet:
LGTM.

PR #767, opened 2021-05-18
Low Memory Transformer's PR - [WIP]

johntremblay:
Just opening the PR for potential comments and showing the progress.

It is still a work in progress [WIP] for now.


johntremblay:
@BassirouNdao As of now, I added the Longformer and Linformer for the LibriSpeech. I suggest we use the current branch as our dev and we can branch out from this one and do PR for ourselves and finally just merge this one into the main SB's repo.

.gitignore
```diff
@@ -143,3 +143,6 @@ dmypy.json
 
 # Mac OS
 .DS_Store
+
+# IntelliJ IDE idea folder
+.idea/
```
Arka161, line 148: I think this diff should be reverted

johntremblay: @Arka161 I tried to find a solution to exclude the ideas folder from pycharm but wasnt able to find something convincing. Do you have a solution in mind?

Arka161: @johntremblay I basically meant something like avoiding any new lines in the `.gitignore`, and checking out this file from `develop` as it's just an IDE change, and not essential for speechbrain. Though it's just a clean coding convention.

johntremblay:
Will be tackled in another PR

PR #776, opened 2021-05-20
Update voxCeleb recipe to provide more flexible options.

zycv:
Generally we will download this data in other recipes, so it would be better to provide a separate option to pass this path. If it is not provided, we will fall back to the Voxceleb directory.


mravanelli:
thank you @zycv!

PR #783, opened 2021-05-22
Minor correction in path instruction

vincentqb:
`hparams/train.yaml` not `train/train.yaml`


mravanelli:
Thank you @vincentqb!

PR #785, opened 2021-05-25
Add Text-To-Speech functionality (DeepVoice, Tacotron, WaveNet)

flexthink:
This pull request aims to add Text-to-Speech (TTS) functionality to the SpeechBrain project with the following models:

- DeepVoice3
- Tacotron2
- WaveNet

DeepVoice3 and WaveNet are based on the R9Y9 implementation, whereas Tacotron2 is based on the NVIDIA implementation.

Authors:
- Artem Ploujnikov
- Aleksandar Rachkov
- Georges Abous-Rjeili


speechbrain/lobes/models/synthesis/tacotron2/model.py
```diff
@@ -0,0 +1,1184 @@
...
+            ]
+        )
+
+    def forward(self, x):
+        for linear in self.layers:
+            x = F.dropout(F.relu(linear(x)), p=0.5, training=True)
+        return x
+
+
+class Postnet(nn.Module):
```
ayush9096, line 230: @flexthink  

- Can we make the code more modular by separating blocks like PostNet Block, Input Text Pre-processing part in different files at common place , so it can be used for FastSpeech as well.

flexthink: Definitely. I believe this one comes from the original NVIDIA Tacotron.

We already have some code sharing across models
For example,

* `speechbrain/lobes/models/synthesis/dataio.py`: shared
* `speechbrain/lobes/models/synthesis/<model>/dataio.py`: specific

So far, we haven't tried sharing NN blocks, but we could create

* `speechbrain/lobes/models/synthesis/model.py`

for anything that's shared

we could even split `model.py` into multiple files for a single model - and for the shared ones too if the files get too big.

anautsch:
@flexthink hope this one is fine too

PR #788, opened 2021-05-27
fixed incorrect argument order in docstring example

leto19:
The order of the function arguments in the docstring example for `prepare_librispeech()` were in the wrong order (`splits` before `save_folder`)  - I have corrected the docstring to reflect the correct order.


mravanelli:
thank you!

geoph9:
Hi, this is very minor but the order in the docstring example is still not correct. It currently is the following:
`prepare_librispeech(data_folder, tr_splits, dev_splits, te_splits, save_folder)`

While it should be:
`prepare_librispeech(data_folder, save_folder, tr_splits, dev_splits, te_splits)`

mravanelli:
Yes, thank you. I just fixed it!

PR #804, opened 2021-06-03
AISHELL with wav2vec

mravanelli:
This small PR adds wav2vec support to the AISHELL-1. Even if the pre-training is done with different languages, it seems to help quite a bit. Our best performance is now CER=5.58%. However, there is still room for improvement to match the best numbers in the literature. For instance, we should add a language model trained on the training transcriptions.


TParcollet (approved):
Nothing to say. Just add the results to the README, if it's not done yet ?

mravanelli:
True, let me add it right after merging it.

PR #812, opened 2021-06-10
PR for MADGRAD + TIMIT Transformer Recipe + Power Normalization

Arka161:
This PR has been developed by Arka, Pavithra, and Valliappan as a part of the Transformer's group for the MILA and University of Montreal course - IFT 6759 - Advanced Projects in Machine Learning. 

This is a sub-component of our project and includes only MADGRAD, Power Normalization, the required YAML files, and a recipe for TIMIT.


danpovey:
I looked at the MADGRAD paper and I wonder whether it might be sufficient to just do Adam with weight shrinkage, but take the base for the weight shrinkage to be the initial parameter value.  If they had such a baseline, I don't remember seeing it.  All that dual-space stuff seems to me a little un-intuitive.

Arka161:
> I looked at the MADGRAD paper and I wonder whether it might be sufficient to just do Adam with weight shrinkage, but take the base for the weight shrinkage to be the initial parameter value. If they had such a baseline, I don't remember seeing it. All that dual-space stuff seems to me a little un-intuitive.

@danpovey Hello, we had done testing with MADGRAD on small datasets (in particular, the TIMIT dataset) and we observed that the model converged a lot better using MADGRAD instead of ADAM. A result of our experiments can be seen here. Hence, the optimizer has been just provided as an option in the YAML with the recipe as it might be better than Adam out of the box for this particular use-case. 

image

mravanelli:
Hi @Arka161, yes this is promising. However, the effectiveness of MADGRAD should be confirmed on larger datasets with different models, tasks, etc. I plan to run some experiments to double-check it. As for this PR, my suggestion is to separate it into two ones (1 part with the MADGRAD only and one part with the power normalization). This way we can assess the two contributions independently. The power normalization part, in particular, requires some changes in the transformer class (which is used in many other recipes in SpeechBrain), and thus we have to double-check carefully possible issues or incompatibilities with existing models.  MADGRAD, instead, is just another optimizer and we can merge that part without trouble if we find it effective.

Arka161:
We will do so. I will close this PR after the clarification, so we will have two PRs:

PR1: Just madgrad dependency and TIMIT recipe (and a sample YAML)
PR2: Just PowerNorm (Main Transformer class and Normalization changes). 

And parallely, we should test MADGRAD with larger datasets? That is it, right? 

Thank you

mravanelli:
Yes, I plan to run some experiments on large datasets.

mravanelli:
@Arka161, one small note: in the yaml file I can see:
```
optim: !name:madgrad.MADGRAD
    lr: 2.54e-4
```
From here people might think that the lr is 2.54e-4. From the code, I can see that what actually decides the lr is the Noam scheduler. Maybe we can put this to zero and add a short comment.

mravanelli:
I just ran an experiment with a Transformer on TIMIT and I can confirm that the convergence looks much faster. I will keep you updated on the other experiments.

Arka161:
@mravanelli Thank you. Will be happy to help in any testing too. 

 I will keep this PR open for the time being for testing, and I will close this when I have the 2 mentioned separate PRs fully ready.

Issue #801, opened 2021-06-01
dataio.read_audio_multichannel num_frames variable with torchaudio


leto19:
I think I've found a bug with ```dataio.read_audio_multichannel``` - if you don't specify a start and stop in the function call i.e:
``read_audio_multichannel({"files":['audio/channel1.wav','audio/channel2.wav']})``
The following error occurs: 

> Traceback (most recent call last):
  File "beam.py", line 32, in <module>
    read_audio_multichannel({"files":['audio/channel1.wav','audio/channel2.wav']})
  File "/home/acp20glc/.conda/envs/speechbrain/lib/python3.8/site-packages/speechbrain/dataio/dataio.py", line 284, in read_audio_multichannel
    audio, fs = torchaudio.load(
  File "/home/acp20glc/.conda/envs/speechbrain/lib/python3.8/site-packages/torchaudio/backend/sox_io_backend.py", line 152, in load
    return torch.ops.torchaudio.sox_io_load_audio_file(
RuntimeError: Invalid argument: num_frames must be -1 or greater than 0.

The source code of ```dataio.read_audio_multichannel``` contains the following:
```
start = waveforms_obj.get("start", 0)

# Default stop to start -> if not specified, num_frames becomes 0,
# which is the torchaudio default
stop = waveforms_obj.get("stop", start)
num_frames = stop - start
```
However the actual torchaudio default for ```num_frames``` as noted in [pyaudio documentation](https://pytorch.org/audio/stable/backend.html#load) is ```-1```, not ```0```:
>num_frames (int) – Maximum number of frames to read. -1 reads all the remaining samples, starting from frame_offset. This function may return the less number of frames if there is not enough frames in the given file.

This can be fixed by changing the line 281:
```    stop = waveforms_obj.get("stop", start) ```
in ```dataio.read_audio_multichannel``` to 
```    stop = waveforms_obj.get("stop", start-1) ```
which ensures that ```num_frames``` is always ```-1``` when no values for ```stop``` and ```start``` are specified. 

I'd make the change and push it myself, but I'm not 100% sure if I'm correct.

TParcollet:
Hey ! Could you do a pull request for that ? Then we can start discussing on the code directly :) Thanks !

PR #814, opened 2021-06-10
change related to issue #801 - dataio.read_audio_multichannel num_fra…

leto19:
…mes variable with torchaudio

Pull request related to issue: https://github.com/speechbrain/speechbrain/issues/801 @TParcollet with my proposed fix.


mravanelli:
Thank you @leto19 for this fix! I tested and everything seems to work now!

Issue #780, opened 2021-05-21
"EnvCorrupt" is not usable outside the training stage


baptistepouthier:
Hello,

Is the following condition necessary? 

https://github.com/speechbrain/speechbrain/blob/97e027ce5234e49e3f4969b218f0e12ee104f2cc/speechbrain/lobes/augment.py#L419

It does not allow the use of data augmentation during the test phase. You may want to test your model in noisy conditions.
IMHO, the "if stage == sb.Stage.TRAIN:" condition in the brain class could suffice.

What do you think?

TParcollet:
Hey, as stated you can easily apply env_corr at test time by removing the if statement in the compute_forward. However, this is not something we want to put in any recipe as models are compared with the clean version of each dataset usually.

mravanelli:
It looks to me that we can even remove that check in the lobe and only keep the check in the recipes. This way we allow using the augmentation even in test modality. Any though @pplantinga?

pplantinga:
Sure, I guess I can see a use for that.

PR #820, opened 2021-06-14
Remove train modality check in augment.py

mravanelli:
This PR removes the check on train/eval modality done internally in augment.py. The check is already done in train.py. With this small change, users might use augmentation also when the system is in eval modality (if they want it for some reason).

This topic was discussed in #780.


mravanelli:
I tested it and everything seems to work as expected. Let me merge it!

PR #836, opened 2021-06-20
Language ID experiments on CommonLanguage

paul-ruban:
Recipes for Language ID on CommonLanguage (CommonVoice dataset for LID) were added and a short description of the dataset. Two models were used for this task:

- X-Vector: pretrained on Speaker ID task on VoxCeleb, the embedding module is part of the model. Cosine Similarity and Additive marginal loss are used to classify languages.
- ECAPA-TDNN: (same as for X-Vectors) fine-tuning of speaker embeddings to LID task. 
- In both cases, batch size and augmentations were limited by compute limitations.
- No results posted as, to find how well the models can perform, more augmentation and higher batch size may be required.
- Overall, the recipe is highly inspired by Speaker ID task on VoxCeleb, as fine-tuning these embeddings proved to outperform other methods.


mravanelli:
Thank you for this contribution @paul-ruban! After the tests, I'm going to merge it and share the pre-trained model on HuggingFace...

PR #845, opened 2021-06-21
fix incorrect comments in ISTFT

wangwei2009:


mravanelli:
Thank you @wangwei2009!

PR #846, opened 2021-06-22
Fix multichannel fbank

wangwei2009:
Filterbank forward can't generate right feature for multi-channel input，I can reproduce the error use the code as below
```python
from speechbrain.processing.features import Filterbank
compute_fbanks = Filterbank()
inputs = torch.ones([10, 101, 201])
output = compute_fbanks(inputs)
inputs_ch2 = torch.stack((inputs,inputs),-1)
output_ch2 = compute_fbanks(inputs_ch2)
torch.sum(output - output_ch2[..., 0])
```

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -539,6 +539,7 @@ class Filterbank(torch.nn.Module):
 
         # Managing multi-channels case (batch, time, channels)
         if len(sp_shape) == 4:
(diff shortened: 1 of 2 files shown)


mravanelli:
Thank you very much @wangwei2009 for this fix! I'm able to reproduce the error and I'm fine with your fix. I just added a unittest for that to make sure this is always checked.

PR #850, opened 2021-06-23
Added LSTM for symbolic music generation

Edward-Son:
Code for symbolic music generation (LSTM). Each dataset has a YAML file related due to different hyperparameters. Training can be run like any other Speechbrain model. Directory contain a generation python script, which will generate a MIDI file in the corresponding dataset results folder.


ycemsubakan:
Thanks for the pull request @Edward-Son . I will try to review this, this week.

ycemsubakan (commented):
Thanks for the PR @Edward-Son. I was able to run the training script, and the code looks great. Before merging however I think we can make some of the things about dataset preprocessing little easier for the users. Here are few comments: 

- [x] In the current organization we name the recipes according to dataset --> task. So, following that organization, I would call the recipe `MAESTRO`, instead of `SymbolicMusic`. And, you can rename the `LSTM` folder under `SymbolicMusic`, `symbolic-generation`.
- [x] It seems you have two different `extra_dependencies.txt` files. One under the top level, and one under the `LSTM` folder. I think you can remove the file `extra_dependencies.txt` file under the `LSTM` folder. 
- [x] It seems like maestro-v2.0.0 is hardcoded as the dataset path. Can we move this to the hparams file so that people can use their own path for the dataset ?
- [x] In the README, it seems like we give a link to download MAESTRO v3.0.0, but in the code MAESTRO v2.0.0 is hard coded. I think we should fix this discrepancy. 
- [ ] I think we can add a small script so that the data is automatically downloaded and processed before running the model. 
- [ ] It seems we are not saving the preprocessed dataset, which causes to rerun the preprocessing script each time we run the model. Is that right? Especially MAESTRO dataset takes very long to preprocess, so it would be great if we could save the preprocessed dataset, and each time we run the script it can be loaded. This way we can avoid the wait time before running the script. 
- [x] We need to add `pandas` to the `extra_dependencies.py`. 
- [ ] I think we move `custom_model.py` under a new folder called `speechbrain/lobes/music_generation_models`.
- [ ] For generation, I think we can move `generate.py` to our general inference file which are used for HuggingFace models. This way we will be able to share our generation models on HuggingFace as well. Our inference interface functions are in the file `speechbrain/pretrained/interfaces.py`. You can use one of the classes to figure out what is needed to create a similar class.

Edward-Son:
Hey @ycemsubakan , thanks for the comments! I agree with your suggestions and will make the changes.

ycemsubakan:
Thanks @Edward-Son ! Please note that I added one more comment about inference interface as well.

ycemsubakan:
Thanks a lot for the updates @Edward-Son!

I have pushed some changes in the .yaml file for the MAESTRO dataset to make things hopefully little bit more compliant with the conventions we follow in the other recipes. I also added a pre-processing bit so that the .csv file with the data is created inside the train file so that we can directly run an experiment by calling `train.py`. 

The remaining things in my opinion are: 

- [ ] Automatic download of the MAESTRO data.
- [ ] updating the .yaml files for the other four datasets (other than MAESTRO) to follow the same variable names as the MAESTRO yaml with my latest updates. 
- [ ] for some reason the linters are failing. The culprit seems to be `./recipes/Voicebank/enhance/segan/train/hparams/train.yaml`, which I don't understand why it is affecting this PR. Any ideas @mravanelli ?

ycemsubakan:
Just added automatic downloading. The paths are not correct, I will that correct in few hours.

ycemsubakan:
Hi @Edward-Son , @mravanelli , 

I have gone over the code, and I have updated the task list above. At the moment the automatic data preparation seems to be working on my side for the MAESTRO data (downloading the .zip file, and unzipping and putting it to the correct path), and for the other datasets (JSB - Musedata, Nottingham, Piano-Midi  -- where we download the .pickle file if it doesn't exist). 

The problem now seems to be at the generation part. For some reason even when I change the path manually in `inferer = SymbolicGeneration.from_hparams(source="correctpath",)`, this is giving the error, `pretrainer = hparams["pretrainer"]`. So we need to update the `.yaml` files to accommodate this. 

Also, I think that we can migrate most of the content in `generate.py`, to `interfaces.py`, so that we can generate a sequence simply with something like: 
```
generator = SymbolicGeneration.from_hparams(source="correctpath",)
generated_wav = generator.generate(seq_len=N)
```

ycemsubakan:
I have updated `generate.py` file so that it's a bit simpler, and takes in a path for the checkpoint folder. I also reverted the commit for the music transformer, as I think we can include it in a new PR, as this one can serve as the basis point.

mravanelli:
I realized that this wasn't merged. @ycemsubakan could you please resume it and finalized when you have some time?

PR #851, opened 2021-06-24
Support ERP decoding

ddavidebb:
With this pull request I implement an ERP decoder based on a compact CNN to detect the P300 event from single electroencephalogram trials.


mravanelli:
Thank you @ddavidebb! Thank you very much for this PR. Let me take this opportunity to announce to the community that we are going to support some other EEG classification tasks soon. Any help from the community is welcome on that front as well.

mravanelli:
Hi @ddavidebb ,
thank you for this PR.
Here are my suggestions (mainly regarding documentation/comments):
1. It is very important to add in `recipes/ERPCore/P3_decoding` a `README.md` file (e.g., see similar files in other recipes).  Here is even more important because we can expect that users aren't familiar with EEG. I would suggest structuring the README in the following way:
**A- Task description** (clarify that here we process EEG brain signals and not speech, describe what P100 is and why it is important, how data are collected, etc. Feel free to add some papers as additional references)
**B- How to run an experiment** => Here you can basically copy and paste what already reported in the docstring of `train.py`.  Maybe you can add few sentences on the k-folding and per subject training. In speech, we train a model that is good for all the speakers, while here we have to train a different model for different subjects (this should be justified). Moreover, since we only have little data for each subject, we need to do a k-fold evaluation. This should be mentioned as well.
**C- Performance** => here you should report the performance. Normally, we also specify the needed computational time and the GPU used for the experiment. 

2. `prepare.py` => add a docstring with a short description of input and output arguments.
3. `extra-depedencies.txt `=> it looks like sklearn is another extra-depenency to add. Please, remove the empty line at the beginning of the file

4. in `p3_decoding.yaml` `sbj_id: !PLACEHOLDER` => add an example of input (e.g sbj_id: !PLACEHOLDER #e.g., sub1)

5. in `p3_decoding.yaml` `# Change 'data_dir' and 'results_dir'` => this looks outdated, right?

6. `train.py` => write short docstrings in all the function that you declare (see for instance https://github.com/speechbrain/speechbrain/blob/develop/recipes/TIMIT/ASR/seq2seq/train.py)
7. `train.py `=> import torch as th. "Th" is not used in speechbrain. I suggest simply using torch-

8. `train.py` => `tmp_idx_valid0`, `tmp_idx_valid1`: I would add a comment to clarify this part. I would also add a comment to clarify that the dataset is  very small and can be stored in the RAM directly (without the need of reading from disk as done in all the other recipes in SpeechBrain)
9. `train.py` => `if epoch is not None` => I'm not sure to understand why this check is needed
10. epoch_loop=> stopper => as discussed privately, maybe you can inherit from the epoch counter. Also, please add a working usage example in the docstring.
11. `linear.py`, `CNN.py` => The new classes requires a docstring formatted in the same way as the other ones (+ it is important to add a usage example)

mravanelli:
Some additional comments:

12. in the docstring of train replace` python3` => `python`
13. remove the train extra folder: `recipes/ERPCore/P3_decoding/train` => `recipes/ERPCore/P3_decoding/*`
14. To do it you can use `git mv file1 file2` and `git rm train` to eliminate the folder once empty 
15. `p3_decoding.yaml` => `train.yaml`
16.` output_folder: !PLACEHOLDER` => `output_folder: !ref results/ERPCore_P3_results/<seed>`

mravanelli:
And here are my last comment after trying the recipe:
16. The code seems to work only with an old version of mpe
17. The file `train_log.txt` is not created because the training logger is not added in `train.py`. You can take a look at other recipes to figure out how to use it (https://github.com/speechbrain/speechbrain/blob/develop/recipes/TIMIT/ASR/CTC/train.py#L92)
18. I suggest adding a comment in `README.md` to describe the content of `metrics.pkl`
19. I would add in the README how the final performance is computed (i.e. averaging the folds and the subjects).

ddavidebb:
Thank you @mravanelli. I reviewed the recipe following your suggestions.

mravanelli:
Excellent work @ddavidebb!
Everything seems to work. I only did some minor modifications (see my latest commits). The main issue with the current version is that it only works with `mne==0.20.7`. With this version, there are so many deprecation warnings that I had to switch off temporarily. The code doesn't look compatible with the latest version of mne (there is an error showing up). I assume most of the users would like our code to work with the latest version of mne, so I would suggest trying to fix this incompatibility issue.

mravanelli:
Excellent work @ddavidebb! I think we can merge it! thank you

PR #859, opened 2021-06-29
Hubert

Moumeneb1:
Hi, 

I added The Hubert hugging face model https://arxiv.org/abs/2106.07447 with a recipe on SLURP as SLU. it's giving some good results for now. It's exactly the same implementation as wav2vec2


TParcollet:
I wonder if we do not want something more generic, so we do not replicate N HuggingFace-based models if the API is the same ... Hum.

TParcollet:
Maybe we just create a HuggingFace lobe and we put all models there, or we just create an HuggingFace folder or we wrap everything in a superclass inheriting from HuggingFace hub.

Moumeneb1:
Actually there is a huggignface class called AutoModel. All models can be loaded from it with AutoModel.from_pretrained('name of the model on the hub'), maybe we can use it.

TParcollet:
do they have all extract_features etc etc ? What I am a bit afraid is to loose the modelName in the yaml instanciation. It is important to show what we are using ..

mravanelli:
@Moumeneb1 any news on that?

Moumeneb1:
An AutoModel can be too generic, and can even load models which are not related to speech. and for the code They use the same processors of wav2vec2 in HubertModel so it's basically the same implementation. 
This is the example code from [HuggingFace's doc](https://huggingface.co/transformers/model_doc/hubert.html?highlight=hubert). 

```python
from transformers import Wav2Vec2Processor, HubertModel
from datasets import load_dataset
import soundfile as sf

processor = Wav2Vec2Processor.from_pretrained("facebook/hubert-large-ls960-ft")
model = HubertModel.from_pretrained("facebook/hubert-large-ls960-ft")

def map_to_array(batch):
    speech, _ = sf.read(batch["file"])
    batch["speech"] = speech
    return batch

ds = load_dataset("patrickvonplaten/librispeech_asr_dummy", "clean", split="validation")
ds = ds.map(map_to_array)

input_values = processor(ds["speech"][0], return_tensors="pt").input_values  # Batch size 1
hidden_states = model(input_values).last_hidden_state
```

TParcollet:
I see, @Moumeneb1 Could you update the table with the results in the README as well? And fix the tests. Then I'll try the recipe.

PR #860, opened 2021-06-29
Load private huggingface models

Moumeneb1:
Users can upload private models to the huggingface hub and download them using the huggingface auth_token CLI. But this is not supported in the current speechbrain code, so I added the ability to specify if users want to use huggingface's private token so they can download private models to speechbrain.


mravanelli:
I tested and it seems to work. Thank you @Moumeneb1!

PR #862, opened 2021-06-29
Refact transformer encoder module

johntremblay:


johntremblay:
Will be tackled in another PR

PR #864, opened 2021-07-01
Add Emotion Recognition Recipe with IEMOCAP

pierreyanni:
Recipe for emotion recognition using the IEMOCAP dataset (using light version of ECAPA-TDNN).


mravanelli:
Thank you @pierreyanni! I did the last changes and replicated your results. I think we can merge it!

PR #871, opened 2021-07-05
Implement VAD

mokleit:
This provides general recipe for training and evaluating a VAD system. It also includes a script for preparing the data prior to the training.


mravanelli:
Hi @mokleit,
thank you for this PR. In my last commit I did some fixes and now everything is working as expected on my side. 
My next step now is to work on the inference part (I would like to add a model on HuggingFace as well). One simple thing that I would ask you is to add short docstrings on all the functions defined in `libriparty_prepare.py`.

mravanelli:
I think everything is done. The VAD now seems to work well. Thank you @mokleit for the PR!

PR #873, opened 2021-07-06
Correct FAR and FRR

IHEII:
Is that
FAR = FPR = FP / (FP + TN)
FRR = FNR = FN / (FN + TP)


mravanelli:
HI @IHEII,
thank you very much for this fix!

PR #877, opened 2021-07-08
Add a guided attention loss

flexthink:
Add a guided attention loss implementation for sequence-to-sequence models where there is a direct correspondence between inputs and outputs, such as Grpaheme-to-Phoneme and Text-to-Speech. 

The implementation is based on this paper
https://arxiv.org/abs/1710.08969

and inspired by the R9Y9 implementation of DeepVoice3:
https://github.com/r9y9/deepvoice3_pytorch

However, unlike the R9Y9 implementation, which contains nested loops and must be optimized with Numba, this proposed implementation relies on PyTorch vectorization instead.


TParcollet:
@mravanelli LGTM, maybe we could merge.

mravanelli:
The only thing that should be done before merging it is to add an example in the docstring.

flexthink:
@mravanelli : Will add shortly

flexthink:
@mravanelli : The example is done

mravanelli:
Thank you for this contribution @flexthink! We should try guided attention for speech recognition as well to figure out if that might help convergence or not.

PR #879, opened 2021-07-08
added SEGAN model. Already corrected based on Mirco's comments/course

carterfrancis:
…peechbrain_dsr


mravanelli:
Thank you for this PR @carterfrancis. I would like to push some small fixes, but it looks I don't have the right permissions. Let me thus merge this PR and open another one with the fixes.

PR #883, opened 2021-07-10
Fixes on SEGAN recipes

mravanelli:
With this PR I propose some fixes to the original PR on SEGAN


carterfrancis (approved):
Looks good, thanks for the input!

mravanelli:
Hi Francis,
I'm still working on that. I might have found a way to replace the
"blocking" part with the fold/unfold functions of pytorch. If everything is
working I will do the change.

PR #891, opened 2021-07-23
Add threshold paramater to verify_files

jaketae:
This PR adds an additional `threshold` parameter to the `verify_files()` function of the `SpeakerRecognition` interface. `verify_files()` internally calls `verify_batch()`, which accepts a `threshold` parameter that determines the decision boundary for cosine similarity. Previously, the internal call only used the default parameter value of 0.25; this PR allows users to use different threshold values when calling `verify_files()`.


mravanelli:
Thank you for the fix!

Issue #902, opened 2021-07-29
Speaker verification EER changes with batch size


SedatDe:
Hi, during experimenting ECAPA verification we noticed that EER changes according to the batch size given. The batch dependency seems to be coming from max calculation done in __amplitude_to_DB_ method of _Filterbank_ class. The max is calculated over whole batch.

`new_x_db_max = x_db.max() - self.top_db`

Is the max calculation over whole batch intended? When I change this line to the following, to calculate max for each sample, the EER stays same between different batch sizes, but the EER becomes higher than previous.

`new_x_db_max = x_db.max(dim=1, keepdim=True).values - self.top_db`

masztalskipiotr:
Noticed the same thing as well. Couldn't get to the exact cause of this problem. Thanks @SedatDe!

mravanelli:
Interesting observation. There might be something to fix here and this
could be connected to the multi-gpu issue as well. We just have to figure
out the right normalization axis which should be the last dimension. In
anycase, we should run again the training part as well to have a network
that matches with that change.

mravanelli:
Thank you for pointing out this issue. It is solved in the latest development branch. I ran again the pre-trained ECAPA-TDNN model with the fix, and the performance looks very similar (or even slightly better):

```
__main__ - Computing EER..
__main__ - EER(%)=0.659382
__main__ - minDCF=0.085088
```

SedatDe:
Thank you for the fix.

PR #903, opened 2021-07-29
Fix FBANK db clipping

TParcollet:
FBANK top db was computed batch-wise. Now, it is computed feature-wise. 

Need to check the impact on the pretrained models and new trained ones:

TIMIT (pre-trained) = 13.78% (before 14.01)
TIMIT (re-train) = 14.18%

Related to #902


mravanelli:
I ran again speaker verification with ECAPA with the pre-trained model.
The performance looks similar or even slightly better:
```
__main__ - Computing EER..
__main__ - EER(%)=0.659382
__main__ - minDCF=0.085088
```
I think we can merge it

Issue #899, opened 2021-07-27
Big difference between training with 1 GPU vs 8 GPUs (via DDP) 


PabloPeso:
Hi all,

I am using the recipe LibriSpeech/ASR/transformer with the hyperparameters saved in hparams/conformer_small.yaml.

I firstly tried training on a single GPU with the following result
```
epoch: 1, lr: 1.76e-04, steps: 4394, optimizer: Adam - train loss: 2.61e+02 - valid loss: 1.41e+02, valid ACC: 1.86e-01
epoch: 2, lr: 3.51e-04, steps: 8788, optimizer: Adam - train loss: 1.48e+02 - valid loss: 1.11e+02, valid ACC: 3.76e-01
epoch: 3, lr: 5.27e-04, steps: 13182, optimizer: Adam - train loss: 67.33 - valid loss: 58.62, valid ACC: 6.64e-01
epoch: 4, lr: 7.03e-04, steps: 17576, optimizer: Adam - train loss: 49.27 - valid loss: 75.51, valid ACC: 6.06e-01
epoch: 5, lr: 8.79e-04, steps: 21970, optimizer: Adam - train loss: 41.83 - valid loss: 21.69, valid ACC: 8.57e-01
epoch: 6, lr: 9.74e-04, steps: 26364, optimizer: Adam - train loss: 37.18 - valid loss: 14.90, valid ACC: 8.99e-01
epoch: 7, lr: 9.02e-04, steps: 30758, optimizer: Adam - train loss: 32.34 - valid loss: 12.56, valid ACC: 9.14e-01
epoch: 8, lr: 8.43e-04, steps: 35152, optimizer: Adam - train loss: 28.62 - valid loss: 15.06, valid ACC: 8.97e-01
epoch: 9, lr: 7.95e-04, steps: 39546, optimizer: Adam - train loss: 25.86 - valid loss: 10.85, valid ACC: 9.24e-01
epoch: 10, lr: 7.54e-04, steps: 43940, optimizer: Adam - train loss: 23.81 - valid loss: 10.45, valid ACC: 9.28e-01, valid WER: 7.79
```

[...]

PR #906, opened 2021-07-30
Fix DDP and DP for transformer and conformer

TParcollet:
Hey, 

This is a PR to investigate and solve the problem raised in #899

For now: 
- Found a problem since all calls in forward are done with self.hparams instead of self.modules for GPUs modules. This destroys DP.
- Found a bug when fixing the call to modules. Masking has a problem when using DP. @popcornell is investigating this as well.


TParcollet:
I will open a cleaner PR.

PR #921, opened 2021-08-05
Fix DER table

lewtun:
This PR fixes a missing column separator in the metrics table for the DER recipes.


mravanelli:
Thank you for the fix!

PR #929, opened 2021-08-12
KenLM as optional dependency

Gastron:
Either use this PR as a reference for how this could be implemented or just merge.

Changes:
- KenLM as optional dependency
- Removed Sentencepiece tokenizer stuff from the KenLM class, that functionality seemed totally orthogonal
- Renamed to KenLMScorer. I am happy to reverse this as well, since KenLM supports any ARPA LM probably? But I think this name is the clearest.

### DIFF
--- a/speechbrain/decoders/scorer.py
+++ b/speechbrain/decoders/scorer.py
@@ -1,7 +1,5 @@
 import torch
 import numpy as np
-import kenlm
-import sentencepiece as spm
 import speechbrain as sb
 from speechbrain.decoders.ctc import CTCPrefixScore
 
@@ -192,9 +190,8 @@ class TransformerLMScorer(BaseScorerInterface):
         return None
 
 
-class NGramLMScorer(BaseScorerInterface):
-    """A ngram LM scorer. Please make sure the tokenizer or token_list
-    matches the tokenizer of the ASR model.
+class KenLMScorer(BaseScorerInterface):
+    """KenLM N-gram scorer
 
     Arguments
     ---------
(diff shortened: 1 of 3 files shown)


30stomercury:
@Gastron, thanks for this PR.
+ I'm okay with KenLMScorer, note that we also need to change yaml files involves ngram scorer. Also, `tokenizer_path` should be removed from the args.
+ Any example of function to parse the token_list given a sentencepiece model path?

speechbrain/decoders/scorer.py
```diff
@@ -192,54 +190,46 @@ def reset_mem(self, x, enc_lens):
...
-                for i in range(vocab_size)
-            ]
-
-        else:
-            raise ValueError(
-                "Please specify the token list or the path to the spm tokenizer."
-            )
+        if len(token_list) != vocab_size:
+            MSG = "The size of the token_list and vocab_size are not matched."
+            raise ValueError(MSG)
```
30stomercury, line 224: Should we also handle other error messages in this way? I think this part should be consistent. Many error messages are not in this format.

Gastron: Here I changed an assert statement, and yes, mostly we should use this if-raise pattern. Asserts are meant for [debugging](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement), like an internal consistency check. In other words, if an assertion error occurs, it should mean that there is a bug in SpeechBrain code, not an error in user code.

30stomercury: I will adapt this rule to other parts.

Gastron:
>     * Any example of function to parse the token_list given a sentencepiece model path?

I added an implementation. In YAML it could be used like:
```yaml
spm_vocab: !apply:speechbrain.tokenizers.SentencePiece.get_spm_tokens
    model_path: !ref <spm_file_path>
```

speechbrain/decoders/scorer.py
```diff
@@ -192,54 +190,46 @@ def reset_mem(self, x, enc_lens):
...
         The path of ngram model.
     vocab_size: int
         The total number of tokens.
-    tokenizer_path : str
-        The path of pre-trained sentencepiece tokenizer.
     token_list : list
         The tokens set.
     """
 
     def __init__(self, lm_path, vocab_size, tokenizer_path=None, token_list=[]):
```
30stomercury, line 206: `tokenizer_path` should be removed here. I can do it on my side.

30stomercury:
Thanks @Gastron, LGTM.

Issue #19, opened 2020-04-29
Update README


Gastron:
The README still describes the old CFG file architecture. The README should be updated.

I think first step would be to remove out-of-date info and add the latest development guidelines.

PR #934, opened 2021-08-14
Add Text-to-Speech functionality (Tacotron only)

flexthink:
This pull request aims to add Text-to-Speech (TTS) functionality to the SpeechBrain project with the Tacotron model:

Authors:

* Georges Abous-Rjeili
* Artem Ploujnikov


mravanelli:
Amazing work @flexthink! This looks really great.
In the following, you can find several small comments that mainly go in the direction of making the recipe more compliant with the others and improving comments, documentation, clarity:

1. In SpeechBrain we always have a data preparation script that prepares the data manifest files (e.g, in json format). On the top of the data manifest the data-io-prep function is created. Note that by design, we would like to have the reading pipeline visible in the train.py directly (see some recipe for an example: https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/seq2seq/train.py).

2. We normally have a README.md for each recipe. It has a pretty standard format (https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/seq2seq/README.md) where we report some information like how to run the experiment, where to find the dataset, performance, link to huggingface for inference (to be added when available), link to google drive (for experiment logs, to be added when available), citing speechbrain etc.

3. Please, remove files in samples. We use those samples only for unit tests integration tests, and some docstring tests.  The goal is to keep the. repository as small as possible. I hope these files won't be added to the .git history when we merge the PR. 

4. We now have two hparams file. We might wanna take one for now and add the phoneme-based one once we have more results with that. 

### Comments on hparams.yaml 
5. Please, add header in docstring with author name + short description (e.g,https://github.com/speechbrain/speechbrain/blob/develop/recipes/TIMIT/ASR/CTC/hparams/train.yaml#L1)

6.`data_folder: ./ #data_folder: !PLACEHOLDER # e.g, /localscratch/fluent_speech_commands_dataset` => `data_folder: !PLACEHOLDER # e.g., /localscratch/LJSpeech`

7. 
```
pretrained_path: !ref <save_folder>
save_for_pretrained: True
```
=> I'm not sure to understand the exact role of it. We already save checkpoints in the fit method + we automatically recover them when needed. This looks pretty much what we need here. What do we add with this extra functionality?

8. For the progress sample part, you might wanna add some short comments in the yaml and train.py to clarify what we are doing (i.e., saving intermediate samples during training)

9. `data_folder_rirs: !ref <data_folder>` => it looks like this part is not used and can be removed.

10. `train_filter: null,  valid_filter: null` => You can briefly explain the role of that with a short comment.

11. `preprocessing_mode: nvidia` => Where is it used and what are the other options? Maybe add a comment

12. `# TODO: Not using !ref because it triggers an odd bug when loading` => is that solved in the end?

13. `text_cleaners: ['english_cleaners'] `=> what are the other options?

14. ` !apply:speechbrain.lobes.models.synthesis.tacotron2.encode_text `=>  `!apply:speechbrain.lobes.models.synthesis.tacotron2.data_io.encode_text` I think it is better to always report the full path. This way users can just go there and check the function.

15.
```
################################
# Audio Parameters             #
################################
sample_rate: 22050
hop_length: 256
win_length: 1024
n_mel_channels: 80
n_fft: 1024
.....
```
=> It is not immediately clear which function/class is using these hyperparameters because they do not appear anymore in the yaml and I cannot find them in train.py. The same happens for other parameters like grad_clip_thresh, batch_size. The users need to find immediately in yaml or train.py who is using the hyperparameters. 


16. ```n_frames_per_step``` => what is that? A short comment might help.

17.
```
   external:
    waveglow: !ref <waveglow> 
```
=> what is the role of that. It looks like it is not used in the yaml and train.py


18. ```model: !new:speechbrain.lobes.models.synthesis.tacotron2.Tacotron2 ```=>``` model: !new:speechbrain.lobes.models.synthesis.tacotron2.model.Tacotron2```
(It is better to be explicit as mentioned before.)

### Comments on Train.py

19. ```Tacotron2Brain(sb.Brain, PretrainedModelMixin, ProgressSampleImageMixin)``` => normally we inherit from sb.Brain only. By design, we don't want to add too many levels of abstraction.  Here things are a bit more complex though that standard recipes, but I'm wondering how bad would be to avoid that and put the needed functions in ```train.py``` instead (or just import the needed methods at the beginning of ```train.py```)

20. ```PROGRESS_SAMPLE_FORMATS``` => what's the role of this and where it is used?

21. ```def __init__(self, *args, **kwargs):``` => I think we can use one of the methods already available in the brain class like ```self.on_fit_start()```

22.
```
        (
            text_padded,
            input_lengths,
            mel_padded,
            gate_padded,
            output_lengths,
            len_x,
            labels,
            wavs,
        ) = batch
```
It looks like the batch variable contains a lot of thing but many of them are not used. Do we need to have a batch variable that complex?

23.
```def batch_to_device(self, batch):``` => I think it would be clearer if you explain in the docstring what we return (e.g., what is x, y, len_x, labels, wavs?)

24. ```# 1#2# ```=> what is this comment?

25. ```self.last_batch = effective_batch``` => you can add a short comment here to explain why it is needed (i.e., we perform inference on the last batch of the validation step)

26. Please, add at leas a short docstring to all the methods/functions (e.g., run_inference_sample, _get_sample_data,  _remember_sample, run_inference_sample)

27. ```self._remember_sample``` => the role if this is not immediately clear, please add a comment to explain.

28. ```def to_device(self, x):``` => why a simple .to(self.device) is not enough? In other recipes, we simply do something like that. I'm not sure calling .contiguous is anymore necessary.  Did you see any problems connected to that?

29.
```
    # If distributed_launch=True then
    # create ddp_group with the right communication protocol
    # sb.utils.distributed.ddp_init_group(run_opts) 
``` 
Why this is commented out? In other recipes we use it even when training is done on a single gpu.

### Comments on /speechbrain/lobes/models/synthesis/tacotron2/model.py

30. ```class Tacotron2(nn.Module)``` => The docstring is not the standard one. Please also add a working example.

31. the docstring of the functions can be improved. For instance, "decode" only describes part of the input arguments.

32. Please, improve the following docstrings:
- class Decoder(nn.Module): => add docstring + example
- Encoder => improve docstring + add example in docstring
- Postnet =>  improve docstring + add example in docstring
- Attention => improve docstring + add example in docstring
- LocationLayer =>  improve docstring + add example in docstring
- ConvNorm => improve docstring + add example in docstring
- LinearNorm=> mprove docstring + add example in docstring
- Prenet => write docstring + example

33. Make sure also support functions have at list a short (e.g 1/2 line) docstring.

34. What's the need of redefining ```LinearNorm```,``` ConvNorm```. It looks like the initialization if different, but does it really play any role to justify a redefinition of it?

35. I have the feeling that the clarity might be improved a bit with some more in-line comments.

36. It looks like the model in this library (e.g., Tacotron2) does not set most of the times the arguments to default values. I think this is a helpful thing because shows directly to the user what could be a reasonable value.

### Comments on speechbrain/nnet/schedulers.py
37. ```IntervalScheduler``` =>  Add docstring example (just like the other ones)

### Comments on /speechbrain/lobes/models/synthesis/tacotron2/data_io.py
38. It looks like many lines are saying "TO DO". We might wanna take this opportunity to do whatever is easy to fix.

39. Improve docstrings and add examples where possible (e.g, ```ExternalModelWrapper)```. Make sure all the functions (even internal support functions) have at least a short docstring (e.g, 1/2 lines).


### Comments on recipes/LJSpeech/TTS/tacotron2/synthesis.py 

40. Add docstring with usage example.
41. ```lobes/models/synthesis/external/waveglow.py```  => It looks like we create a new folder just for a small library that only contains few lines. Is that necessary? 

42. ```speechbrain/utils/progress_samples.py``` => Improve docstrings for classes + docstring example

43. It looks like you still have recipes/TTS with some files inside. You might want to add that function somewhere else. 

Feel free to take you time to address the comments!

**Comment by flexthink** [comment]
Answers to comments that are questions:

11 - the other option is "speechbrain", using the standard text encoder 
13 - basic_cleaners, transliteration_cleaners (see text_to_sequence)

17: This is needed to identify any external models to the pretrained wrapper - it will need to transfer them to the right device.

```python
    def _transfer_external(self):
        """Transfers all external models to the same device as this
        model"""

        external_models = getattr(self.hparams, "external", {})
        for model in external_models.values():
            model.to(self.device)

```

19 - the mixins are shared by multiple Brains on the original branch. One possible way to resolve this is with a different design pattern (e.g. by putting it in hparams)

20 - Progress samples are tidbits that can be saved to monitor a running model, the old-school way - by saving files in the filesystem (spectrograms, etc). By default, the mixin will save everything as images.  If that variable is defined, it will override the format for specific components of the sample - in this particular case, it will save a "batch sample" as a PyTorch file.

If we go with #19, this can go into hparams

22 - I would personally prefer a data structure other than a tuple here (e.g. namedtuple?)

24 - needs to be removed
29 - we may want to uncomment, I don't think it was me that commented it out, although I've only trained with DP, never with DDP
34 - this came from the original nvidia code, we can revise it

43 - the only thing left is the synthesis script, I can put it anywhere else.

PR #937, opened 2021-08-16
[fix] update train.py

fanlu:
change squeeze() to squeeze(dim=1) to avoid only one sample in last batch result in `TypeError: 'int' object is not iterable`


mravanelli:
Makes sense, thank you @fanlu!

PR #940, opened 2021-08-17
In training ECAPA-TDNN，learning rate adjustment

qiny1012:
The learning rate is changed only on the training set and not on the verification set


mravanelli:
Thank you @qiny1012 for the fix. I just fixed the linters and now all the tests should pass.

PR #944, opened 2021-08-19
Support wav2vec training on LibriSpeech

30stomercury:
Hi @mravanelli, @TParcollet,

this is a PR for  wav2vec training on LibriSpeech.
The training is based on `facebook/wav2vec2-large-960h-lv60-self`, and I got **1.90** on test-clean, **3.96** on test-other.
The performance is close to the numbers reported here:
https://huggingface.co/facebook/wav2vec2-large-960h-lv60-self


30stomercury:
Pre-trained model for reproducing the results:
https://drive.google.com/drive/folders/1uKpnllZRM3fs5KVAR7HiFlaNhefGdwLb

@aheba  you may also find this useful.

recipes/LibriSpeech/ASR/seq2seq/hparams/train_with_wav2vec.yaml
```diff
@@ -0,0 +1,160 @@
...
+lr: 0.9
+lr_wav2vec: 0.0001
+sorting: ascending
+auto_mix_prec: False
+sample_rate: 16000
+
+
+# With data_parallel batch_size is split into N jobs
+# With DDP batch_size is multiplied by N jobs
+# Must be 6 per GPU to fit 16GB of VRAM
```
TParcollet, line 44: why is the batch size 8 if it should be 6 to fit 16GB ?

recipes/LibriSpeech/ASR/seq2seq/train_with_wav2vec.py
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/env/python3
+"""Recipe for training a sequence-to-sequence ASR system with librispeech.
+The system employs an encoder, a decoder, and an attention mechanism
```
TParcollet, line 3: There is no attention mechanism right ? Also the recipe should be teleported to /CTC and not /seq2seq, right?

recipes/LibriSpeech/ASR/seq2seq/hparams/train_with_wav2vec.yaml
30stomercury: I use 8 for training. Yep I got OOM in last few iterations. I can change it to six to fit the vram.

recipes/LibriSpeech/ASR/seq2seq/train_with_wav2vec.py
30stomercury: I will change the descriptions. Makes sense, let me move it to `/ctc`.

mravanelli:
Great. @TParcollet, could you replicate this number? Once we replicated that and did the last changes I think we can merge it!

mravanelli:
+ we have to put the model on HuggingFace

mravanelli:
Note that according to the [requirements](https://github.com/speechbrain/speechbrain/blob/develop/requirements.txt#L11) we suggest pytorch 1.8. We actually support pytorch 1.9, which is working totally fine except for some annoying warnings triggered internally to pytorch 1.9.

TParcollet:
I'm on it.

recipes/LibriSpeech/ASR/ctc/hparams/train_with_wav2vec.yaml
```diff
@@ -42,7 +42,7 @@ sample_rate: 16000
 # With data_parallel batch_size is split into N jobs
 # With DDP batch_size is multiplied by N jobs
 # Must be 6 per GPU to fit 16GB of VRAM
-batch_size: 8
+batch_size: 6
```
TParcollet, line 45: Did you try with 6, because I am doing some VRAM tests right now, and it's crazy. BS has to be one to fit a 24GB GPU xD

30stomercury: I used 8 for training and got OOM around last few iterations.

TParcollet:
@30stomercury I tested, it works. Could you please add a README with the results ?

TParcollet:
Once we have the results, I'll merge.

TParcollet:
Ok, now LGTM. Merging after the tests. Thanks @30stomercury

mravanelli:
Ok, can we merge then?

PR #958, opened 2021-08-27
python 3.7 and 3.9 github actions

turian:
If speechbrain setup.py says that speechbrain supports Python >= 3.7, then CI should run speechbrain unittests on python 3.7 - 3.9.

The github actions will not run on this PR unless you give my PR permission to trigger CI. But you can see the CI results here: https://github.com/turian/speechbrain/pull/1

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -32,8 +32,8 @@ jobs:
         name: Tests
         runs-on: ubuntu-latest
         strategy:
-          matrix:
-            python-version: [3.7, 3.8, 3.9]
+            matrix:
+                python-version: [3.7, 3.8, 3.9]
         steps:
             - uses: actions/checkout@v2
             - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -31,15 +31,20 @@ jobs:
     tests:
         name: Tests
         runs-on: ubuntu-latest
+        strategy:
+          matrix:
+            python-version: [3.7, 3.8, 3.9]
         steps:
             - uses: actions/checkout@v2
-            - name: Set up Python 3.8
+            - name: Set up Python ${{ matrix.python-version }}
               uses: actions/setup-python@v1
               with:
-                  python-version: 3.8
+                  python-version: ${{ matrix.python-version }}
             - name: Install libsndfile
               run: |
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you @turian, this is very useful! 
We did an internal discussion and we can take this opportunity to also test different versions of pytorch (e.g, 1.8 and 1.9). However, this might make the tests too time and computationally demanding. We would thus like to take this opportunity to revise the testing pipeline as well. The longest tests are those called "integration tests", where we train a small model on a tiny dataset and we check if we can overfit it. Our idea is to do integration tests only at the end of each PR (e.g., when the tag "ready-to-review" is triggered) and do the other less time-demanding tests (unit tests, docstring tests, linters) after every push. Do you have any suggestions on that?

turian:
@mravanelli my recommendation is sign up for travis which gives you 1K hours of free CI time a month.  Here is an example travis file I use:

https://github.com/torchsynth/torchsynth/blob/main/.travis.yml

I am not sure about travis supporting "ready-to-review" but it has a lot of features, so I assume it does? You can also tell it to do integration tests on main once a week, which is nice, e.g. if the dependencies have updated.

You can ask travis for more free credits as an open source project. Or you can pay.

Lastly, you can configure travis to try different pytorch and python versions, as well as different OS.

Here you can see a huge travis matrix where I try everything under the sun:

https://github.com/torchsynth/torchsynth/pull/280/files

We would someone just merge that PR with main and it would take hours to try every configuration. Then we could see which configuration work, and would update our docs and maybe fix a few bugs with weird versions.

turian:
An alternative is circleci.

This is all assuming that github actions timeout with your CI/CD integration tests. If they don't then just use that every push?

Gastron:
Hey, sorry it took me this long to get to this particular PR. I think this is an important improvement - if we claim to support multiple versions, we should make sure we actually do. I also appreciate the advice about Travis / CircleCI, but as noted, GitHub actually runs our CI just fine.

Regarding running time: The different versions all run in parallel. Interestingly, the Python 3.7 tests take 50% more time (9min vs. 6min).

This PR looks good to me, and in future work we will implement this for different PyTorch versions as well.

Gastron (approved):

turian:
@Gastron Thanks. Consider also adding different pytorch versions to your matrix too. You can see in my PR linked above a huge travis matrix for different archs, ubuntus, python versions, and pytorch versions

PR #969, opened 2021-09-03
Adding missing extra dependencies in recipe/UrbanSound8k

ZiRaymondal:
I've just added pandas & tensorboard which was missing in UrbanSound8k extra_dependencies.txt


mravanelli:
Thank you @ZiRaymondal!

PR #975, opened 2021-09-08
SpeechBrain EEG

ddavidebb:
This is the first commit for a motor imagery decoding recipe based on a benchmark motor imagery dataset (BNCI2014001).
README and other minors are missing.


recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
+import numpy as np
```
mravanelli, line 1: Please, add a short docstring that explain what we are doing here

recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
...
+    return standardized
+
+
+def nth(iterable, n, default=None):
+    """This function returns cross-validation train and test indices of the i-th fold."""
+    return next(islice(iterable, n, None), default)
+
+
+class WithinSession(object):
+    """Within Session """
```
mravanelli, line 25: Here you should improve the docstring with a better description of what we are doing + description  of the arguments (see other docstrings)

recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
...
+        if not isinstance(paradigm, BaseParadigm):
+            raise (ValueError("paradigm must be an Paradigm instance"))
+        self.paradigm = paradigm
+        self.hparams = hparams
+        np.random.seed(hparams["seed"])
+        self.skf = StratifiedKFold(n_splits=hparams["nfolds"])
+        self.iterator_tag = "within-session"
+
+    def prepare(self, dataset):
+        if not (isinstance(dataset, BaseDataset)):
```
mravanelli, line 37: Please,  add a docstring + description of input/output arguments

recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
...
+                        self.iterator_tag,
+                        str(subject).zfill(3),
+                        str(session).zfill(2),
+                        str(sel_fold).zfill(2),
+                    )
+                    yield (exp_dir, datasets)
+
+
+class CrossSession(object):
+    """Cross Session """
```
mravanelli, line 163: Improve docstring (like above)

recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
...
+                    self.hparams["output_folder"],
+                    self.iterator_tag,
+                    str(subject).zfill(3),
+                    str(sel_fold).zfill(2),
+                )
+                yield (exp_dir, datasets)
+
+
+class LeaveOneSubjectOut(object):
+    """Leave one subject out"""
```
mravanelli, line 302: Improve docstring (like above)

recipes/MOABB/MotorImagery_decoding/BNCI2014001/README.md
```diff
@@ -0,0 +1,37 @@
+# Motor imagery decoding from single EEG trials using BNCI2014001 dataset (MOABB benchmark)
+# Task description
+todo
```
mravanelli, line 3: Please, add the description

recipes/MOABB/MotorImagery_decoding/BNCI2014001/README.md
```diff
@@ -0,0 +1,37 @@
...
+# How to run
+Before running an experiment, make sure the extra-dependencies reported in the file `extra_requirements.txt` are installed in your environment.
+Note that this code requires mne==0.20.7.
+
+Perform training (including within-session, cross-session, leave-one-subject-out iterations) with: \
+\>>> python train.py train.yaml --data_folder '/path/to/MOABB_BNCI2014001'
+
+# Results
+
```
mravanelli, line 12: Please add a table with the current results

recipes/MOABB/MotorImagery_decoding/BNCI2014001/README.md
```diff
@@ -0,0 +1,37 @@
...
+  author={Mirco Ravanelli and Titouan Parcollet and Peter Plantinga and Aku Rouhe and Samuele Cornell and Loren Lugosch and Cem Subakan and Nauman Dawalatabad and Abdelwahab Heba and Jianyuan Zhong and Ju-Chieh Chou and Sung-Lin Yeh and Szu-Wei Fu and Chien-Feng Liao and Elena Rastorgueva and François Grondin and William Aris and Hwidong Na and Yan Gao and Renato De Mori and Yoshua Bengio},
+  year={2021},
+  eprint={2106.04624},
+  archivePrefix={arXiv},
+  primaryClass={eess.AS},
+  note={arXiv:2106.04624}
+}
+```
+
+# **Additional References**
```
mravanelli, line 36: Please add

recipes/MOABB/MotorImagery_decoding/BNCI2014001/extra_requirements.txt
```diff
@@ -0,0 +1,3 @@
+sklearn
+moabb
+mne==0.22.1
```
mravanelli, line 3: Is it really needed? I cannot find where we use mne.

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.py
```diff
@@ -0,0 +1,258 @@
+#!/usr/bin/python
+"""
+Recipe for training a compact CNN to decode motor imagery from single EEG trials.
```
mravanelli, line 3: It looks like the train.py is "architecture-agnostic". I suggest removing the CNN mention here and add it in the header of the yaml file instead.

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.py
```diff
@@ -0,0 +1,258 @@
...
+                init.constant_(mod.weight, 1)
+        if hasattr(mod, "bias"):
+            if mod.bias is not None:
+                init.constant_(mod.bias, 0)
+
+
+if __name__ == "__main__":
+    argv = sys.argv[1:]
+    # Temporary switching off deprecation warning from mne
+    import warnings  # noqa
```
mravanelli, line 213: Is that needed? I don't use mne used and I think we solved the warning issue in the previous recipe, right?

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.py
```diff
@@ -0,0 +1,258 @@
...
+    warnings.filterwarnings("ignore")
+    hparams_file, run_opts, overrides = sb.core.parse_arguments(argv)
+    with open(hparams_file) as fin:
+        hparams = load_hyperpyyaml(fin, overrides)
+
+    moabb_dataset = BNCI2014001()
+    # to run on a subset of subjects: moabb_dataset.subject_list = [1, 2, 3, 4]
+    moabb_dataset.download(path=hparams["data_folder"])
+    moabb_paradigm = MotorImagery(
+        n_classes=4,
```
mravanelli, line 224: The number of classes seems hardcoded it. Can we put in the yaml file?

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.yaml
```diff
@@ -0,0 +1,144 @@
+###################################################
```
mravanelli, line 1: Please, write the header. Here is the place where I would mention the EEGNET

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.yaml
```diff
@@ -0,0 +1,144 @@
...
+# Author
+# ------
+# Davide Borra, 2021
+###################################################
+seed: 2602
+__set_seed: !apply:torch.manual_seed [!ref <seed>]
+
+# DIRECTORIES
+data_folder: !PLACEHOLDER  #'/path/to/MOABB_BNCI2014001'
+output_folder: '/home/prometheus/Documents/results/MOABB_BNCI2014001' #!ref results/MOABB_BNCI2014001/<seed>
```
mravanelli, line 12: I think this should be `output_folder: results/MOABB_BNCI2014001/<seed>`. Otherwise, it is hardcoded.

recipes/MOABB/MotorImagery_decoding/BNCI2014001/train.yaml
```diff
@@ -0,0 +1,144 @@
...
+f2: 17  # length of temporal filters in the temporal separable conv. layer
+fp0: 4  # temporal pool size and stride of the first pooling layer
+fp1: 8  # temporal pool size and stride of the second pooling layer
+fptot: !ref <fp0> * <fp1>  # overall pool size
+max_norm_1: 1.  # kernel max-norm constaint of the spatial depthwise conv. layer
+max_norm_out: 0.25  # kernel max-norm constaint of the dense layer
+p_drop: 0.5  # dropout rate
+padding_mode: 'constant'
+
+layer0: !new:speechbrain.nnet.CNN.Conv2d
```
mravanelli, line 62: An alternative here would be to define the model in lobes (e.g. speechbrain/lobes/EEGNET.py). However, defining it every time doesn't look that bad to me in the end.

mravanelli (changes requested):

mravanelli:
Thank you @ddavidebb , great job! This looks already in a good shape.
Beyond the punctual comments done above, let me share a couple of additional suggestions:
1. It looks like` train.py` is dataset-specific (BNCI2014001). However, I think the code should be easily made dataset-independent. For instance, instead of importing the library at the beginning, you can import it right after reading the hparams file. You can add a variable like in the yaml file (e.g., `dataset="moabb.datasets.BNCI2014001"`) and import the library from this string using i`mportlib` like:

```
moabb_dataset= importlib.import_module(hparams["dataset"])
moabb_dataset = moabb_dataset()
```

e.g, https://www.devdungeon.com/content/import-python-module-string-name

If we are able to make the `train.py` dataset independent we can share it across all the datasets that we would like to address. Not sure about that, but maybe we can even create a single `train.py` for all the MOABB tasks (i.e., not only for Motor imagery detection). What do you think? 

2. The `MOABB_dataio_iterators.py` looks something in common to all the MOABB datasets. One possible approach here (already used in other recipes for the data preparation part)  could be to put the script in `recipes/MOABB/ `and add a symboling link in recipes/`MOABB/MOABB/MotorImagery_decoding/BNCI2014001/`

recipes/MOABB/MotorImagery_decoding/BNCI2014001/MOABB_dataio_iterators.py
```diff
@@ -0,0 +1,409 @@
...
+
+def standardize(x, m, s):
+    """This function standardizes input EEG signals (x) using mean value (m)
+    and standard deviation (s)."""
+    demeaned = x - m.reshape((1, m.shape[0], 1))
+    standardized = demeaned / (1e-14 + s.reshape((1, s.shape[0], 1)))
+    return standardized
+
+
+def nth(iterable, n, default=None):
```
fpaissan, line 19: I think this is used on the StratifiedKFold splits...
don't we want to split the data such that we don't have the same sessions of the same subject on train and test?
I think there's a data leakage problem if we do like this. In fact, you already "know" your subject from the train set.
Maybe this is implemented somewhere else and I am just missing it!

recipes/MOABB/train.py
```diff
@@ -131,6 +139,28 @@ def on_stage_end(self, stage, stage_loss, epoch=None):
                     test_stats=last_eval_stats,
                 )
 
+    def check_if_best(
```
ddavidebb, line 149: Here you save the best model for at least one of the specified keys (which are the default ones following your call at line 121 of train.py). Once you have completed the training and you perform the evaluation, the loaded model is not the best model for the 'test_keys' specified in the yaml file. Should we change the call of this method (specifying 'keys=self.hparams.test_keys') or remove the 'test_keys' parameter in the yaml file?

fpaissan: seems like the first option was the initial idea when the test_keys was added in the yaml file

mravanelli:
I'm aware of it and I already fixed it locally. I will push the fix today.

mravanelli:
I'm almost done with the code refactor! Here's a list of the remaining tasks for each team member:

@mravanelli:
1.  - [ ] Complete a full hparam tuning and carefully verify the results to ensure they match the outcomes achieved with the previous code.
2.  - [ ] Implement recipe tests.
3.  - [ ] Address the annoying warnings, such as those prompting users to run pip install braindecode.
4.  - [ ] Share all output folders on Dropbox and provide the necessary links for easy access by collaborators.

@ddavidebb :
1.  - [ ] Investigate and fix the leave one subject out functionality, as it is currently broken.

2.  - [ ] In the utils/data_interators.py file, provide comprehensive descriptions for the arguments used in the prepare_data functions. This will ensure clarity and make it easier for users to understand and utilize the provided hparams.
3.  - [ ] Address the issue with the MSEEGNet, which is currently broken. Refer to the failing docstring example for guidance on identifying and rectifying the problem.
4.  - [ ] If Bioformer and PodNet are integrated, add detailed docstrings that include usage examples for these components.

@fpaissan :
1.  - [ ] Enhance the docstrings in the parse_results.py file. Provide concise yet informative descriptions for both input and output arguments. Expand beyond specifying just the data type to offer a clear understanding of the purpose and relevance of each argument.

mravanelli (approved):

mravanelli:
I'm now going to merge the content of this PR in develop and open a new one in speechbrain/benchmarks with the EEG recipes.

PR #977, opened 2021-09-09
fix for single speaker case

xx205:
In most cases, the first eigenvalue is a constant (zero).
However, leaving this eigenvalue out will make the number of speakers not less than 2.

### DIFF
--- a/speechbrain/processing/diarization.py
+++ b/speechbrain/processing/diarization.py
@@ -909,7 +909,7 @@ class Spec_Clust_unorm:
         if k_oracle is not None:
             num_of_spk = k_oracle
         else:
-            lambda_gap_list = self.getEigenGaps(lambdas[1 : self.max_num_spkrs])
+            lambda_gap_list = self.getEigenGaps(lambdas[0 : self.max_num_spkrs])
 
             num_of_spk = (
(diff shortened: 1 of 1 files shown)


mravanelli:
@nauman-daw, could you please check it?

nauman-daw:
Hi @xx205 ,
Thanks for the PR.  
Actually, we have tried this earlier. If I remember it correctly, in our experiments with AMI datasets, we observe that sometimes the eigen gap between the first (nearly 0 eigenvalue) and the second eigenvalue exceeds other eigen gaps, leading to collapsing in one cluster. Hence to fix this, we remove the first eigenvalue and kept speaker diarization for two or more speakers. Hope I was clear in my explanation. 
Please let me know if you work on any other dataset with 1 speaker and if you found this change to work well in your case. We would be happy to take this forward for your dataset.

Gastron:
In this case, would it make sense to add an argument to allow an estimate of 1 speaker? We'd keep the two or more speakers estimate as the default. Could we even make it a tunable extra threshold / ratio parameter for the first eigen gap (so the first eigen gap has to be x-amount / x-times larger than all others to accept the estimate of just having one speaker)?

Issue #994, opened 2021-09-16
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb2 in position 6: invalid start byte


Jeremiah0425:
I try to train the model on my dataset，There is an error in this code snippet.

 sb.create_experiment_directory(
        experiment_directory=hparams["output_folder"],
        hyperparams_to_save=hparams_file,
        overrides=overrides,
    )

Gastron:
Is your data in UTF-8 or some other encoding? Have you found the text that results in this error?

Jeremiah0425:
> Is your data in UTF-8 or some other encoding? Have you found the file that results in this error?


Thank you for your prompt response, but I think it is not the reason for the dataset.Is it the reason I run the script?

![1631846326](https://user-images.githubusercontent.com/51948524/133715499-7a2bb7db-cc77-42da-bc28-a4b36cb023b9.jpg)

Gastron:
Oh that is interesting, thanks for the wider context, I will fix this.

Jeremiah0425:
> 哦，这很有趣，感谢更广泛的背景，我会解决这个问题。

Ok thanks.I am using the Chinese dataset。http://166.111.134.19:7777/data/thchs30/README.html

Gastron:
Hey @Jeremiah0425 a fix got merged to the develop branch, but would be nice if you can verify if it worked?

PR #996, opened 2021-09-17
Fix Issue #994 /  Fix run_shell for non UTF-8 encoded output

Gastron:
This should fix issue #994.


mravanelli:
Thank you for the fix @Gastron!

PR #997, opened 2021-09-17
Skip CTC-segmentation test if module not available

Gastron:
This simply skips the tests if the extra dependency is not installed.


aheba:
@Gastron,
cool, why we don't just use importorskip() function ?

other thing, I see no use of the CTC_SEG_AVAIL, is it used somewhere ?

Gastron:
Thanks @aheba, looks cleaner now.

TParcollet:
LGTM. Thanks.

PR #1006, opened 2021-09-24
Add the groups parameter to ECAPA_TDNN

goexle:
Add the groups parameter to ECAPA_TDNN. This enables to use the groups parameter in the TDNN conv1d blocks.

Using the groups parameter helps to reduce the number of parameters.

My own experiment showed that the total number of trainable parameters of the speaker verification (training) model can be reduced from 22.2M to 11.8M (by reducing the number of parameters in the ECAPA_TDNN model itself from 20.8M to 10.4M).
The accuracy stayed the same.

These are the parameters which I used:
```

embedding_model: !new:speechbrain.lobes.models.ECAPA_TDNN.ECAPA_TDNN
    input_size: !ref <n_mels>
    channels: [1032, 1032, 1032, 1032, 3096]
    kernel_sizes: [5, 3, 3, 3, 1]
    dilations: [1, 2, 3, 4, 1]
    groups: [1, 3, 3, 3, 3]
```
this is the output of the train_log.txt
```
epoch: 1, lr: 7.46e-04 - train loss: 5.49 - valid loss: 1.52, valid ErrorRate: 4.47e-02     
epoch: 2, lr: 5.07e-04 - train loss: 1.93 - valid loss: 9.08e-01, valid ErrorRate: 2.59e-02     
epoch: 3, lr: 2.39e-04 - train loss: 1.94 - valid loss: 7.75e-01, valid ErrorRate: 2.30e-02     
```


mravanelli:
Thank you @goexle! As for the performance, more than the accuracy on the speaker-id part (which is almost perfect in all the cases) it makes sense to compare the models in the speaker verification task.

goexle:
> Thank you @goexle! As for the performance, more than the accuracy on the speaker-id part (which is almost perfect in all the cases) it makes sense to compare the models in the speaker verification task.

@mravanelli thanks for your positive reply!
Yes, makes sense to me. I assume that I could take the output of the script `speaker_verification_cosine.py` to compare the models? Or do you have something else in mind to compare the models in the speaker verification task?

goexle:
I run the script `speaker_verification_cosine.py` to compute the EER and minDCF.
I got the following result for the ECAPA_TDNN model **with grouped convolution**:
EER(%)=0.765733
minDCF=0.083124

I got the following results for the standard, pretrained model (loaded from [hugging face](https://huggingface.co/speechbrain/spkrec-ecapa-voxceleb)):
EER(%)=0.670017   
minDCF=0.084820    

Given that, the ECAPA_TDNN model with grouped convolution seems to perform a bit worse (although I don't know if it is significant). 

@mravanelli if you prefer, I could revert the changes in the config file `recipes/VoxCeleb/SpeakerRec/hparams/train_ecapa_tdnn.yaml ` s.t. the default model would not be modified. However, I think it is worth adding the group parameter in general so that it can be used.

mravanelli:
I think we can give the users the possibility to use grouped convolution,
as you suggested in this PR. However, by default we can keep the standard
convolution in the recipe because for now it seems to give slightly better
results. We can thus do what you proposed in the previous message.

goexle:
@mravanelli ok, I updated the PR accordingly

PR #1016, opened 2021-09-28
fix the SpecAugment algorithm

NeuroForLunch:
x.shape will never equal 4. x.dim() is required.


Gastron:
Thanks, Looks Good To Me.

PR #1030, opened 2021-10-07
Neural SI-SNR Estimator for the REAL-M dataset

ycemsubakan:
There is still some things to clean-up but I am starting the PR..

### DIFF
--- a/recipes/REAL-M/sisnr-estimation/hparams/pool_sisnrestimator.yaml
+++ b/recipes/REAL-M/sisnr-estimation/hparams/pool_sisnrestimator.yaml
@@ -35,7 +35,7 @@ valid_data: !ref <save_folder>/libri2mix_dev.csv
 test_data: !ref <save_folder>/libri2mix_test.csv
 
 # the root data folder for the whamr dataset
-# the rest of the path is figured out according to sample_rate
+# the rest of the path is figured out automatically according to sample_rate
 # an example path is given below
 whamr_data_folder: /yourpath/whamr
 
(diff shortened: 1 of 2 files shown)


ycemsubakan:
> Isn't better to do a whole new recipe for this ?

Right, we can start a new recipe for REAL-M.

mravanelli:
Hi @ycemsubakan ,
thank you for this PR. I took a look at the code, and I have the following comments:

1. The README file should be more detailed and in the standard format (e.g, see https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/seq2seq/README.md).
It should describe the content of the recipe, how to run it (under a section called "how to run"), report the results, training time, PreTrained Model + Easy-Inference,
reference papers + information about speechbrain (i.e, citing speechbrain + About SpeechBrain). You should also add some information about where to download LibriMix, how to use/prepare WHAMR, and the set of rirs to use!


2. As for the pool system, it looks like few things are hardcoded in the trainer (https://github.com/speechbrain/speechbrain/blob/bb2cb8b84feda58f4871da4fc2ede30132814951/recipes/REAL-M/sisnr-estimation/train.py#L686). I think it is better to put all the models on HuggingFace and fetch the models from there. In the yaml, you can define some objects like (SepFormer, ConvTasnet, DualPathRNN) and a list like "separation_models = [SepFormer, ConvTasnet, DualPathRNN]". In the trainer, you can sample randomly from this list


3. https://huggingface.co/speechbrain/REAL-M-sisnr-estimator => I don't see the inference example here (which is very important for the users that would like to use the estimator as a black box). Moreover, it looks like this repo is very big because it also contains the pool of separated models. My suggestion is to split this into two repositories: one for inference only (the main one) and a support one that contains the pool of separated models.


4. `whamr_dynamic_mixing.py` => Is this file the same as the one in the WHAM/WHAMR recipes? If so, maybe we can add a link here to save code.


_in the yaml file_:
5. b`ase_folder_dm_whamr: /network/tmp1/subakany/wsj0-processed/si_tr_s` => this looks hard coded


_In train.py:_

6. `# added part `=> Maybe you can add here another comment like "computing the SISNR" and later also add some comments when you estimate it


7. Loss => It looks like the loss depends on the SNR compression (L1 with compression, L2 without). This might be weird. I suggest just take the option that works better and clearly justify this choice with a comment.


8. """This script computes the SDR and SI-SNR metrics and saves
        them into a csv file""" => I think this is an outdated comment (SDR is not computed)

PR #1035, opened 2021-10-12
open_rir -> open_rirs

shahad-mahmud:
fixed the typo


Gastron:
First, I guess we should at least standardize this across the whole repo.
But secondly, now that you point this out, I realize that openrir(s) seems to be a name that we have given this resource. Or is there some outside reason to call it that? Looks like it's this resource: [https://www.openslr.org/28/](https://www.openslr.org/28/), which just has the title: Room Impulse Response and Noise Database. It seems to be a collection from many different sources and it is the data used in [this paper](https://danielpovey.com/files/2017_icassp_reverberation.pdf).

mravanelli:
Right, this seems very minor though.

Issue #987, opened 2021-09-12
Idea: make `Pretrained` a subclass of nn.Module?


PiotrDabkowski:
Currently the `Pretrained` class is a parent of all the pretrained modules, however inference is a bit tricky and non-standard as it does not inherit from `torch.nn.Module`.

For example, to move the model to a selected device you would need to specify run_opts during instantiation. Same for the eval mode. Finally, reusing these modules as part of a larger PyTorch model is again not trivial. It is possible to work around all these difficulties, but why?

Simply inheriting from torch.nn.Module would solve these inconveniences:
`Pretrained (torch.nn.Module):`

is there a specific reason why this was not done?

TParcollet:
Hi that is a good question for our very own @Gastron.

Gastron:
Good idea, I see how inheriting from torch.nn.Module could make `Pretrained` classes a little easier to use as part of a larger system. I don't remember if we had any insurmountable reason not to do that in the first place.

There is at least one issue: what to put in the `forward` method of each `Pretrained` subclass interface, since basically all subclasses have multiple methods that do something of interest. I think for all of these we can find a good choice, though. Basically `forward` should work on batch-tensors and apply the model as completely as possible (e.g. encode-decode instead of simply encode).

TParcollet:
I agree, we would also need to do some refactoring on the HuggingFace side. @PiotrDabkowski would you please provide an example with one of the interface? We don't have the resources to put someone on that right now, but we should consider it.

Gastron:
What I like here is that we should be able to make this change without breaking anything existing. I wonder if there could be future `Pretrained` applications where a `torch.nn.Module` interface is not suitable?

TParcollet:
I don't think so, maybe we should set this up for the next SB version. @Gastron what do you think ?

Gastron:
Let's try it out, sure

PR #1042, opened 2021-10-18
Pretrained subclasses nn.Module now

Gastron:
Addresses issue #987.

I've tried to make some sane choice for the forward() method of each subclass. Some models cannot pass gradients through their decode procedure in any meaningful way - is that unexpected for a forward() method? Should we only pick methods that pass gradients?


aheba:
Hi @Gastron , 
let me fork this PR and work on an the proposed solution using import lib and external injection of the forward function

PR #1054, opened 2021-10-22
Support PyTorch 1.10

mravanelli:
I did some tests and everything seems to work. There is only the need to fix some warnings.


mravanelli:
I fixed some warnings. I think we can merge it and test it better the multi-gpu training and the torchaudio transducer loss separately.

PR #1058, opened 2021-10-22
Generic Pretrained interface

Gastron:
This allows adding any custom Pretrained subclass as part of a pretrained model to be fetched.

Like the `.from_hparams()` method on Pretrained, this will download a hyperparams file and look for a pretrainer. In addition, this will download a python module (by default `interface.py`), import it locally, and create an instance of a class (by default a class with the name `Custom`) defined inside that module. The class is passed `modules` and `hparams` as keyword arguments (so this would theoretically work for Brain subclasses too...) and **kwargs of course.


mravanelli:
After an internal discussion, we decided to go ahead with this PR. This interface will give one more option to our users, especially those that do not want to modify our interfaces directly and have a local inference file. We need to improve the documentation and add a section about this in the inference tutorial (@Gastron).

aheba:
Hello @Gastron , @mravanelli,

Do we have a ToDo List ? I can work on this PR

mravanelli:
According to me, what is missing here is:
- improve the documentation with  docstrings + examples
- Write a section in the inference tutorial to let users know about this new option
- Solve the small conflict
As soon as these steps are done, I think we can merge it (and hopefully have it in the upcoming new version of speechbrain)

Gastron:
There's one more question: Should we allow downloading modeling code as well? We can run interface code, but running custom model code might sometimes be also necessary and sometimes it could be a preferrable option in conforming to the existing interfaces.

mravanelli:
As we discussed in the meeting, we might want to manage that with some
specific flag.

Gastron:
I mean a slightly different thing, but I can make an example.

mravanelli:
@Gastron is fine with that. @aheba could you please review it?

aheba:
Thanks you @Gastron for this PR,

Here is an example of the use of external Predictor using the Generic proposed solution with `foreign_class` function.

aheba (approved):

TParcollet:
@aheba did you update the tutorial before merging? I think that this is mandatory given the change.

aheba:
Let me add the tutorial later this week

PR #1062, opened 2021-10-24
wav2vec 2.0 pretraining

TParcollet:
This PR introduces pertaining for wav2vec 2.0 based on the HF interface. The example is given with CommonVoice but works with everything. It has been compared to fairseq -> Better results on the same setup.


mravanelli:
Thank you @TParcollet  for this add. @aheba did you have the chance to try it at least for few epochs?
1. @TParcollet, one thing I noticed is that the code is put in `recipes/self-supervised-learning`. Normally, the current standard is "recipes/<datasetname>" (which is not ideal in some cases, but it is what we did so far)
2. Isn't it better to create it in a recipe using the official "librilight" dataset rather than using commonvoice?  We might not be able to train it now, but it could be useful for the users.
3. I think it could be a good idea to couple this recipe with the webdataset dataloader that @Gastron set up. Also the voxlingua107 recipe (https://github.com/speechbrain/speechbrain/blob/develop/recipes/VoxLingua107/lang_id/hparams/train_ecapa.yaml) that we recently merged is using it. This might speed up everything quite a bit.

TParcollet:
Well for 1. I think it makes no sense to have a SSL recipe under a dataset name, simply because it can't be found as there isn't a dataset that is specific to SSL (which is not true for TTS for instance). 2. We could, but I can't train it as it would cost a LOT. The good thing with CommonVoice is that it already is correctly sized (2K hours, which is enough for a robust SSL model already) and easily downloadable. 3. I can try, yeah. It will take some time though.

mravanelli:
As for point 1, isn't LibriLight the standard dataset used in most
cases for self-supervised learning? It looks like they have different
chunks like small, medium, large.

TParcollet:
Not really :-/ Peoples are using CV, LibriLight, LibriSpeech, VoxPopulli or even LeBenchmark now for French :p I guess we could add it under LibriSpeech ? This would be quick and relatively cheap to train (48H with 20 GPUs) ...

aheba:
Hey @mravanelli , @TParcollet ,

I have 6 GPUs, let me try it for half epochs..

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -161,3 +166,85 @@ def extract_features(self, wav):
...
+    ):
+        super().__init__()
+
+        self.mask_prob = mask_prob
+        self.mask_length = mask_length
+
+        # Download the config of the model from HuggingFace.
+        config = Wav2Vec2Config.from_pretrained(source, cache_dir=save_path)
+        config.gradient_checkpointing = (
+            False  # This cause errors with DDP if True.
```
aheba, line 217: @TParcollet , is it possible to add `run_on_main()` for that part for gradient checkpointing... ?

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
```diff
@@ -0,0 +1,294 @@
...
+Authors
+ * Titouan Parcollet 2021
+ * Yan Gao 2021
+"""
+
+logger = logging.getLogger(__name__)
+
+
+# Define training procedure
+class W2VBrain(sb.core.Brain):
```
aheba, line 43: W2V2 instead?

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
```diff
@@ -0,0 +1,294 @@
...
+    def compute_forward(self, batch, stage):
+        """Forward computations from the waveform batches to the w2v2 loss."""
+
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        wavs, wav_lens = wavs.to(self.device), wav_lens.to(self.device)
+
+        # Normalize wav signal
+        wavs = self.modules.normalize(wavs, wav_lens)
```
aheba, line 52: is this useful ? as far as I remember, I have the same results with or without...

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -161,3 +166,85 @@ def extract_features(self, wav):
...
+        Arguments
+        ---------
+        wav : torch.Tensor (signal)
+            A batch of audio signals to transform to features.
+        """
+        batch_size, raw_sequence_length = wav.shape
+
+        # We must compute the masking before forward. This is used by the loss.
+        sequence_length = self.model._get_feat_extract_output_lengths(
+            raw_sequence_length
```
aheba, line 238: why not using the relative wav_lenght from our data loader ?, this assume that you will use the zero padding ?

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -161,3 +166,85 @@ def extract_features(self, wav):
...
+            False  # This cause errors with DDP if True.
+        )
+        config.output_hidden_states = True  # We want the hidden states as well!
+
+        self.model = Wav2Vec2ForPreTraining(config)
+        self.model.train()
+
+        # We check if inputs need to be normalized w.r.t pretrained wav2vec2
+
+    def forward(self, wav, mask_device):
```
aheba, line 226: why you don't use `length` information...

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
```diff
@@ -0,0 +1,294 @@
...
+
+        # Forward on w2v2 and take the loss.
+        out, mask = self.modules.wav2vec2(wavs, mask_device=self.device)
```
aheba, line 55: why you are not using the length ?
`self.model._get_feat_extract_output_lengths` need this information no ?

speechbrain/lobes/models/huggingface_wav2vec.py
TParcollet: Nope. This is within HuggingFace ... :-/ We must disable it.

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
TParcollet: It is to make it more robust with various input signals. Also the guys from facebook reported that normalising always help. Better to keep it then.

speechbrain/lobes/models/huggingface_wav2vec.py
TParcollet: We decided to use this because we are entirely relying on HF for the loss computation. Using their methods is therefore better to ensure compliance with the next transformer package version. Otherwise, we would have used our own wav_len.

speechbrain/lobes/models/huggingface_wav2vec.py
TParcollet: What do you mean?

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
TParcollet: I am not sure to understand properly.

TParcollet:
@aheba could you add the fix to DDP (local_rank) to this PR ?

aheba:
let me add it as well

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
```diff
@@ -0,0 +1,294 @@
...
+        """Gets called at the end of an epoch."""
+        # Compute/store important stats
+        stage_stats = {"loss": stage_loss}
+        if stage == sb.Stage.TRAIN:
+            self.train_stats = stage_stats
+        else:
+            stage_stats["acc"] = sum(self.acc_metric) / len(self.acc_metric)
+
+        # Perform end-of-iteration things, like annealing, logging, etc.
+        if stage == sb.Stage.VALID and sb.utils.distributed.if_main_process():
```
aheba, line 119: @TParcollet , do you have some bugs here ?
in our `core.py` we have already have the call of `run_on_main`:
https://github.com/speechbrain/speechbrain/blob/68a8b5979dc5d6a885489d9a0428682d5a2ac012/speechbrain/core.py#L1078
let me do a fix of the use of `local_rank` and `rank` in our distributed solution

TParcollet: I guess we can remove it, dunno why this is added here.

aheba:
as discussed offline with @TParcollet , the DDP fix will be in separate PR.

aheba:
@TParcollet , @yan-gao-GY , Thanks you for this PR, 
This work will be very valuable for Spbrain community.

@mravanelli, LGTM.

aheba:
Next work in other PR:
- [ ] add optional arg in `run_on_main`/`if_main_process` to specify local and global run_on_main according to the user. etc (each server need to do data preparation, vs at the end of epoch, only one process do evaluation)...
- [ ] support `torchrun` for next integration with pytorch version

TParcollet:
@aheba looking forward for the first one as I need it to do multinode training properly ahah

TParcollet:
@mravanelli it should be ready for a last try. The model is running on 32 GPUs right now and everything looks fine. Note: I will certainly make an announcement once this is merged because it may interest quite a lot of peoples. @Gastron If you have some spare time, I would love to get some help in building the Webdataset shard script for that ...

TParcollet:
@mravanelli ready. I will just add a wav2vec 2.0 (CV EN) Base pretrained model to show that it works (to the YAML).

TParcollet:
The only thing remaining (that we will maybe handle later with the engineer) is to have a way of saving our pretrained model in a HuggingFace compatible way ... For now it's impossible.

TParcollet:
I started a downstream training to verify that everything is ok on that side as well. Please just try to run the pretraining if you wish.

TParcollet:
@mravanelli now it works. It reverted back to transformers 4.11, you can try the PR again but from my side it worked :-) Things to try : train few epochs and then try to use it with a fine-tuning recipe. I will modify the tutorial to explain the steps as you must create a HF like folder (i.e. just add the config.json and feature_extractor.json with your wav2vec.ckpt)

TParcollet:
Finally fixed. I have to say that this implementation sample negatives from the whole sentence and not only from the masked indices. The latter behaviour was a design choice made by Facebook that is clearly problematic for short sentences that might not have enough mask indices to sample negatives from.

mravanelli:
I tried it on my side and everything is running. Of course, I didn't double checked the big run with 32 GPUs but we can definitely trust you and assume everything is working now. I think we can merge! Thank you @TParcollet

PR #1078, opened 2021-10-28
DE ASR model (seq2seq) from CommonVoice7.0

vitas-ai:
Hey,

we contributed a model to HuggingFace:
https://huggingface.co/speechbrain/asr-crdnn-commonvoice-de

Please find the depending PR here :)

The train_de.yaml is essentially the same as the one for French; substituted the `language` parameter only.

Best wishes,
Andreas


mravanelli:
@vitas-ai, any plan to try the wav2vect + CTC version as well with voxpopuli? All the other languages that we support so far have it and we would need it for German as well.

vitas-ai:
We do not plan to do fine tuning right now, since we like to explore the whole solution space that SpeechBrain offers. VoxPopuli appears critical for a database to use. While the provided ASR model (DE) can clearly be improved, it reached comparable results to e.g. FR already. Hope, that's ok.

mravanelli:
Hi @vitas-ai,
as mentioned privately, the fine-tuning can be done with wav2vec (if you prefer it over voxpopuli). It should work well anyway. For all the languages done so far we have both the fine-tuned and the training from scratch versions (see [here](https://github.com/speechbrain/speechbrain/tree/develop/recipes/CommonVoice/ASR/seq2seq/hparams)). Not having the fine-tuned version for German will introduce some asymmetry difficult to explain for the users. So, I prefer to not merge this PR for now. Instead, I will leave it open to see if there is someone in the community that would like to complete that part.

mravanelli:
Thank you @vitas-ai! @TParcollet will train the fine-tuned model and we will merge it once that part is ready.

mravanelli:
I started one experiment with wav2vec2 fine-tuned with commonvoice-de. It will take some time, but I will add it here when ready.

mravanelli:
Update: I did 15 epochs with wav2vec2 fine-tuned with commonvoice-de. I'm using the same parameters as English, French, and Italian, but the network doesn't converge apparently (WER stays at 50%). My suggestion is to merge this PR and figure out what we can do in another one. Thank you @vitas-ai!

PR #1081, opened 2021-10-29
Remove torch.no_grad in feature lobes

Gastron:
See discussion in [https://speechbrain.discourse.group/t/use-pretrained-speechbrain-model-as-loss/493/8](https://speechbrain.discourse.group/t/use-pretrained-speechbrain-model-as-loss/493/8). Also discussed this offline with Mirco and Titouan.

This change should theoretically be small, but in practice it affects a huge number of recipes, etc. Should maybe do some timing tests as well as run some basic recipe just to make sure there's no regression.

### DIFF
--- a/speechbrain/lobes/features.py
+++ b/speechbrain/lobes/features.py
@@ -131,20 +131,15 @@ class Fbank(torch.nn.Module):
         wav : tensor
             A batch of audio signals to transform to features.
         """
-        with torch.no_grad():
-
-            STFT = self.compute_STFT(wav)
-            mag = spectral_magnitude(STFT)
-            fbanks = self.compute_fbanks(mag)
-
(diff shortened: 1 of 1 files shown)


mravanelli:
Maybe we can take this opportunity to also replace `.no_grad` in `core.py `(https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/core.py#L1062) with `torch.inference_mode()`. What do you think? I'm not sure if this feature was available in pytorch 1.8 though.

mravanelli:
I tested in on TIMIT and it looks like speed and memory is the same. As for the `torch.inference` part, I think this should be eventually changed in another PR. This could be a big change that will break compatibility with pytorch <=1.9.

PR #1091, opened 2021-11-03
Add default checkpoint hooks for Torch Schedulers

Gastron:
Torch schedulers have their own class structure so they need separate default hooks.


Chaanks:
I tried to save and load a torch scheduler, it works well

mravanelli:
Thank you @Gastron for the fix and @Chaanks for the feedback.

PR #1094, opened 2021-11-04
change torchaudio minimal version for compatibility

underdogliu:


mravanelli:
Thank you @underdogliu!

PR #1102, opened 2021-11-08
Pass file as string

### LINKED ISSUE #1101: TypeError: Invalid file: WindowsPath('example_mandarin.wav')
When I tried to run code in WINDOWS
```python
asr_model = EncoderDecoderASR.from_hparams(source="speechbrain/asr-wav2vec2-transformer-aishell",
                                               savedir="pretrained_models/asr-wav2vec2-transformer-aishell")
res = asr_model.transcribe_file('speechbrain/asr-wav2vec2-transformer-aishell/example_mandarin.wav')
```
I got the error:
image

I looked at some of the code located in the file "\speechbrain\pretrained\interfaces.py":
image

and I found that the function `torchaudio.load(path, channels_first=False)` requires a `str`. However, the function `fetch`  in 
 this file return an object with the type  `pathlib.WindowsPath` , so actually I use it instead of an `str`. And this raised the error.

I wonder if this is an incompatibility issue on Windows? I have avoided this problem in some other way, but I sincerely hope it can be resolved.

Gastron:
Fixes #1101

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -148,7 +148,7 @@ class Pretrained(torch.nn.Module):
         """
         source, fl = split_path(path)
         path = fetch(fl, source=source, savedir=savedir)
-        signal, sr = torchaudio.load(path, channels_first=False)
+        signal, sr = torchaudio.load(str(path), channels_first=False)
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you @Gastron!

PR #1110, opened 2021-11-10
Fix slow DDP with dilated Conv

TParcollet:
setting cudnn flags only for DDP is a bad idea, it slows down everything ... This PR removes this.

### DIFF
--- a/recipes/CommonLanguage/lang_id/train.py
+++ b/recipes/CommonLanguage/lang_id/train.py
@@ -2,6 +2,7 @@
 import os
 import sys
 import torch
+import time
 import logging
 import torchaudio
 import speechbrain as sb
@@ -70,12 +71,20 @@ class LID(sb.Brain):
         """
 
         # We first move the batch to the appropriate device.
+        start = time.time()
         batch = batch.to(self.device)
-
+        to_d = time.time()
         # Compute features, embeddings and output
         feats, lens = self.prepare_features(batch.sig, stage)
+        feats_t = time.time()
         embeddings = self.modules.embedding_model(feats)
+        emb = time.time()
         outputs = self.modules.classifier(embeddings)
+        out = time.time()
+        print("to_d " + str(to_d - start))
+        print("feats " + str(feats_t - start))
+        print("emb " + str(emb - start))
+        print("out " + str(out - start))
 
         return outputs, lens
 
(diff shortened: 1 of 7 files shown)


recipes/CommonLanguage/lang_id/train.py
```diff
@@ -296,13 +296,15 @@ def label_pipeline(language):
...
+    torch.backends.cudnn.benchmark = True
+
     # The `fit()` method iterates the training loop, calling the methods
     # necessary to update the parameters of the model. Since all objects
     # with changing state are managed by the Checkpointer, training can be
     # stopped at any point, and will be resumed on next call.
     lid_brain.fit(
         epoch_counter=lid_brain.hparams.epoch_counter,
-        train_set=datasets["train"],
+        train_set=datasets["dev"],
```
mravanelli, line 307: This looks like a temporary chance to train everything faster . It should be datasets["train"], right?

recipes/CommonLanguage/lang_id/train.py
```diff
@@ -296,13 +296,15 @@ def label_pipeline(language):
         checkpointer=hparams["checkpointer"],
     )
 
+    torch.backends.cudnn.benchmark = True
```
mravanelli, line 299: Is that needed only here or in all the recipes using dilated convolution?

recipes/self-supervised-learning/wav2vec2/train.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python3
+
```
mravanelli, line 2: Weird, it look like this PR includes also files from another one

speechbrain/core.py
```diff
@@ -850,12 +847,19 @@ def fit_batch(self, batch):
                 self.scaler.step(self.optimizer)
             self.scaler.update()
         else:
+            # start = time.time()
```
mravanelli, line 850: remove comments

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -19,6 +19,11 @@
     from transformers import Wav2Vec2Model, HubertModel
```
mravanelli, line 19: This is a file from another PR

speechbrain/nnet/CNN.py
```diff
@@ -410,8 +412,9 @@ def forward(self, x):
                 "Padding must be 'same', 'valid' or 'causal'. Got "
                 + self.padding
             )
-
+        # start = time.time()
```
mravanelli, line 415: remove comments

PR #1111, opened 2021-11-10
remove CUDNN flags to fix DDP with dilated conv

TParcollet:
setting cudnn flags only for DDP is a bad idea, it slows down everything ... This PR removes this.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -546,9 +546,6 @@ class Brain:
                         "Only the main process is alive, "
                         "all other subprocess were killed."
                     )
-            # force the models to start and remain synchronized
-            torch.backends.cudnn.deterministic = True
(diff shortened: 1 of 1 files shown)


TParcollet:
From the doc: "Disabling the benchmarking feature with torch.backends.cudnn.benchmark = False causes cuDNN to deterministically select an algorithm, possibly at the cost of reduced performance.

However, if you do not need reproducibility across multiple executions of your application, then performance might improve if the benchmarking feature is enabled with torch.backends.cudnn.benchmark = True." I mean this stuff is terrible... @mravanelli

mravanelli:
I took a look at the voxceleb recipes and the flags were actually set correctly in train.py (`torch.backends.cudnn.benchmark = True`). The issue wa that they were changed later in the ddp calling function. Now, I confirm everything is fine.

PR #1115, opened 2021-11-11
Fix the padding issue

weiwei-ww:
Based on the codes for padding (e.g. `batch_pad_right()` & `pad_right_to()` in `speechbrain.utils.data_utils`), the padding is actually applied to the first dimension, not the last dimension, so I have updated the codes for raising the exception.

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -19,6 +19,7 @@ import torch.nn.functional as F
 from torch.nn.parallel import DistributedDataParallel as DDP
 from speechbrain.utils.data_utils import split_path
 from speechbrain.utils.distributed import run_on_main
+from speechbrain.utils.callchains import lengths_arg_exists
 
 
 class Pretrained(torch.nn.Module):
@@ -2071,7 +2072,7 @@ class SpectralMaskEnhancement(Pretrained):
     ... )
     >>> noisy, fs = torchaudio.load("samples/audio_samples/example_noisy.wav")
     >>> # Channel dimension is interpreted as batch dimension here
-    >>> enhanced = enhancer.enhance_batch(noisy, lengths=torch.tensor([1.]))
+    >>> enhanced = enhancer.enhance_batch(noisy)
     """
 
     HPARAMS_NEEDED = ["compute_stft", "spectral_magnitude", "resynth"]
@@ -2089,7 +2090,7 @@ class SpectralMaskEnhancement(Pretrained):
         feats = self.hparams.spectral_magnitude(feats)
         return torch.log1p(feats)
 
-    def enhance_batch(self, noisy, lengths):
+    def enhance_batch(self, noisy, lengths=None):
         """Enhance a batch of noisy waveforms.
 
         Arguments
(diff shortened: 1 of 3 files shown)


Gastron:
Good catch, I think the check is looking at the wrong dimensions indeed. However, also the tests need to be changed, and the comments around the code block. @popcornell can you comment on this?

popcornell:
If i run: 
`sb.utils.data_utils.batch_pad_right([torch.rand((2, 16)), torch.rand((1, 16))])`
> `OSError: Tensors should have same dimensions except for last one`. 

If I run `sb.utils.data_utils.batch_pad_right([torch.rand((2, 16)), torch.rand((2, 100))])`
The padding is fine but the indexes are not. I get tensor([1., 1.]). 

So the error is somewhere else. 
I think it is here: 

https://github.com/speechbrain/speechbrain/blob/ce3405a3f171cc89ba89aab3fb37b1827f0e6abd/speechbrain/utils/data_utils.py#L418

It should be  valid.append(valid_percent[-1]) because is reversed.

popcornell:
@weiwei-ww will you handle this or should I handle it ?

weiwei-ww:
> If i run: `sb.utils.data_utils.batch_pad_right([torch.rand((2, 16)), torch.rand((1, 16))])`
> 
> > `OSError: Tensors should have same dimensions except for last one`.
> 
> If I run `sb.utils.data_utils.batch_pad_right([torch.rand((2, 16)), torch.rand((2, 100))])` The padding is fine but the indexes are not. I get tensor([1., 1.]).
> 
> So the error is somewhere else. I think it is here:
> 
> https://github.com/speechbrain/speechbrain/blob/ce3405a3f171cc89ba89aab3fb37b1827f0e6abd/speechbrain/utils/data_utils.py#L418
> 
> It should be valid.append(valid_percent[-1]) because is reversed.

I encountered this issue too when fixing this bug. However, the tricky thing is: in the `pad_right_to` function, the list `valid_vals` is actually not reversed, as it is indexed by the variable `j`, not `i`. Therefore, there is no need to modify the code `valid.append(valid_percent[0])`.

weiwei-ww:
> However, the tricky thing is: in the `pad_right_to` function, the list `valid_vals` is actually not reversed, as it is indexed by the variable `j`, not `i`.

The codes are here:

https://github.com/speechbrain/speechbrain/blob/ce3405a3f171cc89ba89aab3fb37b1827f0e6abd/speechbrain/utils/data_utils.py#L343-L358

The `pads` is reversed, as required by `torch.nn.functional.pad`. However, the list `valid_vals` is not reversed.

popcornell:
https://github.com/speechbrain/speechbrain/pull/1118 II think that returning the last pad indx percentage should be enough

popcornell:
I updated the tests check this PR @weiwei-ww  https://github.com/weiwei-ww/speechbrain/pull/1

weiwei-ww:
@popcornell I have accepted this PR which updates the tests:  https://github.com/weiwei-ww/speechbrain/pull/1

popcornell:
Thanks, test are passing LGTM.

popcornell:
we can merge in my opinion

mravanelli:
Thank you @weiwei-ww and @popcornell. I took a look and I think we can merge it.

PR #1118, opened 2021-11-12
return padding percentage for last dim instead of first

popcornell:
@weiwei-ww this should be fine I think.

Also: added better test. we check the lens too now

### DIFF
--- a/speechbrain/utils/data_utils.py
+++ b/speechbrain/utils/data_utils.py
@@ -415,7 +415,7 @@ def batch_pad_right(tensors: list, mode="constant", value=0):
             t, max_shape, mode=mode, value=value
         )
         batched.append(padded)
-        valid.append(valid_percent[0])
+        valid.append(valid_percent[-1])
 
     batched = torch.stack(batched)
 
--- a/tests/unittests/test_batching.py
+++ b/tests/unittests/test_batching.py
@@ -1,5 +1,6 @@
 import pytest
 import torch
+import numpy as np
 
 
 def test_batch_pad_right_to():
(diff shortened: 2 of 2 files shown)


popcornell:
related to https://github.com/speechbrain/speechbrain/pull/1115

popcornell:
drop_freq fails but is unrelated to this fix. Is it just badluck ?

weiwei-ww:
Yes. This is another solution to fix the padding issue, if it is intended to preform padding only to the **_last_** dimension. However, is it more reasonable to perform padding to the **_first_** dimension?  According to the [documentation page](https://speechbrain.readthedocs.io/en/latest/experiment.html), all the tensors should have a format of `tensor=(batch, time_steps, channels[optional])`, so the `time_steps` is the first dimension after `batch`. Therefore, I assume that performing padding to the `time_steps` dimension (first dimension) should make more sense?

popcornell:
IDK, torchaudio for example has [channels, samples] format. 
Plus many recipes (e.g. separation) do have channels first format (which is also faster BTW for convnets). 

The pipeline actually was originally intended to pad arbitrary dimensions: 
https://github.com/speechbrain/speechbrain/blob/ce3405a3f171cc89ba89aab3fb37b1827f0e6abd/speechbrain/utils/data_utils.py#L397

I would leave it as it is right now because this ambiguity will actually go away when we will bring back multidimensional padding. Also we may need to change code in other parts if we are going to pad the first dimension only. WDYT @Gastron ?

Gastron:
What are the usecases for real multi-dimensional padding (e.g. different audio lengths, with different numbers of channels)?

weiwei-ww:
> What are the usecases for real multi-dimensional padding (e.g. different audio lengths, with different numbers of channels)?

I don't see any cases where multi-dimensional padding is needed. However, in my program, I need to perform padding to the computed features, with the shape of (T x C) (time x channels). Therefore, in my case, I hope to perform padding only to the first dimension, not the last dimension.

Gastron:
Yep, I can see padding along the time-dimension as a common thing. I suggest we make that the default operation for this function (basically maintains backwards compatibility but makes it work for a larger set of cases as well). In the future we can implement a multi-dimensional padding function separately. E.G. the PaddedBatch class takes a padding function as argument already.

popcornell:
> What are the use cases for real multi-dimensional padding (e.g. different audio lengths, with different numbers of channels)?

Yes, for multi-mic applications you can have a varying number of microphones, in order to batch examples together you need to pad channels as well. 
Other applications can be images or videos if in the future someone wants to go multimodal. 

Leaving this for future work, now the main question is: do we want to pad the last dimension or the first one ? 
Personally, I prefer the last one because it is coherent with `torchaudio`. 
Also the very `torch.nn.functional.pad` by design encourages padding on the last dimension, and padding on the other dims is generally a little bit slower (talking about few milliseconds but still). 

On the other hand I think most recipes in SpeechBrain have time x channels as default. 
The source separation ones are an exception. 

> However, in my program, I need to perform padding to the computed features, with the shape of (T x C) (time x channels). 
Therefore, in my case, I hope to perform padding only to the first dimension, not the last dimension.

This can always be performed when reading or computing the features via a data-io pipeline and transposing. 

Again, it is a matter of what format people use most: time x channels or channels x time. 
I use channels x times generally, but again, I mostly do source separation. 

Asking also @mravanelli.

Gastron:
In SpeechBrain the convention is `(batch, time[, channels])`

popcornell:
@weiwei-ww I can either perform a PR to your branch or you can move your changes to this branch (and this PR). Your choice, let me know.

weiwei-ww:
> @weiwei-ww I can either perform a PR to your branch or you can move your changes to this branch (and this PR). Your choice, let me know.

If possible, please go ahead to perform the PR to my branch https://github.com/speechbrain/speechbrain/pull/1115. Thanks

PR #1122, opened 2021-11-14
Hyperparameter Fitting Support

flexthink:
In this pull request, an extensible convenience wrapper called `hpopt` was added to facilitate hyperparameter fitting using Oríon or other tools. It makes it possible to easily modify an existing recipe to be used both for training and for hyperparameter fitting with a single file, using command-line switches.

For information on how to use hyperparameter fitting, refer to this tutorial:
https://colab.research.google.com/drive/1b-5EOjZC7M9RvfWZ0Pq0HMV0KmQKu730


mravanelli:
Thank you for this PR @flexthink! I took a look at the tutorial and the code. Everything is clear to me. As mentioned privately, I would consider the following changes:
1- Add examples to docstrings in hpfit.py
2. I think the term "hyperparmeter optimization" is more common that "hyperparametr fitting". I would suggest changing it

Gastron:
Hey, thanks for extensive work and taking the time to make a clear tutorial and all! This is probably a silly question, but is an Orion wrapper really needed, can it not be used directly by just adding `orion.client.report_objective(value)` at the end of a script?

flexthink:
> Hey, thanks for extensive work and taking the time to make a clear tutorial and all! This is probably a silly question, but is an Orion wrapper really needed, can it not be used directly by just adding `orion.client.report_objective(value)` at the end of a script?

The main added value of the wrapper is abstraction, convenience and the handling of optional dependencies.

The wrapper does end up calling `orion.client.report_objective(value)`, **if** Orion is available

It is possible to use Oríon directly; however:
- Importing anything from `orion` will introduce a **hard** dependency on it (i.e. you won't be able to run the recipe without Orion being installed. 
- Currently, `brain.fit(...)` does not return anything. The recipe would also need to take care of collecting the stats to report and passing them on. The wrapper provides a convenience function to "remember" the last result.
- The wrapper makes it possible to use the same recipe both for training and for hyperparameter optimization without duplicating the same conditional logic
- The wrapper also provides a mechanism to use an extra .yaml file for any parameters that are different during hyperparameter optimization. A common use case is to turn off checkpointing or to override the default output path to avoid clashes. 
- The wrapper makes it possible to add support for hyperparameter optimizers other than Oríon in the future

mravanelli:
Thank you @flexthink, I'm happy with this PR now. I think we can merge it!

PR #973, opened 2021-09-06
Add KsponSpeech recipe

kimdwkimdw:
> **KsponSpeech** is a large-scale spontaneous speech corpus of Korean. This corpus contains 965.2 hours of general
> open-domain dialog utterances with transcriptions. More information can be found
> at https://www.mdpi.com/2076-3417/10/19/6936

We added and tested Conformer ASR model recipe with KsponSpeech. /w @ddwkim

Pretrained model and our results are uploaded at [HuggingFace](https://huggingface.co/ddwkim/asr-conformer-transformerlm-ksponspeech)

More detail on [README.md](https://github.com/kimdwkimdw/speechbrain/tree/recipe/ksponspeech/recipes/KsponSpeech)

You can download full corpus via https://aihub.or.kr/aidata/105/download (A pledge is required)

Before PR, we checked conventions from https://speechbrain.readthedocs.io/en/latest/contributing.html

PR #1134, opened 2021-11-17
Simplify .github workflow using pre-commit

kimdwkimdw:
During working on PR https://github.com/speechbrain/speechbrain/pull/973,

I've found `pre-commit workflow on this repository(.github/workflows/pythonapp.yml)` can be simplified.

https://speechbrain.readthedocs.io/en/latest/contributing.html#pre-commit

I've checked all test can pass using this workflow. https://github.com/kimdwkimdw/speechbrain/pull/1

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,16 @@
+name: SpeechBrain pre-commit
+
+on: # yamllint disable-line rule:truthy
+  push:
+    branches: [main, develop]
+  pull_request:
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+      - uses: pre-commit/action@v2.0.3
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -1,4 +1,4 @@
-# SpeechBrain lint and unit tests to ease continuous integration
+# SpeechBrain  unit tests to ease continuous integration
 # NOTE: Caching these offers no speedup
 name: SpeechBrain toolkit CI
 
@@ -9,16 +9,6 @@ on:    # yamllint disable-line rule:truthy
     pull_request:
 
 jobs:
-    linters:
-        name: Linters
-        runs-on: ubuntu-latest
-        steps:
-            - uses: actions/checkout@v2
-            - uses: actions/setup-python@v2
-              with:
-                  python-version: '3.8'
-            - uses: pre-commit/action@v2.0.3
-
     tests:
         name: Tests
         runs-on: ubuntu-latest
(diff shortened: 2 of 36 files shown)


Gastron:
Thanks for the idea, but I think we want to maintain the pre-commit hooks and the GitHub CI tests separately, as we've realized we want to run different things in them.

kimdwkimdw:
> Thanks for the idea, but I think we want to maintain the pre-commit hooks and the GitHub CI tests separately, as we've realized we want to run different things in them.

Something like this?

Gastron:
Sorry, I mean that I think we don't want to run pre-commit on GitHub at all. But perhaps there's something I am not seeing, what are the benefits of this change?

kimdwkimdw:
[From the SpeechBrain document](https://speechbrain.readthedocs.io/en/latest/contributing.html#the-git-pre-commit-hooks), 
> The hooks are only run on the files you git added to the commit. This is in contrast to the CI pipeline, which always tests everything

I thought that SpeechBrain maintainers want to focus on the architecture of a change not a trivial style with [.pre-commit-config.yaml](https://github.com/speechbrain/speechbrain/blob/develop/.pre-commit-config.yaml)

## Changes

So, I just try to simplify these linters from 

```yaml
- name: Lint dependencies
  run: |
      python -m pip install --upgrade pip
      pip install -r lint-requirements.txt
- name: Run linters
  run: |
      flake8 . --count --show-source --statistics
      black --check --diff .
      yamllint .
```

to

```yaml
 - uses: pre-commit/action@v2.0.3
```

However, it makes a lots of changes in Markdown files. If you don't want this change, add `exclude: .md$'` on `.pre-commit-config.yaml`


## the benefits

- Shorter code is usually good.
  - Your linter is working almost same as pre-commit except `*.md`
  - Bridge the gap between .pre-commit0-config.yml and workflow(https://github.com/speechbrain/speechbrain/blob/develop/.pre-commit-config.yaml#L5-L6), config specify removing trailing-whitespace in `text`
- [requirements.txt](https://github.com/speechbrain/speechbrain/blob/develop/requirements.txt) usually don't need linter dependancy.

Gastron (changes requested):
I've thought about this and I want to go ahead with this change. The decoupling of the pre-commit configuration and the GitHub actions config is really only important for the pre-push tests, but we're not touching those here. And this adds the line-ending format thing to GitHub, plus keeps future changes in sync.

I don't want to remove the lint-requirements.txt file, however. I think those are still a valuable part of the development environment. In particular, pytest should hopefully be used by all developers locally, but the other libs can be helpful as well. So could you just add those back?

kimdwkimdw:
Apology for too disruptive changes in commits. I've rollback removal of lint-requirements.txt

Gastron (commented):
Thanks a lot! Now there's some file that seems to conflict with the pre-commit CI test, which failed. Maybe you need to merge the latest develop branch and fix some line-endings or something.

kimdwkimdw:
Right, I pull latest develop and apply all on it.

Gastron (approved):

Gastron:
Ok great, thanks for this contribution! Looks good to me.

Issue #1133, opened 2021-11-17
Define run_opts in hparam.yaml


larsrgit:
It would be useful if there was the possibility to define run_opts in the hyperparameter yaml file.

The split of some parameters into run_opts and hparams is not always intuitive (e.g in my opinion the max_grad_norm should be a hyperparamter since the choice of when to clip gradients is strongly dependent of the model, optimizer, loss and the learning rate). By enabling the user to set the run_opts in the hyperparameter yaml file, the user would not need to handle run-opts and hyperparameters differently.

Currently I just overwrite all run-opts that exist in hparams after loading the yaml. But this disables the ability to change those parameters with overwrites in the command line. (run-opts are not returned as overrides by sb.parse_arguments(), so when overriding the run-opts if they exist in hparams, I can't distinguish if those run-opts where set by default or in the command line).

Gastron:
Actually, run_opts can be set via hparams as well. Looks like this isn't documented.

https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/core.py#L437-L453

larsrgit:
Oh this is nice. But I think this does not work for run_opts that have a default in speech_brain.parse_arguments, because they will always be in the run_opts passed to the brain initialization. 
Maybe there should not be two places in the code with defaults for the run_opts?

Gastron:
Right, we should maybe remove the defaults from the SpeechBrain arguments parsing, and keep the defaults in Brain. We could also have it read the command line if no args are given. I'll make a PR quick.

larsrgit:
Okay, I can understand the reasoning for having two places for defaults.
I still think the priority should always be command line > hparam-file > defaults. The way it is now, for run_opts with a default in parse_args the priority is command line > defaults and the hparam file is ignored.

But it's no big problem for me since there currently is not many defaults in parse-args I want to override in the hparam yaml, so I will just handle them manually.

Gastron:
I agree, that would be the sane priority order, but it's not possible with the current implementation. 

We could maybe load the YAML in the argument parsing though, hmm, gotta think about this.

larsrgit:
If somehow argument parsing would return the information about which run-opts where set in the command line and which by defaults, we could overwrite those which were set by default by the hparams loaded from the yaml.

Gastron:
[...]

PR #1140, opened 2021-11-18
Auto mix prec + noprogressbar override via hyperparams

Gastron:
Auto mixed precision can be controlled via hyperparams.yaml if it has no default in the argument parsing (implementation clarification: `None` values get deleted). Also allow the same for noprogressbar; that seems to do no harm I think.

I noticed a few recipes actually implemented (hacked) this feature already, so it seemed to make sense to change it.

See issue #1133

### DIFF
--- a/recipes/LibriMix/separation/train.py
+++ b/recipes/LibriMix/separation/train.py
@@ -120,7 +120,7 @@ class Separation(sb.Brain):
         if self.hparams.num_spks == 3:
             targets.append(batch.s3_sig)
 
-        if self.hparams.auto_mix_prec:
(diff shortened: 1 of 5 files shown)


mravanelli:
@ycemsubakan, this looks like a small change in the separation part. Could you please quickly make sure everything is fine on your side?

ycemsubakan:
Yes, I will take a look today @mravanelli.

ycemsubakan:
I have checked all of the recipes, and everything seems to work fine.

mravanelli:
Thank you! Feel free to merge

PR #1142, opened 2021-11-19
JITable sepformer

Squire-tomsk:
Sepformer is JITable now via torch.jit.script(<object of speechbrain.pretrained.interfaces.SepformerSeparation>)

Repro:

```
import torch
from hyperpyyaml import load_hyperpyyaml
from speechbrain.pretrained.interfaces import SepformerSeparation

hparams = load_hyperpyyaml(
    open(f'{SPEECBRAIN_PATH}/recipes/LibriMix/separation/hparams/sepformer-libri2mix.yaml'),
    overrides='data_folder: wsj0-mix/2speakers'
)

sepformer = SepformerSeparation(modules=hparams['modules'], hparams=dict(hparams))

class JITableSepformer(torch.nn.Module):
    def __init__(self, modules):
        super().__init__()
        self.encoder = modules['encoder']
        self.decoder = modules['decoder']
        self.masknet = modules['masknet']
    
    def forward(self, mix):
        mix_w = self.encoder(mix)
        est_mask = self.masknet(mix_w)
        mix_w = torch.stack([mix_w] * 2)
        sep_h = mix_w * est_mask

        # Decoding
        est_source = torch.cat([self.decoder(sep_h[i]).unsqueeze(-1) for i in range(2)], dim=-1)

        # T changed after conv1d in encoder, fix it here
        T_origin = mix.size(1)
        T_est = est_source.size(1)
        if T_origin > T_est:
            est_source = torch.nn.functional.pad(est_source, (0, 0, 0, T_origin - T_est))
        else:
            est_source = est_source[:, :T_origin, :]
        return est_source

sepformer_jit = torch.jit.script(JITableSepformer(hparams['modules']))
mix = torch.rand(size=(2, 8000), dtype=torch.float32)
with torch.no_grad():
    assert(torch.allclose(sepformer.separate_batch(mix), sepformer_jit(mix)))
```

### DIFF
--- a/recipes/LibriMix/separation/hparams/sepformer-libri3mix.yaml
+++ b/recipes/LibriMix/separation/hparams/sepformer-libri3mix.yaml
@@ -102,7 +102,7 @@ SBtfintra: !new:speechbrain.lobes.models.dual_path.SBTransformerBlock
     d_model: !ref <out_channels>
     nhead: 8
     d_ffn: 1024
-    dropout: 0
+    dropout: 0.0
     use_positional_encoding: True
     norm_before: True
 
@@ -111,7 +111,7 @@ SBtfinter: !new:speechbrain.lobes.models.dual_path.SBTransformerBlock
     d_model: !ref <out_channels>
     nhead: 8
     d_ffn: 1024
-    dropout: 0
+    dropout: 0.0
     use_positional_encoding: True
     norm_before: True
 
--- a/speechbrain/lobes/models/dual_path.py
+++ b/speechbrain/lobes/models/dual_path.py
@@ -253,12 +253,18 @@ class Decoder(nn.ConvTranspose1d):
                        N = number of filters
                        L = time points
         """
+        # Ensure scriptability
+        # if x.dim() not in [2, 3]:
+        #     raise RuntimeError(
+        #         "{} accept 3/4D tensor as input".format(self.__name__)
+        #     )
+        # x = super().forward(x if x.dim() == 3 else torch.unsqueeze(x, 1))
+        # is not scriptable
 
-        if x.dim() not in [2, 3]:
-            raise RuntimeError(
-                "{} accept 3/4D tensor as input".format(self.__name__)
-            )
-        x = super().forward(x if x.dim() == 3 else torch.unsqueeze(x, 1))
+        if x.dim() != 3:
+            x = torch.unsqueeze(x, 1)
+        output_padding = self._output_padding(x, None, self.stride, self.padding, self.kernel_size, self.dilation)
+        x = F.conv_transpose1d(x, self.weight, self.bias, self.stride, self.padding, output_padding, self.groups, self.dilation)
 
         if torch.squeeze(x).dim() == 1:
             x = torch.squeeze(x, dim=1)
@@ -267,7 +273,7 @@ class Decoder(nn.ConvTranspose1d):
         return x
 
 
-class IdentityBlock:
+class IdentityBlock(nn.Module):
     """This block is used when we want to have identity transformation within the Dual_path block.
 
     Example
@@ -280,7 +286,7 @@ class IdentityBlock:
     def _init__(self, **kwargs):
         pass
 
-    def __call__(self, x):
+    def forward(self, x):
         return x
 
 
@@ -804,8 +810,10 @@ class Dual_Computation_Block(nn.Module):
             self.intra_norm = select_norm(norm, out_channels, 4)
             self.inter_norm = select_norm(norm, out_channels, 4)
 
+        self.intra_linear = IdentityBlock()
+        self.inter_linear = IdentityBlock()
         # Linear
-        if linear_layer_after_inter_intra:
+        if self.linear_layer_after_inter_intra:
             if isinstance(intra_mdl, SBRNNBlock):
                 self.intra_linear = Linear(
                     out_channels, input_size=2 * intra_mdl.mdl.rnn.hidden_size
@@ -951,6 +959,7 @@ class Dual_Path_Model(nn.Module):
         self.conv1d = nn.Conv1d(in_channels, out_channels, 1, bias=False)
         self.use_global_pos_enc = use_global_pos_enc
 
+        self.pos_enc = IdentityBlock()
         if self.use_global_pos_enc:
             self.pos_enc = PositionalEncoding(max_length)
 
@@ -1017,8 +1026,8 @@ class Dual_Path_Model(nn.Module):
         x, gap = self._Segmentation(x, self.K)
 
         # [B, N, K, S]
-        for i in range(self.num_layers):
-            x = self.dual_mdl[i](x)
+        for layer in self.dual_mdl:
+            x = layer(x)
         x = self.prelu(x)
 
         # [B, N*spks, K, S]
@@ -1045,15 +1054,13 @@ class Dual_Path_Model(nn.Module):
 
         return x
 
-    def _padding(self, input, K):
+    def _padding(self, input, K: int):
         """Padding the audio times.
 
         Arguments
         ---------
         K : int
             Chunks of length.
-        P : int
-            Hop size.
         input : torch.Tensor
             Tensor of size [B, N, L].
             where, B = Batchsize,
@@ -1064,15 +1071,15 @@ class Dual_Path_Model(nn.Module):
         P = K // 2
         gap = K - (P + L % K) % K
         if gap > 0:
-            pad = torch.Tensor(torch.zeros(B, N, gap)).type(input.type())
+            pad = torch.zeros(B, N, gap, dtype=input.dtype, device=input.device)
             input = torch.cat([input, pad], dim=2)
 
-        _pad = torch.Tensor(torch.zeros(B, N, P)).type(input.type())
+        _pad = torch.zeros(B, N, P, dtype=input.dtype, device=input.device)
         input = torch.cat([_pad, input, _pad], dim=2)
 
         return input, gap
 
-    def _Segmentation(self, input, K):
+    def _Segmentation(self, input, K: int):
         """The segmentation stage splits
 
         Arguments
@@ -1104,7 +1111,7 @@ class Dual_Path_Model(nn.Module):
 
         return input.contiguous(), gap
 
-    def _over_add(self, input, gap):
+    def _over_add(self, input, gap: int):
         """Merge the sequence with the overlap-and-add method.
 
         Arguments
--- a/speechbrain/lobes/models/dual_path.py
+++ b/speechbrain/lobes/models/dual_path.py
@@ -272,24 +272,6 @@ class Decoder(nn.ConvTranspose1d):
             x = torch.squeeze(x)
         return x
 
-
-class IdentityBlock(nn.Module):
-    """This block is used when we want to have identity transformation within the Dual_path block.
-
-    Example
-    -------
-    >>> x = torch.randn(10, 100)
-    >>> IB = IdentityBlock()
-    >>> xhat = IB(x)
-    """
-
-    def _init__(self, **kwargs):
-        pass
-
-    def forward(self, x):
-        return x
-
-
 class FastTransformerBlock(nn.Module):
     """This block is used to implement fast transformer models with efficient attention.
 
@@ -810,8 +792,8 @@ class Dual_Computation_Block(nn.Module):
             self.intra_norm = select_norm(norm, out_channels, 4)
             self.inter_norm = select_norm(norm, out_channels, 4)
 
-        self.intra_linear = IdentityBlock()
-        self.inter_linear = IdentityBlock()
+        self.intra_linear = nn.Identity()
+        self.inter_linear = nn.Identity()
         # Linear
         if self.linear_layer_after_inter_intra:
             if isinstance(intra_mdl, SBRNNBlock):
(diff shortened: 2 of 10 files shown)


speechbrain/lobes/models/dual_path.py
```diff
@@ -280,7 +286,7 @@ class IdentityBlock:
     def _init__(self, **kwargs):
```
vadimkantorov, line 286: Suggestion: derive IdentityBlock from `nn.Identity`, make `super().__init__()` call in `__init__`, otherwise the module object won't have needed fields, drop the forward method

Squire-tomsk: I replace `IdentityBlock` by `torch.nn.Identity`

mravanelli:
Thank you for this PR!
- @ycemsubakan, could you please make sure everything is fine? (maybe just run few epochs only).
- @Squire-tomsk, could you please write a test in `tests/unittests/` (e.g., [see here](https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_CNN.py#L37)) where we test the jitability of the sepformer? This will ensure that this property will be preserved in future versions. Also, linters test fails due to a tiny flake8 issue.

ycemsubakan:
Yes, I will take a look. But also it seems there is a formatting issue. @Squire-tomsk , are you using the precommit hooks?

Squire-tomsk:
> Yes, I will take a look. But also it seems there is a formatting issue. @Squire-tomsk , are you using the precommit hooks described here: https://speechbrain.readthedocs.io/en/latest/contributing.html ?

Actually no, but now I will

ycemsubakan:
Overall, I am not sure if all of the changes are necessary for Jitability. Also, these changes raises the concern that now the model will not be anymore backward compatible with the pretrained models on github. 

Could you try to verify that,
1) The recipes (WSJ0Mix, WHAMandWHAMR, LibriMix) are able to run?
2) The changes are backwards compatible with the pretrained models?

Squire-tomsk:
> Overall, I am not sure if all of the changes are necessary for Jitability. Also, these changes raises the concern that now the model will not be anymore backward compatible with the pretrained models on github.
> 
> Could you try to verify that,
> 
>     1. The recipes (WSJ0Mix, WHAMandWHAMR, LibriMix) are able to run?
> 
>     2. The changes are backwards compatible with the pretrained models?

Ok I will try 👌

speechbrain/lobes/models/dual_path.py
```diff
@@ -253,12 +253,29 @@ def forward(self, x):
...
+        #     )
+        # x = super().forward(x if x.dim() == 3 else torch.unsqueeze(x, 1))
+        # is not scriptable
+
+        if x.dim() != 3:
+            x = torch.unsqueeze(x, 1)
+        output_padding = self._output_padding(
+            x, None, self.stride, self.padding, self.kernel_size, self.dilation
+        )
+        x = F.conv_transpose1d(
```
mravanelli, line 269: It looks like this is adding convolution step that wasn't there before. Is that correct?

mravanelli:
Also,
it could be great to have in this PR only the minimal amount of changes needed to make the current version jitable.

speechbrain/lobes/models/dual_path.py
Squire-tomsk: It was there. Inside `super().forward(...)` call.

In this module I just have removed exeption raising and replace `super().forward(...)` call by copy of this code https://github.com/pytorch/pytorch/blob/36449ea93134574c2a22b87baad3de0bf8d64d42/torch/nn/modules/conv.py#L770 . Because both operations are not JITable.

Squire-tomsk:
> Also, it could be great to have in this PR only the minimal amount of changes needed to make the current version jitable.

I suppose it is minimal, because I trained Sepformer on code from develop branch and later make every change to fix concrete error.

recipes/LibriMix/separation/hparams/sepformer-libri3mix.yaml
```diff
@@ -111,7 +111,7 @@ SBtfinter: !new:speechbrain.lobes.models.dual_path.SBTransformerBlock
     d_model: !ref <out_channels>
     nhead: 8
     d_ffn: 1024
-    dropout: 0
+    dropout: 0.0
```
vadimkantorov, line 114: Does it break if you keep an integer here?

Squire-tomsk: Yep

ycemsubakan:
Hi @Squire-tomsk , we were wondering if you had the chance to test the backwards compatibility with the existing pretrained models, as well as whether the existing recipes would continue to work with these changes. I was wondering if you have any news on these? Thanks!

Squire-tomsk:
> Hi @Squire-tomsk , we were wondering if you had the chance to test the backwards compatibility with the existing pretrained models, as well as whether the existing recipes would continue to work with these changes. I was wondering if you have any news on these? Thanks!


Hi ) I am in progress now. I have already tested backwards compatibility for shared checkpoints (at least for https://huggingface.co/speechbrain/sepformer-wsj02mix and https://huggingface.co/speechbrain/sepformer-wsj03mix). It works. I will try to test training process at this weekend.

ycemsubakan:
Great, thanks for the update!

Squire-tomsk:
> Great, thanks for the update!

@ycemsubakan I have checked training process with and without this PR. I have used this harams https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriMix/separation/hparams/sepformer-libri2mix.yaml and train model during 25 epochs. Results are almost identical.

ycemsubakan:
> > Great, thanks for the update!
> 
> @ycemsubakan I have checked training process with and without this PR. I have used this harams https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriMix/separation/hparams/sepformer-libri2mix.yaml and train model during 25 epochs. Results are almost identical.

This sounds good. How about backwards compatibility with existing pretrained models on huggingface?

Squire-tomsk:
> > > Great, thanks for the update!
> > 
> > 
> > @ycemsubakan I have checked training process with and without this PR. I have used this harams https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriMix/separation/hparams/sepformer-libri2mix.yaml and train model during 25 epochs. Results are almost identical.
> 
> This sounds good. How about backwards compatibility with existing pretrained models on huggingface?

Fully compatible. Tested on sepformer-wsj02mix and sepformer-wsj03mix.

ycemsubakan:
> > > > Great, thanks for the update!
> > > 
> > > 
> > > @ycemsubakan I have checked training process with and without this PR. I have used this harams https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriMix/separation/hparams/sepformer-libri2mix.yaml and train model during 25 epochs. Results are almost identical.
> > 
> > 
> > This sounds good. How about backwards compatibility with existing pretrained models on huggingface?
> 
> Fully compatible. Tested on sepformer-wsj02mix and sepformer-wsj03mix.

Great, I will test these myself as well, and get back to you. Thanks for updates!

Squire-tomsk:
@ycemsubakan did you check my code?

ycemsubakan:
Hi @Squire-tomsk , I have been busy with several things so I couldn't yet. I will try to do it soon.

ycemsubakan:
I have written on the code also, but I don't understand why are you deleting the IdentityBlock class? It's just a helper class in case you want to skip inter or intra calculations.

Squire-tomsk:
> I have written on the code also, but I don't understand why are you deleting the IdentityBlock class? It's just a helper class in case you want to skip inter or intra calculations.

Because I need to initialize fields `intra_linear` and  `inter_linear` in `Dual_Computation_Block` class constructor. So if I initialize them by `IdentityBlock` I will get error:

```Module 'Dual_Computation_Block' has no attribute 'intra_linear' (This attribute exists on the Python module, but we failed to convert Python type: 'speechbrain.lobes.models.dual_path.IdentityBlock' to a TorchScript type. Only tensors and (possibly nested) tuples of tensors, lists, or dictsare supported as inputs or outputs of traced functions, but instead got value of type IdentityBlock.. Its type was inferred; try adding a type annotation for the attribute.):```

So I decide to use class `nn.Identity` with same functionality, but inherited from `nn.Module`. It fixes this error. Moreover I think that usage of standard PyTorch class is more idiomatic.

**Comment by Squire-tomsk** [comment]
@ycemsubakan up

**Comment by ycemsubakan** [comment]
> @ycemsubakan up

Hi, I did take a look into it, and we discussed this internally. As I have wrote before, this effects other recipes as well, and we need to be careful here.

**Comment by vadimkantorov** [comment]
> As I have wrote before, this effects other recipes as well, and we need to be careful here.

Hi @ycemsubakan ! Would you have any concrete advice on how to ensure the compat or concrete concerns? @Squire-tomsk's changes are minimal. It would be nice to figure out how to make the Sepformer jit-compatible in the core. Would you have any advice on your side @mravanelli?

**Comment by ycemsubakan** [comment]
> > As I have wrote before, this effects other recipes as well, and we need to be careful here.
> 
> Hi @ycemsubakan ! Would you have any concrete advice on how to ensure the compat or concrete concerns? @Squire-tomsk's changes are minimal, and he tested compat with existing checkpoints. It would be nice to figure out how to make the Sepformer TorchScript/JIT-compatible in the core. Would you have any advice on your side @mravanelli?
> 
> On Attention optionally returning weights, there is currently a PR in core PyTorch: [pytorch/pytorch#70055](https://github.com/pytorch/pytorch/pull/70055), but they end up always returning two tensors to not have a static tuple[Tensor, Tensor] return type. @Squire-tomsk would a Union return type work? It seems it might be supported: [pytorch/pytorch#67111](https://github.com/pytorch/pytorch/issues/67111). @mravanelli @stas00 What is core HuggingFace decision on attention returning Union[Tensor, tuple[Tensor, Tensor]] and it being TorchScript-incompatible? It seems that core PyTorch APIs would always return a tuple and potentially place some undefined tensors in the tuple if they are not needed.

Hi @vadimkantorov, as I wrote above we discussed this internally. We would need to test it properly ourselves before we can merge this.

**Comment by vadimkantorov** [comment]
> We would need to test it properly ourselves before we can merge this

Got it! Thanks for outlining the path forward. @Squire-tomsk and I were not sure if there were anything required on our side to help the testing or alternative refactorings.

About the API, in general the union return types are not very nicely handled by TorchScript, so refactoring them out early may be a good idea. Some workarounds: return a dictionary with optional elements or return a tuple with some components filled by UndefinedTensors.

**Comment by stas00** [comment]
> @stas00 What is core HuggingFace decision on attention returning Union[Tensor, tuple[Tensor, Tensor]] and it being TorchScript-incompatible? It seems that core PyTorch APIs would always return a tuple and potentially place some undefined tensors in the tuple if they are not needed.

Vadim, you seem to be asking as if this has already been discussed somewhere at HF Transformers? If it has been do you have the link? If it hasn't if needed please file an Issue and then it can be discussed?

**Comment by Squire-tomsk** [comment]
>  @Squire-tomsk would a Union return type work?

Yes, union actually works, but it breaks scripting process [here](https://github.com/speechbrain/speechbrain/blob/4aa8782f2062979d9f2344ebf1cea00ee5196bc6/speechbrain/lobes/models/transformer/Transformer.py#L327).

```
RuntimeError: 
Union[Tensor, Tuple[Tensor, Optional[Tensor]]] cannot be used as a tuple:
  File "/work/speechbrain/speechbrain/lobes/models/transformer/Transformer.py", line 329
            src1 = src
    
        output, self_attn = self.self_att(
                            ~~~~~~~~~~~~~~
            src1,
            ~~~~~
            src1,
            ~~~~~
            src1,
            ~~~~~
            attn_mask=src_mask,
            ~~~~~~~~~~~~~~~~~~~
            key_padding_mask=src_key_padding_mask,
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            pos_embs=pos_embs,
            ~~~~~~~~~~~~~~~~~ <--- HERE
        )
```

I can refactor to fix it, but I guess it will worse than the current solution.

**Comment by vadimkantorov** [comment]
@stas00 I was wondering whether HF uses `return_attention_weights=True` pattern that alters the return type of the attention module (which is problematic for the JIT and which is the main crux of this issue)

**Comment by stas00** [comment]
```
cd transformers-master
grep -Ir return_attention_weights src
```

so the answer to your question is No.

The flag that does exist is `return_attention_mask`

vadimkantorov:
Thank you! Yep, I meant the pattern where a flag modifies the return type of the method (from what I understand, core PyTorch and maybe fairseq too (don't remember for sure) are now going away from this pattern to facilitate scriptability).

stas00:
It might be tricky due to backward compatibility, and might have to go through a deprecation cycle, but as I mentioned earlier please do file an Issue on the Transformers side and then it'd be easier to think about it on the home grounds once the full context is given, as we can't make any decisions here. Please tag me and I will tag the right people to save you time figuring it out. Does that resonate? 

HF Transformers is definitely gradually improving things to be more JIT'able and friendly to other compilers so this is very important. The main hurdle is backward compatibility.

Adel-Moumen:
Hello @ycemsubakan, 

What is the status of your review please ? 

BTW, @Squire-tomsk, we merged in `develop` the branch `unstable-v0.6` which integrate most of our `unstable` features. You might need to update your PR so that the upstream branch is develop. 

Best, 
Adel

PR #1147, opened 2021-11-22
Aishell1Mix

huangzj421:
This branch adds a new task named Aishell1Mix to the recipes which is similar to the LibriMix but applied to the mandarin AISHELL-1 dataset. Hope to receive your reply. Much thanks.


ycemsubakan:
Thanks for the PR @huangzj421 . I will write a detailed review.

huangzj421:
> Thanks for the PR @huangzj421 . I will write a detailed review.

It's my pleasure.

ycemsubakan:
Hi @huangzj421 , Something we were wondering was that, it seems there are a lot of overlaps with the existing recipes. Especially `train.py` looks very similar. Do you think when possible, could you either import some of the classes from existing recipes, or use symbolic links?  Also, I think you could use the existing `prepare_data.py` for csv creation, and `dynamic_mixing.py` from other recipes. 

In general we think that it would be better to use existing code as much as possible. Otherwise when we want to make new changes, we would need to update a lot of files.

huangzj421:
> Hi @huangzj421 , Something we were wondering was that, it seems there are a lot of overlaps with the existing recipes. Especially `train.py` looks very similar. Do you think when possible, could you either import some of the classes from existing recipes, or use symbolic links? Also, I think you could use the existing `prepare_data.py` for csv creation, and `dynamic_mixing.py` from other recipes.
> 
> In general we think that it would be better to use existing code as much as possible. Otherwise when we want to make new changes, we would need to update a lot of files.

I see. The reason I created this script begins that I saw LibriMix and WSJmix recipes are very similar, just applied to different datasets. But the Mandarin dataset in speech separation is missing, so I created Aishellmix dataset according to the process of Librimix, which means the two scripts are also similar. As for the recipes, I just simply commented out the pre-training model and few more so that it could be trained on the dataset I created. 
As for what you said, I was a little confused. Did you mean that this aishellmix should be a subclass to other recipes like Librimix?

ycemsubakan:
> > Hi @huangzj421 , Something we were wondering was that, it seems there are a lot of overlaps with the existing recipes. Especially `train.py` looks very similar. Do you think when possible, could you either import some of the classes from existing recipes, or use symbolic links? Also, I think you could use the existing `prepare_data.py` for csv creation, and `dynamic_mixing.py` from other recipes.
> > In general we think that it would be better to use existing code as much as possible. Otherwise when we want to make new changes, we would need to update a lot of files.
> 
> Hi @subakan. The reason I created this script begins that I saw LibriMix and WSJmix recipes are very similar, just applied to different datasets. But the Mandarin dataset in speech separation is missing, so I created Aishellmix dataset according to the process of Librimix, which means the two scripts are also similar. As for the recipes, I just simply commented out the pre-training model and few more so that it could be trained on the dataset I created. As for what you said, I was a little confused. Did you mean that this aishellmix should be a subclass to other recipes like Librimix?

I think one way to go about it is to use the existing code for Librimix as much as possible. We can either do this by importing from Librimix, or by symbolic links if we want to use the exact same file. One other option would be to simply add AIshell options to LibriMix, but we think that having a separate recipe for AIshell would increase the visibility.

huangzj421:
> ix, or by symbolic links if we want to use the exact same file. One other option would be to sim

Okay, I got it. I will figure out how to handle it and commit. Thanks for your patience.

huangzj421:
> > > Hi @huangzj421 , Something we were wondering was that, it seems there are a lot of overlaps with the existing recipes. Especially `train.py` looks very similar. Do you think when possible, could you either import some of the classes from existing recipes, or use symbolic links? Also, I think you could use the existing `prepare_data.py` for csv creation, and `dynamic_mixing.py` from other recipes.
> > > In general we think that it would be better to use existing code as much as possible. Otherwise when we want to make new changes, we would need to update a lot of files.
> > 
> > 
> > Hi @subakan. The reason I created this script begins that I saw LibriMix and WSJmix recipes are very similar, just applied to different datasets. But the Mandarin dataset in speech separation is missing, so I created Aishellmix dataset according to the process of Librimix, which means the two scripts are also similar. As for the recipes, I just simply commented out the pre-training model and few more so that it could be trained on the dataset I created. As for what you said, I was a little confused. Did you mean that this aishellmix should be a subclass to other recipes like Librimix?
> 
> I think one way to go about it is to use the existing code for Librimix as much as possible. We can either do this by importing from Librimix, or by symbolic links if we want to use the exact same file. One other option would be to simply add AIshell options to LibriMix, but we think that having a separate recipe for AIshell would increase the visibility.

Hi @ycemsubakan, I have commited a new version. Is it okay?

ycemsubakan:
Hi @huangzj421 , thanks for the updates. From what I understand, the `train.py` script does not have support automatic dataset preparation. I think it would be much more easier for the users if we could automatically download the aishell dataset, and prepare it when a user executes `train.py`. Do you think this would be doable?

huangzj421:
> Hi @huangzj421 , thanks for the updates. From what I understand, the `train.py` script does not have support automatic dataset preparation. I think it would be much more easier for the users if we could automatically download the aishell dataset, and prepare it when a user executes `train.py` (if the dataset does not already exist). Do you think this would be doable?

Hi @ycemsubakan, yeah of course it is doable. But I hope Aishell1mix dataset could be applied separately and widely spread. Thus I suggest if it be generated in prepare_data.py or another new recipe and included in train.py?

ycemsubakan:
@huangzj421 , In `prepare_aishell1mix` function you are currently just creating the csv file right? I was wondering if you could integrate what [this script](https://github.com/huangzj421/Aishell1Mix/blob/main/generate_aishell1mix.sh)   does inside `train.py` ?

huangzj421:
> @huangzj421 , In `prepare_aishell1mix` function you are currently just creating the csv file right? I was wondering if you could integrate what [this script](https://github.com/huangzj421/Aishell1Mix/blob/main/generate_aishell1mix.sh) does inside `train.py` ?

Okay, I will add it into the prepare_aishell1mix fuction to make it generate automatically in train.py. Thank you.

ycemsubakan:
> > @huangzj421 , In `prepare_aishell1mix` function you are currently just creating the csv file right? I was wondering if you could integrate what [this script](https://github.com/huangzj421/Aishell1Mix/blob/main/generate_aishell1mix.sh) does inside `train.py` ?
> 
> Okay, I will add it into the prepare_aishell1mix fuction to make it generate automatically in train.py. Thank you.

Thanks. Also note that we should check if the dataset already exists or not, in order to save time if the dataset already exists in the specified `--data_folder`.

huangzj421:
> > > @huangzj421 , In `prepare_aishell1mix` function you are currently just creating the csv file right? I was wondering if you could integrate what [this script](https://github.com/huangzj421/Aishell1Mix/blob/main/generate_aishell1mix.sh) does inside `train.py` ?
> > 
> > 
> > Okay, I will add it into the prepare_aishell1mix fuction to make it generate automatically in train.py. Thank you.
> 
> Thanks. Also note that we should check if the dataset already exists or not, in order to save time if the dataset already exists in the specified `--data_folder`.

Hi @ycemsubakan , I have integrated it into prepare_data and modified some other files. Thanks.

ycemsubakan:
Thank you @huangzj421. I will take a look tomorrow.

ycemsubakan:
Thanks for the updates @huangzj421 . Here are few comments about data downloading:  

- It's great to have an automatic data preparation option. I however don't think at the current version we are creating the folder for `--data_folder` the specified folder does not exist. Like, what is the behavior if the path specified in `--data_folder` does not exist? Do we create the folder?
- For me `train.py` just hangs at the message `Download Aishell1 into --data_folder`, and because I do not receive any feedback I don't know what's going on as a user. So I think during the download of aishell data I think you can give feedback to the user, so that we see an estimated finish time for the download. For this, you can use the fetch function here: https://github.com/speechbrain/speechbrain/blob/9df49e17c018c9617dfd3161a7db50a75086603e/speechbrain/pretrained/fetching.py#L26

huangzj421:
> * speechbrain/recipes/REAL-M/sisnr-estimation/train.py

For the first one, the system will exit if the specified folder does not exist because the folder needs large space.
For the second one, there is no feedback with the fetch function so I use a simple callback in urllib.urlretrieve fuction to download with percent feedback. Thanks.

ycemsubakan:
> > * speechbrain/recipes/REAL-M/sisnr-estimation/train.py
> 
> For the first one, the system will exit if the specified folder does not exist because the folder needs large space and I prefer users to specify. For the second one, there is no feedback with the fetch function so I use a simple callback in urllib.urlretrieve fuction to download with percent feedback. Thanks.

I am trying to run the script, but for some reason the download is very slow from the provided links (I am seeing 200 kb/s speeds, and I tested both at home and on the server). Do you think you might be able to put these files on a dropbox link (or an alternative) so that the download is faster?

huangzj421:
> > > * speechbrain/recipes/REAL-M/sisnr-estimation/train.py
> > 
> > 
> > For the first one, the system will exit if the specified folder does not exist because the folder needs large space and I prefer users to specify. For the second one, there is no feedback with the fetch function so I use a simple callback in urllib.urlretrieve fuction to download with percent feedback. Thanks.
> 
> I am trying to run the script, but for some reason the download is very slow from the provided links (I am seeing 200 kb/s speeds, and I tested both at home and on the server). Do you think you might be able to put these files on a dropbox link (or an alternative) so that the download is faster?

I am sorry that we don't use dropbox here and I don't have a plus or pro account. [Aishell1](https://www.openslr.org/33) provides many download links and the used one is a mirror of CN. Could you please try other mirrors to download please?

huangzj421:
> > > * speechbrain/recipes/REAL-M/sisnr-estimation/train.py
> > 
> > 
> > For the first one, the system will exit if the specified folder does not exist because the folder needs large space and I prefer users to specify. For the second one, there is no feedback with the fetch function so I use a simple callback in urllib.urlretrieve fuction to download with percent feedback. Thanks.
> 
> I am trying to run the script, but for some reason the download is very slow from the provided links (I am seeing 200 kb/s speeds, and I tested both at home and on the server). Do you think you might be able to put these files on a dropbox link (or an alternative) so that the download is faster?

[Aishell1](https://www.openslr.org/33) provides many download links and I have updated a us mirror link. I have no plus or pro account of dropbox to store such a large file and wish you could download faster with this link.

ycemsubakan:
> > > > * speechbrain/recipes/REAL-M/sisnr-estimation/train.py
> > > 
> > > 
> > > For the first one, the system will exit if the specified folder does not exist because the folder needs large space and I prefer users to specify. For the second one, there is no feedback with the fetch function so I use a simple callback in urllib.urlretrieve fuction to download with percent feedback. Thanks.
> > 
> > 
> > I am trying to run the script, but for some reason the download is very slow from the provided links (I am seeing 200 kb/s speeds, and I tested both at home and on the server). Do you think you might be able to put these files on a dropbox link (or an alternative) so that the download is faster?
> 
> [Aishell1](https://www.openslr.org/33) provides many download links and I have updated a us mirror link. I have no plus or pro account of dropbox to store such a large file so wish you could download faster with this link.

The new links seem to be working much faster. I will let you know how it goes with the rest.

ycemsubakan:
Hi @huangzj421 , The dataset download now works better. But after the download the code crashes with the message:

```
from scripts.create_aishell1_metadata import create_aishell1_metadata
ModuleNotFoundError: No module named 'scripts'
```

Do you run into this error? I have pulled your latest commit. It seems this is due to the fact that scripts is not on the same folder level as `train.py`.

huangzj421:
> Hi @huangzj421 , The dataset download now works better. But after the download the code crashes with the message:
> 
> ```
> from scripts.create_aishell1_metadata import create_aishell1_metadata
> ModuleNotFoundError: No module named 'scripts'
> ```
> 
> Do you run into this error? I have pulled your latest commit. It seems this is due to the fact that scripts is not on the same folder level as `train.py`.

My mistake, sorry.

huangzj421:
Hi @ycemsubakan , I am sorry that I don't understand what's going on with the test error.

ycemsubakan:
> Hi @ycemsubakan , I am sorry that I don't understand what's going on with the test error.

I am not sure either, it seems to be not related to this PR.

ycemsubakan:
I pushed a small change to make it so that we create the `--data_folder`, if that particular folder does not exist.

huangzj421:
> I pushed a small change to make it so that we create the `--data_folder`, if that particular folder does not exist.

Got it, thanks.

ycemsubakan:
For some reason, after the download finishes, the code crashes for me. I am getting a `file not found error for the dataset`. Are you able to make `train.py` work on a completely new datapath? I would like to make it sure that someone new who is trying this would be able to run this without running into any errors.

huangzj421:
> For some reason, after the download finishes, the code crashes for me. I am getting a `file not found` error for the dataset. Are you able to make `train.py` work on a completely new datapath? I would like to make it sure that someone new who is trying this would be able to run this without running into any errors.

I already check it that it was resulted from an small boolean error modified by pre-commit reformatted. It's my mistake that I did not run it again after pre-commit hooks, sorry.

ycemsubakan:
> > For some reason, after the download finishes, the code crashes for me. I am getting a `file not found` error for the dataset. Are you able to make `train.py` work on a completely new datapath? I would like to make it sure that someone new who is trying this would be able to run this without running into any errors.
> 
> I already check it that it was resulted from an small boolean error modified by pre-commit reformatted. It's my mistake that I did not run it again after pre-commit hooks, sorry.

Thanks for the update. I ran into one error during the wham noise creation part. Because the machine I landed on the cluster did not have sox, the code broke down. I think we can add in the README that we need the sox installed. 
```

**Comment by ycemsubakan** [comment]
One other behavior I am not sure of, is what happens if for some reason the data generation is interrupted. At least, can we detect that the dataset is not completely created, and perhaps delete the existing files and restart from scratch?

File: recipes/Aishell1Mix/separation/scripts/create_aishell1mix_from_metadata.py [hunk 1]
```diff
@@ -0,0 +1,534 @@
...
+            )
+            dir_path = os.path.join(mode_path, dir_name)
+            # Create subdir
+            if types == ["mix_clean"]:
+                subdirs = [f"s{i + 1}" for i in range(n_src)] + ["mix_clean"]
+            else:
+                subdirs = (
+                    [f"s{i + 1}" for i in range(n_src)] + types + ["noise"]
+                )
+            # If the files already exist then continue, otherwise remove the dir and regenerate
```
huangzj421, line 152: Hi @ycemsubakan , I have already checked the files were generated properly here.

recipes/Aishell1Mix/separation/scripts/create_aishell1mix_from_metadata.py
```diff
@@ -0,0 +1,534 @@
...
+            sounds = glob.glob(
+                os.path.join(dir_path, "**/*.wav"), recursive=True
+            )
+            if len(md_file) * len(subdirs) == len(sounds):
+                print(
+                    f"Directory {dir_path} already exist. "
+                    f"Files won't be overwritten"
```
huangzj421, line 159: Hi @ycemsubakan , I have already checked the files were generated properly here.

ycemsubakan:
Thanks for the updates @huangzj421 . 

Now I am able to create the dataset, and the experiment runs when I tried without dynamic mixing. I however have few more questions. 
1. It seems we do not to specify the base folder for dynamic mixing specified with `--base_folder_dm`, as it is hard coded in the yaml file as `base_folder_dm: <data_folder>/data_aishell/wav/train`. If we always follow the same folder structure for this, it would be better to write a comment inside the yaml files to say that it shouldn't be changed. The other option would be to hard code it inside `train.py`. What do you think? 

2. I created the data without the `--dynamic_mixing True` option, simply by calling `python train.py hparams/sepformer-aishell1mix2.yaml --data_folder mypath'. The data was created and the training started without error. However, I later on tried to run the same experiment with `python train.py hparams/sepformer-aishell1mix2.yaml --data_folder mypath --dynamic_mixing True`, and I have got the following error: 

```Please, specify a valid base_folder_dm folder when using dynamic mixing```

3. I have noticed that for `sepformer-aishell1mix2-wham.yaml`, and `sepformer-aishell1mix3-wham.yaml`. `use_wham_noise` variables are set to `False`. This is a mistake right?

**Comment by ycemsubakan** [comment]
Hi @huangzj421 , thanks for the quick update. The error message above does not appear now, but during the conversion of sampling rate (for the `base_folder_dm`), I ran into this error:

image

Do you see the same error as well?

File: recipes/Aishell1Mix/meta/preprocess_dynamic_mixing.py [hunk 3]
```diff
@@ -0,0 +1,98 @@
...
+    # add the matlab functions to octave dir here
+
+    files = glob.glob(os.path.join(input_folder, regex), recursive=True)
+    for f in tqdm.tqdm(files):
+
+        audio, fs_read = torchaudio.load(f)
+        audio = audio[0].numpy()
+        audio = signal.resample_poly(audio, fs, fs_read)
+
+        if len(audio) == 0:
```
huangzj421, line 62: Hi @ycemsubakan , aishell1 contains such zero length audio which cause this error so I insert this line to skip. But the aishell1mix generation has filtered out length < 3 audio so don't worry.

ycemsubakan:
Thanks @huangzj421 for all the updates. 

I have test all four .yaml files that you have provided and everything seems to work. If all is okay with @mravanelli as well, we can go ahead and merge the PR I think.

huangzj421:
> changed the experiment names for hparams with wham noise

Okay, much thanks!

huangzj421:
Hi @ycemsubakan @mravanelli , may I ask what's going on now?

ycemsubakan:
Hey @huangzj421 , we need to make one more internal review and then we should be good to merge this.

huangzj421:
> Hey @huangzj421 , we need to make one more internal review and then we should be good to merge this.

Okay, thanks.

ycemsubakan:
Hi @huangzj421 , Sorry for the long wait. 

I have restarted testing the PR, but I noticed that it takes very long for data creation. Please see the screenshot below:
image

Do you see the same behaviour as well?

huangzj421:
> Hi @huangzj421 , Sorry for the long wait.
> 
> I have restarted testing the PR, but I noticed that it takes very long for data creation. Please see the screenshot below: image
> 
> Do you see the same behaviour as well?

No, it's impossible to run so slow. Maybe it's related to the memory space of your servers?

ycemsubakan:
I am not sure what's happening here, I tried to generate the datasets on a desktop machine and it took 3 days. Another thing is, would it be possible to selectively generate the datasets. Like, can we specify ['8kmin', '16min'] if we want 8k min version and 16k min version only?

huangzj421:
> I am not sure what's happening here, I tried to generate the datasets on a desktop machine and it took 3 days. Another thing is, would it be possible to selectively generate the datasets. Like, can we specify ['8kmin', '16min'] if we want 8k min version and 16k min version only?

Hi, now you can specify the generation type in hparams files. Thanks.

ycemsubakan:
I fixed the paths in the yaml files so that the folder names for the datasets are correct. I will merge once the tests pass.

PR #1154, opened 2021-11-24
SentencePiece add_dummy_prefix - keep the default value

aheba:
keep the default value of SentencePiece `add_dummy_prefix=True`


mravanelli:
@aheba, what should be do here? I remember that @Gastron has some concerns on this change.

Gastron:
This is the change I requested. Looks fine to me. @TParcollet is assigned as reviewer.

PR #1163, opened 2021-11-25
Fix bug in InputNormalization

goexle:
change global mean and std only if the model is in training mode. Do not change it in eval mode.

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -1036,36 +1036,47 @@ class InputNormalization(torch.nn.Module):
 
                 spk_id = int(spk_ids[snt_id][0])
 
-                if spk_id not in self.spk_dict_mean:
+                if self.training:
+                    if spk_id not in self.spk_dict_mean:
 
-                    # Initialization of the dictionary
-                    self.spk_dict_mean[spk_id] = current_mean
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you for reporting that. I though this was already solved in the past. Let me investigate a bit and go back to you.

mravanelli:
I took a look at it and we fixed it in #933 (a big PR that will be part of the next major version of speechbrain). I did some tests on TIMIT and LibriSpeech and this, as expected, has no effect on the performance. The count variable is very high after training and the test statistics has a weight that is close to zero.  We can, however, merge this PR to fix that even in the current version.

goexle:
@mravanelli  thank you for reviewing and merging

danpovey:
I am a bit skeptical about how easy it would be to apply this kind of normalization in a product.
In our setup we recently found that simply multiplying the input log-energy features by 0.1 works a little better
than applying feature batch-normalization.  This allows the bias parameter of the 1st layer to be effective
(since optimizers won't usually allow it to reach a large value, like 10, to be comparable to the magnitude of
the input features).

mravanelli:
That's pretty interesting!

danpovey:
Yeah we also found we can remove the batchnorm from the conformer
convolution without degrading the results.

PR #1060, opened 2021-10-23
Tutorial & updates to the DynamicBatchSampler

vitas-ai:
Hi, 

I added a method to reduce unused padding frames, and an alternative computation of the bucket boundaries.
Please take a look at the logging messages if they are fine. The validation set also needs a DynamicBatchSampler on low-VRAM.

Please compare with the shared tutorial draft.

Best wishes,
Andreas

PR #1174, opened 2021-12-01
Dynamic Batching, Updated 

popcornell:
Superseeding https://github.com/speechbrain/speechbrain/pull/1060 done by Andreas since he has no access anymore to that account. 
I am going to close that PR, and polish the code here so we can merge finally this.


recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml
```diff
@@ -48,6 +48,15 @@ batch_size: 8
 lr: 1.0
 ctc_weight: 0.5
 sorting: ascending
+dynamic_batching: False
```
popcornell, line 51: false by default to ensure by default our results are reproducible

recipes/LibriSpeech/ASR/seq2seq/train.py
```diff
@@ -292,7 +293,44 @@ def text_pipeline(wrd):
     sb.dataio.dataset.set_output_keys(
         datasets, ["id", "sig", "wrd", "tokens_bos", "tokens_eos", "tokens"],
     )
-    return train_data, valid_data, test_datasets
+    train_batch_sampler = None
```
popcornell, line 296: i had to add some lines to the train.py of LibriSpeech seq2seq and TIMIT seq2seq recipes. These however are not used by default

popcornell (commented):
Here are some of the changes made in the recipes

recipes/LibriSpeech/ASR/seq2seq/train.py
TParcollet: yeah, this however, might be something to discuss at a high level. Do we wish to duplicate the recipe to avoid having "unused" code, or do we wish to have complete recipes that might appear to be harder to read ? I don't have strong preferences personally.

popcornell: right now we have lot of duplication also for the sorting part. 
Probably we could integrate sorting and dynamic batching and dataset creation from .json and .csv in the toolkit. So most recipes could avoid this boilerplate code

recipes/LibriSpeech/ASR/seq2seq/train.py
```diff
@@ -292,7 +293,44 @@ def text_pipeline(wrd):
...
+        dynamic_hparams = hparams["dynamic_batch_sampler"]
+        hop_size = dynamic_hparams["feats_hop_size"]
+
+        num_buckets = dynamic_hparams["num_quantiles"]
+
+        train_batch_sampler = DynamicBatchSampler(
+            train_data,
+            dynamic_hparams["max_batch_len"],
+            num_buckets=num_buckets,
+            length_func=lambda x: x["duration"] * (1 / hop_size),
```
mravanelli, line 312: I'm not sure to understand why we need to multiply by 1/hope_size. Isn't the duration enough for that? The sampler should operate on waveforms and it should be independent on the feature extraction process.

popcornell: we could avoid that. Personally it is easier for debugging and finding good values for `num_buckets` and `max_batch_len`.  At the first epoch you can look at the logs and intervals are easier to read in terms of feature frames.

mravanelli:
Hi @popcornell and Andreas,
thank you for this PR. I finally reviewed it. TIMIT is working without issue (I can see a clear speed up as well). As for LibriSpeech I had to fix some missing keys (i.e., `max_batch_ex`, `num_quantiles`). In particular,  `max_batch_ex` was expected by train.py but not specified in the yaml one. I set it to 64 (not sure if this makese sense though). @popcornell could you please double check it? As for the speed, could you please report your numbers? I'm not sure I'm able to see any speed improvement with these parameters , but I'm running some experiments to double check it.

popcornell:
`max_batch_ex` will limit the batch size examples with short length but training speed will decrease

popcornell:
With the current parameters full LibriSpeech seq2seq:

1. No dynamic batching but sorted in ascending order: 7:54 hours:mins 
2. Dynamic batching 6:57 hours:mins

mravanelli:
Thank you for this PR. I tried it the latest version and everything is working. The only missing thing is the tutorial, but we can merge this PR while finalizing it.

Issue #1183, opened 2021-12-08
Out of range: piece id is out of range


Jxu-Thu:
Branch：master
I run the SLURP experiments based on HuBERT initialization following https://github.com/speechbrain/speechbrain/tree/develop/recipes/SLURP.

I got the error 

IndexError: Out of range: piece id is out of range. 

The error is the same as the log file provided by official log.txt [checkpoints](https://drive.google.com/drive/folders/1LpcuFldRo_Va1OCGp1bLNdiaC7AQNJOb).

How can I avoid this error?

Moumeneb1:
Hi, I'm running the issue to check, a first thing is that the path speechbrain/SLU-direct-SLURP-wav2vec-enc no longer exists on the speechbrain hub, you have to change :

```yaml
tokenizer_file: speechbrain/SLU-direct-SLURP-wav2vec-enc
```
to 
```yaml
tokenizer_file: speechbrain/SLU-direct-SLURP-hubert-enc
```
and 
```yaml
tokenizer: !ref <tokenizer_file>/SLURP_58_unigram.model
```
to 

```yaml
tokenizer: !ref <tokenizer_file>/tokenizer_58_unigram.model
```

mravanelli:
Can we fix it?

Gastron:
![image](https://user-images.githubusercontent.com/5797214/145474184-29860782-fd9b-432a-94d3-09c5d2fd6e0d.png)

aheba:
Hello @Jxu-Thu , thank you for reporting the issue, let me do a fix,

Moumeneb1:
Hi @Jxu-Thu  The issue is that the tokenizer only generates 51 tokens and not 58. I personally trained the model on real data only, so I didn't noticed this issue. What I found interesting while debugging and it's the thing that is causing this issue is the use of the synthetic data. The issue is that the predicted ids, when using only real data on the first batches, the neurons52 to 57 don't get activated so I didn't encounter it. However, the synthetic data is more perturbating ( at least at first when the layer is not trained). I'll pull request the fix. 
Thanks for the reporting !

aheba:
@Jxu-Thu , thank you for reporting the issue, The PR #1186  is merged :).

Feel Free to share with us any other issues ;)

zhengqing187:
Traceback (most recent call last):
  File "train_with_wav2vec2.py", line 379, in <module>
    valid_loader_kwargs=hparams["dataloader_opts"],
[...]

PR #1186, opened 2021-12-11
Fix slurp tokenizer

Moumeneb1:
A fix for the issue https://github.com/speechbrain/speechbrain/issues/1183#issue-1074106507


aheba:
Hey @Moumeneb1 , where we are in this PR ? 
what next fix is coming ?

mravanelli:
It also looks like there is something minor to fix for the linters (that's why tests are failing)

mravanelli:
Hi @Moumeneb1,
it looks to me that also the current version is not working:

`requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://huggingface.co/speechbrain/SLU-direct-SLURP-hubert-enc/resolve/main/tokenizer_51_unigram.model`

This file is indeed not present (it looks like it should be https://huggingface.co/speechbrain/SLU-direct-SLURP-hubert-enc/blob/main/tokenizer_58_unigram.model, right). It might be that the HF part is not actually updated. Do you confirm it?

aheba:
Thank you @Moumeneb1  for the fix,

LGTM

zhengqing187:
Hi, after running some epochs, I encountered “Loss is not finite and patience is exhausted”  or  “ Out of range: piece id is out of range”, I followed the comments #1183, #1186, #1589 and I modified the output_neurons to 51 as per https://github.com/speechbrain/speechbrain/issues/1183, https://github.com/speechbrain/speechbrain/issues/1186 and https://github.com/speechbrain/speechbrain/issues/1589, but the problem was not solved. How else can I solve this problem?
@aheba

PR #1194, opened 2021-12-14
Implemented GroupNorm in nnet.normalization

gcambara:
I have implemented GroupNorm since I thought it could be useful for writing models using such, like wav2vec2.0 Base LibriSpeech model, for instance. Furthermore, it is one of the norms implemented in the PyTorch library, along with other popular ones like LayerNorm or BatchNorm.


mravanelli:
Hi @gcambara, thank you. Yes, this might be helpful. I just fixed a small style issue and I'm ready to merge it once all tests are done

Issue #1196, opened 2021-12-15
RuntimeError: generator raised StopIteration


KacperKubara:
Hi,

I was running the Transducer model with French CommonVoice data (https://github.com/speechbrain/speechbrain/tree/develop/recipes/CommonVoice/ASR/transducer). I stopped the model training after 1st epoch and then run it again, however, now I am getting `RuntimeError: generator raised StopIteration` when it starts the 2nd epoch. Do you perhaps know why I receive the error?

It was trained with data-parallel-backend option

``` bash
 Available devices: 3
 CUDA Available: True
speechbrain.core - Beginning experiment!
speechbrain.core - Experiment folder: results/cv_transducer/2140
common_voice_prepare - results/cv_transducer/2140/save/train.csv already exists, skipping data preparation!
common_voice_prepare - results/cv_transducer/2140/save/dev.csv already exists, skipping data preparation!
common_voice_prepare - results/cv_transducer/2140/save/test.csv already exists, skipping data preparation!
speechbrain.tokenizers.SentencePiece - Tokenizer is already trained.
speechbrain.tokenizers.SentencePiece - ==== Loading Tokenizer ===
speechbrain.tokenizers.SentencePiece - Tokenizer path: results/cv_transducer/2140/save/1000_unigram.model
speechbrain.tokenizers.SentencePiece - Tokenizer vocab_size: 1000
speechbrain.tokenizers.SentencePiece - Tokenizer type: unigram
speechbrain.core - Info: ckpt_interval_minutes arg from hparam file is used
speechbrain.core - 142.4M trainable parameters in ASR
speechbrain.utils.checkpoints - Loading a checkpoint from results/cv_transducer/2140/save/CKPT+2021-12-09+04-58-56+00
speechbrain.utils.epoch_loop - Going into epoch 2

44178it [00:00, ?it/s]
44178it [00:00, ?it/s]
speechbrain.core - Exception:
Traceback (most recent call last):
  File "/sw/arch/Debian10/EB_production/2021/software/Anaconda3/2021.05/lib/python3.8/site-packages/tqdm/std.py", line 1178, in __iter__
    for obj in iterable:
  File "/home/kubara/repos/speechbrain/speechbrain/dataio/dataloader.py", line 206, in __iter__
    iterator = super().__iter__()
  File "/home/kubara/.local/lib/python3.8/site-packages/torch/utils/data/dataloader.py", line 359, in __iter__
    return self._get_iterator()
[...]

PR #1197, opened 2021-12-15
Catch StopIteration in Fast-Forward

Gastron:
Possible fix to #1196.

When recovering from an intra-epoch checkpoint, the Sampler is fast-forwarded by the saved amount of indices. However, it could be possible to use some kind of Sampler that has some non-deterministic behaviour, and runs out of samples this time. Also might be possible that intra-epoch checkpoint happens after exhausting the whole Sampler and I don't even know, may be there is some edge case where this is legitimate. In any case, let's ignore this (just raise a warning) mismatch between our expectation and what happens.


mravanelli:
Not sure about the best way to test this. For now, I just test regular training/recovery and it works as expected.

PR #1201, opened 2021-12-15
Wav2vec2 finetune quickfix

TParcollet:
@mravanelli this is what I told you about during today' meeting. This PR improves:
- [x] Do not display wav2vec 2.0 parameters as trainable if frozen (CTC and seq2seq)
- [x] Do not use the wav2vec 2.0 optimizer if frozen (CTC and seq2seq)
- [x] Some cleaning of useless hyperparams in yaml. 

PS: Sorry for the large list of commit, this is a bug on my side ... but the files are the right ones.


mravanelli:
I did some quick check to make sure everything is running. I just add the extra requirement file in the CTC folder. I think we can merge it if all tests pass.

PR #1223, opened 2022-01-03
[Fix] Fix init in LibrispeechLmConfig

### LINKED ISSUE #1219: lm_corpus_path - unexpected keyword argument
Hello, I'm trying to start training LibriSpeech language model (transformer).
I manually downloaded the corpus and when I'm trying to run `train.py` on two GPUs:

```
CUDA_VISIBLE_DEVICES=0,1
python train.py hparams/transformer.yaml --data_parallel_backend
```
I'm getting the argument error:
https://paste-bin.xyz/29180
`TypeError: __init__() got an unexpected keyword argument 'lm_corpus_path'`

Is something wrong with the dataset loading script? I tried to find some info about how you should name the argument providing the corpus path but it was unsuccessful.

jamfly:
## Summary
Fix #1219

### DIFF
--- a/recipes/LibriSpeech/LM/dataset.py
+++ b/recipes/LibriSpeech/LM/dataset.py
@@ -16,7 +16,7 @@
 # Lint as: python3
 """
 Librispeech language modeling dataset.
-    this is an extented from huggingface's official implementation to allow the use of train-960 trainscript and lm_corpus for LM training
+    this is an extented from huggingface's officia implementation to allow the use of train-960 trainscript and lm_corpus for LM training
 Authors
  * Jianyuan Zhong 2021
 """
@@ -25,6 +25,8 @@ from __future__ import absolute_import, division, print_function
 
 import datasets
 import re
+from typing import Optional
+
 
 _CITATION = """\
 @inproceedings{panayotov2015librispeech,
@@ -50,11 +52,11 @@ class LibrispeechLmConfig(datasets.BuilderConfig):
     """builder config for LibriSpeech LM
     """
 
-    def __init__(self, lm_corpus_path=None, **kwargs):
-        super().__init__(**kwargs)
-        self.lm_corpus_path = (
-            lm_corpus_path if lm_corpus_path is not None else _DL_URL
-        )
+    lm_corpus_path: Optional[str] = None
+
+    def __post_init__(self):
+        if self.lm_corpus_path is None:
+            self.lm_corpus_path = _DL_URL
 
 
 class LibrispeechLm(datasets.GeneratorBasedBuilder):
--- a/recipes/LibriSpeech/LM/dataset.py
+++ b/recipes/LibriSpeech/LM/dataset.py
@@ -16,7 +16,7 @@
 # Lint as: python3
 """
 Librispeech language modeling dataset.
-    this is an extented from huggingface's officia implementation to allow the use of train-960 trainscript and lm_corpus for LM training
+    this is an extented from huggingface's official implementation to allow the use of train-960 trainscript and lm_corpus for LM training
 Authors
  * Jianyuan Zhong 2021
 """
--- a/recipes/LibriSpeech/LM/dataset.py
+++ b/recipes/LibriSpeech/LM/dataset.py
@@ -25,8 +25,6 @@ from __future__ import absolute_import, division, print_function
 
 import datasets
 import re
-from typing import Optional
-
 
 _CITATION = """\
 @inproceedings{panayotov2015librispeech,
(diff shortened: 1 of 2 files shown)


Gastron:
Thanks for this! Has the HuggingFace datasets library changed recently, is that what causes this? In that case, should we specify some version here: https://github.com/speechbrain/speechbrain/blob/062a17f32c6d67391e73be97b926ec3da3b12a6c/recipes/LibriSpeech/LM/extra_requirements.txt

jamfly:
turns out this approach will still encounter another problem: 
```
raceback (most recent call last):
    File "train.py", line 185, in <module>
        train_data, valid_data, test_data = dataio_prepare(hparams)
      File "train.py", line 115, in dataio_prepare
      datasets = load_dataset(
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/load.py", line 1694, in load_dataset
      builder_instance.download_and_prepare(
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/builder.py", line 595, in download_and_prepare
      self._download_and_prepare(
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/builder.py", line 683, in _download_and_prepare
        self._prepare_split(split_generator, **prepare_split_kwargs)
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/builder.py", line 1085, in _prepare_split
      num_examples, num_bytes = writer.finalize()
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/arrow_writer.py", line 484, in finalize
      self.check_duplicate_keys()
    File "/home/jamfly/miniconda3/envs/sb/lib/python3.8/site-packages/datasets/arrow_writer.py", line 420, in check_duplicate_keys
      raise DuplicatedKeysError(key)
datasets.keyhash.DuplicatedKeysError: FAILURE TO GENERATE DATASET !
  Found duplicate Key: 0
  Keys should be unique and deterministic in nature
```

Hi @Gastron,

After some code tracing, I found that it might be the issue when hugging face update their codebase from [1.6.2](https://github.com/huggingface/datasets/releases?page=2) to [1.7.0](https://github.com/huggingface/datasets/releases?page=2). However, current version is [1.17.0](https://github.com/huggingface/datasets/releases?page=1). I think at this time, we can specifically indicate the version of datasets. But we still need to find the real cause in the future.

So, I will only set datasets' version to 1.6.2 as the workaround.

TParcollet:
Well, more broadly speaking @mravanelli I don't understand why we depend on HuggingFace dataset libs here. This introduces a datasets.py script that seems to be the source of the error and will always be as it couples the recipes with future version of HuggingFace dataset ... :-(

TParcollet:
This really looks like something that MUST be converted to webdataset ... Aku, any chance that you could have the time to look at it ? I could retrain the LM entirely If we are able to switch this to webdataset.

mravanelli:
I think @JianyuanZhong preferred to use the HF dataset here for some reasons. As far as I remember, it was very fast and already set up to work on large text corpora. The standard datasets in speechbrain expects a JSON or a CSV files, while here you might want to train on the big text corpus directly

TParcollet:
Yes, this is why I think Webdataset should be preferred :-/ Or we should revise so it fits to the latest datasets version (1.17 and not 1.6 :/)

mravanelli:
Right, if we are able to remove one extra-dependency I think it is much
better for maintainability.

Gastron:
I think WebDataset is not a good option for text. I think for text something as simple as a few text files as shards could work. Some simple IterableDataset implementation for that would be nice.

mravanelli:
What should we do it here? @jamfly proposed a temporary fix that might work for now. Do you think we have to implement the bigger refactor discussed before in this PR or in another one?

Gastron:
Yeah I think we should merge this version specification change as this is just one recipe. Thanks to @jamfly for taking the time to investigate the bigger picture too!

Moumeneb1:
Hi @mravanelli, @TParcollet I think the issue was that, the values used as keys weren't really unique (values can be duplicated between files(archive paths)) in the speechbrain implementation, and huggingface's dataset previous implementation wasn't checking that the keys were unique. They introduced the check in this new version, which is why we couldn't execute it anymore.

I just made sure that the keys are unique within a set with this small change, and it's working fine with the new version. 

```python
def _generate_examples(self, archive_path):
    """Yields examples."""
    key = 0
    for p in archive_path:
        with open(p, "r", encoding="utf-8") as f:
            for line in f:
                line = re.sub(
                    r"\d+-\d+-\d+\s", "", line
                )  # remove ids in transcripts
                text = line.strip()
                # Skip empty lines.
                # very long sentences (>1000 char) are removed to prevent OOM
                if text and len(text) < 1000:
                    yield key, {"text": text}
                    key +=1
```

PR #1228, opened 2022-01-04
Fix documentation generation

Gastron:
Docs generation was failing.

The import system was somehow getting confused: when a `transformers` module tried to import their local `tokenizers.py`, it instead resolved to importing the SpeechBrain `tokenizers.py` file.

I wanted to fix the whole thing by mocking the extra dependencies. But it turns out, better-apidoc does not mock the imports in autodoc_mock_imports, it just changes ImportErrors to warnings, which does not work for us. So instead I hacked this so that SpeechBrain is added to the end of the path rather than right at the front.

I also fixed the fairseq and huggingface wav2vec2 ImportErrors to properly raise an error instead of printing a message.

### DIFF
--- a/.github/workflows/verify-docs-gen.yml
+++ b/.github/workflows/verify-docs-gen.yml
@@ -1,0 +1,26 @@
+name: Verify docs generation
+
+# Runs on pushes to master and all pull requests
+on:    # yamllint disable-line rule:truthy
+    push:
+        branches: [main, develop]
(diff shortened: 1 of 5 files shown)


Gastron:
I added a GitHub action to test the docs generation so that hopefully this doesn't happen again.

TParcollet:
LGTM.

PR #1229, opened 2022-01-05
Write note about docs-requirements.txt

Gastron:
One more thing :)


Gastron:
Looks good to myself.

PR #1232, opened 2022-01-06
DVoiceSpeechBrain

nairaxo:
We use SpeechBrain to develop an ASR model using the DVoice Dataset for african languages/dialects and Wav2Vec.


mravanelli:
Hi @nairaxo, 
thank you for this PR! I remember we discussed this contribution some time ago in one of our call and I'm happy you find the time for that!  I will review it as soon as I can!

nairaxo:
Thank you @mravanelli for your prompt feedback... I made the necessary changes. I hope it's ok now.

mravanelli:
Hi @nairaxo,
I have a couple of comments:
1. Did you consider putting the data on [Zenodo] rather than GDrive? Grdive is a annoying because it is not "wget" friendly. If you put the data on zenodo, we can download them directly during data preparation like [here](https://github.com/speechbrain/speechbrain/blob/develop/recipes/timers-and-such/prepare.py#L55).
2. It looks like the current data contain a single language but here you report the results for two. Do you think we can have also the second language?
3. You do the augmentation during preparation. In SpeechBrain, instead, we normally do the augmentation on-the-fly during training. See for instance [this example](https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml#L136). Is there a reason for that?

nairaxo:
Hello @mravanelli,
Thank you for your comments!
I just added the dataset for Moroccan Arabic dialect on Zenodo. I'm actually making some process for the Swahili and I'll add it too... For the data augmentation, I performed it during preparation just because this approach helped me before to improve ASR models performances. But I'm experimenting the on-the-fly method which which looks better too. 

I encountered this error while testing...

`Traceback (most recent call last):
  File "/home/n.abdoumohamed/DVoice/speechbrain/recipes/DVoice/ASR/CTC/train_with_wav2vec.py", line 379, in <module>
    asr_brain.evaluate(
  File "/home/n.abdoumohamed/DVoice/speechbrain/speechbrain/core.py", line 1193, in evaluate
    loss = self.evaluate_batch(batch, stage=Stage.TEST)
  File "/home/n.abdoumohamed/DVoice/speechbrain/recipes/DVoice/ASR/CTC/train_with_wav2vec.py", line 142, in evaluate_batch
    loss = self.compute_objectives(predictions, batch, stage=stage)
  File "/home/n.abdoumohamed/DVoice/speechbrain/recipes/DVoice/ASR/CTC/train_with_wav2vec.py", line 94, in compute_objectives
    self.cer_metric.append(ids, predicted_words, target_words)
  File "/home/n.abdoumohamed/DVoice/speechbrain/speechbrain/utils/metric_stats.py", line 287, in append
    scores = wer_details_for_batch(ids, target, predict, True)
  File "/home/n.abdoumohamed/DVoice/speechbrain/speechbrain/utils/edit_distance.py", line 379, in wer_details_for_batch
    return wer_details_by_utterance(
  File "/home/n.abdoumohamed/DVoice/speechbrain/speechbrain/utils/edit_distance.py", line 489, in wer_details_by_utterance
    "WER": 100.0 * sum(ops.values()) / len(ref_tokens),
ZeroDivisionError: float division by zero`

b-med:
Hi Naira,
It should be related to some utterances having an empty reference word sequence.

nairaxo:
Thank you @b-med, it works now!

nairaxo:
Hi @mravanelli!

I hope you're doing well. I maked updates on the repo. I had some issues when inferring the model but i fixed it now. Thank you!

mravanelli:
Thank you for this great PR (and sorry for the late review)! Here are my comments:

1. After the minor modifications that I pushed, I'm able to run the systems for Darija and Swahili. I cannot run the systems for the other languages (those in https://github.com/besacier/ALFFA_PUBLIC ). It looks to me that there is no match between the way the dataset is structured in https://github.com/besacier/ALFFA_PUBLIC and the way the data_preparation script is designed (the preparation expects a folder <data_folder>/wavs). Am I missing something?

2. Normally, we put the full output folder (which contains the models, the logs) in our google drive (see for instance https://drive.google.com/drive/folders/1A2z0cziLSw83YBi5H2ZRiTXTXjV-EXli?usp=sharing). Can you please do the same for DVoice? Please, contact me privately such that I can give you the upload permissions.

3. Normally, we put an audio example (example.wav) in the HuggingFace. This makes it simpler for the users trying the model. Also, are you ok if we upload your systems to the official SpeechBrain HF repository?

nairaxo:
Hi @mravanelli!

Thank you for your reviews and for your last commits. It's really a pleasure that you appreciate the work that has been done. I'm actually making some updates specially in the data-prep task for the other languages when using ALFFA_PUBLIC. And after that I'll get back to you for the uploads.

mravanelli:
Thank you Naira,
please keep me updated!

mravanelli:
Hi @nairaxo,
thank you very much for the updates! I took another review of the code and I did some minor changes. I'm now able to start training with Swahili, Darija,Fongbe, Amharic.
I still see the following issues:
1. The data preparation for Wolof doesn't work. I think this is due to a hard-coded link in dvoice_prepare.py (/data/n.abdoumohamed/dvoice-africa/data-processing/scripts/xo.txt').
2. It is not clear where to find the data for Kabyle (I thus temporarily dropped this recipe). I cannot find the data for Kabyle in https://github.com/besacier/ALFFA_PUBLIC and it is not clear where we can find them.
3. It is also not clear how to run the multi-language recipe. This should be made easier for the users. Some instructions on how to run it should be reported in README.md.

nairaxo:
Hello!
Thank you for these last commits! Sorry for the issue with Wolof. I juste fixed it. I also made some updates to the data_preparation specially for the Multilingual and the Swahili recipes. For these lasts we need both DVoice dataset and ALFFA_PUBLIC data. So I clarified the process with some instructions in the README.

For the Kabyle, I used Common Voice, not ALFFA. I think we can left it aside temporarily as you suggested. I'll come to that right after.

mravanelli:
Thank you @nairaxo, I did some small modifications and now all conditions work. I can merge it now. The very last thing to do is to upload the outpput folder of the experiments (those that contains the checkpoints, training loggers, etc) in the official gdrive folder. I will follow up with you privately for that.

PR #1234, opened 2022-01-07
fix bug: negative_scores calculation in BinaryMetricStats

aluminumbox:
Hi, I have created an issue mentioning a possible bug in utils/metric_stats.py

If it is indeed a bug or typo, I have fixed it in the PR. Hope you can merge it.

Many thanks!

### DIFF
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -402,7 +402,7 @@ class BinaryMetricStats(MetricStats):
                             for i in range(
                                 0,
                                 len(positive_scores),
-                                int(len(positive_scores) / max_samples)
+                                int(len(positive_scores) / max_samples),
                             )
                         ]
                     ]
@@ -414,7 +414,7 @@ class BinaryMetricStats(MetricStats):
                             for i in range(
                                 0,
                                 len(negative_scores),
-                                int(len(negative_scores) / max_samples)
+                                int(len(negative_scores) / max_samples),
                             )
                         ]
                     ]
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -344,7 +344,7 @@ class BinaryMetricStats(MetricStats):
         self.scores.extend(scores.detach())
         self.labels.extend(labels.detach())
 
-    def summarize(self, field=None, threshold=None, beta=1, eps=1e-8):
+    def summarize(self, field=None, threshold=None, max_samples=None, beta=1, eps=1e-8):
         """Compute statistics using a full set of scores.
 
         Full set of fields:
@@ -355,6 +355,7 @@ class BinaryMetricStats(MetricStats):
          - FAR - False Acceptance Rate
          - FRR - False Rejection Rate
          - DER - Detection Error Rate (EER if no threshold passed)
+         - threshold - threshold (EER threshold if no threshold passed)
          - precision - Precision (positive predictive value)
          - recall - Recall (sensitivity)
          - F-score - Balance of precision and recall (equal if beta=1)
@@ -367,6 +368,10 @@ class BinaryMetricStats(MetricStats):
             a dict with all statistics is returned.
         threshold : float
             If no threshold is provided, equal error rate is used.
+        max_samples: float
+            How many samples to keep for postive/negative scores.
+            If no max_samples is provided, all scores are kept.
+            Only effective when threshold is None.
         beta : float
             How much to weight precision vs recall in F-score. Default
             of 1. is equal weight, while higher values weight recall
@@ -380,8 +385,15 @@ class BinaryMetricStats(MetricStats):
             self.labels = torch.stack(self.labels)
 
         if threshold is None:
-            positive_scores = self.scores[(self.labels==1).nonzero(as_tuple=True)]
-            negative_scores = self.scores[(self.labels==0).nonzero(as_tuple=True)]
+            positive_scores = self.scores[(self.labels == 1).nonzero(as_tuple=True)]
+            negative_scores = self.scores[(self.labels == 0).nonzero(as_tuple=True)]
+            if max_samples is not None:
+                if len(positive_scores) > max_samples:
+                    positive_scores, _ = torch.sort(positive_scores)
+                    positive_scores = positive_scores[[i for i in range(0, len(positive_scores), int(len(positive_scores) / max_samples))]]
+                if len(negative_scores) > max_samples:
+                    negative_scores, _ = torch.sort(negative_scores)
+                    negative_scores = negative_scores[[i for i in range(0, len(negative_scores), int(len(negative_scores) / max_samples))]]
 
             eer, threshold = EER(positive_scores, negative_scores)
 
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -380,10 +380,8 @@ class BinaryMetricStats(MetricStats):
             self.labels = torch.stack(self.labels)
 
         if threshold is None:
-            positive_scores = self.scores[self.labels.nonzero(as_tuple=True)]
-            negative_scores = self.scores[
-                self.labels[self.labels == 0].nonzero(as_tuple=True)
-            ]
+            positive_scores = self.scores[(self.labels==1).nonzero(as_tuple=True)]
+            negative_scores = self.scores[(self.labels==0).nonzero(as_tuple=True)]
 
             eer, threshold = EER(positive_scores, negative_scores)
 
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -396,6 +396,7 @@ class BinaryMetricStats(MetricStats):
         self.summary["FAR"] = FP / (FP + TN + eps)
         self.summary["FRR"] = FN / (TP + FN + eps)
         self.summary["DER"] = (FP + FN) / (TP + TN + eps)
+        self.summary["threshold"] = threshold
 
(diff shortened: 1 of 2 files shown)


mravanelli:
I'm wondering if we can extend the unit tests [here] (https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_metrics.py#L51) to test this condition as well. It could also be important to figure out which recipes used this function (I'm not sure we have use it so far, but it is better to double check).  @popcornell could you please review this PR?

aluminumbox:
> I'm wondering if we can extend the unit tests [here] (https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_metrics.py#L51) to test this condition as well. It could also be important to figure out which recipes used this function (I'm not sure we have use it so far, but it is better to double check). @popcornell could you please review this PR?

I have extend the unit test for binary_stats.summarize when threshold=None.
Also, I think it is better to add a max_samples for summarize, because if the postive/negative_scores is a big tensor, it is very easy for EER to cause OOM at https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L448

aluminumbox:
> I'm wondering if we can extend the unit tests [here] (https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_metrics.py#L51) to test this condition as well. It could also be important to figure out which recipes used this function (I'm not sure we have use it so far, but it is better to double check). @popcornell could you please review this PR?

Hi, should I wait for the pending reviewer to review the code? It seems he hasn't respond yet.

mravanelli:
HI @aluminumbox, I talked to @popcornell  and he needs few more days before diving into that.

speechbrain/utils/metric_stats.py
```diff
@@ -367,6 +370,10 @@ def summarize(self, field=None, threshold=None, beta=1, eps=1e-8):
             a dict with all statistics is returned.
         threshold : float
             If no threshold is provided, equal error rate is used.
+        max_samples: float
```
popcornell, line 373: do we really need that to fix the OOM issue in EER computation or is it better to add another argument in EER computation to use a for loop so you trade more memory for less speed ?

speechbrain/utils/metric_stats.py
```diff
@@ -380,10 +387,37 @@ def summarize(self, field=None, threshold=None, beta=1, eps=1e-8):
             self.labels = torch.stack(self.labels)
 
         if threshold is None:
-            positive_scores = self.scores[self.labels.nonzero(as_tuple=True)]
+            positive_scores = self.scores[
```
popcornell, line 390: LGTM

tests/unittests/test_metrics.py
```diff
@@ -63,7 +63,8 @@ def test_binary_metrics(device):
     assert summary["FP"] == 1
     assert summary["FN"] == 2
 
-    summary = binary_stats.summarize()
+    summary = binary_stats.summarize(threshold=None)
+    assert summary["threshold"] >= 0.3 and summary["threshold"] < 0.4
```
popcornell, line 67: needs also a test for max_samples probably

popcornell (commented):
@aluminumbox Thanks for your PR ! 
Sorry for my late response again. 
It looks good. I have only doubts of the max_samples arg. If I understand correctly you are limiting the number of samples used to estimate the threshold because it can fill up your memory. I am wondering if actually we can instead using a for loop in EER computation to achieve the same without any approx. 

EDIT: actually when threshold is None we don't have to compute line https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L448 so maybe the OOM issue can be avoided altogether. EER is not needed.

speechbrain/utils/metric_stats.py
aluminumbox: I am using GTX 2080 with 10G CUDA memory, and limited max samples to 5000. Using the eer/threshold calculated from the 5000 postive + 5000 negative samples. I obtained FAR 0.13936119335841204 and FRR 0.1393614282003676. Not exactly the same, but okay to me. Another way is move postive/negative_socres tensor to cpu and then calculate eer. I agree that there maybe an algorithm trading speed for memory, but I have no clue yet.

aluminumbox: > do we really need that to fix the OOM issue in EER computation or is it better to add another argument in EER computation to use a for loop so you trade more memory for less speed ?

I am using GTX 2080 with 10G cuda memory, and keep 5000 samples for postive/negative scores. Using the eer/threshold from the eer function, I get FAR 0.13936119335841204 and FRR 0.1393614282003676. Not exactly the same, but the precision is okay to me. Another way is to move postive/negative scores to cpu before calculating eer. I agree there is an algorithm trading speed for memory, but I have no clue how to do it yet.

tests/unittests/test_metrics.py
aluminumbox: > needs also a test for max_samples probably

I added a test for max_samples=1. Because there are only 3 postive/negative samples in the test case, in this situation, max_samples only work when max_samples==1, and kept 0.1 /0.2for postive/negative samples.

speechbrain/utils/metric_stats.py
aluminumbox: > do we really need that to fix the OOM issue in EER computation or is it better to add another argument in EER computation to use a for loop so you trade more memory for less speed ?

In EER function, I think maybe we can use a for loop to calcuate far for each candidate threshold https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L451.
But it would require some work, and I am not sure how long it would take when postive/negative_scores are a large tensor.

popcornell: What if you just re-use: 

```python
thresholds, _ = torch.sort(torch.cat([positive_scores, negative_scores]))
thresholds = torch.unique(thresholds)

# Adding intermediate thresholds
interm_thresholds = (thresholds[0:-1] + thresholds[1:]) / 2
thresholds, _ = torch.sort(torch.cat([thresholds, interm_thresholds]))
```

does it go OOM ?

aluminumbox: > https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L451

This thresholds only gives the threshold candidates. We need to calculate all FAR/FRR for all threshold candidates, which is in https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L452 and https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L461. The OOM happens when we try to get a 10000*10000 tensors (for example there are 10000 threshold candidates)

aluminumbox: > 


This thresholds only gives the threshold candidates. We need to calculate all FAR/FRR for all threshold candidates, which is in https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L452 and https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L461. The OOM happens when we try to get a 10000*10000 tensors (for example there are 10000 threshold candidates)

aluminumbox: > What if you just re-use:
> 
> ```python
> thresholds, _ = torch.sort(torch.cat([positive_scores, negative_scores]))
> thresholds = torch.unique(thresholds)
> 
> # Adding intermediate thresholds
> interm_thresholds = (thresholds[0:-1] + thresholds[1:]) / 2
> thresholds, _ = torch.sort(torch.cat([thresholds, interm_thresholds]))
> ```
> 
> does it go OOM ?

This thresholds only gives the threshold candidates. We need to calculate all FAR/FRR for all threshold candidates, which is in https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L452 and https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#L461. The OOM happens when we try to get a 10000*10000 tensors (for example there are 10000 threshold candidates)

popcornell: my bad you are right. What if you do a for loop over 10k instead of having a matrix of 10k ? 
+ I think that all the metrics should be by default in the cpu wdyt @mravanelli this is a terrible waste of VRAM

aluminumbox: > my bad you are right. What if you do a for loop over 10k instead of having a matrix of 10k ? Also I think that all the metrics should be by default in the cpu wdyt @mravanelli this is a terrible waste of VRAM. Your solution is acceptable BTW also. We can merge and I can try to use a for loop

Yes use a for loop can significantly reduce the memory usage, and this step is better calculated on cpu. And the tradeoff between memory and time should be carefully studied.

tests/unittests/test_metrics.py
aluminumbox: Hi @popcornell @mravanelli Should I try to use a for loop in EER function, or we just merge this PR?

mravanelli: I think it is fine avoiding for loop...

mravanelli:
thank you @aluminumbox. @popcornell do you think we can merge it?

popcornell:
Yes, I think we can merge this when checks are passing. @aluminumbox feel free to open another issue/pull request if you try to use a for loop instead. Might try to do it myself also. I ll ping you in that case.

mravanelli:
Before merging it, there it a tiny black issue to fix.

popcornell:
Fixed, thanks again @aluminumbox I am merging this.

PR #1248, opened 2022-01-12
G2P System Improvements

flexthink:
The objective of this PR is to enhance SpeechBrain with a capable neural Grapheme-to-Phoneme system that can take advantage of sentence context when performing the conversion and to disambiguate between homographs.

Capabilities added
* Curriculum training
* CTC support
* Language model support
* Tokenization
* A transformer-based model
* A pertained interface
* A transcription script (batch or interactive)


TParcollet:
Note: I can try to find a French dataset to try it out.

mravanelli:
Hi @flexthink,
thank you for this PR! It implements a very powerful G2P.
My comments mainly regard simplification and documentation:

A) **Simplification**
- The current code supports tons of different options because you did a lot of experiments before finding the best system.  However, such flexibility makes the code hard to follow (the train.py has 1200 lines, while, for instance, the librispeech seq2seq one that is still a rather complex system has only 400 lines).

- I suggest keeping only the minimal code needed to implement the best systems.
We can maybe keep only the best RNN and the best Transformer.  We can also remove orion-related files (e.g, hpfit.yaml).

B) **Documentation**
- B.1 We have to rewrite a README.md file (mainly with the content already available in the header of train.py). 
- B.2 Make sure to add a table with the best results for each released system.
- B.3 IMO, it is not clear which data we have to use for training the system. There are a couple of links, but I'm still not sure what I have to use. What about putting all the text data somewhere and downloading them directly in the data preparation?
- B.4 I suggest putting more effort into adding proper examples (at least where is feasible) in the docstrings. This might help users a lot.  For instance:
- In `speechbrain/dataio/sampler.py` `class BalancingDataSampler(ReproducibleWeightedRandomSampler)` => example in the docstring.
- `speechbrain/lobes/models/g2p/dataio.py` => clean_pipeline,  grapheme_pipeline, phoneme_pipeline, char_range, build_token_char_map, flip_map
- `lobes/models/g2p/homograph.py` => add docstring examples whehre possible for instance in SubsequenceLoss
- `speechbrain/nnet/schedulers.py` => add example for InverseSquareRootScheduler (just like the other schedulers) 
-` speechbrain/utils/metric_stats.py` => ClassificationStats add example + docstrings for support functions
-` speechbrain/wordemb/util.py`,  speechbrain/wordemb/transformer.py and everywhere you think is feasible.

c)  **Others**
- C.1 Make sure all the systems are running. For instance, when I try `python train.py hparams/hparams_attnrnn_lexicon.yaml` => "ValueError: The reference "N_epochs" is not valid". 
- C.2 https://github.com/speechbrain/speechbrain/blob/8b3f9dbce94ace199e869af0d9d415116e234e79/speechbrain/lobes/models/g2p/dataio.py#L387 => It looks like the arguments described in the docstring do not always correspond to the ones actually in the input
- C.3 The tensorboard should be options (usually we have an in yaml something like use_tensor_board)
- C.4 We have to upload our best models on HuggingFace and in the official Google Drive of SpeechBrain.
- C.5` /speechbrain/pretrained/training.py `=> not sure to understand why we need this

flexthink:
C.3 This is already the case:
use_tensorboard: True

C.5: Right now all it does is load the best checkpoint and dumps it to a folder - to load to the pretrainer, excluding anything used only in training. There are a few known caveats:

* Dependencies aren't exported
* Hyperparameters aren't exported (needs manual tweaks). The hyperparameter files output by training will contain absolute references to file paths.

flexthink:
Thanks a lot @mravanelli  for the detailed review. I will be addressing the comments.

flexthink:
@mravanelli re: the conflict, my version no longer has a file named `train.yaml` (which was a lexicon-only file) - instead, there are files for the different models. If there were changes on the main branch to that file, that would cause a conflict.

Do we need to bring back the original file? 

There would have been other conflicts previously - those are resolved. At the time you reviewed, the branch was far behind develop.

flexthink:
@mravanelli : Please don't merge just yet. There appears to have been a recent nonlinear merge to the `develop` branch, causing conflicts. I am manually reviewing all changed files.

flexthink:
@mravanelli: FYI, my force-push was into my branch only - and it is expected to simply undo the previous attempt at a merge.

mravanelli:
Thank you,
here are my final comments:
1. Add a section on train lm README
2. Make sure that all the phases of G2P training are actually working (I see weird errors when setting one epoch for each phase, but I need to investigate more).
3. You can upload the best models (one for each of the released hparam files here:https://drive.google.com/drive/folders/1OFz4Ov1wwooIu9eI7fjvWUfg2bUylEyo?usp=sharing).
4. Make sure to update the README file to mention the output folders uploaded on gdrive
5. You can create the official HF repo. You can do it with your account and then we can upload it on the official speechbrain one. Make sure the model description is aligned with other models available on the SpeechBrain HF. Also, only upload the best model (no need to have more than one).

mravanelli:
Thank you @flexthink for this great G2P system!
As mentioned privately, we can merge it and do the last minor changes in a next PR.

PR #1251, opened 2022-01-13
Add gradient accumulation and optimizer step limits to Brain class

kokamido:
This PR is based on [discourse conversation](https://speechbrain.discourse.group/t/feature-request-iteration-number-in-training-loop/545). The PR provides a possibility to stop training depending on optimizer step count. For some cases (listed in the discourse topic) it's more handy than the epoch counter.
This PR also adds a gradient accumulation inspired by the [CommonVoice ASR transformer recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/transformer/train.py#L140). It's easier to configure training with optimizer step limit while gradient accumulation is enabled. The existing approach (based on epoch counter) requires calculations every time you want to tweak batch size keeping the same number of optimization steps.
I've also made a branch with tests: https://github.com/kokamido/speechbrain/tree/grad_acc_and_iter_stop_tests
Tests run deterministic training pipeline with various combinations of gradient accumulation factor and optimizer step limit then check that logged validation losses are the same.
You can run the tests as
```
cd speechbrain/tests/grad_acc_test/
sh test_grac_acc.sh
```
Two GPUs are required for testing due to DP/DDP scenarios.
I hope you’ll find this convenient

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -256,6 +256,16 @@ def parse_arguments(arg_list=None):
         help="Amount of time between saving intra-epoch checkpoints "
         "in minutes. If non-positive, intra-epoch checkpoints are not saved.",
     )
+    parser.add_argument(
+        "--grad_accumulation_factor",
+        type=int,
+        help="Number of batches to accumulate gradients before optimizer step",
+    )
+    parser.add_argument(
+        "--optimizer_step_limit",
+        type=int,
+        help="Number of optimizer steps to run. If not passed, all epochs are run.",
+    )
 
     # Accept extra args to override yaml
     run_opts, overrides = parser.parse_known_args(arg_list)
@@ -435,6 +445,8 @@ class Brain:
             "nonfinite_patience": 3,
             "noprogressbar": False,
             "ckpt_interval_minutes": 0,
+            "grad_accumulation_factor": 1,
+            "optimizer_step_limit": None
         }
 
         for arg, default in run_opt_defaults.items():
@@ -558,6 +570,7 @@ class Brain:
         # Prepare iterating variables
         self.avg_train_loss = 0.0
         self.step = 0
+        self.optimizer_step = 0
 
         # Add this class to the checkpointer for intra-epoch checkpoints
         if self.checkpointer is not None:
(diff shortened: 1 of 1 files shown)


TParcollet:
Hey! Thank you so much, these directions look cool and may interest @Gastron and @pplantinga

Gastron:
The looks good to me, but @TParcollet do you think we should update all the recipes that use gradient accumulation to this?

mravanelli:
I'm fine with these changes. 
- As for the gradient accumulation, we already considered adding it in the `fit_batch` function of core.py.  However, as you can see there is often the need to override the `fit_batch` method within the recipe script. For instance, in the [CommonVoice Transformer recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/transformer/train.py#L150), we need to override it because we have a learning rate annealing done at every update (not supported by the version in `core.py`). 
In this case, we have to re-implement the gradient accumulation again in train.py.

- As it is basically very hard to write a` fit_batch` function that meets everyone's needs, our approach was to keep in `core.py` a basic version of it and leave users all the freedom to customize  ` fit_batch`  based on their needs in train.py. Gradient descends only require a couple of lines of code, so it shouldn't be too hard to add.  Said that I'm fine adding it `core.py` as well if this is considered useful by the users.

So @Gastron, there is not need to update the existing recipes because they already have to override fit_batch.

Gastron:
Right! I think this makes sense to include, and I like how this PR deals with the two different but connected improvements. Merging!

RuABraun:
One slight issue I want to point out is that if one is comparing builds with different amounts of gradient accumulation then in, the current implementation, the loss will be scaled differently making it harder to compare.

The correct thing to do is to sum the loss across the different forward-backward steps (until one does one optimizer step). But I can understand avoiding that because it makes the code more ugly.

Gastron:
@RuABraun could you elaborate a little, I'm not sure if I understand. I think with this current scaling, with gradient accumulation factor `x`, the loss will be the sum of `x` batches' losses, each divided by `x`, which I think should keep the same scale as training without gradient accumulation. (I suppose this assumes loss reduction is mean-based or some such.)

RuABraun:
You're write sorry! misread the code

Gastron:
Phew, well thanks for double checking this!

speechbrain/core.py
```diff
@@ -839,24 +852,29 @@ def fit_batch(self, batch):
...
             self.optimizer.zero_grad()
             with torch.cuda.amp.autocast():
                 outputs = self.compute_forward(batch, Stage.TRAIN)
                 loss = self.compute_objectives(outputs, batch, Stage.TRAIN)
-            self.scaler.scale(loss).backward()
-            self.scaler.unscale_(self.optimizer)
-            if self.check_gradients(loss):
-                self.scaler.step(self.optimizer)
-            self.scaler.update()
+            self.scaler.scale(loss/self.grad_accumulation_factor).backward()
```
danpovey: If you might be using DDP you might want to consider no_sync().
https://github.com/pytorch/pytorch/issues/43201

Gastron: Thanks, looks like that would speed things up! @TParcollet I think you've had some recipes using both gradient accumulation and DDP, got time to add this?

Issue #1253, opened 2022-01-14
Speechbrain downloads pretrained models even if local files are passed


gabrielziegler3:
I noticed that when I ran the example for using ECAPATDNN's pretrained model after downloading the files, speechbrain will read the first two files from `filesystem` and then will start to request the other files from https://huggingface.co/speechbrain/spkrec-ecapa-voxceleb

To reproduce, run:

```python
import torchaudio
from speechbrain.pretrained import EncoderClassifier
classifier = EncoderClassifier.from_hparams(source="path/to/downloaded/model/dir")
signal, fs =torchaudio.load('samples/audio_samples/example1.wav')
embeddings = classifier.encode_batch(signal)
```

This behaviour may be seamless for most users, but if you do not have internet in an environment and just wants to load the files from filesystem you will not be able to do so with the `from_params` method.

I figured what was making speechbrain fallback for requests and I am opening a PR to fix this issue.

Gastron:
Actually I think the issue is in the `hyperparams.yaml` file in that particular repo (and unfortunately this pattern may occur elsewhere): https://huggingface.co/speechbrain/spkrec-ecapa-voxceleb/blob/main/hyperparams.yaml#L54-L58

There `hyperparams.yaml` explicitly specifies the location of the files. This overrides any default_source given to the Pretrainer. There is no need to explicitly specify the location, it would be more flexible if the paths weren't specified. @mravanelli can you remove those lines?

wbj0110:
When I load local file,It occur OSError: [Errno 62] Too many levels of symbolic links: 'pretrained_models/EncoderClassifier-3077685261458028297/embedding_model.ckpt',

`
 verification = SpeakerRecognition.from_hparams(source="/speaker_verification/tdnn/model/CKPT+2022-04-17+20-14-58+00",savedir="pretrained_models/CKPT+2022-04-17+20-14-58+00")

score, prediction = verification.verify_files("speechbrain/spkrec-ecapa-voxceleb/example1.wav", "speechbrain/spkrec-ecapa-voxceleb/example2.flac")
print(score,prediction)
 score, prediction = verification.verify_files("speechbrain/spkrec-ecapa-voxceleb/example1.wav", "speechbrain/spkrec-ecapa-voxceleb/example1.wav")
 print(score,prediction)
`

Adel-Moumen:
Hello, 

Any news regarding this issue please? Thanks. 

Adel

PR #1254, opened 2022-01-14
Fix bug where speechbrain would request files from HF even with local files provided

gabrielziegler3:
#1253  Signed-off-by: Gabriel Ziegler <gabrielziegler3@gmail.com>

### DIFF
--- a/speechbrain/utils/parameter_transfer.py
+++ b/speechbrain/utils/parameter_transfer.py
@@ -168,10 +168,11 @@ class Pretrainer:
         loadable_paths = {}
         for name in self.loadables:
             save_filename = name + PARAMFILE_EXT
+            filename = None
             if name in self.paths:
                 source, filename = self.split_path(self.paths[name])
-            elif default_source is not None:
-                filename = save_filename
+            if default_source:
+                filename = save_filename if not filename else filename
                 source = default_source
             else:
                 raise ValueError(


mravanelli:
@Gastron

gabrielziegler3:
I believe that this change alone doesn't fix the whole issue, but the lines I changed were the one making `fetch` fallback to requesting from `Hugging Face`.

mravanelli:
I'm fine with these changes. 
- As for the gradient accumulation, we already considered adding it in the `fit_batch` function of core.py.  However, as you can see there is often the need to override the `fit_batch` method within the recipe script. For instance, in the [CommonVoice Transformer recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/transformer/train.py#L150), we need to override it because we have a learning rate annealing done at every update (not supported by the version in `core.py`). 
In this case, we have to re-implement the gradient accumulation again in train.py.

- As it is basically very hard to write a` fit_batch` function that meets everyone's needs, our approach was to keep in `core.py` a basic version of it and leave users all the freedom to customize  ` fit_batch`  based on their needs in train.py. Gradient descends only require a couple of lines of code, so it shouldn't be too hard to add.  Said that I'm fine adding it `core.py` as well if this is considered useful by the users.

So @Gastron, there is not need to update the existing recipes because they already have to override fit_batch.

mravanelli:
What do you think of this PR @Gastron?

Adel-Moumen:
Up @Gastron, could you please have a look ?

Gastron:
Ooh this is an old PR, has somehow slipped past me, thanks for bringing it up again.

Gastron (commented):
This changes the if/else logic, and would unfortunately break the case where `default_source` is not specified.

It seems to me that this PR just stems from path specification that did not work. I wonder if it is a path like `"./filename.ckpt"`, perhaps that is currently broken?

PR #1257, opened 2022-01-17
fix tokenizer

Moumeneb1:
Hi, 

We employ a manual count of lines used in training in the current speechbrain tokenizer. In addition, the condition check has a minor flaw. Even if we modify parameters like the num_sequences after the txt file has already been generated one time, the changes will not take effect because we check if the file exists or not (which is already there in this case)

https://github.com/speechbrain/speechbrain/blob/e14d2c15874a22df018edf441af62b105b33ed39/speechbrain/tokenizers/SentencePiece.py#L160

SentencePiece provides built in parameters : 
**--input_sentence_size** (maximum size of sentences the trainer loads)  type: int32 default: 0
**--shuffle_input_sentence** (Randomly sample input sentences in advance. Valid when --input_sentence_size > 0)  type: bool default: true

We can sample more conveniently and randomly just using these parameters. I added these two parameters and removed num_sequences in this PR

@lorenlugosch @aheba

### DIFF
--- a/speechbrain/tokenizers/SentencePiece.py
+++ b/speechbrain/tokenizers/SentencePiece.py
@@ -66,9 +66,11 @@ class SentencePiece:
         If False, allow the sentencepiece to extract piece crossing multiple
         words. This feature is important for : Chinese/Japanese/Korean.
         (default: True)
-    num_sequences : int
-        If not none, use at most this many sequences to train the tokenizer
-        (for large datasets). (default: None)
+    input_sequences : int 
+        if > 0 input_sequences will be loaded from the dataset else all the dataset is loaded. (default: 0)
+    shuffle_input_sentence : bool 
+        Randomly sample input sentences. Valid when --input_sentence_size > 0.
+        default: true
     annotation_list_to_check : list,
         List of the annotation file which is used for checking the accuracy of
         recovering words from the tokenizer.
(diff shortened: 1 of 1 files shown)


lorenlugosch:
The reason I added `num_sequences` was that some recipes were running out of memory because we were using too many sequences to train the tokenizer. (Maybe the AISHELL recipe?)

I think I remember the problem still occurred when using SentencePiece's built-in parameter, but maybe I'm remembering that wrong. If your change works, let's go with it.

(And you're right that we should be shuffling!)

anautsch:
@Moumeneb1 any updates?
Please run `pre-commit run --all-files` - maybe a pull of the latest develop branch is necessary as well.

Adel-Moumen:
Dear @Moumeneb1,

Thank you for your contribution to our project through this pull request. Unfortunately, we have decided to close this pull request due to its staleness.

Since the initial submission, our project has undergone significant updates, and we've released several new versions. As a result, we kindly request you to create a new pull request with the latest commits to ensure compatibility and integration with the current codebase.

Please sync your fork with the latest changes from the main repository and address any potential conflicts that may arise during this process. We appreciate your understanding and continued interest in contributing to our project.

If you have any questions or encounter difficulties while creating the new pull request, feel free to reach out. We value your contributions and look forward to seeing an updated pull request from you.

Best regards,
Adel Moumen

PR #1261, opened 2022-01-18
Fix #1260: Return avg loss from evaluate() as promised

### LINKED ISSUE #1260: Inconsistency between documentation and the code of evaluate() function in the brain class
Hi I've noticed that there are some inconsistencies in the code and docs of the evaluate() method. 

The code shown here does not have a return value: https://speechbrain.readthedocs.io/en/latest/_modules/speechbrain/core.html#Brain.evaluate

And in the docs https://speechbrain.readthedocs.io/en/latest/API/speechbrain.core.html#speechbrain.core.Brain, it is mentioned that it returns average loss.

Gastron:
Fixes #1260

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1233,6 +1233,7 @@ class Brain:
                 self.on_stage_end, args=[Stage.TEST, avg_test_loss, None]
             )
         self.step = 0
+        return avg_test_loss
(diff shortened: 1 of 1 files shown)


Gastron:
Let's just merge this one line change.

PR #1263, opened 2022-01-18
CuPy Implementation of LiGRU (faster by more than 2x)

Adel-Moumen:
Hello, 

This PR aims to integrate a new implementation of the LiGRU with CuPy - a framework like Numba to write custom cuda kernels. 

I decided to only implement a LiGRU cell because it's the main bottleneck of the LiGRU currently. Implementing more than just the cell would have led to less customization of the neural network. 

I did not implement the forward with CuPy because it was leading to nan values during training.

I used CuPy instead of Numba because Numba was not a suitable option for this exact type of problem. I also tried Triton but the performances were not great because of a problem related to the for loop. (large CPU overhead of a kernel launch)

For the moment, three activations functions are supported: 
1. ReLU
2. Leaky ReLU
3. Tanh

In terms of results I got the following ones:
image
image
image

Also, I trained the model on Librispeech 960, in descending order, for **58 batches**:
1. LiGRU Vanilla: **3 mins**
2. LiGRU CuPy: **1 min and 15 secs** (speed up of more than **2x**)

and in ascending order, for **4011 batches**:
1. LiGRU Vanilla: **28 mins and 55 secs**
2. LiGRU CuPy: **21 mins and 36 secs**

PS: the choice of the number of batch is when the LiGRU Vanilla suffers from the instability (when the model has a loss of nan) 

According to my previous screenshot, the CuPy LiGRU is faster by **~2.4x** than the Jitted LiGRU and by **~4x** the vanilla LiGRU. 

Finally, I trained the CuPy version on TIMIT and I got a PER of **13.99**. 

This work has been followed by Assoc. Prof Titouan Parcollet. 

If you have any questions/query I will try to answer them as fast as possible and hopefully I did not any mistakes in the process of PR because it's my first time :) 

Best,
Adel 

PS: Happy new year

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ SoundFile; sys_platform == 'win32'
 torch>=1.8.0,<=1.10.1
 torchaudio>=0.8.0,<=0.10.1
 tqdm>=4.42.0
+cupy==10.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ SoundFile; sys_platform == 'win32'
 torch>=1.8.0,<=1.10.1
 torchaudio>=0.8.0,<=0.10.1
 tqdm>=4.42.0
-cupy==10.0.0
+cupy>=10.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 -r lint-requirements.txt
+cupy>=10.0.0
 huggingface_hub>=0.0.6
 hyperpyyaml>=0.0.1
 joblib>=0.14.1
@@ -11,4 +12,3 @@ SoundFile; sys_platform == 'win32'
 torch>=1.8.0,<=1.10.1
 torchaudio>=0.8.0,<=0.10.1
 tqdm>=4.42.0
-cupy>=10.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 -r lint-requirements.txt
-cupy>=10.0.0
 huggingface_hub>=0.0.6
 hyperpyyaml>=0.0.1
 joblib>=0.14.1
--- a/speechbrain/nnet/RNN.py
+++ b/speechbrain/nnet/RNN.py
@@ -1249,6 +1249,10 @@ class LiGRU_Layer(torch.nn.Module):
         ---------
         wx : torch.Tensor
             Linearly transformed input.
+        ht : torch.Tensor
+            Hidden state at timestep - 1.
+        drop_mask : torch.Tensor
+            dropout masks.
         """
         hiddens = []
 
@@ -1266,7 +1270,17 @@ class LiGRU_Layer(torch.nn.Module):
         return h
 
     def _ligru_cell(self, w, ht):
-        
+        """Returns the hidden states for each time step.
+        Choose the appropriate method to call based on the device of 
+        each tensors. 
+
+        Arguments
+        ---------
+        wx : torch.Tensor
+            Linearly transformed input.
+        ht : torch.Tensor
+            Hidden state at timestep - 1.
+        """
         # Sampling dropout mask
         drop_mask = self._sample_drop_mask(w)
 
--- a/speechbrain/nnet/RNN.py
+++ b/speechbrain/nnet/RNN.py
@@ -1279,7 +1279,7 @@ class LiGRU_Layer(torch.nn.Module):
         if w.is_cuda == ht.is_cuda == drop_mask.is_cuda == True:
             from speechbrain.nnet.ligru.cuda_ligru_cell import _ligru_cell_cupy
             
-            return _ligru_cell_cupy.apply(self._ligru_cell_jit, w, self.u.weight, ht, drop_mask, self.act)
+            return _ligru_cell_cupy.apply(self._ligru_cell_jit, w, self.u.weight, ht, drop_mask)
         else:
             return self._ligru_cell_cpu(w, ht, drop_mask)
 
--- a/speechbrain/nnet/RNN.py
+++ b/speechbrain/nnet/RNN.py
@@ -1032,7 +1032,6 @@ class LiGRU(torch.nn.Module):
         self.bidirectional = bidirectional
         self.reshape = False
 
-
         # Computing the feature dimensionality
         if len(input_shape) > 3:
             self.reshape = True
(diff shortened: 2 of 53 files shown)


Adel-Moumen:
**Update:** Within the advice of Mr. Titouan Parcollet, I investigate the forward pass of my LiGRU and found out that I can leverage the use of Jit even if I use Autograd. By doing so, I speed up the LiGRU even more. 

These are the results that I got using my own benchmark: 
image

image-2

image-3

On LibriSpeech 960: 
- 1 min 11 secs for 58 batches in descending 
- 21 min 25 secs for 4011 batches in ordering

requirements.txt
```diff
@@ -11,3 +11,4 @@ SoundFile; sys_platform == 'win32'
 torch>=1.8.0,<=1.10.1
 torchaudio>=0.8.0,<=0.10.1
 tqdm>=4.42.0
+cupy>=10.0.0
```
TParcollet, line 14: This is not correct. It has to be added to an extra_requirements.txt to any recipe using the LiGRU, but not on the main requirements. We do not want the users to depend on CuPy if they don't want to.

speechbrain/nnet/RNN.py
```diff
@@ -1199,13 +1196,16 @@ def __init__(
...
         if nonlinearity == "tanh":
             self.act = torch.nn.Tanh()
-        elif nonlinearity == "sin":
-            self.act = torch.sin
         elif nonlinearity == "leaky_relu":
             self.act = torch.nn.LeakyReLU()
         else:
             self.act = torch.nn.ReLU()
 
+        from speechbrain.nnet.ligru.cuda_ligru_cell import _ligru_cell_jit
```
TParcollet, line 1204: Hum, enforcing Jit is not following the philosophy of SB. In practice, using jit or not is defined in the YAML, see for instance [this recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/TIMIT/ASR/seq2seq/hparams/train.yaml). Here, it should be the same. If the encoder is passed to the (jit_module_keys) paremeter in the yaml THEN your stuff should be jitted, otherwise, nope. Maybe it's already the case and you don't have to add this wrapper.

speechbrain/nnet/RNN.py
```diff
@@ -1230,11 +1230,14 @@ def forward(self, x, hx: Optional[Tensor] = None):
             w_bn = self.norm(w.reshape(w.shape[0] * w.shape[1], w.shape[2]))
             w = w_bn.reshape(w.shape[0], w.shape[1], w.shape[2])
 
+        # Sampling dropout mask
+        drop_mask = self._sample_drop_mask(w)
```
TParcollet, line 1234: Why is the drop mask here now ?

speechbrain/nnet/RNN.py
```diff
@@ -1243,31 +1246,44 @@ def forward(self, x, hx: Optional[Tensor] = None):
 
         return h
 
-    def _ligru_cell(self, w, ht):
+    def _ligru_cell_cpu(self, w, ht, drop_mask):
```
TParcollet, line 1249: Why do we have a cpu and a non-cpu function? It's more of a normal forward or a cupy forward, no ? Naming might be misleading here

speechbrain/nnet/ligru/__init__.py
```diff
@@ -0,0 +1,2 @@
+"""Package containing LiGRU Cell with CuPy.
```
TParcollet, line 1: Add a bit more of context. Why do we need this package, what is he bringing and under what conditions? Do we need to install something first?

speechbrain/nnet/ligru/cuda_ligru_cell.py
```diff
@@ -0,0 +1,223 @@
+"""This library implements a LiGRU cell needed for the LiGRU.
+
+LiGRU is a Gated Recurrent Unit introduced by Assoc. Prof Mirco 
+Ravanelli in 2018 (see: https://arxiv.org/pdf/1803.10225.pdf).
```
TParcollet, line 4: Please add some infos on what must be installed on the header of this file as well.

speechbrain/nnet/ligru/cuda_ligru_cell.py
```diff
@@ -0,0 +1,223 @@
...
+            Recurrent weight. 
+        ht : torch.Tensor
+            Hidden state. 
+        drop_mask : torch.Tensor
+            Dropout mask. 
+        act : nn.Module
+            Activation Function. 
+        """
+        # save values for backward
+        hiddens        = []
```
TParcollet, line 79: I also prefer this style of alignement with Python. HOWEVER, this is not compliant with what SpeechBrain has adopted :-P "=" symbols must have one space before and after

TParcollet (changes requested):

mravanelli:
@TParcollet, as fas as I remember there is another implementation of the fast LiGRU that is under development, right?

TParcollet:
Indeed. @Adel-Moumen started a new internship. First, SLT submission. Then, we refactor and merge this code.

Adel-Moumen:
Hi @mravanelli, 

We (@TParcollet and I) developed a new Li-GRU called Li-GRU 2.0 which is a new version of the Li-GRU that is stable against the gradient exploding problem. Indeed, the Li-GRU 2.0 can be trained on medium to large ASR datasets such as LibriSpeech or CommonVoice while the Li-GRU was not even able to do one epoch of training due to the gradient exploding problem. Our work is currently being finalized and will be submitted at SLT 2022. 

As you know, I previously implemented the Li-GRU with CuPy which lead to a faster neural network in inference and backward than the jitted version. Unfortunately, this implementation is not compatible with our new Li-GRU 2.0 for some specific reasons. This is why I developed a CUDA implementation of the Li-GRU 2.0 that is five times faster than the Li-GRU, and faster than the custom implementation that I previously did with everything in c++ and bind with PyTorch thanks to a "custom op" (see: https://pytorch.org/tutorials/advanced/cpp_extension.html).

We were thinking about replacing the Li-GRU in SpeechBrain with our new Li-GRU 2.0 for the stability and the non-negligible gain of performance which would increase the adoption of the Li-GRU in the ASR community. Hence, we chose to close this PR and we will open a new one that would propose this new CUDA implementation when it will be ready. 

Let me know what you think about that 😁

Adel-Moumen:
Hi, to give you an example of what it might look like, in fairseq they did something similar: https://github.com/facebookresearch/fairseq/tree/main/fairseq/modules/lightconv_layer

PR #1269, opened 2022-01-21
Use of hashlib

stachu86:
The result of buil-in `hash` function depends on the value of environment variable [PYTHONHASHSEED](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED). If this variable is not set, by default random seed values will be used between executions, which results in creating multiple unnecesary copies in `./pretrained_models` direcotry.

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -80,7 +80,7 @@ def foreign_class(
         An instance of a class with the given classname from the given pymodule file.
     """
     if savedir is None:
-        savedir = f"./pretrained_models/{classname}-{hashlib.md5(source.encode('UTF-8')).hexdigest()}"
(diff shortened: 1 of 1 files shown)


speechbrain/pretrained/interfaces.py
```diff
@@ -327,7 +328,7 @@ def from_hparams(
         """
         if savedir is None:
             clsname = cls.__name__
-            savedir = f"./pretrained_models/{clsname}-{hash(source)}"
+            savedir = f"./pretrained_models/{clsname}-{hashlib.md5(source.encode('UTF-8')).hexdigest()}"
```
Gastron, line 331:
```suggestion
            savedir = f"./pretrained_models/{clsname}-{hashlib.md5(source.encode('UTF-8', errors='replace')).hexdigest()}"
```

Gastron (commented):
Thanks, good catch, I want to merge this PR, but now that we're here I guess we can use add a tolerant error handler for the string encoding.

Gastron:
Thanks again, LGTM

PR #1283, opened 2022-01-29
add HifiGAN vocoder for LJSpeech TTS

BenoitWang:
Hi @mravanelli ,

This is a hifi-gan vocoder PR as discussed which was started by @Chaanks. In this PR : 
 
1. An LJSpeech recipe is added with hifi-gan training scripts.
2. HifiGAN model added in lobes.models
3. Functions added in TensorBoardLogger to log audios and spectrograms (matplotlib imported for spectrogram)
4. A scalarize function added in data_utils

@Chaanks please feel free to check my modifications too.


recipes/LJSpeech/TTS/vocoder/hifi_gan/hparams/train.yaml
```diff
@@ -0,0 +1,179 @@
...
+save_folder: !ref <output_folder>/save
+train_log: !ref <output_folder>/train_log.txt
+epochs: 100
+keep_checkpoint_interval: 10
+
+#################################
+# Data files and pre-processing #
+#################################
+data_folder: !PLACEHOLDER # e.g, /datasets/ljspeech
+train_csv: !ref <save_folder>/train.csv
```
flexthink, line 16: If using CSV, training on phonemes will be more challenging - phoneme labels consist of more than one character.

speechbrain/utils/train_logger.py
```diff
@@ -200,3 +209,19 @@ def log_stats(
             self.run.log({**logs, **stats_meta}, step=step)
         else:
             self.run.log({**logs, **stats_meta})
+
+
+def plot_spectrogram(spectrogram, ap=None, fig_size=(16, 10), output_fig=False):
+    import matplotlib
+
+    matplotlib.use("Agg")
+    import matplotlib.pyplot as plt
```
flexthink, line 218: We may want to be careful about using the global `plt` here in a reusable library - it is better to draw on the axis object.

flexthink (commented):
Please see my comments

BenoitWang:
Hi @mravanelli @flexthink, I just added the tacotron part. 

1. tacotron recipe with text_to_sequence.py which is a text preprocessing pipeline from nvidia
2. Tacotron2 into lobes.models which contains the tacotron model, the TextMelCollate function and a tacotron logger
3. an attribute "half_life" added to StepScheduler in nnet.schedulers, together with a new "IntervalScheduler"
4. Tacotron and HifiGan share a same ljspeech_prepare.py, a same seed can be used to train the two in series

@flexthink Thanks and I'm going to modify the code according to your hifi-gan comments.

mravanelli:
Dear @BenoitWang, @flexthink, @Chaanks 
thank you for this excellent contribution!
I reviewed the code and already modified few minor things. The main issues that I see are the following:
1. I cannot run the tacotroon2 recipe. I got the following error:

```
  0%|                                                                                                                                                           | 0/184 [00:00<?, ?it/s]/network/scratch/r/ravanelm/py110_1/lib/python3.9/site-packages/torch/functional.py:568: UserWarning: torch.meshgrid: in an upcoming release, it will be required to pass the indexing argument. (Triggered internally at  ../aten/src/ATen/native/TensorShape.cpp:2228.)
  return _VF.meshgrid(tensors, **kwargs)  # type: ignore[attr-defined]
  0%|                                                                                                                                                           | 0/184 [00:13<?, ?it/s]
speechbrain.core - Exception:
  File "/network/scratch/r/ravanelm/speeechbrain_hifigan/recipes/LJSpeech/TTS/tacotron2/train.py", line 382, in <module>
    tacotron2_brain.fit(
  File "/network/scratch/r/ravanelm/speeechbrain_hifigan/speechbrain/core.py", line 1054, in fit
    loss = self.fit_batch(batch)
  File "/network/scratch/r/ravanelm/speeechbrain_hifigan/recipes/LJSpeech/TTS/tacotron2/train.py", line 78, in fit_batch
    result = super().fit_batch(batch)
  File "/network/scratch/r/ravanelm/speeechbrain_hifigan/speechbrain/core.py", line 872, in fit_batch
    (loss / self.grad_accumulation_factor).backward()
  File "/network/scratch/r/ravanelm/py110_1/lib/python3.9/site-packages/torch/_tensor.py", line 363, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "/network/scratch/r/ravanelm/py110_1/lib/python3.9/site-packages/torch/autograd/__init__.py", line 173, in backward
    Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [64, 80, 839]], which is output 0 of AsStridedBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

I don't know if this is relevant, but I'm working with the latest version of pytorch. Any idea?  I guess you @BenoitWang and @flexthink  were able to train it.

2. The HiFiGAN seems to start at least. However, it looks like at the end of the recipe there is an error. If I run:
`python train.py hparams/train.yaml --data_folder /localscratch/LJSpeech-1.1 --debug`
I see this error due to `sig_out = self.modules.generator.inference(x)`.

 ```
 File "/network/scratch/r/ravanelm/speeechbrain_hifigan/speechbrain/nnet/CNN.py", line 973, in remove_weight_norm
    self.conv = nn.utils.remove_weight_norm(self.conv)
  File "/network/scratch/r/ravanelm/py110_1/lib/python3.9/site-packages/torch/nn/utils/weight_norm.py", line 129, in remove_weight_norm
    raise ValueError("weight_norm of '{}' not found in {}"
ValueError: weight_norm of 'weight' not found in ConvTranspose1d(512, 256, kernel_size=(16,), stride=(8,), padding=(4,))
```

Note that here I'm using the debug modality, which simply means that we train the model with one batch over 1 epoch and then we move to test directly. Are you able to replicate this error?

3.` speechbrain/lobes/models/HifiGAN.py` =>  some docstrings are missing and some others are not in the expected format (see an example of expected docstring [here]( https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/nnet/CNN.py#L23)). For the main classes (e.g., HifiganGenerator, HifiganGenerator) I would suggest adding the example as well. This helps users a lot and we test it every time with the doctests.

4. It looks like the tensorboard is mandatory, while it should be optional (by default there should be `use_tensorboard=False`).

5. `vocoder/hifi_gan/train.py`=>It is not clear to me why we are not reusing the forward computations in fit_batch (it looks like we are redoing the same computations and we are not sing compute_forward). Am I missing something? It is also not clear to my why we say that "compute_forward" is "is only used for inference".

6.` LJSpeech/TTS/tacotron2/text_to_sequence.py` => some docstrings are missing and some others are not in the expected format (see an example of expected docstring [here]( https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/nnet/CNN.py#L23)).

mravanelli:
There are a couple of more things I noticed:
7. The HiFiGAN recipe does not create the `train_log.txt`. This is very useful for debugging and we create it in all the other recipes.
8. I started the training of Tacotroon, but the losses are in a scale very different from those that I can find in the model previously trained (see google drive):

old model
- `Epoch: 1, lr: 1.00e-03 - train loss: 1.29, train mel_loss: 1.26, train gate_loss: 5.65e-03, train attn_loss: 1.70e-02`

Current model (trained with pytorch 1.11)
- `Epoch: 1, lr: 1.00e-03 - train loss: 48.63, train mel_loss: 48.03, train gate_loss: 4.81e-01, train attn_loss: 1.13e-01`

It looks to me that this might be something related to a normalization thing. Any idea @BenoitWang?

BenoitWang:
Hi @mravanelli, thanks a lot for the comments. I have fixed some code according to your comments, please see some details as below:

issue1 & issue8 : Fixed. The problem comes from torch 1.11，it's not really clear to me, I think that there's a mask operation considered as an inplace operation, fixed [here](https://github.com/speechbrain/speechbrain/pull/1283/commits/f79b77f1431ecf617c70bc2fac2531d01f7d6820), for more similar discussions, an example can be found here https://github.com/NVlabs/FUNIT/issues/23#issuecomment-549099948

issue2 : Fixed in https://github.com/speechbrain/speechbrain/pull/1283/commits/3fb18475fd191995504d6f9f43bbe8a00ba508c6. The issue is that weight normalisation is added for the training stage and is removed for inference in the validation stage, and it should be re-added to the training stage for the next epoch. The original implementation does not do inference at the end of each epoch, but we do. So during the inference, a `deepcopy()` is used.

issue3 & issue6 : Fixed, docstrings added and put into correct format.

issue5 : Fixed, the idea was to show that the training is adversarial, but you are right that's a bit strange, so it's added in https://github.com/speechbrain/speechbrain/pull/1283/commits/9e52637f79b973ab2bfdfd6209bfaac6a05d9a6d

issue4 & issue7 : working on them now

BenoitWang:
Hi @mravanelli, sorry for the delay, I've done some fixes as below and normally it's ok for merging now.

1. change `tensorboard_logger` into an option by using a `use_tensorboard` flag :
    if `use_tensorboard: True`, tensorboard_logger is created and tensorboard should be installed
    else `use_tensorboard: False`, target and synthesized audios are saved in samples/epoch/ 
2. add  train_log.txt
3. use 500 epochs

mravanelli:
Thank you @BenoitWang for the last changes! Everything looks good to me now.  Thank you to all the other contributors (@flexthink @Chaanks). I will put it on Huggingface soon as well.

a897456:
> 4. It looks like the tensorboard is mandatory, while it should be optional (by default there should be `use_tensorboard=False`).

I don't think so. The default value in ‘train.yaml’ is false. When the epoch=26 is run, the checkpoint and sample wave are still not saved, wasting 10 hours. Is there something wrong with my operation?

PR #1290, opened 2022-02-06
Update Arabic language normalization rules.

alisafaya:
Current rules replaces diacritics with spaces. This causes the splitting of words with diacritics. The proposed simple modification prevents this.


TParcollet:
Okay ! Thanks.

TParcollet:
Okay ! Thanks.

PR #1293, opened 2022-02-07
Adding support for the pre-trained model WavLM

alisafaya:
This pull request provides support for the WavLM pre-trained model by microsoft.

### DIFF
--- a/speechbrain/lobes/models/huggingface_wav2vec.py
+++ b/speechbrain/lobes/models/huggingface_wav2vec.py
@@ -1,7 +1,9 @@
-"""This lobe enables the integration of huggingface pretrained wav2vec2 models.
+"""This lobe enables the integration of huggingface pretrained wav2vec2/hubert/wavlm models.
 
 Reference: https://arxiv.org/abs/2006.11477
 Reference: https://arxiv.org/abs/1904.05862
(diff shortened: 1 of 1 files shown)


TParcollet:
Hi and thanks. Does WavLM need special input-output norm ? As you can see in this class, we apply normalization in a "bad" way for now as it is done over the padding (will be changed later).

alisafaya:
Regarding the input norm, I believe you are fetching this config from the huggingface/transformers config file here:

https://github.com/speechbrain/speechbrain/blob/9d56d50809d8745cc7dafa930aec554145be4028/speechbrain/lobes/models/huggingface_wav2vec.py#L120

And in the original config file this option is disabled, as it is in the original implementation [here](https://github.com/microsoft/unilm/blob/d33e3b489cbad81a823c4bc1bbf102a24837190e/wavlm/WavLM.py#L177)

For the normalization on the output side, seems like they don't apply any normalization on that end. Check [here](https://github.com/microsoft/unilm/blob/d33e3b489cbad81a823c4bc1bbf102a24837190e/wavlm/WavLM.py#L364)

TParcollet:
I see, if you can update the code so the tests pass I'll try the PR :-) Thanks. We will also soon revise the HF integration for something better.

PR #1305, opened 2022-02-17
fixing model eval on list instead of class attribute

Hguimaraes:
### The problem
.eval() function is applied on a list object (**model** object)

### Solution
To solve this, similar to the class *FairseqWav2Vec2*, it should be applied to the class attribute *self.model*

It is a very small change but it will fix the usage of the wav2vec model based on the fairseq implementation.

### DIFF
--- a/speechbrain/lobes/models/fairseq_wav2vec.py
+++ b/speechbrain/lobes/models/fairseq_wav2vec.py
@@ -250,7 +250,7 @@ class FairseqWav2Vec1(nn.Module):
         self.model = model
         self.model = self.model[0]
         if self.freeze:
-            model.eval()
+            self.model.eval()
 
         # Randomly initialized layers if pretrain is False
         if not (pretrain):


TParcollet:
Hi ! I am not sure to understand what it implies. In practice it doesn't change anything, right?

Hguimaraes:
Hi @TParcollet,

It does change and avoids an error. I will try to explain better and give you the context to replicate the error.
The object `model` returned from `fairseq.checkpoint_utils.load_model_ensemble_and_task` in line 246 is a list, and one of its elements (The first) is actually the Pytorch sequential container (The model itself).

If you observe lines 250 and 251, it is created a class attribute `self.model = model`, and then it is re-assigned as `self.model = self.model[0]`.

In lines 252 and 253, the following operation is performed:
```python
if self.freeze:
    model.eval()
```
If you observe, it is applied on the original list, not in `self.model` (which is `model[0]`).
Since `.eval()` is not a method from the list object, it throws an error. For this reason, I changed to:

```python
if self.freeze:
    self.model.eval()
```
This change also makes this class compatible with the FairseqWav2Vec2.

mravanelli:
@TParcollet , what should we do with this PR?

TParcollet:
We merge it. Fairseq support will be dropped in the next version of SpeechBrain.

PR #1308, opened 2022-02-20
fixed a bug in the augmentation

SkuratovichA:
* Using `wavs` makes no sense because `wavs`  is not used anymore and SpecAugment() is applied on features and returns features, not wave signal.
A * small typo in the comment was fixed


TParcollet:
Indeed, thanks !

PR #1311, opened 2022-02-25
Fix libresspech lm dataset

Moumeneb1:
Hi, @TParcollet  it's the fix for using the dataset library in the LM librispeech experiment. We can now use the latest datasets library with no issues.


TParcollet:
@jamfly as you did the latest change, could you please verify if this works well for you as well ?

TParcollet:
Ok good thanks !

mravanelli:
What should we do with this PR?

TParcollet:
Based on @jamfly comment, we should not do anything and web for someone removing the need to use dataset from HF ...

PR #1312, opened 2022-02-25
wav2vec2 pretraining implemented with speechbrain

RuABraun:
(this is not meant for merge yet but to share the code)

This allows one to pretrain a wav2vec2 model without relying on fairseq or huggingface. It follows the fairseq implementation though there are various differences.

Here is a plot showing WER after finetuning vs different number of pretraining steps and comparing this implementation to fairseq (as well as without and with the quantisation). This is on italian commonvoice 7.0, using validated as the pretraining set, train for finetuning

pretrain_wer_comparison

I made a wrapper around the whole w2v2 object to make it easy to hold connecting objects like the masking tensor and various projection layers. The feature extractor and encoder are arguments that can be overridden. Using vector quantisation is toggle-able. The implementation follows the most recent fairseq implementation (but not the very recent conformer stuff) so uses normalise_before=True and layer norm is used instead of group norm.

One can use the existing finetuning script inside `/ASR/CTC` to train the pretrained model.

Minor notes and TODOs:
- The queue at my institute favors short jobs so I added a run_opts option "train_time_hours" so I can stop training after x hours (I have another wrapper that then takes care of resubmitting)
- Overloaded fit() for logging
- Overloaded update_average() because original relies on self.step which is reset every epoch plus I don't get the formula why not just use EMA
- Should I actually put the recipe in CommonVoice/self-supervised-learning ? Think some classes could maybe be moved to a better location
- Generally needs more docstrings etc.
- Made a change to make logging output more informative
- I use wandb so there is an option to use that, should change that later to use existing wandb-logger (I saw it but skipped using that for now)


RuABraun:
Latest commit is result of merging mine and guille's implementations. Best to ignore commonvoice files as those are not up-to-date and I'd like to focus on getting the librespeech implementation done before we update commonvoice.

recipes/LibriSpeech/SSL/hparams/train_wav2vec.yaml
```diff
@@ -0,0 +1,95 @@
...
+  num_vars: 320
+  temp_tuple: (2.0, 0.25, 0.999995,)
+  groups: 2
+  vq_dim: !ref <final_dim>
+
+model: !new:speechbrain.lobes.models.wav2vec.Wav2Vec2
+  embedding_dim: !ref <embedding_dim>
+  final_dim: !ref <final_dim>
+  quantiser: !ref <quantiser>
+  latent_extractor: !ref <extractor>
```
TParcollet, line 73: Do we really need to store the extractor and quantiser in the Wav2vec2 class ? I.e can't we compose the forward explicitly in the compute_forward function instead ? If this is not too messy it's usually better.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
+#!/usr/bin/env python3
+"""Recipe for pretraining wav2vec2. See config file for model definition.
+
```
TParcollet, line 3: Authors.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+
+PYTHON_VERSION_MAJOR = 3
+PYTHON_VERSION_MINOR = 7
+
+
+logger = logging.getLogger(__name__)
+
+
+@sb.utils.checkpoints.register_checkpoint_hooks
```
TParcollet, line 42: Why is this needed ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+        """Computes forward pass through wav2vec model and returns encoded and
+        target embeddings as well as other metrics of interest.
+        """
+        batch = batch.to(self.device)
+        wavs, wav_lens, padding_mask, mask = batch.sig
+        wavs, wav_lens = wavs.to(self.device), wav_lens.to(self.device)
+        padding_mask, mask = padding_mask.to(self.device), mask.to(self.device)
+
+        # normalisation already done in dataloader
+        results = self.modules.model(wavs, wav_lens, padding_mask, mask, is_pretraining=True,
```
TParcollet, line 54: See, it might look nice because we only have one line. However, the purpose of compute_forward is to actually expose the forward computation. Again, if this is possible, maybe it would be great to decompose the wav2vec 2 forward here instead.

TParcollet: For instance, we have absolutely no idea what are the outputs because of that.

recipes/LibriSpeech/SSL/hparams/train_wav2vec.yaml
```diff
@@ -0,0 +1,95 @@
...
+  latent_encoder: !ref <encoder>
+  dropout_context_input: 0.1
+  dropout_targets: 0.
+
+modules:
+  model: !ref <model>
+
+loss: !new:speechbrain.lobes.models.wav2vec.ContrastiveLoss
```
TParcollet, line 81: move the loss to the speechbrain.nnet.losses file ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+                    "temp": predictions["temp"],
+                }
+            )
+        return objectives
+
+    def fit_batch(self, batch):
+        should_step = self.step % self.grad_accumulation_factor == 0
+        # Managing automatic mixed precision
+        if self.auto_mix_prec:
+            with self.modules.model.no_sync() if not should_step else ExitStack():
```
TParcollet, line 91: This I don't understand, what is the matter ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+                    )
+                (backprop_loss / self.grad_accumulation_factor).backward()
+                objectives["total_loss"] = backprop_loss.detach()
+                if should_step:
+                    self.optimizer.step()
+                    self.optimizer.zero_grad()
+                    self.optimizer_step += 1
+
+        if should_step:
+            if isinstance(
```
TParcollet, line 142: Why is this needed ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+            ):
+                w2v_model = self.modules.model.module
+            else:
+                w2v_model = self.modules.model
+            if w2v_model.quantiser is not None:
+                w2v_model.quantiser.update_temp(
+                    self.optimizer_step
+                )
+
+            if (
```
TParcollet, line 153: Ok I understand the need for logging here. What I propose is to introduce something bigger directly in the core.py of SpeechBrain. I think that logging at every batch totally is relevant for many recipes. What about creating a callback in the official Brain class that is called at the end of fit_batch ? Then you could override it here to manage the logging.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+                hasattr(self.hparams, "log_interval")
+                and self.optimizer_step % self.hparams.log_interval
+                == 0
+            ):
+
+                log_dct = {
```
TParcollet, line 159: I guess that this is mandatory for your logging, but it's very hard to follow. Is it mandatory due to wandb?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+                        },
+                    )
+
+            self.hparams.lr_scheduler(
+                self.optimizer, self.optimizer_step
+            )
+
+        return objectives["loss"].detach()
+
+    def evaluate_batch(self, batch, stage):
```
TParcollet, line 196: why is it redefined ? I don't see a difference with the standard process of Brain.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+            )
+            if self.hparams.use_wandb:
+                wandb.log(
+                    {
+                        "valid_acuracy": stage_loss,
+                    },
+                    step=self.optimizer_step,
+                )
+
+    def update_average(self, loss, avg_loss):
```
TParcollet, line 214: what is this function necessary for ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+        if avg_loss == 0.0:
+            avg_loss = loss.item()
+        else:
+            avg_loss = 0.99 * avg_loss + 0.01 * loss.item()
+        return avg_loss
+
+
+def sample_negatives(y, num_neg):
```
TParcollet, line 222: This is something that could be used by things outside of wav2vec2 and this recipe. What about creating a new file in speechbrain.processing like speechbrain.processing.sequence_management or something like that. I can't think of a better name right now, but this function must be integrated to the SpeechBrain library itself

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+    @sb.utils.data_pipeline.provides("sig")
+    def audio_pipeline(wav):
+        info = torchaudio.info(wav)
+        sig = sb.dataio.dataio.read_audio(wav)
+        assert sig.dim() == 1, sig.dim()
+        resampled = torchaudio.transforms.Resample(
+            info.sample_rate, hparams["sample_rate"],
+        )(sig)
+        with torch.no_grad():
+            resampled = F.layer_norm(resampled, resampled.shape)
```
TParcollet, line 273: Isn't this a bottleneck cpu-wise ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+        return resampled
+
+    sb.dataio.dataset.add_dynamic_item(datasets, audio_pipeline)
+
+    sb.dataio.dataset.set_output_keys(datasets, ["id", "sig"])
+
+    train_sampler = DynamicBatchSampler(
+        train_data,
+        hparams["seconds_per_batch"],
+        num_buckets=70,
```
TParcollet, line 283: All paremeters -> hparams.yaml :-)

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+        length_func=lambda x: x["duration"],
+        shuffle=True,
+        batch_ordering="random",
+    )
+
+    W2VPaddedBatch_partial = functools.partial(W2VPaddedBatch,
```
TParcollet, line 289: What is that ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
...
+    hparams_file, run_opts, overrides = sb.parse_arguments(sys.argv[1:])
+
+    sb.utils.distributed.ddp_init_group(run_opts)
+
+    with open(hparams_file) as fin:
+        hparams = load_hyperpyyaml(fin, overrides)
+    hparams.update(run_opts)
+
+    # Gets hyperparams
+    dct = {
```
TParcollet, line 323: What is that ?

speechbrain/core.py
```diff
@@ -266,6 +266,12 @@ def parse_arguments(arg_list=None):
         type=int,
         help="Number of optimizer steps to run. If not passed, all epochs are run.",
     )
+    parser.add_argument(
```
TParcollet, line 269: Hum, this must be discussed internally, but I don't see any reason to have this. Stopping with respect to optimisation steps, ok, but not training time.

speechbrain/lobes/augment.py
```diff
@@ -529,3 +531,60 @@ def _prepare_csv(folder, filelist, csv_file, max_length=None):
                         )
     finally:
         sb.utils.distributed.ddp_barrier()
+
+
+class AddNoiseSimple(torch.nn.Module):
```
TParcollet, line 536: More context needed on what this is. Especially since it introduces a new dependency on turtle ?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -420,13 +422,16 @@ def __init__(
             ]
         )
         self.norm = sb.nnet.normalization.LayerNorm(d_model, eps=1e-6)
+        self.layerdrop_prob = layerdrop_prob
+        self.rng = np.random.default_rng(0)  # for DDP so same layer dropped
```
TParcollet, line 426: Replicability here: Should this be set with the seed of the yaml instead ? Always bad to have some constant in the code.

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+
+from speechbrain.lobes.models.transformer.Transformer import (
+    TransformerEncoder,
+    PositionalEncoding,
+)
+from speechbrain.utils.data_utils import batch_pad_right
+from torch.utils.data._utils.collate import default_convert
+from speechbrain.utils.data_utils import mod_default_collate
+from speechbrain.utils.data_utils import recursive_to
+from torch.utils.data._utils.pin_memory import (
```
TParcollet, line 19: Why do we need this ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+        self,
+        out_channels=[512, 512, 512, 512, 512, 512, 512],
+        kernel_sizes=[11, 3, 3, 3, 3, 3, 3],
+        strides=[5, 2, 2, 2, 2, 2, 2],
+        dropout=0.0,
+        init='kaiming',
+        extractor_mode="layer_norm",
+    ):
+        super().__init__()
+        if extractor_mode != "layer_norm":
```
TParcollet, line 48: What is the point of having a parameter if we only have one implementation ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+        """
+        if normalize_signal:
+            x = F.layer_norm(x, x.shape[1:])
+
+        results = {}
+        latents = self.latent_extractor(x)
+
+        latent_l2 = latents.float().pow(2.).mean()
+
+        with torch.cuda.amp.autocast(enabled=False):
```
TParcollet, line 249: Why ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+                    feats,
+                    self.mask_emb,
+                    padding_mask,
+                    mask_prob,
+                    mask_length,
+                    mask_channel_prob,
+                    mask_channel_length,
+                )
+            else:
+                with torch.cuda.amp.autocast(enabled=False):
```
TParcollet, line 276: Why ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+    def freeze_latent_extractor(self):
+        for param in self.latent_extractor.parameters():
+            param.requires_grad = False
+
+    def unfreeze_latent_extractor(self):
+        for param in self.latent_extractor.parameters():
+            param.requires_grad = True
+
+    @contextmanager
+    def no_sync(self):
```
TParcollet, line 344: ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+        """Dummy method to avoid if/else statements depending on if DDP used."""
+        yield
+
+
+class ContrastiveLoss(nn.Module):
```
TParcollet, line 349: Move to speechbrain.nnet.losses

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+
+def make_padding_mask(inp_shape, abs_lens):
+    """False when to be used, following speechbrain convention """
+    padding_mask = (
+        torch.arange(inp_shape[1], device=abs_lens.device)[None, :] >= abs_lens[:, None]
+    )
+    return padding_mask
+
+
+def compute_mask(shape, padding_mask, mask_prob, mask_length):
```
TParcollet, line 400: Could this also be a more general util function ? I don't think it is specific to w2v2. ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+            arange = np.delete(arange, mask_idc)
+            extra_indcs = np.random.choice(
+                arange, num_mask_missing, replace=False
+            )
+            mask[i, extra_indcs] = True
+        mask[i, mask_idc] = True
+    return mask
+
+
+def apply_mask(
```
TParcollet, line 446: The entire masking pipeline is hard to follow, we will need lot of documentation

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+        super().__init__()
+        self.num_pad_remove = 1 if num_conv_pos_embeddings % 2 == 0 else 0
+
+    def forward(self, hidden_states):
+        if self.num_pad_remove > 0:
+            hidden_states = hidden_states[:, :, : -self.num_pad_remove]
+        return hidden_states
+
+
+PaddedData = collections.namedtuple("PaddedData", ["data", "lengths", "padding_mask"
```
TParcollet, line 492: ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+def get_output_lengths(input_lengths, kernel_sizes, strides):
+    def _conv_out_length(input_length, kernel_size, stride):
+        return torch.floor((input_length - kernel_size) / stride + 1)
+
+    for kernel_size, stride in zip(kernel_sizes, strides):
+        input_lengths = _conv_out_length(input_lengths, kernel_size, stride)
+    return input_lengths.to(torch.long)
+
+
+class W2VPaddedBatch(PaddedBatch):
```
TParcollet, line 505: What is that ?

speechbrain/nnet/CNN.py
```diff
@@ -351,6 +351,7 @@ def __init__(
         bias=True,
         padding_mode="reflect",
         skip_transpose=False,
+        conv_init='',
```
TParcollet, line 354: Would, None be better ?

speechbrain/utils/distributed.py
```diff
@@ -118,19 +118,19 @@ def ddp_init_group(run_opts):
...
-                "experiment.py hyperparams.yaml --distributed_launch=True "
-                "--distributed_backend=nccl"
-            )
-        else:
-            if run_opts["local_rank"] + 1 > torch.cuda.device_count():
-                raise ValueError(
-                    "Killing process " + str() + "\n"
-                    "Not enough GPUs available!"
-                )
+        # if "local_rank" not in run_opts:
```
TParcollet, line 121: ?

speechbrain/utils/logger.py
```diff
@@ -100,7 +100,7 @@ def setup_logging(
         recursive_update(config, overrides)
         logging.config.dictConfig(config)
     else:
-        logging.basicConfig(level=default_level)
+        logging.basicConfig(level=default_level, format='%(asctime)s | %(module)s-%(lineno)d | %(levelname)s | %(message)s')
```
TParcollet, line 103: Why is this needed ?

TParcollet (changes requested):
Thanks for the huge and awesome work ! Please find my major comments on this review :-)

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,398 @@
+#!/usr/bin/env python3
+"""Recipe for pretraining wav2vec2. See config file for model definition.
+
+To run this recipe call python train.py hparams/train_it_wav2vec.yaml --find_unused_parameters --max_grad_norm 0.0
```
gcambara, line 4:
```suggestion
To run this recipe call python train.py hparams/train_wav2vec.yaml --find_unused_parameters --max_grad_norm 0.0
```

recipes/LibriSpeech/librispeech_prepare.py
```diff
@@ -284,6 +284,9 @@ def create_csv(
     """
     # Setting path for the csv file
     csv_file = os.path.join(save_folder, split + ".csv")
+    if os.path.exists(csv_file):
+        logger.info('Csv file %s already exists, not recreating.' % csv_file)
+        return
 
     # Preliminary prints
     msg = "Creating csv lists in  %s..." % (csv_file)
```
gcambara, line 292: We should push the symbolic link to "librispeech_prepare.py" under SSL, otherwise this script won't be found by default.

recipes/LibriSpeech/SSL/hparams/train_wav2vec.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+train_splits: ["train-clean-100", "train-clean-360", "train-other-500"]
+dev_splits: ["dev-clean"]
+test_splits: ["test-clean", "test-other"]
+train_csv: !ref <output_folder>/train.csv
+valid_csv: !ref <output_folder>/dev-clean.csv
+test_csv:
+   - !ref <output_folder>/test-clean.csv
+   - !ref <output_folder>/test-other.csv
+skip_prep: true
```
gcambara, line 15:
```suggestion
skip_prep: False
```

Otherwise won't work plug and play.

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+    """Wraps transformer and positional encoder.
+    """
+
+    def __init__(
+        self,
+        embedding_dim=768,
+        num_layers=8,
+        d_ffn=3072,
+        nhead=8,
+        dropout=0.1,
```
gcambara, line 146:
```suggestion
        dropout=0.1,
        activation=nn.GELU,
```
Now the encoder is using ReLU, but the default activation in the wav2vec2.0 paper is GELU.

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+    ):
+        # TODO: attention_dropout not used
+        super().__init__()
+        self.embedding_dim = embedding_dim
+        self.encoder = TransformerEncoder(
+            num_layers=num_layers,
+            nhead=nhead,
+            d_ffn=d_ffn,
+            d_model=embedding_dim,
+            dropout=dropout,
```
gcambara, line 159:
```suggestion
            dropout=dropout,
            activation=nn.GELU,
```

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+class W2VLatentEncoder(nn.Module):
+    """Wraps transformer and positional encoder.
+    """
+
+    def __init__(
+        self,
+        embedding_dim=768,
+        num_layers=8,
```
gcambara, line 143:
```suggestion
        num_layers=12,
```
It is the number of layers in the wav2vec2.0 BASE model.

recipes/LibriSpeech/SSL/train.py
RuABraun: Yes after thinking about it I completely agree with you! The quantiser is not necessary for inference so it makes more sense to keep it seperate from the wav2vec object I think

recipes/LibriSpeech/SSL/train.py
RuABraun: When doing DDP the gradients are synced on the backward pass, but when doing gradient accumulation every K steps then the syncing is only necessary every K steps and it's wasteful to do it on every backward pass. dan povey pointed it out in this PR https://github.com/speechbrain/speechbrain/pull/1251

the with .. if .. else ExitStack() is syntax so that the `with` statement is only executed when the condition is true

recipes/LibriSpeech/SSL/train.py
RuABraun: when wrapping a module with `DistributedDataParallel` one does not have access to the module's attributes anymore, so depending on the whether DDP is used or not the API changes from `normal_module.attribute` to `ddp_module.module.attribute`

It's ugly I agree but don't know of a better approach https://discuss.pytorch.org/t/access-to-attributes-of-model-wrapped-in-ddp/130572

recipes/LibriSpeech/SSL/train.py
RuABraun: Do you mean it's unclear because it's not clear why I'm doing this? The idea here is I just log everything (it's not needed because of wandb). Another option is to be more explicit like `logger.info(f'Step: {self.optimizer_step} - loss: {objectives["loss"]} - prob_perplexity: {objectives["prob_perplexity"]} - latent_l2: {objectives["latent_l2"} - lr: {lr} - temp: {temp} ....)`

But that is a lot more verbose and less robust if I want to add or remove something from logging.

recipes/LibriSpeech/SSL/train.py
RuABraun: I think this is something worth discussing, at the moment `evaluate_batch` is assumed to only return one value, but I would actually want to return both the loss and the accuracy (but this breaks because core.py will then try and use the return value to calculate the average loss)

you're right this change at the moment is pointless, when I was overriding `fit()` I had some changes here that I now removed.

speechbrain/lobes/models/wav2vec.py
gcambara: Since there are other wav2vec2.0 variants that use Group Norm, we use that argument as a placeholder. But I agree that maybe we could just erase that for simplicity, until we finally implement Group Norm or other variants.

recipes/LibriSpeech/SSL/train.py
RuABraun: The default `update_average` [link](https://github.com/speechbrain/speechbrain/blob/a9b373a9fef3fdfc4e3b4a41478f7b283866186f/speechbrain/core.py#L1239) has two issues for me:

1. it uses `self.step` which is reset every epoch which seems wrong to me (why would you want to have such a reset after every epoch?)
2. it's not clear to me why/how the implementation works

so I used a simple exponential moving average

recipes/LibriSpeech/SSL/train.py
RuABraun: will change this is a mistake, only commonvoice needs this

speechbrain/lobes/augment.py
RuABraun: mistake, was trying some stuff out and forgot to exclude

recipes/LibriSpeech/SSL/train.py
RuABraun: So I noticed a significant speedup from moving the mask creation into the dataloader. This is what huggingface does and is I suspect the major reason why you find it to be faster than fairseq.

But for the mask creation I need to know some information about the model like the size and number of kernels+strides. So I need to pass that information to a custom collate class (`W2VPaddedBatch`) of the dataloader. `functools.partial` allows me to pass that information to the class without actually initialising it (by setting some specific arguments)

speechbrain/lobes/models/wav2vec.py
RuABraun: yea this is kind of silly you're right

speechbrain/utils/logger.py
RuABraun: By default the output from a message "bla bla" will look like:
```
INFO:root:bla bla
```
With this change a lot more useful information (notice it also prints the module name and line number) is added to the output, now it looks like:
```
2022-04-05 12:14:15,431 | test-4 | INFO | bla bla
```

recipes/LibriSpeech/SSL/train.py
RuABraun: I was overwriting _save and _recover before, think it can be removed now, will do

recipes/LibriSpeech/SSL/train.py
RuABraun: Yes I like that idea (I agree putting logging in fit_batch is awkward), will do

recipes/LibriSpeech/SSL/train.py
RuABraun: (going to do these changes locally and then make a big commit)

done.

recipes/LibriSpeech/SSL/hparams/train_wav2vec.yaml
RuABraun: good catch, done

recipes/LibriSpeech/SSL/train.py
gcambara: I'd like to discuss a few things about this.

1. I understand about the advantages of exposing the forward in compute_forward, but at the same time I like the idea that the pretrain forward logic is self-contained within the wav2vec2 class. Someone could simply import our SpeechBrain wav2vec2 module into another framework and easily do the pretrain forward with our "is_pretraining" flag. In any case, I see pros/cons in both approaches, what do you think?

2. Although the quantizer is not needed for inference, I'd suggest keeping it when we release our models, whether it is stored in the wav2vec object or not. I find that it is interesting to enable the community to play around with it, especially for research purposes (I have done explorations on HuggingFace's w2v2 quantizers myself).

recipes/LibriSpeech/SSL/train.py
RuABraun: done

speechbrain/lobes/augment.py
RuABraun: done

speechbrain/lobes/models/wav2vec.py
RuABraun: done

recipes/LibriSpeech/SSL/hparams/train_wav2vec.yaml
RuABraun: done

recipes/LibriSpeech/SSL/train.py
RuABraun: for my wandb will remove

speechbrain/lobes/models/wav2vec.py
RuABraun: need for W2VPaddedBatch answered elsewhere

speechbrain/lobes/models/wav2vec.py
RuABraun: had some issues in the past without, might not be necessary, need to investigate more

speechbrain/lobes/models/wav2vec.py
RuABraun: done

speechbrain/lobes/models/wav2vec.py
RuABraun: when wrapped in DDP this method exists. added this method so that when I call it I don't have to do an awkward check for if the module is DDP or not

speechbrain/lobes/models/wav2vec.py
RuABraun: answered elsewhere

speechbrain/nnet/CNN.py
RuABraun: done

recipes/LibriSpeech/librispeech_prepare.py
RuABraun: done

speechbrain/core.py
RuABraun: Will remove this later

RuABraun:
I've refactored a lot, cleaned things up, added more docstrings, think it's worth going through another review.

Some things I still plan to do:
- I think the `W2VLatentEncoder` object is unnecessary, plan to remove and instead just use `TransformerEncoder` directly
- not 100% sure the changes I made to `DynamicBatchSampler` are necessary, need to look into it more
- remove the train_time_hours option I added
- probably should add more info to the README

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+    -------
+    wavs_padded : torch.Tensor, shape (B, T)
+        Audio arrays with right-sided padding.
+    wav_lens : torch.Tensor, shape (B,)
+        For each sample the percentage of the array that is not padding.
+    mask : torch.Tensor, shape (B, T)
+        Boolean mask to mask frames.
+    """
+    wav_lst, latent_length_lst = [], []
+    # If the sample is longer than `max_dur` create several shorter samples from it.
```
TParcollet, line 387: Wait, the fact that it will split the data will affect the batch size, correct ? So it bypasses Dynamic Batching, in the sense that the number of batches won't correspond ?

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+        (bs, batch_time_len,), latent_length_lst, mask_prob, mask_length
+    )
+    return (
+        torch.as_tensor(wavs_padded),
+        torch.as_tensor(wav_lens),
+        torch.as_tensor(mask, dtype=torch.bool),
+    )
+
+
+def _setup_wandb(hparams):
```
TParcollet, line 428: Did you try this with multinode training ? If not, we should, because this often break when multinode training is used.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+            logger.info(
+                f"Creating new ID {id} and name {name} for wandb, putting in file {id_file}"
+            )
+            with open(id_file, "w") as fh:
+                fh.write(f"{id} {name}")
+
+
+def main():
+    logger.setLevel(logging.INFO)
+    print(sys.argv[1:])
```
TParcollet, line 464: remove the print :p

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+            "dev_splits": hparams["dev_splits"],
+            "te_splits": hparams["test_splits"],
+            "save_folder": hparams["output_folder"],
+            "merge_lst": hparams["train_splits"],
+            "merge_name": "train.csv",
+            "skip_prep": hparams["skip_prep"],
+        },
+    )
+
+    if hparams["use_wandb"]:
```
TParcollet, line 495: Can we make wandb compliant with offline mode here ? I mean, 99% of large cluster don't have access to internet on the compute nodes

speechbrain/core.py
```diff
@@ -1078,36 +1096,41 @@ def fit(
                             self._save_intra_epoch_ckpt()
                         last_ckpt_time = time.time()
 
-            # Run train "on_stage_end" on all processes
```
TParcollet, line 1081: What are all these changes on the core.py, I don't understand? This is an absolutely critical place on SB.

speechbrain/core.py
```diff
@@ -1078,36 +1096,41 @@ def fit(
...
-                        if self.debug and self.step == self.debug_batches:
-                            break
-
-                    # Only run validation "on_stage_end" on main process
-                    self.step = 0
-                    run_on_main(
-                        self.on_stage_end,
-                        args=[Stage.VALID, avg_valid_loss, epoch],
-                    )
+            train_time = time.time() - last_ckpt_time
```
TParcollet, line 1130: This feature shouldn't be part of the PR, please remove :-) We can discuss it elsewhere, but it requires to consult the whole team.

speechbrain/core.py
```diff
@@ -1236,6 +1259,36 @@ def evaluate(
         self.step = 0
         return avg_test_loss
 
+    @contextmanager
+    def no_sync(self, use=True):
```
TParcollet, line 1263: Is this affecting the training time? If so, it also looks like something that should be put on an other PR because it affects ALL the recipes and not just this one.

speechbrain/dataio/sampler.py
```diff
@@ -10,6 +10,7 @@
 """
 import torch
```
TParcollet, line 11: Ok, could you elaborate on why this is necessary? I remember the issue, but at the same time, we have something that wraps the sampler automatically to prevent this to happen, no ? In which use cases our wrapping isn't sufficient ?

speechbrain/lobes/models/convolution.py
```diff
@@ -133,6 +138,8 @@ def __init__(
                 stride=stride if i == num_layers - 1 else 1,
                 dilation=dilation,
                 layer_name=f"conv_{i}",
+                bias=conv_bias,
+                padding=padding,
```
TParcollet, line 142: I am not sure about the padding. Padding is managed by SpeechBrain, this is why we don't provide it. No ?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -420,6 +422,8 @@ def __init__(
             ]
         )
         self.norm = sb.nnet.normalization.LayerNorm(d_model, eps=1e-6)
+        self.layerdrop_prob = layerdrop_prob
+        self.rng = np.random.default_rng(0)  # for DDP so same layer(s) are dropped
```
TParcollet, line 426: can't we use a given seed instead? Problem is that the same layer will ALWAYS be dropped -> find a way to use self._seed instead ?

speechbrain/lobes/models/wav2vec.py
```diff
@@ -0,0 +1,315 @@
+import math
```
TParcollet, line 1: Function like negative sampling and collate function can go there, if you think that they are only linked to w2v2

speechbrain/nnet/losses.py
```diff
@@ -1128,3 +1128,47 @@ def nll_loss_kd(
     # Loss averaging
     loss = torch.sum(loss.reshape(N_snt, max_len) * mask) / torch.sum(mask)
     return loss
+
+
+class ContrastiveLoss(nn.Module):
```
TParcollet, line 1133: Where did you get your inspiration from ? If the code is taken from Fairseq or HF, we must mention it.

speechbrain/nnet/quantisers.py
```diff
@@ -0,0 +1,107 @@
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+
+class GumbelVectorQuantizer(nn.Module):
```
TParcollet, line 6: Where did you get your inspiration from ? If the code is taken from Fairseq or HF, we must mention it.

speechbrain/utils/distributed.py
TParcollet: Still ?

TParcollet:
Also please, resolve the conflict so that I can run the tests :p

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+            )
+            logger.info(log_str)
+
+            if self.hparams.use_wandb:
+                run_on_main(
+                    wandb.log,
+                    kwargs={"data": log_dct, "step": self.optimizer_step,},
+                )
+
+    def evaluate_batch(self, batch, stage):
```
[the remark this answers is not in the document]
RuABraun: Because accuracy is more meaningful and seems to correspond better to actual performance.
w2v_loss_acc

In the image above it is the model with the better accuracy (and worse loss) that achieves better WER after finetuning.

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+
+    def update_average(self, loss, avg_loss):
+        if avg_loss == 0.0:
+            avg_loss = loss.item()
+        else:
+            avg_loss = 0.99 * avg_loss + 0.01 * loss.item()
+        return avg_loss
+
+
+def sample_negatives(y, num_neg):
```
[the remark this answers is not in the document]
RuABraun: will do

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+        collate_fn=w2v_mask_collate_fn_partial,
+        num_workers=hparams["test_dataloader_options"]["num_workers"],
+        batch_size=hparams["test_dataloader_options"]["batch_size"],
+        pin_memory=True,
+    )
+
+    return train_loader, valid_loader
+
+
+def w2v_mask_collate_fn(
```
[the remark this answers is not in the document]
RuABraun: I really think that for this it makes more sense to keep it specific to w2v. We can always move it if another recipe does end up using it.

recipes/LibriSpeech/SSL/train.py
RuABraun: I always do multinode training

recipes/LibriSpeech/SSL/train.py
RuABraun: will do

recipes/LibriSpeech/SSL/train.py
RuABraun: what do you mean? if this is false then wandb wont be used (so that should be fine for offline mode)

speechbrain/core.py
RuABraun: will do

recipes/LibriSpeech/SSL/librispeech_prepare.py
```diff
@@ -0,0 +1 @@
+../librispeech_prepare.py
```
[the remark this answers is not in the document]
RuABraun: will do

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+        """Computes forward pass through wav2vec model and returns encoded and
+        target embeddings as well as other metrics of interest.
+        """
+        wavs, wav_lens, mask = batch
+        wavs, wav_lens, mask = (
+            wavs.to(self.device),
+            wav_lens.to(self.device),
+            mask.to(self.device),
+        )
+        B = wavs.size(0)
```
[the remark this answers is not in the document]
RuABraun: will change

speechbrain/lobes/models/convolution.py
RuABraun: This follows fairseq where the input to the feature extractor is not padded. I followed that because I think it makes sense to not bother doing SSL training with representations that are on the "edge".

speechbrain/lobes/models/transformer/Transformer.py
RuABraun: Does it really make sense to have a seed variable? It doesn't seem to me like one would ever want to change it (so a constant seems fine to me).

speechbrain/lobes/models/wav2vec.py
RuABraun: okay will move the collate function

speechbrain/nnet/losses.py
RuABraun: will do

speechbrain/nnet/quantisers.py
RuABraun: will do

TParcollet:
Hi @RuABraun what are the next steps to follow in your view?

RuABraun:
There's the PR #1449 that needs to be merged first. 

I recently figured out one thing that was causing bad performance, but there's still one more issue remaining (get significantly better performance training with 2 GPUs vs with 4). Hope to figure that out in the next ~2 weeks (was busy with interviews in the last ~6 weeks but thankfully that's over now).

Got a few interesting results from trying out different things, will post about them in the slack soon.

speechbrain/nnet/losses.py
RuABraun: done

speechbrain/nnet/quantisers.py
RuABraun: done

RuABraun:
waiting on #1518

recipes/LibriSpeech/SSL/train.py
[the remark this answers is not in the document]
RuABraun: done

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+        results = self.modules.latent_encoder(
+            latents,
+            mask=mask,
+            wav_lens=wav_lens,
+            disable_halfprec=disable_halfprec,
+        )
+
+        embeddings = results["embeddings"]
+
+        embeddings = embeddings[mask]
```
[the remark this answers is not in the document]
RuABraun: Can you explain why you think one can't use boolean masks?

recipes/LibriSpeech/SSL/librispeech_prepare.py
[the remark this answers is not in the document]
RuABraun: done

recipes/LibriSpeech/SSL/train.py
```diff
@@ -0,0 +1,518 @@
...
+        w2v_model.quantiser.update_temp(self.optimizer_step)
+
+        self.hparams.lr_scheduler(self.optimizer, self.optimizer_step)
+
+        if (
+            hasattr(self.hparams, "log_interval")
+            and self.optimizer_step % self.hparams.log_interval == 0
+        ):
+
+            log_dct = {
```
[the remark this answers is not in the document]
RuABraun: done

recipes/LibriSpeech/SSL/train.py
RuABraun: done

RuABraun:
have updated the PR, before merging I need to attach outputs of a build, think it would be nice if someone else could do the pretraining to confirm it works!

TParcollet:
Hi @RuABraun I would remove this logging or put it only on the log.log file (changing the logger). Also, please bring the progress bar back :-) 
Capture d’écran 2022-08-11 à 17 19 21

RuABraun:
@TParcollet  I think the current approach is best for the following reasons.

About the logging:
There are several relevant metrics (loss, accuracy, diversity loss at the minimum) that one wants to track. Is is important for when trying things out to frequently see what these are. Logging these once per epoch is not enough, everybody serious trains on at least thousands of hours, waiting until the end of an epoch to get metrics would take too long.

About the progressbar:
Most people who have the resources to train this will be training on a queue (sungrid engine, slurm etc.). This is the case both in industry and in academia. When you submit jobs on a queue you don't have an interactive commandline interface anymore, instead all the output you would normally see in your terminal will be redirected to a file. Progressbars do not work well with files as output. Additionally, progress-bars cannot display all the different metrics that you care about.

Therefore, I think it is best to have the current behaviour as default.

TParcollet:
Hi @RuABraun I hear what you say and partly agree. I think we should keep this intra-step logging, but not in the current form. Can we make it more human-readable i.e. not displaying as a list or dict? Also, note that the progress bar works if you cat the file instead of "vim"-it :p At least, this is what we do on our supercomputer. 

When building a toolkit, it is important to remain consistent, always and forever. All our recipes have progress bars. Now, in this particular case, I agree with you. If we can make the per_step logging more readable, I will be happy with it. Note: we should also bring the epoch_percentage back to the log if the progress bar is removed.

RuABraun:
Ah I didn't know about the vim/cat difference, cheers. I like your point about an epoch_percentage!

Not sure how to make it more human-readable though.. I could make it so the decimal numbers are rounded (instead of 0.63242835256...). Not sure if that's enough?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -440,6 +442,8 @@ def __init__(
             ]
         )
         self.norm = sb.nnet.normalization.LayerNorm(d_model, eps=1e-6)
+        self.layerdrop_prob = layerdrop_prob
+        self.rng = np.random.default_rng()
```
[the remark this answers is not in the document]
RuABraun: So we do want this randomness. If the seed were set then the layers dropped will be the same across nodes and that will cause a large drop in performance (notice I do not have a seed in the yaml).

Even if we were to take the rank (node ID) into account while setting the seed, this leads to problems when restarting a build because you are going to be using the same layerdrop and dropout masks as when you started training. Properly setting the seed for DDP training that allows restarting training is not trivial, a hacky solution is possible but I preferred keeping the code clean.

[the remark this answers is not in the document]
TParcollet: I understand, but a huge point with SB is that everything should be replicable. Can we replace numpy with torch.rand so users don't have to install numpy ?

TParcollet:
Rounded numbers and a proper formatted string ? I didn't check but is it also printed in the log.log file ? I guess I could do all this once I am back to work in October:-)

RuABraun:
No problem I can do that!

recipes/LibriSpeech/ASR/CTC/train_sb_wav2vec.py
```diff
@@ -0,0 +1,397 @@
...
+        hparams=hparams,
+        run_opts=run_opts,
+        checkpointer=hparams["checkpointer"],
+    )
+
+    # We dynamicaly add the tokenizer to our brain class.
+    # NB: This tokenizer corresponds to the one used for the LM!!
+    asr_brain.tokenizer = label_encoder
+
+    pretrained_model_path_extract = (
```
TParcollet, line 367: This loading seems a bit unusual compared to what we do. Can't we just only use the checkpointer ? Or PreTrainer ?

TParcollet:
Left to be done before merging:
— Finish the training of the BASE on libri and verify perf (add to YAML)
— Upload the model
— Verify the logging of the new refactored version
— Verify pretraining of refactored version
— Verify fine-tuning of refactored version

TParcollet (approved):
Let's go, only need to test now!

TParcollet:
@RuABraun All green! Now, just need to verify that I did not break everything :D

RuABraun:
Awesome! :) let's get it finished

RuABraun:
Btw I think we could actually just use the scheduler from #1537 as it seems to me it's equivalent.

TParcollet:
Well it's a two steps scheduler vs 3 steps. I guess all have reasons to exist ?

RuABraun:
Ah nevermind you're right, for some reason I thought it was 3 step.

TParcollet:
What an amazing PR! Thank you so much @RuABraun and @gcambara! We are finally done!

RuABraun:
@TParcollet Noting two things down here to look at in the future for better performance:

1. No weight decay on biases and layernorm parameters

2. Finetuning on our model just modifies the transformer encoder, the huggingface model finetuning seems to also adjust the convolutional model.

PR #1337, opened 2022-03-21
VAD: fix edge cases, starting speech at end frames

### LINKED ISSUE #1108: VAD.get_boundaries RuntimeError: shape '[-1, 2]' is invalid for input of size 5
Hello,
the code:
```
VAD = VAD.from_hparams(source="speechbrain/vad-crdnn-libriparty", savedir="pretrained_models/vad-crdnn-libriparty")
prob_chunks = VAD.get_speech_prob_file(audio_file)
prob_th = VAD.apply_threshold(prob_chunks, activation_th=0.5, deactivation_th=0.25).float()
boundaries = VAD.get_boundaries(prob_th)
VAD.save_boundaries(boundaries, audio_file=audio_file)
```
reports runtime error in `boundaries = VAD.get_boundaries(prob_th)`,
`indexes = (prob_th == 1).nonzero()[:, 1].reshape(-1, 2)
RuntimeError: shape '[-1, 2]' is invalid for input of size 5`

I found that when reporting this error, the number of nonzero elements in (prob_th==1) is odd (which should be even) , and the last element of `prob_th` is 1.     Should the last element be set to 0 when there're odd nonzero elements?

The audio prob figure is below:
[
6850070556935454723_2_prob
](url)

And the audio threshold figure is below:
[
6850070556935454723_2_threshold
](url)

Best regards!

ferugit:
I faced some problems while using VAD interface to get speech segments. In fact, the problem was already detected [here](#1108). Cases, when speech starts in final frames, are not well managed. The **get_boundaries** method returns an even number of bounds, so if speech starts in final frames (start bound) the code was not able to give an end bound. For more details go to issue #1108. The next images show an example that used to break the execution:

Output probability:
image
Threshold-based decision:
image

With this PR, the following lines no longer break the execution.

```python
import torch
from speechbrain.pretrained import VAD

vad = VAD.from_hparams(
    source="speechbrain/vad-crdnn-libriparty",
    savedir="data/vad",
)
torch.manual_seed(2022)

for i in range(100):
    probs_th = (torch.randn(1, 100, 1) > 0.5).int()
    vad.get_boundaries(probs_th)
```

close #1108

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -1286,12 +1286,12 @@ class VAD(Pretrained):
         prob_th[:, 0, :] = (prob_th[:, 0, :] >= 1).int()
         prob_th[:, -1, :] = (prob_th[:, -1, :] >= 1).int()
 
-        # Fix edge cases (when a new sentence starts in the last frame)
-        if prob_th[:, -1, :] == 1 and prob_th[:, -2, :] == 1:
-            prob_th[:, -2, :] = 2
-
-        if prob_th[:, -1, :] == 1 and prob_th[:, -2, :] == 0:
-            prob_th[:, -2, :] = 1
+        # Fix edge cases (when a speech starts in the last frames)
+        if (prob_th == 1).nonzero().shape[0] % 2 == 1:
+            prob_th = torch.cat(
+                (prob_th, torch.Tensor([1.0]).unsqueeze(0).unsqueeze(2)), 
+                dim=1
+                )
 
         # Where prob_th is 1 there is a change
         indexes = (prob_th == 1).nonzero()[:, 1].reshape(-1, 2)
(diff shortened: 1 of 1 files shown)


mgoldey:
You probably want to set the torch random seed to ensure you are testing the same data before and after your fix
https://pytorch.org/docs/stable/notes/randomness.html

That said, I think it improbable that 100 random samples wouldn't surface the issue

ferugit:
Hello @mgoldey, 

You are right, I corrected the code snippet. Thank you for pointing this out!

Nevertheless, I tested this PR with a huge amount of real data, the data that helped me to discover the issue. I sadly cannot share that data here.

TParcollet:
Thanks @ferugit ! This is something for @mravanelli, I hope he will be able to have a look :-)

mravanelli:
Thank you very much for the fix (and sorry for the late review). I tried it and everything seems to work!

PR #1353, opened 2022-04-01
Add `cspell`

calvinballing:

Possible future work, if cspell is added: working more on the custom dictionary.  Currently it has 5 sections, the last of which could be sorted into the rest of the dictionary.:


mravanelli:
Thank you @calvinballing. What do you think about it @anautsch?

anautsch:
Hi @calvinballing and @mravanelli - it's evident that we need it. 

Yet, integrating it now will cause too many issues with ongoing PRs. Perhaps it would be good to have it in the pre-commit as well, so everyone can check their spelling before PR merger. I put it into the v0.6 track. 

At the time, I was thinking to do spell checks just before releasing v0.6 and merge this PR then. Now, I'm wondering if we should merge it right away into the `unstable` branch to kick-off v0.6 development with spellchecking already in-place (but open a later PR to include it into pre-commit checks). It's part of formatting our files, and if we migrate to current versions of flake8 and black as well, we will have another big PR impacting everything. I haven't had sorted out what would be a good procedure. Merging it into `unstable` and getting a feeling for how to use spell best while developing v0.6 there could enrich our experience how to integrate it best for automatic checks in future PRs.

@mravanelli what's your take?

anautsch:
We could also:
* make a `develop-v0.5.12` branch
* map all current PR's origins there
* merge this one into develop

This way, all future PRs would use cspell directly. Later on, we would merge the `develop-v0.5.12` back into `develop`.
It seems like an overhead workaround though for that almost all files must be changed by this PR.

mravanelli:
Hi,
I think merging it into unstable is fine. Is there a way to integrate the spell checker into the tests that we run? This way we are sure all future codes are compliant with that as well.

anautsch:
Ok. Merging this one into it after the #751 PR is on unstable.

Adel-Moumen:
Hello @calvinballing,

Thanks for this PR! 

May I ask you if you are still up to work a bit on that? Since your PR has been created in 2022, there's been a lot of changes in the current `unstable-v0.6` branch which ultimately leads to many merge conflicts. Do you mind if you could update your PR with the latests commits in `unstable-v0.6` ? 

You could also create a new PR which would be easier I think. Let me know what you'll do. I'd like to work on this PR during the week so if you are not available (which is totally fine) I will have to create myself the PR (but the credits will goes to yours in the files no worries :)) 

Best,
Adel

calvinballing:
> You could also create a new PR which would be easier I think. Let me know what you'll do. I'd like to work on this PR during the week so if you are not available (which is totally fine) I will have to create myself the PR (but the credits will goes to yours in the files no worries :))

I've started work on a new PR here: https://github.com/speechbrain/speechbrain/pull/2176

Issue #1319, opened 2022-03-02
I tried to run a librispeech transformer recipe with 8 GPU but a word error rate remains very large.


ken57:
I tried to run a librispeech transformer recipe with 8 GPU by using DDP (https://github.com/speechbrain/speechbrain/blob/develop/recipes/LibriSpeech/ASR/transformer/train.py) but a word error rate remains very large(around 100%) in spite of 10 epochs.
```
epoch: 1, lr: 1.00e+00, steps: 4056, optimizer: Adam - train loss: 2.50e+02 - valid loss: 1.32e+02, valid ACC: 1.97e-01
epoch: 2, lr: 1.17e-04, steps: 12845, optimizer: Adam - train loss: 2.11e+02 - valid loss: 1.27e+02, valid ACC: 2.24e-01
epoch: 3, lr: 1.97e-04, steps: 21634, optimizer: Adam - train loss: 2.04e+02 - valid loss: 1.25e+02, valid ACC: 2.43e-01
epoch: 4, lr: 2.07e-04, steps: 30423, optimizer: Adam - train loss: 1.98e+02 - valid loss: 1.23e+02, valid ACC: 2.58e-01
epoch: 5, lr: 1.82e-04, steps: 39212, optimizer: Adam - train loss: 1.93e+02 - valid loss: 1.21e+02, valid ACC: 2.67e-01
epoch: 6, lr: 1.65e-04, steps: 48001, optimizer: Adam - train loss: 1.89e+02 - valid loss: 1.21e+02, valid ACC: 2.71e-01
epoch: 7, lr: 1.51e-04, steps: 56790, optimizer: Adam - train loss: 1.85e+02 - valid loss: 1.21e+02, valid ACC: 2.70e-01
epoch: 8, lr: 1.41e-04, steps: 65579, optimizer: Adam - train loss: 1.82e+02 - valid loss: 1.22e+02, valid ACC: 2.67e-01
epoch: 9, lr: 1.32e-04, steps: 74368, optimizer: Adam - train loss: 1.79e+02 - valid loss: 1.23e+02, valid ACC: 2.64e-01
epoch: 10, lr: 1.25e-04, steps: 83157, optimizer: Adam - train loss: 1.76e+02 - valid loss: 1.24e+02, valid ACC: 2.61e-01, valid WER: 96.31
```
I ran the following command.
```
[...]

PR #1356, opened 2022-04-05
Fix for transformer recipe

kenro515:
**Problem:**
I ran the script of transformer ASR, but the error occurred and stopped training.

**Solution:**
It seems that some typos were made in the recipe of transformer ASR.
(By fixing them, I could run this script in my environment.)

ps: see as https://github.com/speechbrain/speechbrain/issues/1319#issuecomment-1080201256.

### DIFF
--- a/recipes/LibriSpeech/ASR/transformer/train.py
+++ b/recipes/LibriSpeech/ASR/transformer/train.py
@@ -3,22 +3,29 @@
 The system employs an encoder, a decoder, and an attention mechanism
 between them. Decoding is performed with (CTC/Att joint) beamsearch coupled with a neural
 language model.
+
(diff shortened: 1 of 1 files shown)


TParcollet:
Hi, this should not cause any issue. Are you on the latest SpeechBrain develop ?

TParcollet:
This error will disappear once we release a new SB version. But we will merge it if the tests pass so we don't have any issues anymore

kenro515:
Hi !
I am apologize that I am too late for reply.

I checked it the version of SpeechBrain in my environment, and reinstall developer version.

I could run this script without any modification.
I am sorry for bothering you.
Thanks.

PR #1361, opened 2022-04-07
Some bugs in speaker verification  on Voxceleb

qiny1012:


mravanelli:
Thank you for the fix!

Issue #1339, opened 2022-03-25
Incorrect use of torchaudio's rnnt_loss


csukuangfj:
It is suprising that no one has noticed that the current code is feeding the output of 
log_softmax to torchaudio's rnnt_loss, which is not correct.

torchaudio's rnnt_loss accepts only `logits` as input.

I am wondering whether anyone has succeeded in training a transducer model
with the current usage of torchaudio's rnnt_loss using speechbrain.

---

The relevant code is listed below:

https://github.com/speechbrain/speechbrain/blob/24720c1532355dfeeda78304939dcc80283244bf/recipes/CommonVoice/ASR/transducer/train.py#L75

https://github.com/speechbrain/speechbrain/blob/24720c1532355dfeeda78304939dcc80283244bf/recipes/CommonVoice/ASR/transducer/train.py#L136-L138

https://github.com/speechbrain/speechbrain/blob/24720c1532355dfeeda78304939dcc80283244bf/recipes/CommonVoice/ASR/transducer/hparams/train_fr.yaml#L183-L185

https://github.com/speechbrain/speechbrain/blob/24720c1532355dfeeda78304939dcc80283244bf/speechbrain/nnet/losses.py#L59-L77

vsokolov00:
I'm trying to train an RNN-T model... Have you tried to train a model with this error fixed? How much was the difference?

csukuangfj:
No, I have not used it to train a model. I just spotted this bug while looking at the code.

aheba:
Hello @csukuangfj , thank you for reporting the issue, 
let me do a fix

vsokolov00:
It's strange but I didn't notice any improvements during the training process, after fixing this bug it produces even worse results during the validation.. I'll make the update once I decode it on the test set

vsokolov00:
It's strange but I didn't notice any improvements during the training process, after fixing this bug it produces even worse results during the validation.. I'll make the update once I decode it on the test set

vsokolov00:
softmax outputs - `Epoch loaded: 30 - test loss: 31.53, test WER: 55.05, test CER: 41.22`
(FIXED) logits - `Epoch` loaded: 30 - test loss: 30.48, test WER: 54.91, test CER: 41.20`

Adel-Moumen:
Hello, 

This issue has been solved with #1368, so I'm closing this issue. 
Thanks again for reporting this problem! :-)

PR #1368, opened 2022-04-10
fix RNN-T loss bug - use logits instead of log_probs

aheba:
Hi @TParcollet ,
as reported in #1339, thanks to @csukuangfj,

Only torchaudio is using logits as input of the RNNT loss function. (other lib doesn't, see https://github.com/1ytic/warp-rnnt, https://github.com/HawkAaron/warp-transducer/tree/master/pytorch_binding)

Here is a fix to the RNN-T loss using torchaudio,

I will use this PR to add to our Numba code:
- [ ] // computation for alpha & beta.
- [ ] add fastemit_lambda (see https://arxiv.org/abs/2010.11148)

### DIFF
--- a/recipes/CommonVoice/ASR/transducer/train.py
+++ b/recipes/CommonVoice/ASR/transducer/train.py
@@ -71,8 +71,7 @@ class ASR(sb.Brain):
         joint = self.modules.Tjoint(x.unsqueeze(2), h.unsqueeze(1))
 
         # Output layer for transducer log-probabilities
-        logits = self.modules.transducer_lin(joint)
-        p_transducer = self.hparams.log_softmax(logits)
+        logits_transducer = self.modules.transducer_lin(joint)
 
         # Compute outputs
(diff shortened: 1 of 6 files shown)


csukuangfj:
> Only torchaudio is using logits as input of the RNNT loss function.

The advantage of using `logits` as input is that it can reduce memory consumption since you don't need to allocate space for `log_probs` and its gradients. Furthermore, the gradient of `logits` can share the same memory space as `logits`.

There are other implementations that take `logits` as input. See 
- https://github.com/danpovey/fast_rnnt/
- https://github.com/csukuangfj/optimized_transducer

aheba:
Thanks for highlighting that point, let me take a look on :

https://github.com/csukuangfj/transducer-loss-benchmarking

mravanelli:
@aheba, what's the status of this PR?

mravanelli:
Thank you  @aheba  for the fix and @csukuangfj for the comments. I tested it and everything seems to work.

PR #1370, opened 2022-04-11
Allow for resampling noises/reverb at different rates from clean

### LINKED ISSUE #901: Data augmentation at 8kHz
I intend to train a speaker recognition model with 8kHz data, but it looks like the data augmentation implementation is somewhat restricted to work with the original RIRS noise sampling rate, which is 16kHz. More specifically, the EnvCorrupt class does not allow specifying the sampling rate.
I considered using the "rir_scale_factor" param to resample the reverb impulse response, but I suppose the resulting response might be affected by aliasing. With regard to noise addition, I didn't find a way to work with a different sampling rate.
At last, I observed that the TimeDomainSpecAugment class does not expose some parameters such as drop_freq_low, drop_freq_high, drop_width, which need to be set in accordance with the target sampling rate. 
Please let me know if I am missing something.

pplantinga:
This PR closes #901 

Basically, our AddNoise and AddReverb classes weren't set up to allow resampling corruptions that were sampled at a different rate from the clean signal. While this could technically be handled by creating a copy of the corruptions resampled to the correct rate, this diverges from SpeechBrain philosophy of "do all transformations on-the-fly".

### DIFF
--- a/speechbrain/lobes/augment.py
+++ b/speechbrain/lobes/augment.py
@@ -21,7 +21,6 @@ from speechbrain.processing.speech_augmentation import (
     AddBabble,
     AddNoise,
     AddReverb,
-    Resample,
 )
 from speechbrain.utils.torch_audio_backend import check_torchaudio_backend
 
--- a/speechbrain/lobes/augment.py
+++ b/speechbrain/lobes/augment.py
@@ -21,6 +21,7 @@ from speechbrain.processing.speech_augmentation import (
     AddBabble,
     AddNoise,
     AddReverb,
+    Resample,
 )
 from speechbrain.utils.torch_audio_backend import check_torchaudio_backend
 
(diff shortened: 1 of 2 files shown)


pplantinga:
I haven't yet tested this, I'll need to scrounge up a 8kz signal and see if this works on that.

speechbrain/lobes/augment.py
```diff
@@ -338,6 +338,8 @@ class EnvCorrupt(torch.nn.Module):
         If ``0 < rir_scale_factor < 1``, the impulse response is compressed
         (less reverb), while if ``rir_scale_factor > 1`` it is dilated
         (more reverb).
+    reverb_sample_rate : int
+        Sample rate of input audio signals (rirs) used for reverberation.
```
pplantinga, line 342: Missing `noise_sample_rate` and `clean_sample_rate`

pplantinga:
Okay, this is tested with an 8k signal, seems to work as expected, ready to review.

mravanelli:
Great fix @pplantinga! I tested it and everything seems to work as expected.

PR #1385, opened 2022-04-22
Profiling & benchmark decorators

anautsch:
To measure real-time performance, tensorboard profiling and PyTorch benchmarking seem useful. Yet, they offer so much more than just measuring real-time. While getting more familiar with profiling & benchmarking, I started to create polymorphic decorators that to be used in different ways:

To export benchmarks for  `__init__`; `fit()`, and `evaluate()` functions of Brain:s (after a scheduled warm-up).
```
@export
@schedule
@profile
class SimpleBrain(Brain):
    def compute_forward(self, batch, stage):
        return self.modules.model(batch[0])

    def compute_objectives(self, predictions, batch, stage):
        return torch.nn.functional.l1_loss(predictions, batch[1])
```

Or let's say, we have some pre-trained model:
```
class SimpleBrainUntracked(Brain):
    def compute_forward(self, batch, stage):
        return self.modules.model(batch[0])

    def compute_objectives(self, predictions, batch, stage):
        return torch.nn.functional.l1_loss(predictions, batch[1])

pretrained = SimpleBrainUntracked({"model": model}, lambda x: SGD(x, 0.1), run_opts={"device": device})
# de/serialise as usual
```
Then, one could run:
```
# hook the model to profiling
scheduled_profiler = schedule(profile)
scheduled_profiler(pretrained)

# call as usual
pretrained.evaluate(test_set=test_set)
```

An `@export` decorator directs outputs to a `log` folder.

- [x] polymorphic `@profile` decorator to wrap PyTorch profiling & benchmark for Brain classes & train/inference (fit/evaluate) functions
- [x] polymorphic `@schedule` decorator to handle warm-ups
- [x] polymorphic `@export` decorator to store trace(s) from profiling: open tensorboard(s) in chrome://tracing and export FlameGraphs
- [x] use case support: there are two competing implementations -> which is faster? There's a diff for EventList:s to filter out akin events (e.g. same number of call) and print for each implementation, what it did more/less than the other one
- [x] unit tests
- [x] `@profile_details` decorator as benchmark with some standard config
- [x] aggregate traces for multiple interactions with a Brain model (two types of EventLists are supported by torch.profiler, yet, only one is directly tractable)
- [x] template with README and mock-up example to measure real-time performance
- [x] pre-commit clean-ups


anautsch:
Created two pre-configured decorators (instead of one): `@profile_analyst` and `@profile_optimiser`. The analyst logs everything but creates overheads because of that. The optimiser logs memory & modules (without overheads - according to torch.profiler documentation).

Currently, FLOPs are only reportable for matrix multiplication and conv2 operations. Limited its use hence to the analyst configuration.

anautsch:
@TParcollet please take a look

TParcollet:
LGTM

PR #1391, opened 2022-05-05
Add recipe Binaural-WSJ0Mix

huangzj421:
This is a new separation task that applied to binaural-channel WSJ0Mix. Please check out the paper [Real-time binaural speech separation with preserved spatial cues](https://ieeexplore.ieee.org/abstract/document/9053215) for details. Much thanks.

### DIFF
--- a/recipes/Binaural-WSJ0Mix/extra-dependencies.txt
+++ b/recipes/Binaural-WSJ0Mix/extra-dependencies.txt
@@ -1,0 +1,2 @@
+mir-eval==0.6
+pyroomacoustics==0.3.1
--- a/recipes/Binaural-WSJ0Mix/separation/README.md
+++ b/recipes/Binaural-WSJ0Mix/separation/README.md
@@ -1,0 +1,80 @@
+# Speech separation with Binaural-WSJ0Mix
+This folder contains some recipes for the Binaural-WSJ0Mix task (2/3 sources). Please refer to [Real-time binaural speech separation with preserved spatial cues](https://ieeexplore.ieee.org/abstract/document/9053215) [1] for details.
+
+
+Additional dependency:
+```
+pip install mir_eval, pyroomacoustics
+```
+
+To run it:
+
+```
+python train-parallel.py hparams/convtasnet-parallel.yaml --data_folder yourpath/binaural-wsj0mix/2speakers
+```
+Note that during training we print the negative SNR instead of SI-SNR because the scale-invariance property of SI-SNR makes it insensitive to power rescaling of the estimated signal, which may fail in preserving the ILD between the outputs.
+
+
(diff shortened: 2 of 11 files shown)


TParcollet:
I suppose that this is for @ycemsubakan to review ? :P

ycemsubakan:
Sure.. I will do a review. Thanks for the PR!

mravanelli:
@ycemsubakan, could you please review this PR?

ycemsubakan:
Yes @mravanelli I will do it this weekend.

ycemsubakan:
Hi @huangzj421, again sorry for the delay, and thanks for your PR. I wrote several inline comments, please take a look. In addition those here are some of the other things I noticed:

- An important takeaway from the inline comments is that it is important not to change the functions / classes when possible in order not to break other recipes. 
- I do not understand why we need to have `train-binaural.py`, `train-parallel.py`, `train-cross.py` separately. I find this to be a little convincing. Normally we only have one `train.py` in the recipes. Is it possible to merge everything in one `train.py'? 
- Related to the comment above, it seems you have separate `.yaml` files for binaural, parallel and cross. It is okay to have separate `.yaml` files, but the differences should be explained in the README. 
- As AIShell1-Mix recipe that you pushed, I have a comment about data generation. It would be much more convenient for the user to automatize the data generation. Currently what you have requires the user to manually create the data themselves, separately. Similar to the AIshell-1Mix recipe, can you make it so that `train.py` automatically generates the dataset? 
 - It seems there are conflicts now with the develop branch. Can you pull the most recent develop branch?

huangzj421:
> Hi @huangzj421, again sorry for the delay, and thanks for your PR. I wrote several inline comments, please take a look. In addition those here are some of the other things I noticed:
> 
> * An important takeaway from the inline comments is that it is important not to change the functions / classes when possible in order not to break other recipes.
> * I do not understand why we need to have `train-binaural.py`, `train-parallel.py`, `train-cross.py` separately. I find this to be a little convincing. Normally we only have one `train.py` in the recipes. Is it possible to merge everything in one `train.py'?
> * Related to the comment above, it seems you have separate `.yaml` files for binaural, parallel and cross. It is okay to have separate `.yaml` files, but the differences should be explained in the README.
> * As AIShell1-Mix recipe that you pushed, I have a comment about data generation. It would be much more convenient for the user to automatize the data generation. Currently what you have requires the user to manually create the data themselves, separately. Similar to the AIshell-1Mix recipe, can you make it so that `train.py` automatically generates the dataset?
> * It seems there are conflicts now with the develop branch. Can you pull the most recent develop branch?

Hi @ycemsubakan. I am sorry for the old version of develop branch. I am gonna close it and reopen again for a more suitable review.

PR #1394, opened 2022-05-09
Dynamic batching for wav2vec2 pertaining

TParcollet:
Dynamic Batching was missing for the pretraining so I added it !

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
     hooks:
       - id: black
         types: [python]
-
+        additional_dependencies: ['click==8.0.4']
   - repo: https://gitlab.com/pycqa/flake8.git
     rev: 3.7.9
     hooks:
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,4 +1,5 @@
 black==19.10b0
+click==8.0.4
 flake8==3.7.9
 pycodestyle==2.5.0
 pytest==5.4.1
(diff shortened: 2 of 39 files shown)


TParcollet:
@anautsch I know that you already have a lot on your plate, but please run this recipe somewhere with and without dynamic batching. on my side, it works. If it works on your side too, we can merge :-)

mravanelli:
@TParcollet, what is the status of this PR? It is ready to be reviewed?

TParcollet:
@anautsch  is on it. We are investigating the reason why Dynamic Batching seems to be slow ..

anautsch:
lgtm - tested on 'it' subset, one epoch is now (slightly enough) faster with dynamic batching than without. After the last 2 commits, the num_workers had not as much of an impact as the max_batch_size. I used two nodes with 4 V100 GPUs each, so `num_workers = 2`. For dynamic batching, `max_batch_size` is a way to give a tell on available VRAM - how much it is used, depends on the data though.

```
w/o 1954/1954 [13:23<00:00,  2.43it/s, train_loss=7.3e+3]
w/  1173/1173 [15:25<00:00,  1.27it/s, train_loss=1.29e+4]  // max_batch_size: 64
w/  1173/1173 [12:12<00:00,  1.60it/s, train_loss=1.33e+4]  // max_batch_size: 128
w/  1173/1173 [12:15<00:00,  1.59it/s, train_loss=1.29e+4]  // max_batch_size: 256
```

PR #1403, opened 2022-05-22
Handle ROCm pytorch

jnordberg:
This fixes a crash where the logger unexpectedly encountered `None` when running with a ROCm backed pytorch.


mravanelli:
Thank you @jnordberg

PR #1404, opened 2022-05-22
Testing consistency between YAML and training script

mravanelli:
With this PR I wrote a simple test that checks the YAML files and their corresponding python scripts. The test detects those variables declared in the yaml file but never used in the training script. 
Here are my remarks:
1. I think we can take the opportunity of the minor revision to fix all these inconsistencies (which are quite annoying because might create confusion). There are currently tons of them detected by the current version of the test. You can run `pytest tests/unittests/test_yaml.py` to double-check it.
2. There are some inconsistencies that are recurrent. For instance, the variable `ckpt_interval_minutes` is declared in the yaml file and used in `core.py`. IMO this is not transparent. I suggest adding this variable as an optional input to the brain class. There are other examples that require little modifications of the code.
3. I put the test in `tests/unittests/test_yaml.py` and the related library in `speechbrain/utils/check_yaml.py`. I'm not sure this is the best place. If you have a better suggestion, please let me know.

Please, share your feedback!


speechbrain/utils/check_yaml.py
```diff
@@ -0,0 +1,112 @@
...
+                # The pattern can be ["key"] or ".key"
+                if '["' + var + '"]' in line:
+                    if var not in detected_var:
+                        detected_var.append(var)
+                if "." + var in line:
+                    if var not in detected_var:
+                        detected_var.append(var)
+
+    # Check which variables are declared but not used
+    unused_vars = list(set(var_lst) - set(detected_var))
```
anautsch, line 102: This filters for the hparam keys declared but unused in train.py - how about the keys asked for in train.py but not defined in hparams? It's just for an overview of which default variables are used - so no error needs to be thrown.

`default_vars = list(set(detected_var) - set(var_lst))`

// just a thought on exhaustivity

anautsch:
On 3. looks like a good place for me - anyone adding a recipe gets confronted with their mishaps pre-commit or on PR auto checks.

On 2. What do you mean by 'not transparent'? Just not to miss out. The result listing works two ways: declare the missing OR hey, that might be a good default variable (if the recipe user neglects it anyways, better to default on it).

On 1. Yep, that will consume a bit of time & thought. The tests are quite revealing, thank you for surfacing the issues this way!

---

Just to preserve of what the log output is returning at the moment —

Recipes w/o warning:
* `recipes/VoxCeleb/SpeakerRec/hparams/verification_plda_xvector.yaml`
* `recipes/Fisher-Callhome-Spanish/Tokenizer/hparams/train_bpe_1k.yaml`
* `recipes/CommonLanguage/lang_id/hparams/train_ecapa_tdnn.yaml`
* `templates/speech_recognition/Tokenizer/tokenizer.yaml`
* `templates/enhancement/train.yaml`

Missing recipes from enlisted ones:
```
File recipes/DNS/enhance/spectral_map/hparams/params_CNN.yaml not found!
File recipes/TIMIT/ASR/seq2seq_knowledge_distillation/save_teacher.py not found!
File recipes/TIMIT/ASR/transducer/train_with_wav2vec2.py not found!
```

Part of the log on tamplates:
```
Checking templates/speech_recognition/LM/RNNLM.yaml...
	WARNING: variable "blank_index" not used in templates/speech_recognition/LM/train.py!
	WARNING: variable "ckpt_interval_minutes" not used in templates/speech_recognition/LM/train.py!
	WARNING: variable "output_neurons" not used in templates/speech_recognition/LM/train.py!
	WARNING: variable "pretrainer" not used in templates/speech_recognition/LM/train.py!
Checking templates/speech_recognition/Tokenizer/tokenizer.yaml...
Checking templates/speech_recognition/ASR/train.yaml...
	WARNING: variable "ckpt_interval_minutes" not used in templates/speech_recognition/ASR/train.py!
	WARNING: variable "unk_index" not used in templates/speech_recognition/ASR/train.py!
Checking templates/enhancement/train.yaml...
Checking templates/hyperparameter_optimization_speaker_id/train.yaml...
	WARNING: variable "ckpt_interval_minutes" not used in templates/hyperparameter_optimization_speaker_id/train.py!
	WARNING: variable "tdnn_channels_out" not used in templates/hyperparameter_optimization_speaker_id/train.py!
	WARNING: variable "tdnn_channels" not used in templates/hyperparameter_optimization_speaker_id/train.py!
	WARNING: variable "trial_id" not used in templates/hyperparameter_optimization_speaker_id/train.py!
Checking templates/speaker_id/train.yaml...
	WARNING: variable "ckpt_interval_minutes" not used in templates/speaker_id/train.py!
```

Summary (status)
* 117 recipes with warnings -> 78/117 are fully characterized by these bullet points
* 49x `ckpt_interval_minutes`
* 22x `use_speedperturb_sameforeachsource`; `progressbar`
* 22x `experiment_name`
* 17x `unk_index`; `jit_module_keys`
* 15x `experiment`
* 14x `train_annotation`; `valid_annotation`;  `test_annotation`; `token_type`; `eos_threshold`
* 12x `train_log`; `ctc_weight_decode`
* 10x `data_folder_rirs`
* 8x `valid_csv`
* 6x `models`; `gradient_clipping`; `vocab_size`; `compute_error`
* 5x `blank_index`; `activation`
* 4x `asr_model`; `sample_rate`
* 3x `character_coverage`; `compute_istft`; `pad_index`; `skip_prep`
* 2x `waveform_target`; `mean_var_norm`; `csv_read`; `dynamic_batching`; `dynamic_batch_sampler`; `wav2vec2_opt_class`; `random_chunk`

(I skipped some keys here if they're only relevant to a database/recipe series.)

Some of them are copy/paste related, from one recipe to another.
Some of them help to identify where we missed out to avail a new feature (see dynamic batching).
@mravanelli  could some keys be a top-level hparam variable used only within that hparams for !ref like `experiment` or `experiment_name` ?

Recipes with particular keys:
* recipes/Google-speech-commands
* recipes/Voicebank/MTL
* recipes/LibriMix
* recipes/LibriSpeech/ASR/transformer
* recipes/CommonVoice/ASR/CTC
* recipes/Aishell1Mix
* recipes/WHAMandWHAMR/enhancement
* recipes/ERPCore/P3_decoding/train.yaml (has many top-level keys)

(I skipped some recipes that had smaller issues only.)

pplantinga:
>> There are some inconsistencies that are recurrent. For instance, the variable `ckpt_interval_minutes` is declared in the yaml file and used in `core.py`. IMO this is not transparent. I suggest adding this variable as an optional input to the brain class. There are other examples that require little modifications of the code.

> On 2. What do you mean by 'not transparent'? Just not to miss out. The result listing works two ways: declare the missing OR hey, that might be a good default variable (if the recipe user neglects it anyways, better to default on it).

Here's the context for this as I remember it: the `run_opts` are a set of options that we wanted users to be able to specify in a few different ways:

1. By command line (without changing any code)
2. By configuration file (so they don't have to specify it each time if its not going to change)
3. By code (though I guess it might be a bit non-obvious how to do it here, you'd have to create or edit the `run_opts` dict)

The `run_opts` dict is the solution we came up with to do all of this, though I agree that it is quite confusing in its current form. And more specifically about `ckpt_interval_minutes` I remember discussing what we wanted the default to be, and came down on the side of "its pretty dataset/setup dependent, lets not specify a global default". Maybe @Gastron has more memory of this discussion.

One slight improvement could be to unpack the `run_opts` dict as separate options, then merge with parameters found in hparams: i.e.

```
param_file, run_opts, overrides = parse_arguments(sys.argv[1:])

...

run_opts = update_run_opts(hparams)
...

asr_brain = ASR(
    modules=hparams["modules"],
    hparams=hparams,
    **run_opts
)
```

This would be a major change. Perhaps better to wait for 0.6

speechbrain/utils/check_yaml.py
mravanelli: I think we can potentially add this check as well. However, If train.py uses a variable not declared in the yaml file, there will be an error at runtime. So, we can still detect these errors with the recipe tests that we will hopefully setup. As far as I remember, some recipes define new hparams (like `self.hparams.new_varable=value`) in the train script directly. This is not a good practice, but it was needed for some reason I don't recall. It might be a bit trickier to set up this check.

mravanelli:
@anautsch,
let me clarify why I think the current way to manage some variables (e.g., `ckpt_interval_minutes`) is not totally transparent IMO. If we declare a variable in the yaml file (e.g., `ckpt_interval_minutes: 15`), it would be great to see it used in `train.py `somewhere (otherwise, users might not have any idea about where this is actually used). The `ckpt_interval_minutes`, instead, is used in `core.py` directly.

I don't think we have a solution for that now, because we have some "special" variables that can be declared both in the yaml file and as running options from the command line (run_opts dict). A simple way to circumvent this issue is to remove `ckpt_interval_minutes` from the hparams file (users can set it from the command-line by setting the running options).

In the long-term, maybe we can summarize the default running options in an external yaml file and include it in the main yaml file like (include: run_opts.yaml). However, I 'm not sure this will add more transparency. What do you think @pplantinga?

anautsch:
@mravanelli yep, agreed. 

Some hparams are used in the deep depths and there is no documentation to it on the higher level, so how to make use of them? One has to be an expert in using/editing the tool already. Lot's of initial effort is required to get there (beyond, click & it works). That is one of the things I see in the complex "roundrip testing" - but testing w/could easily lead to overkill in testing everything all the time. If each subsystem works (hparams where they are used - and it is known what their purpose is there) and the the way any larger system could combine these subsystems, then that could be a more elegant solution (like it is currently with doctests, unittests & integration tests).

Gut feeling: everything that provides more transparency, will be a change. We could think also about splitting hparams into two/multiple parts by their purpose (model, run opts, ...). In #1407 the remark is on deriving a pretrained model yaml file, and I get the intuition behind it: shouldn't it be there already? ;-)

@pplantinga 
As one design choice is: recipes are one big yaml file - any refactoring here would bring lots of changes. For testing, a modularised treatment of hparams files could help. For this in turn, namespaces to parameter could also be of benefit (to avoid too many yaml files flying around). 

e.g. `ckpt_interval_minutes` w/could be namespaced as
```yaml
run_opts:
    ckpt_interval_minutes: 15
```

Wondering if default yaml configs could be part of a doctest then, so that one sees with the namespace also the definition of the yaml block and its default parameters and how they are used. Another script could then aggregate the graph of all hparams involved. That's more v0.6 stuff.

As for now, targeting a potential minor version, all we care about is consistency. For that, my suggestions would be:
* if keys are reported as missing in the train.py but are !ref used in other parts of the hparam, they are ok
* if keys are reported but do have default values somewhere, they are also ok

I'll look into getting a list of default keys from declared modules - maybe it's easier as I think of it right now.

speechbrain/utils/check_yaml.py
anautsch: Should the `self.hparams.new_varable=value` bad practice be checked for?

btw (untested)
```python
default_run_opt_keys = ["debug", "debug_batches", "debug_epochs", "device": "cpu", "data_parallel_backend", "distributed_launch", "distributed_backend", "find_unused_parameters", "jit_module_keys", "auto_mix_prec", "max_grad_norm", "nonfinite_patience", "noprogressbar", "ckpt_interval_minutes", "grad_accumulation_factor", "optimizer_step_limit"]
unused_vars = list(set(var_lst) - set(detected_var) - set(default_run_opt_keys))
```

anautsch:
The default run_opts are not importable outside core.py
https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/speechbrain/core.py#L434

```python
        run_opt_defaults = {
            "debug": False,
            "debug_batches": 2,
            "debug_epochs": 2,
            "device": "cpu",
            "data_parallel_backend": False,
            "distributed_launch": False,
            "distributed_backend": "nccl",
            "find_unused_parameters": False,
            "jit_module_keys": None,
            "auto_mix_prec": False,
            "max_grad_norm": 5.0,
            "nonfinite_patience": 3,
            "noprogressbar": False,
            "ckpt_interval_minutes": 0,
            "grad_accumulation_factor": 1,
            "optimizer_step_limit": None,
        }
```

Sh/could they become part of a static function `get_default_run_opts()` ?
Such static default functions could provide hparam default snippets at each function level where they are used.
Then, a full hparams file could be generated from there - and existing files be checked against.

speechbrain/utils/check_yaml.py
anautsch: The `progressbar` logic causes 22x warnings - while it is in hparams as `"noprogressbar"`

https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/speechbrain/core.py#L1022

```python
        if progressbar is None:
            progressbar = not self.noprogressbar
```

=> simply add "progressbar" to the above default list?

anautsch:
Just going through the repo with grep and the keys throwing warnings
* `json_path=hparams[f"{dataset}_annotation"]` issues on `train_annotation; valid_annotation; test_annotation`
* `token_type` is used/consumed directly in yaml `!ref <token_type>` (as many other keys)

```yaml
recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml:models: !include:../models/MetricGAN.yaml
recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml:    generator: !ref <models[generator]>
recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml:    discriminator: !ref <models[discriminator]>
recipes/Voicebank/enhance/MetricGAN/hparams/train.yaml:        model: !ref <models[generator]>
```

* in some recipes `train_log` is consumed by a later train_logger - it seems too much to check that automatically, but should there be a grep -R run after the script identified missing keys to visualise how/if that key was used in other recipes?

`$ grep -R train_log recipes`
```
...
recipes/SLURP/NLU/hparams/train.yaml:train_log: !ref <output_folder>/train_log.txt
recipes/SLURP/NLU/hparams/train.yaml:train_logger: !new:speechbrain.utils.train_logger.FileTrainLogger
recipes/SLURP/NLU/hparams/train.yaml:    save_file: !ref <train_log>
...
```

* `vocab_size` is quite a candidate - as a parameter read only from CommonVoice hparams ( [example](https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/recipes/CommonVoice/ASR/transformer/train.py#L409) ) and otherwise it is passed into the function call during train.py (same for `character_coverage`)

```yaml
tokenizer: !name:speechbrain.tokenizers.SentencePiece.SentencePiece
   model_dir: !ref <output_folder>
   vocab_size: !ref <token_output>
   annotation_train: !ref <train_csv>
   annotation_read: !ref <csv_read>
   model_type: !ref <token_type> # ["unigram", "bpe", "char"]
   character_coverage: !ref <character_coverage>
   annotation_list_to_check: [!ref <train_csv>, !ref <valid_csv>]
```

(https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/recipes/AISHELL-1/Tokenizer/train.py#L52)

https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/speechbrain/tokenizers/SentencePiece.py#L105

https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/speechbrain/tokenizers/SentencePiece.py#L108

So if it is a nested key and argument of a class constructor/function - then, it is also ok. There's some polymorphology going on here :D

* perhaps a debugging relict: `gradient_clipping` (transformers); `compute_error` (xvecs)

anautsch:
@mravanelli the script should be in a good shape now. As of now, I haven't made a commit with changed yaml & py files - but the test works out on my end. I'll push the recipe & template changes in a minute.

I did not run `./tests/.run-recipe-tests.sh`.

PR #1405, opened 2022-05-23
Automatic checking URLs

mravanelli:
In the repo, we have tons of links. The goal of this PR is to automatically extract the URLs (from the python, txt, and md files) and make sure they aren't broken.  Some comments:
1. Checking all the links is very slow, so we cannot do this check for every commit. Instead, this is a check that we can do periodically (like the recipe tests).
2. The current version is not working well. For some reason, it detects as DOWN  several links that are not DOWN. Any help to fix it is welcome!
3. Once solved this issue, we have to fix all the files containing broken links.


anautsch:
contains files from #1404 ;-)

anautsch:
After changing the url call to a simple print (curious about which URLs & such), some reporting of what we are dealing with:
* 790 URLs are mentioned in 227 SpeechBrain files
* 19 files have at least 10 URLs
* 4 files have at least 15 URLs
* 1 file has 20+ URLs (it's 37 links)

Where to?
* 129x huggingface.co
* 116x github.com
* 94x arxiv.org
* 73x drive.google.com
* 71x www.openslr.org
* 59x speechbrain.github.io --> +3x other github.io
* 23x zenodo.org; ed.ac.uk
* 17x mozilla.org
* 15x ldc.upenn.edu
* 12x colab.research.google.com
* 11x aihub.or.kr; openslr.org
* 9x wikipedia.org; readthedocs.io
* 8x www.dropbox.com; urbansounddataset.weebly.com
* 7x pytorch.org
* 6x www.robots.ox.ac.uk; doi.org
* 5x bark.phon.ioc.ee; wham.whisper.ai; qut.edu.au
* 4x tensorflow.org; img.shields.io; citi.sinica.edu.tw
* 3x docs.python.org; creativecommons.org; osf.io
* 2x stanford.edu; docs.pytest.org; stackoverflow.com; www.aclweb.org; mdpi.com; azurewebsites.net; sail.usc.edu; www.freesound.org; twitter.com; speechbrain.discourse.group

... and 24 other unique URLs

Maybe there's another way to ensure what the URL alive check is intended for.

mravanelli:
As for the first comment, I think you are right. I'm not sure why we often have links that are detected as Down while they aren't down, but one possible reason is that this is detected as some sort of attack. But, I might be wrong and maybe there is a way to avoid it. As  mentioned, this is not a test we should run for every commit (also because it is very slot), but maybe once a week only.

mravanelli:
As for the files from the other PR, I removed them from this one.

mravanelli:
As for the numbers, there looks pretty impressive. When you say "`Maybe there's another way to ensure what the URL alive check is intended for.`" Do you have something specific in mind?

anautsch:
The problem here is to have content available and not to promote dead links. How do we act if there are dead links? What is the process that follows? Maybe there's another way to support the process diligently than boxing good/bad on each all the time. Just writing my thoughts out; no idea yet.

Let's go for a first draft attempt on classifying the link for the purpose critical to SpeechBrain:
a. URLs on the README.md (30+)
b. URLs in our control: `speechbrain.github.io; speechbrain.discourse.group; https://twitter.com/SpeechBrain1`
c. Archive URLs (change = other trouble): `github.com; arxiv.org; openslr.org; doi.org; creativecommons.org; wikipedia.org; stackoverflow.com`
d. Data exchange platforms (somewhat stable, changes depend on their business though; at their cycles' mercy): `huggingface.co;osf.io; zenodo.org; ldc.upenn.edu; freesound.org; aihub.or.kr`
e. Documentation URLs (change on their revisions; this implies dependency updates): `pytorch.org; docs.python.org; tensorflow.org; docs.pytest.org; readthedocs.io`
f. University URLs (somewhat stable; no researcher's profile page): `sail.usc.edu; ed.ac.uk; robots.ox.ac.uk; bark.phon.ioc.ee; qut.edu.au; citi.sinica.edu.tw; stanford.edu`
g. Changes in the providers' environment to be expected: `drive.google.com; mozilla.org; colab.research.google.com; dropbox.com; wham.whisper.ai; urbansounddataset.weebly.com; aclweb.org; mdpi.com; azurewebsites.net; github.io`
h. links to external pictures

From there, one way could be to assign different time-scales when these links are checked (e.g.: weekly, monthly or quarterly). Or if checking is required at all.

Manual checking of links seems pointless if it is done all over again (no need to make a person feel like a monkey operating production line). Yet, hosts of these links might sense an attack schematic in whatever rhythm. One easy way could be to `time.sleep(1.4)` for a bit (1.4s example or with some randomness) - and to group URLs in bulks of different hosts (so that their penetration is somewhat distributed).

URLs could be grouped so that each batch of URLs has different domains only. That way, hosts are stressed less. This is not only receiving some streaming data, one is going around town to knock on many doors.

mravanelli:
Thank you for the comments @anautsch! I refactored the code based on your comments. In particular:
1. Now, we first create a list of unique urls and only after that we check it they are broken or not.
2. I added a sleep of 1 second between once check and another.
3.I'm now able to detect better the links that are actually broken with a better check on the status (I noticed that the links that are actually broken have status=404 or > 499).
4. I manually fixed the broken links.

Unless you have other comments, I think we can now merge it. This is a test that we can do periodically (e.g, ones a week).

anautsch:
lgtm @mravanelli - added a progressbar and shortened a link (now, it's 330 URLs)
> 331/331 [11:21<00:00,  2.06s/it]

anautsch:
ran it once more, all links work :mechanical_arm:

PR #1408, opened 2022-05-25
Binaural-WSJ0Mix

huangzj421:
This is a new separation task that applied to binaural-channel WSJ0Mix. Please check out the paper [Real-time binaural speech separation with preserved spatial cues](https://ieeexplore.ieee.org/abstract/document/9053215) for details. Much thanks.


huangzj421:
Here are some responeses to the @ycemsubakan review:
1. I have merged into one 'train.py' and clarified the differences between three yaml files in the Results of README. These indeed are three different model structures and I have distinguished them in 'experiment_name' of .yaml.
2. I know it's better not to change the speechbrain/ files but I have tried not to impact other recipes. The changes are summed up into two--one is that add cal_snr function like cal_si_snr in the losses.py and the other is that make the dim=-2(binaural channels) work well in the training.
3. I don't think the datasets cannot be generated automatically as the Aishell1Mix do  because wsj0 is not open-source.
Much thanks.

huangzj421:
> Thanks for the updates @huangzj421 !
> 
> Regarding, automatic data generation, yes WSJ0 is not open source. However many source separation researchers have it. Can we add an automatic generation option, and clearly indicate in the README that this automatic data generation script requires WSJ0? If the person does not have WSJ0, they will not be able to run the recipe in any case..

Hi @ycemsubakan , 

I have tried to move all the files in https://github.com/huangzj421/Binaural-WSJ0Mix to this recipe folder but encountered this commit failure. What should I do?
截屏2022-05-25 23 48 21

ycemsubakan:
I see, for this what you can do would be to download these files inside the script. You can use the `fetch` function from speechbrain for this. Please see some examples here: https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/recipes/REAL-M/sisnr-estimation/train.py#L759

huangzj421:
> I see, for this what you can do would be to download these files inside the script. You can use the `fetch` function from speechbrain for this. Please see some examples here:
> 
> https://github.com/speechbrain/speechbrain/blob/d6736adaabf0c6a1021dd1970056f9416bcdcd62/recipes/REAL-M/sisnr-estimation/train.py#L759

Thanks! I have pushed one version. Please check it.

ycemsubakan:
@huangzj421 I will take a look tomorrow. Thanks!

ycemsubakan:
Hi @huangzj421 ,

When try to run my script with, 

```python train.py hparams/convtasnet-parallel.yaml --data_folder /my_data_path/binaural_wsj0mix```

I get the following error: 

```ValueError: Interpreted https://github.com/huangzj421/Binaural-WSJ0Mix/archive/refs/heads/main.zip as web address, but could not download.
```

Is it working fine for you?

huangzj421:
> Hi @huangzj421 ,
> 
> When try to run my script with,
> 
> `python train.py hparams/convtasnet-parallel.yaml --data_folder /my_data_path/binaural_wsj0mix`
> 
> I get the following error:
> 
> ```
> ValueError: Interpreted https://github.com/huangzj421/Binaural-WSJ0Mix/archive/refs/heads/main.zip as web address, but could not download.
> ```
> 
> Is it working fine for you? (I cloned your fork `https://github.com/huangzj421/speechbrain/tree/Binaural-WSJ0Mix`, and I am working on the branch `Binaural-WSJ0Mix`.)

I am sorry for the error. Now you can generate the datasets correctly with: python train.py hparams/convtasnet-parallel.yaml --data_folder yourpath/binaural-wsj0mix/2speakers --wsj_root yourpath/to/wsj/. Thanks.

huangzj421:
> @huangzj421 , thanks for the updates.
> 
> I have noticed few things that should be fixed with data generation:
> 
> 1. It seems like you are using the `prepare_data.py` from the WSJ0-Mix recipe. The `prepare_wsjmix` function in there gives an assertion error if the chosen `--data_folder` contains the string `wsj`. I would suggest adding your own `prepare_data.py` and put it under the `Binaural-WSJ0Mix` folder.
> 2. The data generation script generates all of the combinations from the lists `['min, 'max']` and `['8k', '16k']`. Can we make it similar to the AIShell1Mix recipe where we selectively generate the subset we choose?
> 3. It would be more user friendly to have a !PLACEHOLDER for the parameter `wsj_root`.
> 
> Another thing that we ask from new recipes would be to have your checkpoint files from your best model. Can you share the model folder that speechbrain generates with a publicly accessible link? (This folder should contain all the things that speechbrain generate. You can just copy paste the folder that is called `convtasnet-parallel/` under the `results/` folder. ) A Dropbox link would work.
> 
> Otherwise, things seems to be functional and I was able to run the script.

Hi @ycemsubakan , I have sent a email with the google driver link of the training folder to @mravanelli because I don't know your email address, sorry...

ycemsubakan:
@huangzj421 @mravanelli , I fixed few things. The most important thing was, after updating `prepare_data.py` there was an inconsistency in the generated and read `.csv' file names. I also renamed the recipe to remove the hyphen, as that causes import errors, and moved `prepare_data.py`, one level up for consistency with other separation recipes. There was also an if-else error. 

I tested the three provided .yaml files, and everything seems to be working. 

If everything is okay with @mravanelli also, I think we can merge..

huangzj421:
> @huangzj421 @mravanelli , I fixed few things. The most important thing was, after updating `prepare_data.py` there was an inconsistency in the generated and read `.csv` file names. I also renamed the recipe to remove the hyphen, as that causes import errors, and moved `prepare_data.py`, one level up for consistency with other separation recipes. There was also an if-else error.
> 
> I tested the three provided .yaml files, and everything seems to be working.
> 
> If everything is okay with @mravanelli also, I think we can merge..

Thank you for your hard work! @ycemsubakan

ycemsubakan:
@huangzj421 , one more thing, Do you think it would be possible to also share some separated audio examples from the test set for the three options (parallel, cross, independent)?

recipes/BinauralWSJ0Mix/separation/hparams/convtasnet-parallel.yaml
```diff
@@ -0,0 +1,183 @@
...
+test_data: !ref <save_folder>/binaural_wsj0-2mix_tt.csv
+skip_prep: False
+
+
+# Experiment params
+auto_mix_prec: False # Set it to True for mixed precision
+test_only: False
+num_spks: 2 # set to 3 for wsj0-3mix
+progressbar: True
+save_audio: True # Save estimated sources on disk
```
huangzj421, line 44: Yes, it's okay. I have set the save_audio true in three .yamls and saved 10 items in the inference. These results are in the save/audio_results folder that I shared with you through Google driver. Thanks.

ycemsubakan: Great, thanks!

mravanelli:
Thank you @huangzj421 for this PR and @ycemsubakan for the review. I did a final review as well and everything seems to work. My only suggestion before merging this PR is to avoid using sound file as this adds an additional dependency that we prefer to avoid (as we already have torchaudio). Do you think you can replace soundfile with torchaudio?

huangzj421:
> Thank you @huangzj421 for this PR and @ycemsubakan for the review. I did a final review as well and everything seems to work. My only suggestion before merging this PR is to avoid using sound file as this adds an additional dependency that we prefer to avoid (as we already have torchaudio). Do you think you can replace soundfile with torchaudio?

Okay, I have replaced soundfile with scipy.io.wavfile in the utils.py of dataset automatical generation scripts. Thanks.

PR #1412, opened 2022-05-29
Update check to work with webdataset > 0.1

### LINKED ISSUE #1400: WebsetDataset version when using Voxlingua107 recipe sample
Hi guys. I have used recipe sample of Voxlingua107 doing language detection research.

When I run the script, I found that the WebDataset version of this script should be used as the latest one (https://github.com/speechbrain/speechbrain/commit/4846380c5b60216a8ead0e34ee206e80665c605e).   
However the dataload function of speechbrain dataio (https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/dataio/dataloader.py) still using the function `wds.dataset.Composable` which failed import from the latest version WebDataset.

**Would you please suggest the correct version of WebDataset and Speechbrain?**

Here is the details of my machine environment:
Python==3.8.11.   
Speechbrain==0.5.11 (Installed by pip).  
Webdataset==0.1.62 (The version above this do not include the function wds.dataset would cause training failed).

pplantinga:
Closes #1400 

This should allow use of WebDataset 0.2

Still needs testing

### DIFF
--- a/speechbrain/dataio/dataloader.py
+++ b/speechbrain/dataio/dataloader.py
@@ -52,8 +52,15 @@ from speechbrain.utils.checkpoints import (
 # Optional support for webdataset
 try:
     import webdataset as wds
+    from importlib_metadata import version
 
     WDS_AVAILABLE = True
+
+    # Use appropriate class based on webdataset version
+    if version("webdataset")[0:4] == "0.1.":
+        WDS_CLASS = wds.dataset.Composable
+    else:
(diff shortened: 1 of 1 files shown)


mravanelli:
@Gastron, could you please take a look at it?

pplantinga:
I've checked this with webdataset 2.5 and 1.0.103 and it doesn't crash anymore. I think we're good to merge!

mravanelli:
Thank you @pplantinga!

PR #1416, opened 2022-05-30
fix SepformerSeparation:separate_file

BuxianChen:
return signal should be normalized to [-1, 1]


ycemsubakan:
@BuxianChen Thanks for fixing this bug!

PR #1419, opened 2022-05-31
Clean up of  samples and integration tests

mravanelli:
With this PR, I cleaned up the sample folder. It wasn't well organized and contained files never used.  This will also make it easier implementing  he recipe tests (that I will implement in another PR after merging this).


anautsch:
lgtm - tests on my laptop work - on github, running the same test 3x fails on two different ends. The transducer loss < 90.0 assertion is not met (github: 92.3 value - laptop: 78.6). Another failed run on `assert end_loss < start_loss` for the core module - which did not happen on my laptop, neither in re-runs. Could this be a github issue, potentially?

Issue #1415, opened 2022-05-30
IEMOCAP emotion recognition results


WillQuCD:
Hi Guys,

thank you very much for the contribution and the great project! 

I have one question regarding the speech emotion recognition (SER) results in table 1 of paper
https://arxiv.org/pdf/2111.02735.pdf

![image](https://user-images.githubusercontent.com/28244700/170935124-646cca07-49fe-40d9-af69-13c27ce63b70.png)

Frozen-w2v-base[15] reports their results in the speaker-independent settings (no speaker overlap between training and test sets) with 5-fold cross validation;
(see the dataset preparation codes: https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/emotion/IEMOCAP_preprocess.py)

Attention Pooling[32] reports their results in a 10-fold cross-validation using a leave-one-out strategy in a speaker-independent setting as well. (see the paper https://kar.kent.ac.uk/67453/1/Atten_pooling_v4.pdf)

However, in the train.json and test.json in the following link, I found maybe your results are performed with speaker-dependent (speaker overlap between training and test sets) cases?
https://drive.google.com/drive/folders/1TslAr9NQ_nAjLeyoJ-f6xabk7n8C-HO8

Please feel free to correct me if I am wrong.

Thank you again for the great project!

WillQuCD:
Hi @BenoitWang , 
thank you very much for the quick reply. 

Yes, IEMOCAP is a very small dataset. The leave-two-session-out strategy will loss a lot of data. 

That would be great to build a standard benmark for the community by this excellent project.

thank you again for your contribution!

Adel-Moumen:
Hello,

It seems that everything is settled now. Therefore, I am closing this issue. Feel free to reopen if needed! Thanks.

PR #1429, opened 2022-06-05
Add speaker-independent splits for IEMOCAP

BenoitWang:
Hi, this PR adds a speaker independent split option for IEMOCAP preparation as mentioned in #1415, where the main modifs are:
1. leave-two-session-out -> leave-two-speaker-out, seeing that leave-two-session-out causes a huge loss of training data. (PS: For IEMOCAP, each session contains 2 speakers.) 
2. A `test_spk_id` is added and can be changed manually to perform 10-fold cross-validation.
3. Instead of creating an intermediate transformed dataset, the new preparation function reads directly the IEMOCAP_full_release folder (original dataset) and creates json splits from it.


mravanelli:
Thank you @BenoitWang for the fix. I finally was able to run it and everything works.

Issue #1399, opened 2022-05-17
DynamicBatchSampler


bofenghuang:
Hi,

I'm trying to understand the implementation of `DynamicBatchSampler`.
I would like to know, in [`_get_boundaries_through_warping`](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/dataio/sampler.py#L497), why would you use the lognorm of `s=1` to get the quantiles and linearly scale up to `max_batch_length`. What about using `lognorm.fit` ?

cc: @popcornell 

Thanks in advance !

TParcollet:
cc @anautsch maybe as well

anautsch:
Hi @bofenghuang yeah, that warping idea is on me. It's resolving time resolution in the latent statistical space rather than needing to define it explicitly for every audio collection that one fancies at a time. I write the following bulk just to make sure we are on the same page - from your message I assume you know it intuitively already, just so others can follow when reading this when archived.

Please take a look at our [tutorial](https://colab.research.google.com/drive/1mypqbHDrusZaIbqPoiEGY-WIbnpMHa2I?usp=sharing) for context. The prequel to your question: 
* long audios are a burden, random batches run easily oom for limited VRAM
* reducing batch size to 1 takes for ever in training, it wastes time and power consumption (lots of padding also)
* so, how do we create batches that need the least padding possible while maximising training efficiency?
* the go-to approach was: let's create buckets that have properties and we treat these properties then - words - so - one bucket takes only one long audio, others some medium audios, and others all the small ones - to reduce padding, tiny audios are put in here & there to reduce padding -> note: VRAM determines what's the max length of a bucket with the longest audio

Now, how to get these buckets with their fancy properties?
Without the latent space/warping approach, one would need to define: 0 to 0.2; 0.2 to 0.6; 0.6 to 1.2; ... or to script some exponential growth there. Rationale for exponential growth: most datasets are log-normal distributed when it comes to audio duration. So, the buckets are projected to treatment in a linear space. Therefore, the log-normal distribution is used.

About your question.
Which log-normal distribution to use is perhaps somewhat arbitrary. ;-)

Here, the goal was to get an initial handle on:
* `max_batch_length` which represents the VRAM limit
* `num_quantiles` which represents the targeted resolution in the latent space

For this latent space, your question is about:
* why use standard parameters for the assumed distribution?
* why not using a good fit of the actual duration distribution?

[...]

PR #1431, opened 2022-06-06
New DynamicBatchSampler

bofenghuang:
Hello @anautsch ,

As discussed in https://github.com/speechbrain/speechbrain/issues/1399, this is the PR for the new version of `DynamicBatchSampler`

- [x] Distribute data into buckets by `log-norm` distribution fitted on the datasets (by default)
- [x] Generate buckets by k-means (optional)
- [ ] Create a tool to run your pre-tests with a table output -> based on that, one could make a choice on which dynamic batching method to choose with which parameters etc.

### DIFF
--- a/speechbrain/dataio/sampler.py
+++ b/speechbrain/dataio/sampler.py
@@ -422,12 +422,21 @@ class DynamicBatchSampler(Sampler):
         seed: int = 42,
         epoch: int = 0,
         drop_last: bool = False,
+        bucket_by_kmeans: bool = False,
         verbose: bool = False,
     ):
         self._dataset = dataset
-        self._ex_lengths = {}
         ex_ids = self._dataset.data_ids
         self.verbose = verbose
+
+        self._max_batch_length = max_batch_length
+        self._shuffle_ex = shuffle
+        self._batch_ordering = batch_ordering
+        self._seed = seed
+        self._drop_last = drop_last
+        if max_batch_ex is None:
+            max_batch_ex = np.inf
+        self._max_batch_ex = max_batch_ex
 
         # We do not put a default on num_buckets to encourage users to play with this parameter
         if num_buckets is None and len(bucket_boundaries) == 0:
@@ -438,18 +447,15 @@ class DynamicBatchSampler(Sampler):
 
         if lengths_list is not None:
             # take length of examples from this argument and bypass length_key
-            for indx in range(len(lengths_list)):
-                self._ex_lengths[str(indx)] = lengths_list[indx]
+            self._ex_lengths = np.asarray(lengths_list)
         else:
             # use length func
             if not isinstance(dataset, DynamicItemDataset):
                 raise NotImplementedError(
                     "Dataset should be a Speechbrain DynamicItemDataset when using length function"
                 )
-            for indx in range(len(self._dataset)):
-                self._ex_lengths[str(indx)] = length_func(
-                    self._dataset.data[ex_ids[indx]]
-                )
+            _ex_lengths = [length_func(self._dataset.data[ex_ids[indx]]) for indx in range(len(self._dataset))]
+            self._ex_lengths = np.array(_ex_lengths)
 
         if len(bucket_boundaries) > 0:
             if not all([x >= 0 for x in bucket_boundaries]):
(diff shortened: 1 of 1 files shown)


bofenghuang:
There are also some other changes including:
- Change `_ex_lengths` from dict to np array
- Isolate the function `_get_bucket_ids` which distributes the examples into buckets by bucket boundaries, to be compatible with K-means solution
- `_bucket_lens` is now the division of `max_batch_length` by length of the longest example in the bucket (not the upper boundary of the bucket), see `_get_bucket_lens`
- Change the size of `_bucket_lens` from n+1 to n
- Modify debugging messages

anautsch:
@bofenghuang don't forget to put yourself in the author list, you are making good improvements here!

At some point, we might need to revice the tutorial as well.

anautsch:
Hi @bofenghuang - thanks for coming back to this!

Well, naturally, bad WERs are a bummer. As we wrote on Slack, the WERs are all reported for 1 epoch as is the original recipe also running exactly for one epoch with a 1.90% test clean WER. The kmeans WERs are off, the other two appear somewhat ok. With more epochs in the training, I'd resort to argue about fewer batches mean less randomness and variation as the composition of batches should vary lots between epochs. @popcornell  ;-)

BUT - we are discussing exactly one epoch here!! This pokes a particular batch composition - and this can be the outcome to any random drawing of batches, if variable batch sizes would be ok. So, what we can see is that limiting batch size compared to dynamic batch size, preserves somewhat a treatment of unknown confounding factors - a randomised controlled trial (RCT) underlying to statistic aggregation for gradient updates. @TParcollet I hope I'm not jumping off cliffs here :sweat_smile: 

Allow me to approach this issue from a more philosophical perspective (I try to limit my nonesense bits), duration is a proxy to phonetic and ambient diversity / quality of statistical aggregation. Kmeans (or any duration sampling as a strategy) somewhat distorts the sampling of unknown confounding factors - your experiments demonstrate that we move away from the RCT paradigm of DNN gradient updates. So, what happened with the original dynamic batching strategy in comparison?

The lognormal sampling ensured that magnitudes of different lengths could be placed in alike categories / buckets. The magnitude of duration served as a proxy to a series of confounding factors - not all (but remaining ones are perhaps treated balanced nonetheless). Now, the issue might be kmeans by itself which is not looking for relating magnitudes of duration but directly their values. It's telling the clustering not to consider relative factors but to discriminate absolute alikeness instead. Kmeans uses Gaussian clusters, and we look at 1D values (scalars) for clustering - is there either a kmeans with lognormal distribution instead - OR - how about feeding `exp(duration)` or `log(duration)` values instead of `duration` to kmeans?
The log(duration) values to kmeans could do the trick - for the exp(duration) values, I'm just curious :thinking: 

@mravanelli pointed out that a kmeans implementation without extra dependencies would be a nice to have. Is there a native kmeans clustering in pytorch? (by any chance)

popcornell:
I am not aware of any native k-means implementation on torch but there are tons out-there. 
E.g. look at https://github.com/lucidrains/vector-quantize-pytorch/blob/d380862ea2c01093f72b6c2e884e2d611e6b8552/vector_quantize_pytorch/vector_quantize_pytorch.py#L51

anautsch:
Hi @bofenghuang any news here?

Please fetch the latest develop branch and run the pre-commit for fixing the linters remarks.

danpovey:
If the models you are using have BatchNorm in them, non-i.i.d. selection of utterance durations could be a problem.

anautsch:
@mravanelli what do you have in mind?

@bofenghuang was working on:
* reporting estimates
* using another, kmeans based, strategy 

for dynamic batch sampling. The old sampler and the other one we merged last year (that halped in wav2vec2 training), would still remain compatible. Yet, he ran into some issue on the performance side - while training speed improved lots, WER dropped. The above discussion helped pointing out possible reasons. @danpovey made good points - what's next in this PR, is up to @bofenghuang - faster training speed with dynamic batching, there are several options to work on ... but it will be time-intensive. As @popcornell pointed out, there are implications to the entire learning process as well. Potential for sth. beautiful to speed up training & classifier performance, yet, it seems to be effort demanding.

bofenghuang:
hi @anautsch, sorry for the late response. Yeah the reason I didn't continue the PR is that this selection of examples non-i.i.d may impact the training process in an unexpected way. And I saw in some recipes utterances are sorted by duration before feeding into batches, this can't be faster.

https://github.com/speechbrain/speechbrain/blob/c1d209360d2cf742e9ea18f1384a9a5466619fba/recipes/LibriSpeech/ASR/CTC/hparams/train_sb_wav2vec.yaml#L41

I will close this issue if you agree with that @anautsch @mravanelli.

mravanelli:
Thank you @bofenghuang for your attempt! I think we can close it then.

PR #1438, opened 2022-06-11
delete unused params in yaml for TTS

BenoitWang:
Hi @mravanelli, here's a fix as discussed, and also some docstrings are fixed.


mravanelli:
Thank you for the fix @BenoitWang! I tested it and everything seems to work!

PR #1439, opened 2022-06-14
[WIP] Speaker ID - Updates on VoxCeleb speaker recognition recipe

underdogliu:
So this PR is based on earlier discussion with many great people on current problems on scoring and computing metric stats for speechbrain, along with minor fix on buggy stuff when running. This PR aims for opening such discussion and related contributions by fixing some obvious ones.

Two problems are handled here first:

- Score normalization. Well actually the implementation of speechbrain for that is not wrong. The problem lies in the algorithm themselves. There are three algorithms: z-norm, t-norm and s-norm. The latter two acquires test mean/std statistics, which may not be legitimate for voxceleb’s 1-to-1 trial list. But based on my testing, score normalization itself at least has trivial impact on the EER(%) performance and undramatic impact on DCF. Also, z-norm is valid to be applied here because it only aggregates enrollment statistics. **But just to be safe, the default score normalization methods in param file is set to be 'none'.**

- Embedding normalization. This is accordingly more problematic because we aggregates mean/var statistics across the batch. This is essentially not legitimate because we are not supposed to aggregate diff. statistics across the utterance for 1-to-1 trial lists. This can be solved by either wiping out the normalization step in the code or set the batch size to 1 when doing the inference. I removed such embedding normalization step in the scripts.

The performance are shown here, along with the old statistics:

| Model        | Version | EER, minDCF |
| ------------- |:-------------:| -----:|
| ECAPA-TDNN, cosine | Old | 0.69%, 0.08258 |
| ECAPA-TDNN, cosine | Updated | 0.90%, 0.11045 |
| TDNN, PLDA | Old | 3.2%, - |
| TDNN, PLDA | Updated | 3.23%, 0.34713 |

@mravanelli @anautsch There are also things which I would like to confirm:

1. I did not find any score normalization step in PLDA scoring, including base code. Is that correct?
2. Feature mean/std normalization. I assume this one is same as cepstral mean subtraction and variance normalization (CMVN) which is quite familiar to most speech processing folks. Is that correct or there are some batch-wise operations? (Although I do not think this is problematic for now)

We can update the statistics after approving and fixing all changed noted here, including README and hugging face READMEs. We then can move on to further discussion on optimization, as well as other speaker verification examples.


mravanelli:
Thank you for this PR @underdogliu! I would encourage all the speaker recognition experts that would like to give a contribution to help with this PR. We would like the speaker recognition community to be happy with the baselines we propose, but we need your help to make sure this is the case.

TParcollet:
Hi, starting from this branch, someone got this error in Avignon. That's weird, isn't Cyclic scheduler hooked properly? 
Capture d’écran 2022-06-24 à 08 57 50

mravanelli:
I can try to replicate. Weird, everything was working in the past..

TParcollet:
I solved the issue by just registering the Scheduler. Indeed, it is not registered (no decorator) in the code.

mravanelli:
Could you please push this change?

mravanelli:
Thank you for this PR @underdogliu!

- As for the s-norm, it looks like it relies on a cohort-based on the training data only. I might miss something, but it looks correct to me, no?
In the current version, I left it activated, but we can rediscuss it off-line (and revise that in another PR). Meanwhile, I think we can merge it to make sure at least this fix is part of the upcoming release.

PR #1443, opened 2022-06-15
RE-SepFormer

ycemsubakan:
Cleaned up the implementation of RE-SepFormer and SkiM that I had in my fork.


mravanelli:
Thank you for this PR. I will try to run the experiment tomorrow. Meanwhile, I have the following comments:

1. in `lobes/models/resepformer.py`, the docstrings of `MemLSTM`, `SegLSTM`, `ResourceEfficientSeparationPipeline`,  `ResourceEfficientSeparator` are not in the expected format. ` SBRNNBlock`, instead, is formatted correctly.
2. Make sure all the functions have a  docstring (including the forward function where you can describe the input arguments). 
3. The `README.md` should be update readme with the new models.
4. You can put the models in gdrive and mention them in the `README file`.
5. I would suggest putting the best resepformer on HF as well (in this case we have to mention it in the README as well).

mravanelli:
Thank you for this PR. I will try to run the experiment tomorrow. Meanwhile, I have the following comments:

1. in `lobes/models/resepformer.py`, the docstrings of `MemLSTM`, `SegLSTM`, `ResourceEfficientSeparationPipeline`,  `ResourceEfficientSeparator` are not in the expected format. ` SBRNNBlock`, instead, is formatted correctly.
2. Make sure all the functions have a  docstring (including the forward function where you can describe the input arguments). 
3. The `README.md` should be update readme with the new models.
4. You can put the models in gdrive and mention them in the `README file`.
5. I would suggest putting the best resepformer on HF as well (in this case we have to mention it in the README as well).

mravanelli:
I did the last tests and everything is working. We can merge it! Thank you @ycemsubakan!

PR #1445, opened 2022-06-16
Improve yaml/training script matching

mravanelli:
The goal of this PR is to further improve the matching between yaml/training script. The idea is that every hyperparameter declared in the yaml file must appear in train.py. This is almost always the case, and it is an important feature we introduced to make everything more transparent and easier to debug. However,  in some cases we slightly deviate from this good practice (and our `check_yaml.py` raises some warnings). 

Cases to fix:

- [x] Cases where we loop over the datasets with [f"{dataset}_annotation"]
- [x] Cases where the full hparam dictionary is passed to an external function
- [x] Cases where we declare or change hyperparameters in the train.py.


mravanelli:
Everything seems ok. I think we can merge it. This way I can move on with the other tests I plan to code:
1- Check if all the functions have a docstring
2- Check if all the README files on HF contain runnable code
3- Check if all the tutorials are working.

@anautsch might give a comprehensive review on the new tests right before the new release.

PR #1449, opened 2022-06-19
Adding no_sync and on_fit_batch_end method to core

RuABraun:
The wav2vec PR would change some things to core.py, this is a new PR to integrate the changes that have a wider impact (are in `core.py`) first.

The main change is a new `no_sync()` method, this is for when one is doing DDP and gradient accumulation, in which case it is wasteful to sync gradients on every forward-backward pass. One would use the `nn.module.no_sync()` context-manager to change that, but this is very awkward to do in SB since one tends to have multiple modules. This method simplifies things: You can just (in fit_batch) do `with self.no_sync(not should_step):` and everything will be taken care of.

Two other minor things are:
1. Adding a method `on_fit_batch_end()`, upon suggestion from @TParcollet this is to keep the logging separate from the training related stuff happening in `fit_batch()`. Not sure if some arguments should be passed from `fit_batch()` by default?
2. `clip_grad_norm_` is only called if `max_grad_norm != 0.`

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -883,11 +883,11 @@ class Brain:
                     self.optimizer.step()
                 self.optimizer.zero_grad()
                 self.optimizer_step += 1
-        if should_step:
-            self.on_fit_batch_end()
+
+        self.on_fit_batch_end(batch, outputs, loss, should_step)
         return loss.detach().cpu()
 
-    def on_fit_batch_end(self):
+    def on_fit_batch_end(self, batch, outputs, loss, should_step):
         """Called after ``fit_batch()``"""
         pass
 
@@ -929,7 +929,7 @@ class Brain:
                 return False
 
         # Clip gradient norm
-        if self.max_grad_norm != 0.0:
+        if self.max_grad_norm > 0.0:
             torch.nn.utils.clip_grad_norm_(
                 (p for p in self.modules.parameters()), self.max_grad_norm
             )
@@ -1300,7 +1300,7 @@ class Brain:
         Arguments
         ---------
         use : bool
-            If set to `False` will still sync gradients.
+            If set to `False` will still sync gradients, useful to make behaviour togglable.
         """
         if use:
             old_values_list = []
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -20,6 +20,7 @@ import pathlib
 import argparse
 import tempfile
 import warnings
+from contextlib import contextmanager
 import speechbrain as sb
 from datetime import date
 from enum import Enum, auto
@@ -882,8 +883,13 @@ class Brain:
                     self.optimizer.step()
                 self.optimizer.zero_grad()
                 self.optimizer_step += 1
-
+        if should_step:
+            self.on_fit_batch_end()
         return loss.detach().cpu()
+
+    def on_fit_batch_end(self):
+        """Called after ``fit_batch()``"""
+        pass
 
     def check_gradients(self, loss):
         """Check if gradients are finite and not too large.
@@ -923,9 +929,10 @@ class Brain:
                 return False
 
         # Clip gradient norm
-        torch.nn.utils.clip_grad_norm_(
-            (p for p in self.modules.parameters()), self.max_grad_norm
-        )
+        if self.max_grad_norm != 0.0:
+            torch.nn.utils.clip_grad_norm_(
+                (p for p in self.modules.parameters()), self.max_grad_norm
+            )
 
         return True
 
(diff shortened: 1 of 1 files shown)


speechbrain/core.py
```diff
@@ -882,9 +883,14 @@ def fit_batch(self, batch):
...
                 self.optimizer_step += 1
-
+        if should_step:
+            self.on_fit_batch_end()
         return loss.detach().cpu()
 
+    def on_fit_batch_end(self):
+        """Called after ``fit_batch()``"""
+        pass
+
```
pplantinga, line 893: I like this hook, since it makes it less likely people will have to override `fit_batch`, which has grown somewhat in complexity since we first implemented it. Would it make sense to have a similar function at the beginning, `on_fit_batch_start`?

speechbrain/core.py
```diff
@@ -923,9 +929,10 @@ def check_gradients(self, loss):
                 return False
 
         # Clip gradient norm
-        torch.nn.utils.clip_grad_norm_(
-            (p for p in self.modules.parameters()), self.max_grad_norm
-        )
+        if self.max_grad_norm != 0.0:
```
pplantinga, line 932: Since it wouldn't make sense to have a negative norm, perhaps a better comparison would be:
```suggestion
        if self.max_grad_norm > 0.0:
```

speechbrain/core.py
```diff
@@ -1281,6 +1288,38 @@ def update_average(self, loss, avg_loss):
...
         return avg_loss
 
+    @contextmanager
+    def no_sync(self, use=True):
+        """Copies pytorch's implementation for doing no_sync across all modules.
+
+        Explanation: nn.module.no_sync() is a context manager for when one does
+        not want to sync gradients, which happens when using both DDP and gradient accumulation.
+        Speechbrain brain's class can contain multiple modules and calling no_sync on these
+        individually would be very awkward, therefore this contextmanager exists.
```
pplantinga, line 1298: This seems like a useful utility! I'm curious though, is there a reason we can't use the PyTorch version of `no_sync` and have to roll our own?

speechbrain/core.py
```diff
@@ -882,9 +883,14 @@ def fit_batch(self, batch):
                     self.optimizer.step()
                 self.optimizer.zero_grad()
                 self.optimizer_step += 1
-
+        if should_step:
+            self.on_fit_batch_end()
```
pplantinga, line 887: I'm wondering if we ought to always call `on_fit_batch_end` and perhaps pass `should_step` as an argument. Someone might want to do something there, even if `should_step` is `False`.

pplantinga (changes requested):
This will be a great addition once the comments are addressed!

speechbrain/core.py
RuABraun: I guess so, although I imagine the point of `on_fit_batch_start` is for people to do stuff with results like the loss etc. that have been calculated, what would `on_fit_batch_start` take as input?

speechbrain/core.py
RuABraun: Makes sense will do

speechbrain/core.py
RuABraun: Imagine your brain class has modules: `[module1, module2, module3]`

Then you would have to call:
```
with module1.no_sync():
    with module2.no_sync():
        with module3.no_sync():
            fit_batch contents
```
which I don't think is nice..

pplantinga: Aha, this makes sense

speechbrain/core.py
pplantinga: I think passing the loss would likely make sense too.

speechbrain/core.py
pplantinga: If there's no obvious use case for `on_fit_batch_start` then I guess we should wait to add it until its needed under the YAGNI principle.

pplantinga: Do we need any significant effort to migrate metric computation code to the on_fit_batch_end?

speechbrain/core.py
pplantinga: If one were going to compute e.g. WER in the `on_fit_batch_end`, they would need both the `predictions` and `batch` as well.

speechbrain/core.py
RuABraun: done

speechbrain/core.py
RuABraun: I don't think so, since AFAIK there is no metric computation stuff in core.py ?

speechbrain/core.py
RuABraun: pushed, `on_fit_batch_end()` is now passed `batch, outputs, loss, should_step`

let me know what you think

RuABraun:
I could also add a `with no_sync(not should_step)` to fit_batch, the way it's implemented it will only not sync if DDP is used and should_step is False, which is the desired behaviour, what do you think @pplantinga ?

speechbrain/core.py
pplantinga: I guess it doesn't need to be part of this PR, but I believe we should start another once this is in to migrate the `templates` at the very least to do metric computation in this function, as that would be much cleaner.

speechbrain/core.py
RuABraun: true, will do :+1:

TParcollet:
Hum, I have no idea what No_sync will do outside of DDP though.

RuABraun:
It won't do anything, it will just be an empty context manager.

TParcollet:
Then let's have it only for DDP.

pplantinga:
> Then let's have it only for DDP.

I'm not sure exactly what you mean, there isn't an easy way to apply this only when DDP is active. Code without:

```
loss = self.compute_objectives(outputs, batch, Stage.TRAIN)
(loss / self.grad_accumulation_factor).backward()
```

Code with:
```
loss = self.compute_objectives(outputs, batch, Stage.TRAIN)
with self.no_sync(not should_step):
  (loss / self.grad_accumulation_factor).backward()
```

I think this is a good addition because it speeds up computation by 10-20% in cases where DDP and grad accumulation are used.

RuABraun:
What's our conclusion here?

pplantinga:
I'd say lets go ahead with the `self.no_sync(not should_step):` feature

RuABraun:
Okay great, will add the final touches today/tomorrow

speechbrain/core.py
RuABraun: done

pplantinga (approved):
Looks good!

PR #1453, opened 2022-06-20
Docstring Checks

mravanelli:
This PR adds a test to make sure all the functions/classes have a docstring. This is another contribution towards more tests.


mravanelli:
All tests are passing now. I think we can merge that such that I can move on with the rest of the tests I would like to implement. @anautsch will take a look at all the new test right before releasing the new version of speechbrain.

PR #1456, opened 2022-06-21
Remove redundant checks for None

SkuratovichA:
lines 84-95 ensure, that {train, dev, test}_tsv_file != None.


mravanelli:
Thank you for this PR @SkuratovichA! I tested and it works. I only fixed the style with black. It looks like the tests are failing, due to a change done on HF (but this is not related to this PR).

PR #1458, opened 2022-06-21
Fix use of `huggingface_hub.cached_download`

SolomidHero:
issue #1459

### DIFF
--- a/speechbrain/pretrained/fetching.py
+++ b/speechbrain/pretrained/fetching.py
@@ -9,6 +9,7 @@ import urllib.error
 import pathlib
 import logging
 import huggingface_hub
+from requests.exceptions import HTTPError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


mravanelli:
It looks like this is due to an interface change in the latest version of HuggingFace Hub. We can fix it in the development branch, but people using the main branch will see this error. Also, it seems your fix doesn't solve the problem:

SolomidHero:
Should we also update requirements:
https://github.com/speechbrain/speechbrain/blob/786ec3252658253009010a0bcf87db77bb892a96/requirements.txt#L2
to the moment this hf_hub_download function was proposed (0.0.19 as latest patch version)

TParcollet:
I hope it won't break other things to do this ...

mravanelli:
Thank you @SolomidHero @julien-c for the help. I think we can merge it!

PR #1464, opened 2022-06-22
Fix consistency issue with threshold

aviasd:
In lines 496 and 566, the predicted positives are defined as the scores that are strictly greater than the threshold.
In line 425, there was a consistency issue because the predicted positives were defined as scores greater than or equal to the threshold.
I fixed it and removed the equal sign, so now there is consistency between the definitions of predicted positives throughout the file.


mravanelli:
LGTM! Thank you  @aviasd!

PR #1466, opened 2022-06-22
Conformer fix

popcornell:


mravanelli:
As mentioned privately, we can merge this PR and make it part of the upcoming release. I'm training and experiments and it looks everything is fine. I will update the readme file when the experiment will end.

TParcollet:
@popcornell @mravanelli should we scale this to see if we can reach ESPNET?

mravanelli:
Right, that might be the case. Currently the conformer is pretty small...

PR #1468, opened 2022-06-23
Changed EER function

AlmogAmiga:
Made it more memory efficient, the FAR and FRR aren't saved in tensors anymore but in separate variables.

### DIFF
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -496,13 +496,12 @@ def EER(positive_scores, negative_scores):
 
         # Finding the threshold for EER
         if (FAR - FRR).abs().item() < abs(final_FAR - final_FRR) or i == 0:
-           min_index = i
-           final_FRR = FRR.item()
-           final_FAR = FAR.item()
+            min_index = i
+            final_FRR = FRR.item()
+            final_FAR = FAR.item()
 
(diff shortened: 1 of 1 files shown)


mravanelli:
Hi @AlmogAmiga, thank you for this PR! Indeed, the vectorized EER computation can be demanding. I compared the two implementations on voxceleb by running:

`python speaker_verification_cosine.py hparams/verification_ecapa.yaml --data_folder=/localscratch/voxceleb1/`

When I compare the current vectorized implementation and the proposed non-vectorize one, the results are slightly different:

Vectorized:
```
__main__ - EER(%)=0.696605
__main__ - minDCF=0.076422
```

Non-vectorized
```
__main__ - EER(%)=0.701922
__main__ - minDCF=0.076103
```
Any idea? It could be great if you can investigate more before doing the switch.

AlmogAmiga:
Hello @mravanelli, that's odd, because at the time I wrote this code, I've also checked if it gives the same results as the original version (Vectorized) and it gave exactly the same numbers. What was different in my case though is that I've checked it with the plda verification code on ecapa. Maybe that's the reason. I'll try to look into it and let you know if I find anything.

mravanelli:
I checked it again and everything is fine. Thank you for your contribution @AlmogAmiga!

AlmogAmiga:
Great, happy to hear 😀

PR #1471, opened 2022-06-24
add decorator to CyclicLRScheduler

mrouvier:
Add decorator to CyclicLRScheduler.


TParcollet:
Huge thanks! @mravanelli dunno how this did not triggered an error before btw.

TParcollet:
@underdogliu FYI.

Issue #1469, opened 2022-06-23
Failing tests


GAN-pie:
Hi, I have fresh installed speechbrain from 0.5.8 to the last release and I spotted 4 tests that are failing with weird assertions. I'm not 100% sure but I do not think these tests were failing on the older install.

Here is the tests output:

```
FAILED tests/unittests/test_features.py::test_features_multimic[cuda] - AssertionError: assert tensor(0.0008, device='cuda:0') < 5e-05
FAILED tests/unittests/test_profiling.py::test_profile_class[cuda] - assert 3 == 2
FAILED tests/unittests/test_profiling.py::test_scheduler[cuda] - assert 1 == 0
FAILED tests/unittests/test_profiling.py::test_aggregated_traces[cuda] - assert (4222 == 4209 or 4209 == 2993)
```

anautsch:
Hi @GAN-pie thanks for reaching out! I can assist on the tests for profiling. We provided a wrapper to benchmark VRAM peaks and real-time demands. The reported errors are from mismatching number of nested function calls. We have had similar issues over the past weeks with test_profiling as the number of nested function keeps changing with operating systems, library versions etc. - sorry for that. We use pytorch's profiling tools and wrap them. I've tested on cpu only - I see you run tests on cuda. 

Which pytorch version are you running? (curiosity question)

@TParcollet these issues with the profiling tests start to get more annoying than that they help checking on the profiling tool - mainly as everything underlying keeps changing like a moving target. That in essence makes harmonised profiling & benchmarking questionable by itself since it really depends on everything then that is circumstantial to the actual module under test. It's all of the code, hardware, ... that is under test with profiling. Comparability across sites gathering such data can only be achieved on categorical/nominal scales.

Will update with reducing profiling test assertions -

GAN-pie:
@anautsch 2 out 3 test_profiling.py seem to work properly now, but still one that is messing with the assert 1 == 0 statement.

`FAILED tests/unittests/test_profiling.py::test_scheduler[cuda] - assert 1 == 0`

`E       assert 1 == 0
E        +  where 1 = len([<FunctionEvent id=0 name=cudaDeviceSynchronize device_type=DeviceType.CPU node_id=-1 cpu_time=8.000us start_us=805 en...ead=0 input_shapes=[] cpu_memory_usage=0 cuda_memory_usage=0 is_async=False is_remote=False seq_nr=-1 is_legacy=False>])
[...]

PR #1476, opened 2022-06-27
fix for #1469, minimal testing for profiling

anautsch:
@TParcollet regarding #1469 -> I commented out all assertions that are out of my control for testing the profiling wrapper. There are too many changes on how many nested functions are called for changes of package installation, libraries, OS, ... - this way testing is more slim but shouldn't raise issues in the future. let's see


TParcollet:
lgtm

PR #1487, opened 2022-06-30
test clean-ups

anautsch:
Made a few clean-ups on tests:
* doctest & unittest -> temporary dirs outside of repo's root folder (testing is less noisy in the file system)
* most files created by tests in `tests/tmp` (if not in other temporary folders outside of the local repo)
* load-yaml-tests -> a. cpu-friendly scripts only | b. rir download to `tests/tmp/rir` (aim: regardless of which test runs, it's 1.3 Gb each time for the same archive - so, recipes can refer to the same download destination and skip if existing already)

Some recipes were adjusted (minor edits).

#1469 - test for profiling adjusted: CUDA presence issues recording of events, although PyTorch profiler with warming-up scheduler schould not record anything (but it does) ... during testing, these events were simply removed from the list to not interfer with PyTorch profiler's doings - we provide a wrapper only.


mravanelli:
Thank you @anautsch, thank you for this improvement. I re-run all the tests and everything seems to work as expected.

PR #1488, opened 2022-06-30
Added README link to interactive sandbox

mernit:
I was playing around with speechbrain and wanted an easy way to deploy the inference pipeline, so I hosted a basic app for the pretrained model on this [Slai Sandbox](https://www.slai.io/s/ss0u0). All the required libraries are installed, there's a sample audio file included, and the pipeline can be quickly customized or deployed as an API.


anautsch:
@mernit thank you for suggesting—we discussed internally and might for now not refer to the Slai Sandbox in the README.

PR #1504, opened 2022-07-14
Remove torch maximum compatible version

### LINKED ISSUE #1493: Torch 1.12 not compatible?
working to install speechbrain 0.5.12, and getting the error that "speechbrain 0.5.12 requires torch<=1.11,>=1.7, but you have torch 1.12.0 which is incompatible." read elsewhere that it should work with >=1.7.

pplantinga:
Closes #1493 

As PyTorch is more stable, I think it makes sense to remove the maximum compatible version.

Is anyone aware of why the minimum versions were different between `setup.py` and `requirements.txt`? I updated `setup.py` to reflect the more stringent requirement in `requirements.txt` but anyone with more info could chime in if there is a good reason for it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
         "packaging",
         "scipy",
         "sentencepiece",
-        "torch>=1.7,<=1.11",
+        "torch>=1.9",
         "torchaudio",
         "tqdm",
         "huggingface_hub",
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,6 @@ pre-commit>=2.3.0
 scipy>=1.4.1
 sentencepiece>=0.1.91
 SoundFile; sys_platform == 'win32'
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -37,7 +37,7 @@
         "packaging",
         "scipy",
         "sentencepiece",
-        "torch>=1.7,<=1.11",
```
TParcollet, line 40: Wouldn't this cause an error ? 1. nothing after

pplantinga: The `7` is replaced by a `9` to try and synchronize the `requirements.txt` and `setup.py` to have the same minimum version of torch. I'm not totally sure why it was different to begin with, if you know it was intentional we can leave it as `1.7`

TParcollet: dunno ...

TParcollet (approved):

PR #1505, opened 2022-07-16
Decay gumbel temperature over training steps

thadeous:
Updated the Common Voice / self-supervised learning / wav2vec training recipe to decay the gumbel temperature at the end of each update.  Added three values to the corresponding yaml file in support:
gumbel_temp_max: 2.0
gumbel_temp_min: 0.5
gumbel_temp_decay: 0.999995 # Decay of gumbel temperature during training
The default values are the same used within the huggingface trainer: https://github.com/huggingface/transformers/blob/v4.20.1/examples/research_projects/wav2vec2/run_pretrain.py#L275-L282

### DIFF
--- a/recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
+++ b/recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
@@ -123,6 +123,10 @@ class W2VBrain(sb.core.Brain):
 
                 # anneal lr every update
                 self.hparams.noam_annealing(self.optimizer)
+                # decay gumbel temperature until it reaches the minimum value
+                self.modules.wav2vec2.module.model.set_gumbel_temperature(
+                    max(self.hparams.gumbel_temp_max * .99995 ** self.step // self.hparams.gradient_accumulation,
+                        self.hparams.gumbel_temp_min))
 
         return loss.detach()
 
--- a/recipes/CommonVoice/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml
+++ b/recipes/CommonVoice/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml
@@ -41,6 +41,9 @@ number_of_epochs: 100
 lr_adam: 2.0 # This will get reduced by the training scheduler
 weight_decay: 0.01
 d_model: 768  # Needed by the scheduler. 768 is for the BASE w2v2
+gumbel_temp_max: 2.0
+gumbel_temp_min: 0.5
+gumbel_temp_decay: 0.999995 # Decay of gumbel temperature during training
 sorting: ascending
 auto_mix_prec: False
 sample_rate: 16000


TParcollet:
Hi @thadeous and thanks for the PR! Could you please fix the pre-commit tests before I proceed with reviewing? You can install it following the documentation, it's very quick :-)

TParcollet:
@RuABraun may I ask you to tell me if this is correct to your understanding of the decay in w2v2? :-) Thanks folks!

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
```diff
@@ -123,6 +123,10 @@ def fit_batch(self, batch):
 
                 # anneal lr every update
                 self.hparams.noam_annealing(self.optimizer)
+                # decay gumbel temperature until it reaches the minimum value
+                self.modules.wav2vec2.module.model.set_gumbel_temperature(
+                    max(self.hparams.gumbel_temp_max * self.hparams.gumbel_temp_decay ** self.step //
+                        self.hparams.gradient_accumulation, self.hparams.gumbel_temp_min))
```
RuABraun, line 129: It's not clear to me, will this work correctly for when training with 1 GPU and when training with multiple? In the huggingface code there is a check for this.

RuABraun (commented):
Other than the one comment LGTM

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
thadeous: The paper has this line: "The Gumbel softmax temperature tau is annealed from 2
to a minimum of 0.5 for BASE and 0.1 for LARGE by a factor of 0.999995 at every update.", but I can can checkout the fairseq implementation.  For my hack, I have the values hardcoded running on a single dgx across the 8 V100s with the data_parallel_backend flag (I didnt properly use the yaml file, but I can test tomorrow).  I suspect its uncommon to use that data_parallel_backend flag with the recipe since the recipe receives an array of loss values and erroneously attempts to solve the gradients.  I updated to take the mean prior.

thadeous:
Absolutely!  The paper has this line: "The Gumbel softmax temperature tua is annealed from 2
to a minimum of 0.5 for BASE and 0.1 for LARGE by a factor of 0.999995 at every update.", but I will check out the fairsq repo.  I simply jammed the hardcoded value into the recipe to evaluate but I will test the yaml piece tomorrow.   I run on a single dgx across the 8 v100s.  I should note that I use the data_parallel_backend flag.  I suspect its uncommon to use that flag with this recipe since it will receive an array of loss values across the gpus and attempts to erroneously solve the gradients.  I just modified to take the mean prior to provide a scalar.

RuABraun:
Okay so the pytorch documentation has said for years that DistributedDataParallel is always better than DataParallel (even when running on just one node). So I think it's important to make the code work for that.
If you look at the huggingface code in the [link](https://github.com/huggingface/transformers/blob/v4.20.1/examples/research_projects/wav2vec2/run_pretrain.py#L275-L282) from your initial comment, you can see there's a check for when one is using multiple GPUs. This is because when you use DDP the model gets wrapped into a DDP instance, which changes the interface for accessing the pytorch modules of the model (hence the if/else in the huggingface code). I'm pretty sure you need to do this as well otherwise it will crash when using DDP.

TParcollet:
Hi @thadeous indeed, it would be lovely if you can make it work with DDP as well. By the way, I encourage you to use DDP, especially with DGX machines, as I would expect a good boost in performance :p

anautsch:
@thadeous there is a linters error blocking the merger.

Please fetch the latest version of the develop branch into your PR branch, (just to be up to date) and run `./tests/.run-linters.sh`. There's a line in `recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py` that needs reformatting with the `black` tool. All other tests, on Jul 19, were passing.

recipes/CommonVoice/self-supervised-learning/wav2vec2/train.py
anautsch: @thadeous any updates/comments?

Adel-Moumen:
Hello @thadeous, 

Thanks for your PR.

We recently deprecated the support of Data Parallel (DP) in favour of Distributed Data Parallel (DDP). Now, it should be straightforward to modify your code to support DDP. Could you please confirm me if you are still willing to finish this PR? Sorry for the delay... 

Best, 
Adel

Issue #1459, opened 2022-06-21
[Bug] huggingface_hub.cached_download() takes 1 positional argument but 2 were given


SolomidHero:
Hi!

Followed [this model](https://huggingface.co/speechbrain/asr-crdnn-transformerlm-librispeech)
`speechbrain/asr-crdnn-transformerlm-librispeech`
from huggingface models hub.

## Error
code:
```py3
EncoderDecoderASR.from_hparams(
    source=asr_model_source,
    savedir=to_absolute_path('./models'),
    run_opts={'device': torch.device('cpu')}
)
```

trace (catched in pytest):
```tb
score.py:101: in main
    asr_model = EncoderDecoderASR.from_hparams(
../miniconda3/lib/python3.9/site-packages/speechbrain/pretrained/interfaces.py:335: in from_hparams
    pymodule_local_path = fetch(


>               fetched_file = huggingface_hub.cached_download(url, use_auth_token)
E               TypeError: cached_download() takes 1 positional argument but 2 were given

../miniconda3/lib/python3.9/site-packages/speechbrain/pretrained/fetching.py:116: TypeError
```

# Versions

`pip freeze | grep -E hugging|speech`:
```
huggingface-hub==0.8.1
speechbrain==0.5.11
```

version of `huggingface-hub<0.8` works fine

## Solution

use_auth_token should be a keyword argument

Here is my contribution #1458

camilotorron:
I had the same issue this morning. I downgraded hugging-face version to 0.7.0 but still not working like yesterday. When performing "verification.verify_files()" this error appears:

'types.SimpleNamespace' object has no attribute 'mean_var_norm_emb'

TParcollet:
@mravanelli shouldn't this be fixed now ?

Arnold-git:
`from speechbrain.pretrained import SpeakerRecognition
verification = SpeakerRecognition.from_hparams(source="speechbrain/spkrec-ecapa-voxceleb", savedir="pretrained_models/spkrec-ecapa-voxceleb")
score, prediction = verification.verify_files("speechbrain/spkrec-ecapa-voxceleb/example1.wav", "speechbrain/spkrec-ecapa-voxceleb/example2.flac")

print(prediction, score)`

error

`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-4-4e6723d977df>](https://localhost:8080/#) in <module>()
[...]

PR #1508, opened 2022-07-18
Remove limit for HF hub as it does not work with colab

TParcollet:
Google colab is failing because HF hub bellow 0.7 is not managed properly due to some tensor flow version mismatch. We remove the maximum HF HUB version to avoid this

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 -r lint-requirements.txt
-huggingface_hub>=0.6.0,<=0.8.0
+huggingface_hub>=0.7.0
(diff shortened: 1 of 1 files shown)


pplantinga:
Could someone try the latest version of `huggingface_hub` with `fetch()`? We want to make sure nothing breaks here.

For reference, these are the relevant issues/PRs regarding the limit:

#1458 , #1459

pplantinga (approved):
Tried `pretrained` model with `huggingface_hub==0.8.1` and it works.

Issue #1509, opened 2022-07-20
Use specific version of a model hosted on Hugginface 


hbredin:
`huggingface_hub.hf_hub_download` supports passing a `revision` parameter to `hf_hub_{url|download}` to load a specific version (understand: commit, tag, or branch head) of models hosted on Huggingface Hub:

```python
from huggingface_hub import hf_hub_download
hf_hub_url(repo_id, filename, revision=..., ...)
```

It would be nice to expose this `revision` parameter in `speechbrain` so that one can pin a model to a specific version (and ensure that a given code will still behave the same even if the model is updated upstream).

The API could look like this:

```python
from speechbrain.pretrained import EncoderClassifier
encoder = EncoderClassifier.from_hparams(
    source="speechbrain/spkrec-ecapa-voxceleb",
    revision="5c0be3875fda05e81f3c004ed8c7c06be308de1e",
)
```

or maybe like this that @julien-c [implemented](https://github.com/pyannote/pyannote-audio/blob/2decda1d0482578ae1f64aa98cbb6b5fbe32b6be/pyannote/audio/core/model.py#L773-L782) for `pyannote.audio`:

```python
from speechbrain.pretrained import EncoderClassifier
encoder = EncoderClassifier.from_hparams(
    source="speechbrain/spkrec-ecapa-voxceleb@5c0be3875fda05e81f3c004ed8c7c06be308de1e")
```

TParcollet:
Done https://github.com/speechbrain/speechbrain/pull/1510.

PR #1510, opened 2022-07-20
Add revision to hub download

TParcollet:
This is a feature requested in #1509. We now can request a specific revision when using our interfaces. 

@hbredin for some reasons, I can't add you as a reviewer so I asked someone from the team to do it. Please feel free to try the PR and let us know if it works well for you!

### DIFF
--- a/speechbrain/pretrained/fetching.py
+++ b/speechbrain/pretrained/fetching.py
@@ -30,6 +30,7 @@ def fetch(
     overwrite=False,
     save_filename=None,
     use_auth_token=False,
+    revision=None,
 ):
     """Ensures you have a local copy of the file, returns its path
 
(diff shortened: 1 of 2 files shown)


hbredin:
Thanks @TParcollet! 

LGTM though I must confess I like the `source=repoId@revisionId` API syntax better than `source=repoId, revision=revisionId` but that is probably just me.

pplantinga (approved):
I tried this and everything seems to work! I was wondering if there were other instances of `fetch()` that ought to be updated, and came across `speechbrain.pretrained.interfaces.foreign_class` which doesn't seem to be referenced anywhere? And I also looked at the `speechbrain.utils.parameter_transfer` which doesn't seem straightforward to modify.

All this to say, LGTM.

mravanelli:
Thank you all!

PR #1511, opened 2022-07-21
Fix transducer loss + --distributed_launch=True in recipes

Adel-Moumen:
Hi,

I added a small fix to raise an error when non-cuda devices are fed into the transducer loss because the current implementation relies on Numba which only allows cuda arrays inputs.

The second one is just a replacement of distributed_launch=True with --distributed_launch in each recipe as discussed in https://github.com/speechbrain/speechbrain/pull/1498 .


TParcollet:
Thanks @Adel-Moumen.

PR #1518, opened 2022-07-26
Fixing shuffle argument for distributed sampler in core.py

### LINKED ISSUE #1495: `make_dataloader` always sets shuffle=False when DistributedSampler is employed.
In speechbrain.core.Brain class, when DistributedSampler is employed, it sets shuffle to False. Is there any reason to do so?
image

RuABraun:
Currently shuffling of the sample indices is not done before distributing the indices across nodes, this means the distribution will be the same for each epoch. This causes a small but noticeable decrease in performance. 

Fixes #1495 

I don't think there's ever a reason to have it set to False, therefore hardcoded to True.

I also took the liberty of adding a tiny bit of extra info to why `DistributedSamplerWrapper` needs to exist.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -755,7 +755,7 @@ class Brain:
             elif loader_kwargs.get("batch_sampler") is None:
                 # no sampler and batch-sampler
                 self.train_sampler = DistributedSampler(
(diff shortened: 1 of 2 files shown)


TParcollet (approved):
LGTM!

TParcollet:
If the tests pass, we can merge @RuABraun

PR #1519, opened 2022-07-27
Conformer separation

ycemsubakan:
I added a wrapper class to use conformer for separation. 

Also a small fix to remove the 'optional' from `pos_embs` argument of `ConformerEncoder` forward function, as it requires specifying the positional embeddings (it is not optional).


speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -355,9 +356,18 @@ def forward(
...
         src_key_padding_mask : torch.Tensor, optional
             The mask for the src keys per batch.
-        pos_embs: torch.Tensor, torch.nn.Module, optional
+        pos_embs: torch.Tensor, torch.nn.Module,
             Module or tensor containing the input sequence positional embeddings
+            If custom pos_embs are given it needs to have the shape (1, 2*S-1, E)
+            where S is the sequence length, and E is the embedding dimension.
         """
+
+        if self.attention_type == "RelPosMHAXL":
```
popcornell, line 365: Looks good to me!

popcornell (approved):

recipes/WSJ0Mix/separation/hparams/sepformer-conformerintra.yaml
```diff
@@ -0,0 +1,170 @@
...
+
+# e.g. '/yourpath/wsj0-mix/2speakers'
+# end with 2speakers for wsj0-2mix or 3speakers for wsj0-3mix
+data_folder: !PLACEHOLDER
+
+# the path for wsj0/si_tr_s/ folder -- only needed if dynamic mixing is used
+# e.g. /yourpath/wsj0-processed/si_tr_s/
+# you need to convert the original wsj0 to 8k
+# you can do this conversion with the script ../meta/preprocess_dynamic_mixing.py
+base_folder_dm: /yourpath/wsj0-processed/si_tr_s/
```
anautsch, line 22: I like the example in the comments
`!PLACEHOLDER` will serve well here

anautsch (approved):
lgtm - there's a small thing to change in the yaml file.

I couldn't run a test since I've no access to LDC data.
Is there a comparable mini dataset that I could use to run the recipe in `--debug` mode ?

recipes/WSJ0Mix/separation/hparams/sepformer-conformerintra.yaml
ycemsubakan: @anautsch , I prefer not having the `!PLACEHOLDER` here, as I think it's better not to have code crash in case where we do not use dynamic mixing.

anautsch: good to know ! :D

PR #1525, opened 2022-08-01
Fix to use output of unsqueeze() in Tacotron2 parse_decoder_outputs()

jqug:
Hi, this is a fix for an issue where an extra dimension of gate_outputs was not being added (since unsqueeze() doesn't operate inplace). In case the Tacotron2 model generates only a single frame of output, this would cause a failure downstream when parsing the model output.

Thanks for all the work on this great project!


anautsch:
Hi @jqug please fetch the latest develop branch into your PR branch. 

The merger is blocked on our side by a testing error
```
  File "/opt/hostedtoolcache/Python/3.8.13/x64/bin/pytest", line 8 in <module>
/home/runner/work/_temp/cd68f463-66e5-4420-bbfe-1ac2fb187ee6.sh: line 1:  2012 Aborted                 (core dumped) pytest --doctest-modules speechbrain
speechbrain/processing/PLDA_LDA.py 
```

It goes without saying, this error has another cause—other PRs faced it also (in short, it was for updates in dependencies); we fixed it some weeks ago. Yet, all PRs need to fetch that fix into their branch; for which, please fetch the latest develop branch. Then, we can re-run the tests and get much closer to an "ok" for merger :)

BenoitWang (approved):

jqug:
Thanks @BenoitWang and @anautsch - I've pulled the recent changes, let me know if any other action is needed.

anautsch (approved):

Issue #1522, opened 2022-07-29
No module named 'torch.autograd.profiler_util' error


fosimoes:
I am getting the following error when importing speechbrain.

    import speechbrain as sb
speechbrain/__init__.py:4: in <module>
    from .core import Stage, Brain, create_experiment_directory, parse_arguments
speechbrain/core.py:35: in <module>
    from speechbrain.utils.distributed import run_on_main
speechbrain/utils/__init__.py:11: in <module>
    from . import *  # noqa
speechbrain/utils/profiling.py:11: in <module>
    from torch.autograd.profiler_util import (  # pytorch v1.10.1
E   ModuleNotFoundError: No module named 'torch.autograd.profiler_util'

The error occurs with speechbrain v0.5.12, but v0.5.11 works fine.
I tested both torch 1.10.0 and 1.11.0, with similar results.
Any idea what it might be?

anautsch:
@fosimoes sorry to hear that - and thanks for reporting. 

We introduced profiling based on pytorch's profiler in v0.5.12 - I looked into it and tested with pytorch 1.10 and 1.11 - earlier pytorch versions didn't had `torch.autograd.profiler_util` (hence the comment); looks like they removed it afterwards again (althought a legacy pytorchh profiler library existed already with pytorch 1.11, if I remember correctly). sigh. Until now, it has caused more trouble than good, as everything changes between systems and their environment configurations (e.g., the number of nested function calls).

Wondering if it would be best to take some/all profiling features out - or - to only activate it for certain pytorch versions. Or perhaps another way...

The methods that are imported from `torch.autograd.profiler_util` are only relevant for some 'nicer' reporting. Their imports and the depending code could also be simply removed for the time being.

@mravanelli @TParcollet what are your thoughts? Would it also make sense to have a requirement check/dependency testing if functions we intend to call are still present for all sb requirements and their version ranges?

mravanelli:
Hi Andrea,
maybe we can take the profiling test out and run them only periodically
(just like what we do with other tests like the recipe tests, etc). What do
you think?

fosimoes:
update: 
Despite using a dockerfile based on pytorch:1.11.0, I was forcing torchaudio to v0.9.0 (due to previous conflicts), which caused pytorch to downgrade to v1.9.0.  So my tests were actually run on pytorch 1.9.0, which caused the profiler to fail.
So I changed my dockerfile to force torchaudio to v0.11:
[...]

PR #1526, opened 2022-08-01
fix 1522

anautsch:
see #1522 

Imports in `core.utils.profiling.py` were too depending on a particular pytorch version (were dropped later). This caused errors or downgrades on user site. Depending tests were commented out as well.


anautsch:
@mravanelli please take a look. 

If you want, I can put the profiling tests in one test function (right now, each test function call takes the same amount of time of some 0.5s which could indicate fixed setting up/tearing down costs are repeated - otherwise, the tests demonstrate profiling is more expensive as a whole activity than just running scripts; which should be the case, too).

mravanelli (approved):

PR #1535, opened 2022-08-09
Pass overrides_must_match in from_hparams to set it in load_hyperpyyaml

AlonKellner:
I tried to load one of the models on huggingface with the `from_hparams` method, when trying to use the `overrides` variable to set the `device` hparam I got an error, the original hyperparams.yaml did not have `device` set in it.  
After considering it, it seems trivial that when `overrides` is set then `overrides_must_match` should be able to be set as well, rendering it much more flexible.


pplantinga:
This change seems reasonable to me, however, I would note that the intended way to change `device` is through the `run_opts` parameter. For example:
```python
Pretrained("speechbrain/model", run_opts={"device": "cpu"})
```

AlonKellner:
> This change seems reasonable to me, however, I would note that the intended way to change `device` is through the `run_opts` parameter. For example:
> 
> ```python
> Pretrained("speechbrain/model", run_opts={"device": "cpu"})
> ```
> 
> This could probably be better documented.

Cool to know, thanks :)

mravanelli:
@pplantinga, do you think we shoud close this PR?

pplantinga (approved):
This is a simple change that will allow more flexibility with hparams, looks good to me!

mravanelli:
One small issue is that the new argument `overrides_must_match` is not described in the docstring. @AlonKellner, could you please add a short description before merging?

anautsch:
@AlonKellner what's the status on this PR?

This PR looks approved, with minor comments though.

Adel-Moumen:
Hello @mravanelli and @pplantinga, what do you think of this PR? I see that there's a lack of docstring. Do you think the PR is still relevant? If so, I will update it myself to speedup a bit the process.

pplantinga:
I merged develop and added a docstring, this is ready to review again.

PR #1546, opened 2022-08-18
updating scipy version in requirements.txt.

nauman-daw:
Hi @TParcollet ,
As we discussed over chat, there are some issues with the latest version of spicy (1.9) and PLDA code. I tested the code and it works with the latest version of scipy on MacOS but fails on Linux.  I made changes to the requirements.txt to restrict scipy to 1.8.1 till we find a reason for this behavior. Meanwhile, it should be safe to merge PRs that do not make use of scipy.


TParcollet (approved):
LGTM

Issue #1514, opened 2022-07-25
topk argument should also return top k hypothesis


KacperKubara:
For all classes inheriting S2SBeamSearcher, if I set the **topk** argument only top k scores are returned. Snippet from the last part of the forward() function of this class

``` Python
        ) = self._get_top_score_prediction(hyps_and_scores, topk=self.topk,)
        # pick the best hyp
        predictions = topk_hyps[:, 0, :]
        predictions = batch_filter_seq2seq_output(
            predictions, eos_id=self.eos_index
        )

        if self.return_log_probs:
            return predictions, topk_scores, log_probs
        else:
            return predictions, topk_scores
```
As you can see, only topk scores are returned, not top k hypothesis. 

That it is incorrect, as according to the documentation: `topk  – The number of hypothesis to return. (default: 1)`.

KacperKubara:
Hi @TParcollet, do you know if there is any way to return the hypothesis without forking a Speechbrain repo and changing the source code?

TParcollet:
Hi, I am unsure, as the expected behavior would be to really return the predictions in "predictions", maybe we did something wrong. If so, we would be glad to see someone opening a PR for that. HR ressources are currently drained by the next version ...

KacperKubara:
Ok no worries, I have created a fix on my fork so I can propose the PR, maybe even today.

KacperKubara:
PR in https://github.com/speechbrain/speechbrain/pull/1550

30stomercury:
Related to #1550:

I actually have addressed the issue at
https://github.com/speechbrain/speechbrain/blob/102f9f785f13f96110de02cae7a00df217d4e247/speechbrain/decoders/seq2seq.py#L733-L741
It's not been merged yet as it involves several major changes.
I also remember similar issue was raised before, I think the changes will be done in the future (refactored) version.

Adel-Moumen:
Hello, 

As mentioned by @30stomercury, a PR is already addressing this issue. 

We have some major steps to finish before merging the latter PR but It will solve your issue!

PR #1550, opened 2022-08-24
Hypotheses scores

KacperKubara:
Related to https://github.com/speechbrain/speechbrain/issues/1514. Top-k argument in the S2SBeamSearcher is described as _The number of hypotheses to return_. However, if we don't specify the return_log_probs argument, only final scores for each hypothesis are returned (without a probability path for each hypothesis). This is somewhat confusing from the point of the user. 

Hereby, I propose to change slightly S2SBeamSearcher to return either log_probs, or probs, depending whether return_log_probs is True or False.

_I am not sure if that change might be breaking some stuff as a number of returned elements is changed (please check!)_. In that case, we should make just crystal clear in the documentation that multiple probability paths will be returned only if topk > 1 and return_log_probs = True. If that's the case, let me know so I will just change the docs strings accordingly.

### DIFF
--- a/speechbrain/decoders/seq2seq.py
+++ b/speechbrain/decoders/seq2seq.py
@@ -832,7 +832,7 @@ class S2SBeamSearcher(S2SBaseSearcher):
         if self.return_log_probs:
             return predictions, topk_scores, log_probs
         else:
-            return predictions, topk_scores
+            return predictions, topk_scores, torch.exp(log_probs)
 
     def ctc_forward_step(self, x):
         """Applies a ctc step during bramsearch."""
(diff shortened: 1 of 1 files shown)


TParcollet:
Thanks !!! @KacperKubara @30stomercury could you guys take a look / review ?

30stomercury:
Please see my reply here #1514.

Adel-Moumen:
As mentioned by @30stomercury and me, we already have a PR that will solve your issue. 

I don't think that we should introduce any new modifications to the beam search algorithm before merging the new PR. (or at least new critical ones) 

Thanks for your contribution though always great to have people working to improve SpeechBrain!

Btw you can take a look at the new PR if you want! Feedbacks are open.

Issue #988, opened 2021-09-14
RuntimeError: Boolean value of Tensor with more than one value is ambiguous


srdfjy:
Run the default sepformer recipe, stand-alone 4 gpu cards, batch_size is 2, the following error occurred

--------------------------------------------------------------------

2021/09/14 10:46:23     separator.fit(
2021/09/14 10:46:23   File "train.py", line 620, in <module>
2021/09/14 10:46:23     avg_valid_loss = self.update_average(
2021/09/14 10:46:23  99%|█████████▊| 4035/4091 [16:06<00:13,  4.05it/s, train_loss=-1.32]Traceback (most recent call last):
2021/09/14 10:46:23   File "/speechbrain/speechbrain/core.py", line 1068, in fit
2021/09/14 10:46:23   File "/speechbrain/speechbrain/core.py", line 1213, in update_average
2021/09/14 10:46:23 RuntimeError: Boolean value of Tensor with more than one value is ambiguous

YUCHEN005:
Hello, I met this same issue, how do you solve it? Thank you.

srdfjy:
I didn't solve it.

poor1017:
Hi, I met the same issue, can you reopen this issue?

PR #1555, opened 2022-08-28
Fix return value of batch_evaluation for separation recipes

z-wony:
In case of 'batch_size' is not 1, evaluate_batch() returns tensor array (in separation recipes).
It occurs exception in _fit_valid() (https://github.com/speechbrain/speechbrain/issues/988)
So, this commit fixes return values as API reference of [evaluate_batch()](https://speechbrain.readthedocs.io/en/latest/API/speechbrain.core.html#speechbrain.core.Brain.evaluate_batch)

But, I expect there are a lot of same issues in other recipes.
So, I'd like to suggest changing update_average() in core.py.
If 'list of tensors' are allowed to input argument, solution is more simple.

### DIFF
--- a/recipes/BinauralWSJ0Mix/separation/train.py
+++ b/recipes/BinauralWSJ0Mix/separation/train.py
@@ -283,7 +283,7 @@ class Separation(sb.Brain):
             else:
                 self.save_audio(snt_id[0], mixture, targets, predictions)
 
-        return loss.detach()
+        return loss.mean().detach()
 
     def on_stage_end(self, stage, stage_loss, epoch):
         """Gets called at the end of a epoch."""
--- a/recipes/LibriMix/separation/train.py
+++ b/recipes/LibriMix/separation/train.py
@@ -211,7 +211,7 @@ class Separation(sb.Brain):
             else:
                 self.save_audio(snt_id[0], mixture, targets, predictions)
 
-        return loss.detach()
+        return loss.mean().detach()
 
     def on_stage_end(self, stage, stage_loss, epoch):
         """Gets called at the end of a epoch."""
--- a/recipes/WHAMandWHAMR/separation/train.py
+++ b/recipes/WHAMandWHAMR/separation/train.py
@@ -207,7 +207,7 @@ class Separation(sb.Brain):
             else:
                 self.save_audio(snt_id[0], mixture, targets, predictions)
 
-        return loss.detach()
+        return loss.mean().detach()
 
     def on_stage_end(self, stage, stage_loss, epoch):
         """Gets called at the end of a epoch."""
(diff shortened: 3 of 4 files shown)


z-wony:
My callstack is below. (with batch_size: 2)

```
speechbrain.core - Exception:
  File "train.py", line 629, in <module>
    separator.fit(
  File "/home/jwkim/github/speechbrain/speechbrain/core.py", line 1154, in fit
  File "/home/jwkim/github/speechbrain/speechbrain/core.py", line 1057, in _fit_valid
    avg_valid_loss = self.update_average(loss, avg_valid_loss)
  File "/home/jwkim/github/speechbrain/speechbrain/core.py", line 1303, in update_average
    if torch.isfinite(loss):
RuntimeError: Boolean value of Tensor with more than one value is ambiguous

```

In the core.py
```
1045     def _fit_valid(self, valid_set, epoch, enable):
1046         # Validation stage
1047         if valid_set is not None:
1048             self.on_stage_start(Stage.VALID, epoch)
1049             self.modules.eval()
1050             avg_valid_loss = 0.0
1051             with torch.no_grad():
1052                 for batch in tqdm(
1053                     valid_set, dynamic_ncols=True, disable=not enable
1054                 ):
1055                     self.step += 1
1056                     loss = self.evaluate_batch(batch, stage=Stage.VALID)
1057                     avg_valid_loss = self.update_average(loss, avg_valid_loss)

```

evaluate_batch() returns `tensor([19.2439, 24.9512], device='cuda:0')` to <b>loss</b>
So, `if torch.isfinite(loss):` in update_average() throws exception.

z-wony:
Dear @mravanelli . Could you review this?

ycemsubakan:
@z-wony Sorry for my late reply. I will take a look soon. Thanks for the PR.

z-wony:
ping? : )

ycemsubakan:
Guys, I'll take a look this week. Very sorry, I am swamped with several things.

ycemsubakan:
Alright, I just tried this branch, it seems this doesn't cause anything else to break. We can merge. 

@z-wony  Do you know when this started to break the eval loop? Because originally this wasn't causing any issues. Like do you know which commit started to cause this issue? (Just out of curiosity)

ycemsubakan (approved):

z-wony:
> Alright, I just tried this branch, it seems this doesn't cause anything else to break. We can merge.
> 
> @z-wony Do you know when this started to break the eval loop? Because originally this wasn't causing any issues. Like do you know which commit started to cause this issue? (Just out of curiosity)

@ycemsubakan Thank you for review and sorry I have no idea about your question.
But some recipes separate batch_size configuration for evaluation and set to 1 in their hparam. (distinct with train_batch_size)
Thus, issue not be raised.

mravanelli:
Cem, do you have some time to review this as well?

PR #1559, opened 2022-09-01
sync unstable-v0.6 branch

anautsch:
To move ahead with the PR 751 which is based on the `unstable-v0.6` branch. Another PR should follow that fetches this update into the depending ctc branch.


ga1aXy-pog (commented):
LGTM!

PR #1560, opened 2022-09-01
Syncing develop branch to PR 751 branch

anautsch:
Second PR to merge latest develop branch into PR 751 branch - it took a d-tour via the `unstable-v0.6` branch (keeping git tree intact and as clean as possible).


anautsch:
This is particular...
The error is: `ERROR tests/unittests/test_ctc_segmentation.py::test_CTCSegmentation - Import...`

Many files changes; this one did not.

anautsch:
Also; the errors are equivalent to the latest errors on the 751 branch; compare:
https://github.com/speechbrain/speechbrain/runs/7116413103?check_suite_focus=true

TParcollet:
We can now open PR on HuggingFace, do you think it would make sense to do so to adapt?

anautsch:
I'll mark the HF related tests to be skipped by pytest.
If we update YAML files now; all will crash. HF updates need to be at the same time, as v0.6 is released (put into develop and onto main). There is the general legacy-breaking issue we will face. Alternatively, can we make branches on the HF repos? Then, we could have a v0.5 and below branch and a 'current' one.

anautsch:
Ok, so after skipping HF related tests through two commits on the 751 branch/PR - we have the following situation:
* tests here fail
* tests there work

Compare 751 workflow:
https://github.com/speechbrain/speechbrain/runs/8132425462?check_suite_focus=true

Here, the base is the develop/unstable-v0.6 branch. In another recent PR, we observed that getting the PR branch up to speed can be a critical issue, too. As in the changed files list here, the tests that are failing were not skipped - although they are skipped in the targeted branch. Thus, I'd simply merge this PR into the 751 branch:
1. the result is not affected by the errors, since they relate to the `unstable-v0.6` (we could update there but it is our guinea pig develop branch)
2. merging this PR should neither affect the refactoring nor the tests skipping

Earlier, major syncing between the 1+ year old 751 PR and develop was made; the current updates are small changes since then (+ `requirements.txt`).

Issue #1277, opened 2022-01-26
TIMIT Recipe Missing uppercase option


picheny-nyu:
in the seq2seq recipe for timit in train.py the keyword uppercase is missing so if the data is uppercase directory format it can't read it if uppercase is in the yaml file (which is where I assume it should be). It's a trivial change to add it....

mravanelli:
Sure, we can add it (I took a note on that). Meanwhile, you can add this
option in train.py when calling timit prepare:

    run_on_main(
        prepare_timit,
        kwargs={
            "data_folder": hparams["data_folder"],
            "save_json_train": hparams["train_annotation"],
            "save_json_valid": hparams["valid_annotation"],
            "save_json_test": hparams["test_annotation"],
            "skip_prep": hparams["skip_prep"],
            *"uppercase": True*
        },
    )

Adel-Moumen:
Hello, 

The issue has been solved with #1564, I'm closing this issue as a consequence.

PR #1564, opened 2022-09-07
Fix Issue #1277 timit recipe missing uppercase option 

Adel-Moumen:
Small fix for #1277

### DIFF
--- a/recipes/TIMIT/ASR/CTC/hparams/train.yaml
+++ b/recipes/TIMIT/ASR/CTC/hparams/train.yaml
@@ -19,6 +19,7 @@ train_annotation: !ref <output_folder>/train.json
 valid_annotation: !ref <output_folder>/dev.json
 test_annotation: !ref <output_folder>/test.json
 skip_prep: False # Skip data preparation
+uppercase: False # Must be True when the TIMIT dataset is in the upper-case version
 
 # Training parameters
 number_of_epochs: 50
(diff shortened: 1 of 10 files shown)


anautsch:
lgtm - some thoughts on an adjacent issue.

This was not detected by yaml tests, although they consider data preparation scripts - why?
https://github.com/speechbrain/speechbrain/blob/424e7921531b0ea6523557ef0fd6ca249936bd26/tests/consistency/test_yaml.py#L33

While the recipe overview has the data preparation script referenced, it plays a litter role for the yaml testin.
https://github.com/speechbrain/speechbrain/blob/develop/tests/recipes.csv
https://github.com/speechbrain/speechbrain/blob/424e7921531b0ea6523557ef0fd6ca249936bd26/speechbrain/utils/check_yaml.py#L129

Should testing include that through YAML one could specify parameters of any subfunction? 
(e.g., retraversal of uses; namespaces, and if all function arguments are settable then; their default is changeable through YAML.)

Adel-Moumen:
> This was not detected by yaml tests, although they consider data preparation scripts - why?

Hmm according to the doc string of https://github.com/speechbrain/speechbrain/blob/424e7921531b0ea6523557ef0fd6ca249936bd26/speechbrain/utils/check_yaml.py#L129 this is "normal" that the tests have failed to catch this problem because the tests are only checking if a variable is unused in the python files while in our case we just forgot to declare it and use it. (which is a different problem)

Not sure to understand clearly your last paragraph.

anautsch:
> this is "normal" that the tests have failed to catch this problem because the tests are only checking if a variable is unused in the python files while in our case we just forgot to declare it and use it. (which is a different problem)

You got it; don't worry. I'd see that as a part of functionality for that testing script, whereas it is currently not. It's literally up to us to make it function/or not. Yet the existence of this PR demonstrates: it should be part of function. Question is: what's the least effort here - writing a new testing script; reusing the old one; some hybrid approach - or simply fixing this instance and waiting until it pops up elsewhere again :wink:

anautsch (approved):

PR #1536, opened 2022-08-10
bug-fix: fixed OPEN_RIR data preparation process conflict.

xin-w8023:
https://github.com/speechbrain/speechbrain/blob/598f6eda70f9b0c9ad49b393114ff483add1fd25/speechbrain/lobes/augment.py#L470
None main process might skip the following block if main process already got in and created the folder.

PR #1566, opened 2022-09-07
fix distributed namespace

anautsch:
Issue resulting from #1536 - the `utils` namespace is missing.

Running the template:
```python train.py tokenizer.yaml```

throws: `AttributeError: module 'speechbrain' has no attribute 'distributed'`

Not catched by our current unit & doc tests.


Adel-Moumen (approved):
lgtm.

PR #1567, opened 2022-09-08
feat: use member field instead of hard-code

### LINKED ISSUE #1451: Code Style Fixing
https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/metric_stats.py#:~:text=(self.-,labels,-%3D%3D%201).
image
It's better to use `self.labels == self.positive_label` than hardcode to `self.labels == 1`.

xin-w8023:
fix #1451 It's better to use member filed `self.positive_label` than hard code to 1.

### DIFF
--- a/speechbrain/utils/metric_stats.py
+++ b/speechbrain/utils/metric_stats.py
@@ -389,10 +389,10 @@ class BinaryMetricStats(MetricStats):
 
         if threshold is None:
             positive_scores = self.scores[
-                (self.labels == 1).nonzero(as_tuple=True)
+                (self.labels == self.positive_label).nonzero(as_tuple=True)
             ]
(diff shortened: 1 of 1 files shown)


xin-w8023:
@Adel-Moumen Hi this PR aims to address issue #1451

anautsch (commented):
@Adel-Moumen lgtm

The proposed change relates to lines 321-323, where `self.positive_label` is initialized - subsequently, it can also be used later on, as demonstrated in this PR.

```python
    def __init__(self, positive_label=1):
        self.clear()
        self.positive_label = positive_label
```

anautsch (approved):

Adel-Moumen:
Lgtm.

PR #1570, opened 2022-09-09
Output hiddens states from all the transformer layers of huggingface_wav2vec

BenoitWang:
Hello, this PR intends to add an option to ouput all the layers of `HuggingFaceWav2Vec2`. Some works have shown that different layers of wav2vec/HuBert/WavLM have different impact on different speech tasks. So this change may allow to study this effect.  

An attribute `output_all_hiddens` (default: `False`) is added:
 If True: outputs the hidden states from all transformer layers
 If False: outputs the hidden states only from the last transformer layer


Adel-Moumen (changes requested):
Thanks for the PR!

Please fix some minor issues. 

Note: I will refactor the hugging face interface so it might be possible that this code will change in the future.

BenoitWang:
Hi @Adel-Moumen, thanks for the review, the fix is done.

Adel-Moumen (approved):
lgtm ! Thanks for the pr. 🤗

PR #1573, opened 2022-09-10
add coloring in tqdm progress bar

sangeet2020:
I initiate a PR to change the color of tqdm progress bar during training, validation and test.

The main aim is to visually distinguish between train, valid and test during the script run phase.


anautsch (commented):
@sangeet2020 thanks for making this a smaller PR !
(and sorry for the delay on our side)

@mravanelli was asking internally if we could have the SB color of the new logo.
The tqdm allows for color codes to be passed too. With a color picker on the SB logo svg file, I got two color values `328ce7` (and a darker `4c99e8`).

As @pplantinga mentioned - customization is key here - the run opts could be useful here. 

How about having an option between setting three different colors -or- overriding all of them with one argument? (having both options available, is it feasible?)

mravanelli:
@sangeet2020, could you please do the modifications suggested by @anautsch?

sangeet2020:
Hi, 
First of all, apologies for the big delay in progressing this PR.

I agree with all the suggestions @pplantinga  and @anautsch mentioned. It's better to have an override that can suppress the colored progress bar and rather use the default color- white.

Here is the override
```python
"--white_tqdm", 
default=False, 
action="store_true",
 help="Enable default (white) color for tqdm progress-bar. If this is "
 "not passed a multi-colored progress-bar is used.",
```
and within `run_opt_defauls`, I added
```python
"white_tqdm": False,
 "tqdm_progressbar": {
    "train": "GREEN",
     "valid": "MAGENTA",
     "test": "CYAN",
},
```
so whenever, `--white_tqdm` is passed, it sets all colors to an empty string (empty string implies tqdm would use the default color )
```python
# Force default color for tqdm progrressbar
if self.white_tqdm:
    self.tqdm_progressbar = dict.fromkeys(self.tqdm_progressbar, "")
```
Let me know if this is an acceptable optimal solution. 

If you guys are okay with this, I shall further change the currently hard-coded colors to use the colors present in SpeechBrain's logo as mentioned by @anautsch (`328ce7` and a darker `4c99e8`). But I have a concern that most of us use a dark terminal, so a dark-color progress bar may not be our first choice. What do you think?

sangeet2020:
Hi @anautsch, is there any update on this one? Do you recommend any other color combination?

speechbrain/core.py
```diff
@@ -269,6 +269,13 @@ def parse_arguments(arg_list=None):
         type=int,
         help="Number of optimizer steps to run. If not passed, all epochs are run.",
     )
+    parser.add_argument(
+        "--white_tqdm",
+        default=False,
```
anautsch, line 274: True?

speechbrain/core.py
```diff
@@ -455,6 +462,12 @@ def __init__(  # noqa: C901
             "ckpt_interval_minutes": 0,
             "grad_accumulation_factor": 1,
             "optimizer_step_limit": None,
+            "white_tqdm": False,
```
anautsch, line 465: True

speechbrain/core.py
```diff
@@ -269,6 +269,13 @@ def parse_arguments(arg_list=None):
         type=int,
         help="Number of optimizer steps to run. If not passed, all epochs are run.",
     )
+    parser.add_argument(
+        "--white_tqdm",
```
anautsch, line 273: why white? not just "default"

anautsch (commented):
Hi @sangeet2020 thank you for pushing this again.

I like the idea, don't worry. For me, this is not about discussing colors. Wondering if we can figure a terminal color out to use then a matching color scheme. (Way beyond the scope of this PR.)

The default setting should be as it is now for everyone, in a way—but: when we have a new release, switching that to a SB color would also be a good touch-up. (thinking out loud)

In a way, it is funny "not discussing colors" but then also "default: white".
When people use another terminal color scheme, we have no way to control what is light/dark in terms of colors we put forth. Then, whatever their default is, it will be white or another color, as to their default preference. (please rename the variable -> not all defaults are white; maybe it's just me but I start having absurd discussions with myself here)

I spotted this PR on some dark mode issues:
`https://github.com/tqdm/tqdm/issues/842`
we will not be able to control all environments, in which people will use SpeechBrain (terminal, server console, some editor, colab, jupyter, ...) neither the type of shell (bash, fish, zsh, ...) nor their (non-linux) OS. This just makes me think, when I use dark mode while browsing, some websites are not able to handle it well at all.

In a way, having nice colors matching to the terminal settings must be a tqdm concern with a good amount of discussion done already. Is there more to it?

I don't know what an optimal solution is here, can only talk it through to understand it better.

Came across this one too
`https://github.com/MicrosoftDocs/terminal/blob/main/TerminalDocs/customize-settings/color-schemes.md`
should or can we set the tqdm background as well, if needed?

Please let me know if these questions help you or appear bogus - there's no need to make this PR bigger than it needs to be.

speechbrain/core.py
anautsch: tqdm_default_color?

if we can specify between text & bg color, maybe we need to disinguish that too

sangeet2020:
Thanks for your detailed thoughts.

- Default value `tqdm_default_color=True`. Changes made as suggested above.
- It is true that we are not capable to control all environments- be it a white terminal, jupyter, green console etc.; Tqdm adjusts the color of its progress bar automatically. Therefore, it's wiser to let tqdm use its default settings.
- > In a way, having nice colors matching to the terminal settings must be a tqdm concern with a good amount of discussion done already. Is there more to it?
  -  True, we should let tqdm handle color of its progress bar in an unknown terminal setting. I don't think we should put extra thought than this.
- I could not find any relevant documentation on setting a background color for tqdm progress bar.

In my opinion, we should make it simple- use the default colored progress bar, or use a multi-colored progress bar and if that's a problem- switch back to the default setting and let tqdm handle  what color to use in particular that environment setting.

anautsch:
> In my opinion, we should make it simple- use the default colored progress bar, or use a multi-colored progress bar and if that's a problem- switch back to the default setting and let tqdm handle what color to use in particular that environment setting.

Sounds good!

I ran a test with
```
python recipes/LibriSpeech/ASR/seq2seq/train.py recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml --data_folder=tests/samples/ASR/ --train_csv=tests/samples/annotation/ASR_train.csv --valid_csv=tests/samples/annotation/ASR_train.csv --test_csv=[tests/samples/annotation/ASR_train.csv] --number_of_epochs=10 --skip_prep=True --data_folder_rirs=tests/tmp --device="cpu" --debug --tqdm_default_color
```
and the color did not change—when I took `--tqdm_default_color` out, the color did not change either.

Do you have a minimal example that I could use to verify it works?
Maybe the issue is on my side...

When I commented above, I thought it would be possible to provide a `--tqdm_default_color=True` or `--tqdm_default_color=False` - but since this is a flag, this is not possible. I'm still indecessive on what would be better (default: we take tqdm default or default: we change the colors).
// Maybe we are asking too much with having the color dict to be also handed over as a parameter via "--args=..." style. If it is not possible to pass on a dict, a list could do as well (it's still ordered data).

sangeet2020:
Hi,
I made a few changes- change of variable names for easy understanding. (Please pull again)

Basically, not passing any arguments would enable default setting for tqdm progress bar. While if `--tqdm_colored_bar` is passed,  a colored progress bar is enabled. You could find two snap shot showing these setups.
- Default tqdm- would work as has always been; no colors
image

- colored  tqdm- one would need to pass `--tqdm_colored_bar` to enable this.
image

Let me know if you find this suitable or you wish some changes.

Thank You

anautsch (approved):
works!
THANK YOU :)

business as usual
```
python recipes/LibriSpeech/ASR/seq2seq/train.py recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml --data_folder=tests/samples/ASR/ --train_csv=tests/samples/annotation/ASR_train.csv --valid_csv=tests/samples/annotation/ASR_train.csv --test_csv=[tests/samples/annotation/ASR_train.csv] --number_of_epochs=1 --skip_prep=True --data_folder_rirs=tests/tmp --device="cpu" --debug
```

colored progress bars as suggested—there's a clear visual difference between speechbrain.STAGES (!) with `--tqdm_colored_bar` (as you described)
```
python recipes/LibriSpeech/ASR/seq2seq/train.py recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml --data_folder=tests/samples/ASR/ --train_csv=tests/samples/annotation/ASR_train.csv --valid_csv=tests/samples/annotation/ASR_train.csv --test_csv=[tests/samples/annotation/ASR_train.csv] --number_of_epochs=1 --skip_prep=True --data_folder_rirs=tests/tmp --device="cpu" --debug --tqdm_colored_bar
```

one can override cli params for different colours by `--tqdm_barcolor={"train": "BLUE", "valid": "RED", "test": "YELLOW"}`
```
python recipes/LibriSpeech/ASR/seq2seq/train.py recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml --data_folder=tests/samples/ASR/ --train_csv=tests/samples/annotation/ASR_train.csv --valid_csv=tests/samples/annotation/ASR_train.csv --test_csv=[tests/samples/annotation/ASR_train.csv] --number_of_epochs=1 --skip_prep=True --data_folder_rirs=tests/tmp --device="cpu" --debug --tqdm_colored_bar --tqdm_barcolor={"train": "BLUE", "valid": "RED", "test": "YELLOW"}
```

or put custom colors into hparams
```yaml

tqdm_barcolor:
   train: BLUE
   valid: "#328ce7"
   test: "#4c99e8"
```
=> regarding SpeechBrain colour discussion: they all look "BLUE" to me.

I played a bit with https://colorbrewer2.org/ to compare different colour schemes - the current one put forth by @sangeet2020 is close to colorblind safe schemes and as tqdm default colour names are used, I've a good feeling regarding accessibility! We can update later if necessary.

LGTM!

PR #1578, opened 2022-09-19
Fix bugs of update_learning_rate

wangxin22:
Fix bugs of update_learning_rate, in which param_group doesn't work as expected:
When the input optional param param_group is NOT None, [0] or [0,1] for example, it will throws Exception because of None groups.


anautsch (approved):
good catch - lgtm

PR #1584, opened 2022-09-26
In filter_ctc_output(), remove redundant filtering

olvb:
Hi,
I noticed that in `filter_ctc_output()` we had 2 blocks of code performing the same task: merging repeated tokens.
I kept the `groupby()` rather than the list comprehension as it appears to be faster.


anautsch (approved):
Well spotted, @olvb !
(it's in there since the first version of this file)

lgtm

PR #1592, opened 2022-10-04
read_audio fixes and docs cleanup

AsuMagic:
When using the `dict` variant of `read_audio`, omitting the `"stop"` key would fail due to a breaking change in torchaudio `0.8.0` (SB currently requires any version newer than `0.9.0`).

While this was undocumented, the code seemed to intend it as possible.

This PR does several things:

- The function was cleaned up, with the call to `torchaudio.load` fixed, keeping backwards-compatibility.
- Added sanity checks, which raise an exception when the start-stop range seems incorrect. Let me know if these are overkill.
- It attempts to improve the documentation for `read_audio` significantly, documenting edge cases and previously undocumented behavior, namely

  - Keys for the dict variant are documented properly, and it is explained which ones are optional.
  - Behavior for multi-channel files is now explained properly.
  - General cleanups and formatting improvements.

Following this PR, all of these now work:

```python
# always has worked
a = read_audio("/path/to/file.wav")

# always has worked
b = read_audio({
    "file": "/path/to/file.wav",
    "start": 8000,
    "stop": 16000
})

# was failing, now works
c = read_audio({
    "file": "/path/to/file.wav"
})

# was failing, now works
d = read_audio({
    "file": "/path/to/file.wav",
    "start": 8000
})
```

Tests were added accordingly.

## Root cause

Previously, the `num_frames` parameter defaulted to `0`, which meant "load from `frame_offset` through the end of file".  
This is not the case anymore: `num_frames` now defaults to `-1`, with the same meaning. However, passing `0` now fails.

## Implementation details

In order to best match the past intended behavior, if `start == stop`, this PR leaves the `num_frames` parameter unspecified, which I feel is more intuitive than the previous code here.

### DIFF
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -7,6 +7,7 @@ Authors
  * Ju-Chieh Chou 2020
  * Samuele Cornell 2020
  * Abdel HEBA 2020
+ * Sylvain de Langen 2022
 """
 
 import os
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -164,26 +164,41 @@ def read_audio(waveforms_obj):
     Expected use case is in conjunction with Datasets
     specified by JSON.
 
-    The custom notation:
+    The parameter may just be a path to a file:
+    `read_audio("/path/to/wav1.wav")`
 
-    The annotation can be just a path to a file:
-    "/path/to/wav1.wav"
+    Alternatively, you can specify more options in a dict, e.g.:
+    ```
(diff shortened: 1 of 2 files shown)


mravanelli:
@Adel-Moumen, could you please review it when you have time?

Adel-Moumen (changes requested):
Hello, 

Thanks for the PR. Everything looks nice. I tried your PR and everything works as expected. 
Could you please add your name at the top of the file?

Thanks again!

Adel-Moumen (approved):
LGTM!

PR #1595, opened 2022-10-05
Fix documented min python version to 3.7

### LINKED ISSUE #1593: Conflicting documentation on supported Python versions
Both 3.7 and 3.8 are mentioned as minimal Python versions in different places, which could be confusing to users and contributors. So, which one is it?

Those are all the mentions to Python versions I've found after a quick lookup in the codebase:
- `README.md` mentions Python 3.8+
- `docs/contributing.md` mentions Python 3.8+
- Test actions are run for Python 3.7, 3.8, 3.9
- `setup.py` requires Python 3.7+
- `docs/installation.md` mentions Python 3.7+

FWIW, if you want, I can make a PR to solve the issue depending on your answer.

AsuMagic:
Fixes #1593: The minimal Python version SB targets is 3.7, not 3.8.

Those were the only mentions to 3.8 being the minimal requirement I could find in the codebase. 

FWIW,

- readthedocs is configured for 3.8. I just noticed that TParcollet edited the version there from 3.9 to 3.8 a while back.
- pre-commit checks are run on the CI using 3.8.
- the PyPI publish step is run using 3.8.

Should anything be done about these?

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ SpeechBrain can be installed via PyPI. Moreover,  a local installation can be us
 
 ## Install via PyPI
 
-Once you have created your Python environment (Python 3.8+) you can simply type:
+Once you have created your Python environment (Python 3.7+) you can simply type:
 
 ```
 pip install speechbrain
@@ -178,7 +178,7 @@ import speechbrain as sb
 
 ## Install with GitHub
 
-Once you have created your Python environment (Python 3.8+) you can simply type:
+Once you have created your Python environment (Python 3.7+) you can simply type:
 
 ```
 git clone https://github.com/speechbrain/speechbrain.git
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -50,7 +50,7 @@ These will automatically check the code when you commit and when you push.
 
 ## Python
 ### Version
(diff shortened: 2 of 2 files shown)


anautsch:
Below are some comments to respond more broadly about your well brought up concerns (no idea at the moment what would be a good 'instruction').
* We run github tests on python 3.7; 3.8 & 3.9.
* readthedocs needs to be configured to one Python version. 3.8 (to me) appears as the best bridge between these versions—yet, at some point, python 3.11 release candidates might deprecate 3.7-3.10 - after a period for security updates, the whole pytorch ecology will have no chance but to migrate to the new space - end of life: 3.7 (27 Jun 2023); 3.8 (14 Oct 2024); 3.9 (05 Oct 2025)—when 3.8 turns deprecated, why even bother to adapt python 3.10 ?
* likewise, pre-commit checks need to operate on one specific version—the github checks cover 3.7 to 3.9 versions (if there'd be an issue, it comes clear there—haven't seen that happening thus far though)
* pypi step also needs to use a specific version, I'd assume—3.8 seems a reasonable choice for the time being

You have some good points—and pretty good catches !

anautsch (approved):

PR #1596, opened 2022-10-05
Refactoring of SB to  HuggingFace interface

anautsch:
@TParcollet requested a more flexible interface to use HuggingFace than only with wav2vec2. 

In another conversation, @Moumeneb1 pointed me to `AutoModel` of the HuggingFace transformers library.

---

Upfront a worthwhile side note—recently released:
* transformers 4.22.2 on Sep 27, 2022
* datasets 2.5.2 on Oct 5, 2022
* huggingface-hub 0.10.0 on Sep 28, 2022

Should we opt for a minimum HF library support ?
(edit: the hub is now at 0.7.0 in `requirements.txt` - the other two are optional; should they remain so ? )

---

What has happened at the opening of this Draft PR:
* availed handling of HuggingFace `cache_dir` throughout (so it can be put outside of `$HOME/.cache`)
  => minor impact to `speechbrain/pretrained/fetching.py`
* removed unnecessary imports and dicts (handled via AutoModel)
* placed static functions outside of a class
  * `_check_model_source(path, save_path)` // has changes to check if source is downloaded already
* created helper static function - some are used implicitly by the pre-existing two classes
  * `config_return_hidden_states(config)`
  * `model_set_spectral_augmentation(model, apply_spec_augment)`
  * `modify_state_dict_wav2vec2(path)`
  * `default_forward(model, data)`
  * `wav2vec2_forward(model, data, output_all_hiddens)`
  * `wav2vec2_pretraining_forward(model, data, mask_prob, mask_length)`
  > These functions are intended to be used as partials - use them -or- plug-in your own :)
* new class: `HuggingFaceModel(nn.Module)` to handle all interfaces with HuggingFace transformers - with init doing:
  * determine `AutoConfig` (adjust if wanted)
  * create/download model from `AutoModel` (adjust if wanted)
  * prepare forward function abstraction
    * set input layer norm flag
    * assign inner forward function from given/default partial Callable 
      <i>(e.g., default_forward; wav2vec2_forward; wav2vec2_pretraining_forward)</i>
    * set output layer norm flag
    * output of a variable -or- tuple
    > Wrapper: forward() -> _forward() -> self.forward_partial_fn(data=data)
  * handle Freezing
* `HuggingFaceWav2Vec2` inherits now from `HuggingFaceModel` and is reduced to a super().__init__ call
* same for `HuggingFaceWav2Vec2Pretrain`; different init parameterization (serves here as proof-of-concept)
* docstring examples for the three classes were working on my end

Drafting status:
* [x] initial PR (docstring examples & linters)
* [x] create integration test folder with YAML examples
* [x] whether/not pythonapp workflow integration tests should install `transformers>=4.22.2` (or: skip their integration examples)
* [x] resolve TODO comments
* [x] check on single GPU if nothing breaks & on DDP for wav2vec2 training
* [x] minimize online communication overheads (once downloaded, that's it)

---

Edit (2022-10-11).
* [x] dissolve current file & create a nested folder structure with main interface & helper functions
* [x] drop normalization functions (note: they have been migrated correctly BUT were ontologically superfluous in the starting code prior to this PR)
* [x] expedite further auto-general use features provided by HF
* [x] explore to provide further hub examples (beyond w2v2)
* [x] expand briefly the existing tutorial for how to make use of this PR

---

Edit (2022-12-13).
* [x] merge testing from #1600 
* [x] re-test HF pretrained models & apply fixes
* [x] fix failing recipes (when transformers integration of this PR is the issue)


speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+    """
+    model.config.apply_spec_augment = apply_spec_augment
+    return model
+
+
+def modify_state_dict_wav2vec2(path):
+    """Function to be used in HuggingFaceModel._load_sb_pretrained_parameters.
+
+    A custom loading ensures SpeechBrain compatibility for Pretrain and model
+    de/serialization. Here, '.wav2vec2' is removed before loading.
```
Adel-Moumen, line 123: do you think this problem can happen with other HF models? If so, could we instead have something that is more high-level? E.g the function takes a string in input that is removed from the key of the model (e.g, here it would be "model.wav2vec2."

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+    # We remove the .wav2vec2 in the state dict.
+    for key, params in orig_state_dict.items():
+        if "wav2vec2." in key:
+            save_key = key.replace("model.wav2vec2.", "")
+            modified_state_dict[save_key] = params
+
+    return modified_state_dict
+
+
+def default_forward(model, data):
```
Adel-Moumen, line 137: Do you think we should add an example to better illustrate/have an automatic test?

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+    """
+    out = model(data)
+    norm_shape = out.shape
+    return out, norm_shape
+
+
+def wav2vec2_forward(model, data, output_all_hiddens):
+    """Takes an input waveform and return its corresponding wav2vec encoding.
+
+    Arguments
```
Adel-Moumen, line 163: "output_all_hiddens" is missing here

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+def wav2vec2_pretraining_forward(model, data, mask_prob, mask_length):
+    """Takes an input waveform and return its corresponding wav2vec encoding.
+
+    Parameters
+    ----------
+    model : transformers.AutoModel
+        A valid HuggingFace transformers model.
+    data : torch.Tensor (signal)
+        A batch of audio signals to transform to features.
+    mask_prob : float (default: 0.65)
```
Adel-Moumen, line 199: no default value in the function

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+        Probability of masking a given frame. Default is taken from the paper.
+    mask_length : float (default: 10)
```
Adel-Moumen, line 201: no default value in the function

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+        out = torch.stack(list(out.hidden_states), dim=0)
+        norm_shape = out.shape[-3:]
+    else:
+        out = out.last_hidden_state
+        norm_shape = out.shape
+
+    return out, norm_shape
+
+
+def wav2vec2_pretraining_forward(model, data, mask_prob, mask_length):
```
Adel-Moumen, line 190: example is relevant here?

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
-    def _from_pretrained(self, source, config, model, save_path):
+            if freeze_nested_models_their_calls is not None:
+                if type(freeze_nested_models_their_calls) is list:
+                    freeze_nested_models_their_calls = [
+                        freeze_nested_models_their_calls
+                    ]
+                for nested_freeze_call in freeze_nested_models_their_calls:
+                    eval(f"self.model.{nested_freeze_call}()")
+
+    def _from_pretrained(
```
Adel-Moumen, line 432: documentation on args

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+                ckpt_full_path,
+                modify_state_dict_partial_fn=modify_state_dict_partial_fn,
+            )
         else:
             self.model = model.from_pretrained(source, cache_dir=save_path)
 
-    def _load_sb_pretrained_w2v2_parameters(self, path):
-        """Loads the parameter of a w2v2 model pretrained with SpeechBrain and the
-        HuggingFaceWav2Vec2Pretrain Object. It is necessary to perform a custom
+    def _load_sb_pretrained_parameters(
```
Adel-Moumen, line 466: same

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -208,101 +492,125 @@ def _load_sb_pretrained_w2v2_parameters(self, path):
...
 
-    def extract_features(self, wav):
-        """Takes an input waveform and return its corresponding wav2vec encoding.
+        return self._forward(data)
 
-        Arguments
-        ---------
-        wav : torch.Tensor (signal)
-            A batch of audio signals to transform to features.
+    def _forward(self, data):
```
Adel-Moumen, line 508: doc args

Adel-Moumen (changes requested):

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: It should be provided by the user.
Maybe that needs to come out more clearly. This is just an example for how others can tackle it in their functions. This function is tasked only with wav2vec2 - we are not engineering around HF interfaces, as these revealed a moving target.

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: it's there - please see ;-)
```
        mask_prob=0.65,
        mask_length=10,
```

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: It's not intended for use in that way. The concept works here more contextual - infiltrate with the forward outer wrapper with the right snippet to insert for modification of function behaviour. It's a guided insertion attack :yum:

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: is given already, see below

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: yep, more documentation coming. It's a different concept of how to program than what usually happens.

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: got what you mean; just copy pasted to docstring - well, then I'd remove the default here

anautsch: handling a default value there is too much; this function will get called way too often for that

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: actually, it is within a class - we have barely that detail level there on funcs in other classes...

anautsch: Compare https://github.com/speechbrain/speechbrain/blob/261cecf162766e04148cbd9c33eb7d58064f03e0/speechbrain/lobes/models/EnhanceResnet.py#L114

so, no args documentation here as well

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: same - internal function of class - no args docs

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
-    "hubert": HubertModel,
-    "wavlm": WavLMModel,
-}
 
-HF_config = {
-    "wav2vec2": Wav2Vec2Config,
-    "hubert": HubertConfig,
-    "wavlm": WavLMConfig,
-}
+def _check_model_source(path, save_path):
```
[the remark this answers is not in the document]
anautsch: done.

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+                is_sb = False
+                return is_sb, checkpoint_filename
 
-    The model can be used as a fixed feature extractor or can be finetuned. It
-    will download automatically the model from HuggingFace or use a local path.
+    err_msg = f"{path} does not contain a .bin or .ckpt checkpoint !"
+    raise FileNotFoundError(err_msg)
+
+
+def config_return_hidden_states(config):
```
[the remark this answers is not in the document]
anautsch: done.

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+    """Sets `output_hidden_states = True` for a transformer config.
+    """
+    config.output_hidden_states = True  # We want the hidden states as well!
+    return config
+
+
+def model_set_spectral_augmentation(model, apply_spec_augment):
```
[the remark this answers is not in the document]
anautsch: done.

speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -39,162 +40,445 @@
...
+    """Sets `model.config.apply_spec_augment` the flag to a specific value.
+    """
+    model.config.apply_spec_augment = apply_spec_augment
+    return model
+
+
+def modify_state_dict_wav2vec2(path):
```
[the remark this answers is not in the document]
anautsch: done.

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: done.

Testing is in the TODO list of the PR - yaml files etc - the default here migt not be called for the specific nature of each HF transformers model.

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: done. good catch !

speechbrain/lobes/models/huggingface_wav2vec.py
anautsch: done. Check out the documentation for the default forward above

speechbrain/lobes/models/huggingface_wav2vec.py
Adel-Moumen: OK. I see. 

I reread the python file and it makes sense.

anautsch:
The above pre-review was resolved entirely in commit [fb80a9d](https://github.com/speechbrain/speechbrain/pull/1596/commits/fb80a9d41a83268413d50b0889f5dc5b2aea6dc3).

Since then, the following above TODOs have been completed:
* create integration test folder with YAML examples
* resolve TODO comments (note: those in py files)
* check on single GPU if nothing breaks & on DDP for wav2vec2 training 
  --> instead: on cluster cpu node (since the rest is covered through other tests)
* minimize online communication overheads (once downloaded, that's it) 
  --> cpu node ran offline

The four test yamls target the existing interfaces:
* `speechbrain.lobes.models.huggingface_wav2vec.HuggingFaceWav2Vec2`
* `HuggingFaceWav2Vec2` immediately mapped from yaml through its new base `HuggingFaceModel`
* `speechbrain.lobes.models.huggingface_wav2vec.HuggingFaceWav2Vec2Pretrain`
* `HuggingFaceWav2Vec2Pretrain` immediately mapped from yaml through its new base `HuggingFaceModel`

Tested with
```
transformers       4.22.2
huggingface-hub    0.10.0
datasets           2.5.2
```

anautsch:
Upfront - the provided integration tests are limited in their scope;
Not: throughout tests if all possible wav2vec2 settings would work or not.
Goal: does anything break immediately, crash, or not run to the end of a "minimal" example.

A list for potential legacy testing of HF integration:
```
pip install huggingface_hub==0.7.0 datasets==2.0.0 transformers==4.18.0
2022: May 24 - Mar 15 - Apr 6

pip install huggingface_hub==0.4.0 datasets==1.18.2 transformers==4.16.0
2022: Jan 11 - Jan 28 - Jan 27

pip install huggingface_hub==0.1.0 datasets==1.15.0 transformers==4.12.3
2021: Nov 2 - Nov 2 - Nov 3

pip install huggingface_hub==0.0.1 datasets==1.1.3 transformers==4.0.0
2020: Dec 23 - Nov 19 - Nov 30
```

As for the warning messages about changes in transformers v5, I'd recommend sticking with:
`pip install huggingface_hub>=0.10.0 datasets>=2.5.2 "transformers>=4.22.2<5.0"`

@TParcollet what's your take ?

---

Note: when jumping between the HF versions, on legacy testing, lots of cache-related errors appeared—one of the logs showed:
```
The new cache file layout looks like this
```
Locking the latest versions would be another option:
`pip install huggingface_hub==0.10.0 datasets==2.5.2 transformers==4.22.2`

anautsch:
A question for the scope of this PR. Thus far, it was to make more transformers model available. In that sense, future PRs could explore more. Yet, what is more to be explored? :hugs: 
* data preparation: `datasets` is considered feature-complete
* streaming (HF-tested on w2v2 fine-tuning), see: https://huggingface.co/docs/datasets/stream

HF/datasets related pointers:
https://huggingface.co/docs/datasets/audio_load
https://huggingface.co/datasets/mozilla-foundation/common_voice_11_0
https://huggingface.co/blog?tag=audio

I'd see these features different to the nature of this PR.

Then, there is the transformer Auto~ and Pipelines.
* pipeline for inference (comparable to SB architectures); see: https://huggingface.co/docs/transformers/pipeline_tutorial
* AutoClass classes beyond `AutoModel` (currently the only one provided in this PR): `AutoTokenizer`, `AutoFeatureExtractor`, `AutoProcessor` & `AutoModel`; see: https://huggingface.co/docs/transformers/autoclass_tutorial

While we already do the former in SB (but in the SB way), the latter could have a broader impact: from which of these AutoClasses to choose from?
// perhaps an init check for alternatives to `AutoModel` would be a go-to here; structurally it appears of same use style.

anautsch:
@Adel-Moumen Just noticed that we will need to update w2v2 related YAMLs on HF when this PR gets merged. What is the best way forward in your view?

For now, I'll put back in the `speechbrain.lobes.models.huggingface_wav2vec` interfaces.
Their use could be resolved when we do the v0.6 migration (which is becoming some sort of mantra...).

anautsch:
Status: added a before/after testing for changing HF yamls of depending wav2vec2 repos. All of them use input norm & output norm. When substituting output norm with a layer norm (manually), only some of the one-short examples were still working.

anautsch:
With a bit of re-naming, in-/output norms can be substituted using this yaml snippet
```yaml
wav2vec2: !new:speechbrain.nnet.containers.LengthsCapableSequential
    w2v2_in: !new:speechbrain.processing.features.InputNormalization
        norm_type: sentence
        dim: -1
    w2v2: !ref <w2v2>
    w2v2_norm: !new:speechbrain.processing.features.InputNormalization
        norm_type: sentence
        dim: (-2, -1)
```

Until v0.6, both ways to de/activate wav2vec2 in/out norms should be available. Another PR should drop the legacy one (that PR would be to the unstable branch).

anautsch:
Added two files for testing whether/not HF interface refacotrings break anything for uploaded pretrained models
```
tests/integration/HuggingFace_transformers/refactoring_checks.py
tests/integration/HuggingFace_transformers/overrides.yaml
```
with which one can also test a before/after refactoring of HF `hyperparams.yaml` files (interface refactorings might require these to change as well). They are provided with comment detail level at the one of other script files in `tests/integration` (i.e.  none).

What can they do:
* test single files (from HF repo) as specified in another branch (yaml files not even visible to the `develop` branch)
* test with pre-prepared test partitions of specified/related recipe datasets to check with specified testers

Example of one reference "to-test/new" YAML files (as of now, the original - PR 1623 seeks into their updating):
https://github.com/speechbrain/speechbrain/blob/testing-refactoring/updates_pretrained_models/asr-wav2vec2-librispeech/hyperparams.yaml

Example of updating PR changes to `hyperparams.yaml` files: 
https://github.com/speechbrain/speechbrain/pull/1623/files

Example of test specification:
https://github.com/speechbrain/speechbrain/blob/testing-refactoring/updates_pretrained_models/asr-wav2vec2-librispeech/test.yaml

For testing some single file from its repo:
```yaml
sample: example.wav
cls: EncoderASR
fnx: transcribe_batch
```

For testing with a dataset test partition, add:
```yaml
dataset: LibriSpeech
recipe_yaml: recipes/LibriSpeech/ASR/CTC/hparams/train_hf_wav2vec.yaml
overrides:
  output_folder: !ref tests/tmp/<dataset>
dataio: from recipes.LibriSpeech.ASR.CTC.train_with_wav2vec import dataio_prepare
test_datasets: dataio_prepare(recipe_hparams)[2]
test_loader: test_dataloader_opts
performance:
  CER:
    handler: cer_computer
    field: error_rate
  WER:
    handler: error_rate_computer
    field: error_rate
predicted: predictions[0]
```
They relate to the depending recipe hparams file.

Example of testing with custom interface of pretrained models hosted on HF:
https://github.com/speechbrain/speechbrain/tree/testing-refactoring/updates_pretrained_models/asr-wav2vec2-ctc-aishell

The `tests/integration/HuggingFace_transformers/overrides.yaml` file allows to specify a glob filter:
```yaml
# Filter HF repos (will be used in a local glob dir crawling)
# glob_filter: "*wav2vec2*"
# glob_filter: "*libri*"
glob_filter: "*"
```

As for now, legacy YAMLs are supported as well as new interface. I'd like to merge this PR on `develop` and to later drop the legacy support, as further YAML interface changes will come into play (e.g. with PR 751 and alikes).

anautsch:
Tutorial update:
https://colab.research.google.com/drive/1Ggjw5sao7dFh2GSNwe7hTJFWzkGBzpJB#scrollTo=psdnsUwfsGgf

speechbrain/lobes/models/huggingface/transformers.py
```diff
@@ -0,0 +1,274 @@
...
+                cache_dir=cache_dir,
+            )
+
+        # Adjust model as desired
+        if override_hf_model_partial_fn is not None:
+            self.model = override_hf_model_partial_fn(self.model)
+
+        # Set inner forward function
+        if forward_partial_fn is not None:
+            if isinstance(forward_partial_fn, partial):
```
TParcollet, line 170: Why is this check necessary?

speechbrain/lobes/models/huggingface/transformers.py
```diff
@@ -0,0 +1,274 @@
...
+            )
+            # We transfer the parameters from the checkpoint.
+            self._load_sb_pretrained_parameters(
+                ckpt_full_path,
+                modify_state_dict_partial_fn=modify_state_dict_partial_fn,
+            )
+        else:
+            self.model = model.from_pretrained(source, cache_dir=save_path)
+
+    def _load_sb_pretrained_parameters(
```
TParcollet, line 234: Well this was something that I created specifically for wav2vec2 pertaining. I have NO IDEA if it will work for others. xD

TParcollet (changes requested):

TParcollet:
Alright, shouldn't we also update all the recipes, even if it's backward compatible? Because the example are using the new yaml notation for wav2vec2, but not the recipes. This could be confusing?

speechbrain/lobes/models/huggingface/lib_deser.py
```diff
@@ -0,0 +1,127 @@
+"""Helpers library providing de/serialization and HuggingFace <> SpeechBrain conversion mechanics.
```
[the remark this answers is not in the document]
anautsch: please make a specific suggestion for change.

speechbrain/lobes/models/huggingface/transformers.py
anautsch: either it is harmonised for all the interface interactions;
or put to another pretrained file;
or we put some handle here & put this function declaration to another lib_-style file.

wav2vec2 pretrainer - should it be handled through the wav2vec2 transformer interface - through another one - or completely different - what do you want here?

anautsch:
> Alright, shouldn't we also update all the recipes, even if it's backward compatible? Because the example are using the new yaml notation for wav2vec2, but not the recipes. This could be confusing?

Yes — recipes need updating, too.
No — there are bigger issues to get to first.

Meanwhile this yes/no situation is there
* These two files show-case legacy-preserved usability:
https://github.com/speechbrain/speechbrain/blob/d3264fcba7097b14246e01439de961a5e73a4ed9/tests/integration/HuggingFace_transformers/wav2vec2_for_pretraining.yaml
https://github.com/speechbrain/speechbrain/blob/d3264fcba7097b14246e01439de961a5e73a4ed9/tests/integration/HuggingFace_transformers/wav2vec2_from_pretrained.yaml

* and these two demonstrate — for early adapters — the new interface that will be breaking the legacy at v0.6:
https://github.com/speechbrain/speechbrain/blob/d3264fcba7097b14246e01439de961a5e73a4ed9/tests/integration/HuggingFace_transformers/wav2vec2_for_pretraining_immediate.yaml
https://github.com/speechbrain/speechbrain/blob/d3264fcba7097b14246e01439de961a5e73a4ed9/tests/integration/HuggingFace_transformers/wav2vec2_from_pretrained_immediate.yaml

Why do I provide both?
> => The current SB testing is `non-functional` to this type of refactoring. 

SB is a merger of YAML and python - on git & on HF - at the core modules; in the recipes; in the pretrained models - the SB ecosystem was not set up for this type of refactoring PR to be easy.

---

First, SB needs to be made ready to have tools that provide testing for continuity of purpose & consistency, especially for multi-domain refactoring operations, like this PR.

Second, SB needs to transition through the refactoring phase. For this, either it is one PR (business as usual) which does not work that well as seen with PR #751 already since 15+ months. OR - we transition in a two phases: I. legacy-preserving & II. legacy-breaking. Here, the attempt is to roll out the latter strategy for two reasons:
* improve testing while at it (see below)
* do the major changes, so the legacy-breaking will hurt less, in a follow-up PR 
  > // that then will be painful in combination with all PRs that are legacy-breaking for v0.6

---

Why not be legacy-preserving in the first place, when you can? [Type I error]
> => HF changed their interfaces and avail more services, especially through the AutoClass - to open up SB services, one approach is to provide a general interface to HF with SB-compliance. The implication is simple: what is old in SB will be put into comparison with many more new functions of alike types, such that sticking on one particular interface to one particular transformer, the wav2vec2, becomes of a conservative artifact if we "conserve" it for too long. 

Why not ado with it at once then, either? [Type II error]
> => As mentioned before, SB is a multi-domain, distributed & multi-language ecosystem. Automated testing is limited to:
> A. Changes to function body &/or interface invocation via YAML
> but not:
> B. Changes to function signature (interface), legacy-preserving
> C. Changes to function signature (interface), legacy-breaking

Details are explained here (the file is under development):
https://github.com/speechbrain/speechbrain/blob/127e509e93f26cb38cebda365f0986b40385e724/tests/consistency/COVERAGE.md

This PR is also concerned with providing tools for B.
Please take a look at the most important file of this PR:
https://github.com/speechbrain/speechbrain/blob/d3264fcba7097b14246e01439de961a5e73a4ed9/tests/integration/HuggingFace_transformers/refactoring_checks.py

It might need better placement, as it is the basis to check refactorings for YAML files that are hosted on HF to have the pretrained models there to be functional with the changes of this PR. It tests the changes put forth in PR #1623 which is concerned with the updating of the HF yaml files for pre-trained models only.

> Updating recipe yamls can be done here, or later.

PR #1600 needs to be wrapped-up first, so that we then can check the recipes to be functional after their refactoring. At the moment I have this knowledge only because of the `tests/integration/HuggingFace_transformers/refactoring_checks.py` just mentioned above, and this one works only for the HF repos - not for the recipes. (That's why 1600 is necessary.)

I can update the recipe yamls here after:
* other fixes we need to do to this interface (not redoing that too often)
* the recipe testing ++ PR is finished, otherwise reviewers have belief but no tools

---

As testing is present at the moment:
> recipe YAMLs will change; tests will say: you used all variables & all objects instantiate, bravo

What current testing will not say: 
1. data flow is not broken and the overall pipeline still works
2. the refactored recipe/pipeline gives the same result

Point 2. is a given when the other two points before are consolidated. 

This & the 1600 PRs add testing tools that are needed to ensure point 1. 

This PR can serve only as an intermediate refactoring, otherwise one does not have control of consistency to the last develop version before/after. Legacy-breaking is one; continuity of consistency breaking is another.

As such, the updating of recipe YAMLs is a choice of When - in this PR on in its follow-up PR.

speechbrain/lobes/models/huggingface/transformers.py
anautsch: good catch!

speechbrain/lobes/models/huggingface/lib_deser.py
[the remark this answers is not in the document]
anautsch: getting back to this. I re-organise the file structure, so there is `overrides.py`; `forward.py` and `transformer.py` - open for alternative name suggestions. will upload when we have somewhat a clue on what to go for

speechbrain/lobes/models/huggingface/transformers.py
anautsch: also getting back here, just spotted the reason more clearly.

the outer part
```python
if forward_partial_fn is not None:
   ...
else:
    self.forward_partial_fn = partial(default, model=self.model)
```
is there to put a default forward.

The inner part
```python
  if isinstance(forward_partial_fn, partial):
      self.forward_partial_fn = forward_partial_fn
      self.forward_partial_fn.keywords["model"] = self.model
  else:
      self.forward_partial_fn = partial(
          forward_partial_fn, model=self.model
      )
```
sets the model as a parameter. If the function is a partial, then the model needs to be added to the keywords dict, so when the partial is called later on, this parameter is set correctly. If `forward_partial_fn` is not a partial (but a function), a partial is created.

I put two comments in (will come with the next push).

speechbrain/lobes/models/huggingface/transformers.py
anautsch: How about making the entire `_from_pretrained` to another partial and the if/else construct could be two different "default" functions that are placed in the `overrides.py`?

Would prefer to have a generic interface with some "default" specifications/overrides, so that users have full flexibility here to use HF as they need and as HF progresses through their next versions.

anautsch:
> Alright, shouldn't we also update all the recipes, even if it's backward compatible? Because the example are using the new yaml notation for wav2vec2, but not the recipes. This could be confusing?

My main concern is the testing tool that this PR provides. In a way, that tool could be "copied" to the 1600 PR. Then, this PR could go directly to unstable and I'd start refactoring the recipes also. Didn't see that option 2-3 weeks ago.

anautsch:
As a reminder for next year.

One could retrain all transformer recipes, since there is a correct feature norm in-place now. One could also refer to the SpeechBrain git commit ID under which a model was trained and re-run performance checks on the testing sets with existing pretrained models (otherwise performance reporting wouldn't be there in the first place). That way, one spares the hassle while having the latest results. Rightfully, one might argue that it is not a 1:1 match between training & testing, which is true. Yet, the added feature norm concerns the final part of an utterance that is padding. As such, we are rather removing artefacts during scoring than we are treating major junks of speech in an unbalanced fashion compared to model training. This change on feature norm concerns the padding which was unaccounted for before, such that now, we expect improvements (generally speaking).

---

Should we run tests of the first recipe training stages if close/exact same results are yielded?
Due to the nature of DNN training, this option might be unavailable, though (randomness is a key element).

---

A future PR might look into training with one epoch & testing then (one-shot) for improvements to zero-shot (testing with pretrained model as-is).

Merge procedure:
1. PR #1600 -> develop
	0. Read https://github.com/anautsch/speechbrain/blob/refactor-recipe-testing/tests/coverage/README.md
	1. test all recipes are working
	2. use tests/utils/recipe_tests.py to gather results before refactoring (single-audio interface results)
2. Read PR #1623 -> testing-refactoring
	1. should the branch be pruned from any non-relevant files?
	2. some changes are about the nature of testing, should they be pushed directly before PR merger?
	3. develop an idea of the changes to come for the HF repos (performance results on model card & interface + github READMEs)
3. This PR #1596 -> unstable-v0.6
	1. push latest develop to unstable-v0.6 (this mimics stopping pushes to develop)
	2. run HF tests for (a) single-audio interface & (b) entire test partitions
	3. check working tests -> are file changes making sense?
	4. for failing tests -> are there obvious flaws in the changed files?
4. retrain 1x wav2vec2 or whisper model to see whether/not this is going as expected (one-shot)

---

Please keep in mind that here, commits to PR 1600 are included as well (68 commits w/ +3,965 -923 changes). The shown 115 commits with +6,403 -1,761 changes are not as "dramatic". I merged in a recent develop version to unstable-v0.6 and rebased this PR twice so the commits are managed properly by git.

I prepared a tutorial update - but there was another update on whisper, so I'll tackle it next year.

All changes to existing yamls are conceptually aligned to this way:
```
[ wav -> w2v2_input_norm -> w2v2 -> w2v2_output_norm -> Dense ] = encoder
         \--------------- wav2vec2 ----------------/
```
What was `wav2vec2` before is now a `w2v2` hparam. Benefit: a correct input feature norm is used.

anautsch:
Intermezzo notes:
* `115 commits with +6,403 -1,761 changes` is now `49 commits with +2,498 -972 changes`
* PR #1623 is now #1801 (cleaned-up before 1600/testing refactoring was merged)
* meanwhile, whisper came in & changed
* meanwhile, w2v2 was hotfixed
* meanwhile, the tutorial changed

Wrapping this one up over the next few weeks.

anautsch:
To follow-up on the merge procedure described in:
https://github.com/speechbrain/speechbrain/pull/1596#issuecomment-1354727118

For how to use the advanced test tools, when interface refactoring touches upon pretrained model interfaces (e.g. YAML files on HuggingFace), please read:
https://github.com/speechbrain/speechbrain/blob/develop/tests/utils/README.md

To-date, the following SpeechBrain branches & PRs are related:

| Branch | Purpose | PR |
| ---  | ---  | ---  |
| develop | v0.5.14 | to-come v0.6 |
| [unstable-v0.6](https://github.com/speechbrain/speechbrain/tree/unstable-v0.6) | v0.6.0 | #1596 flexible transformer integration |
| [hf-interface-testing](https://github.com/speechbrain/speechbrain/tree/hf-interface-testing) | keeping track of HF interfaces (YAMLs & custom.py) | #1868 adds two recent HF repos (original interfaces) |
| hf-interface-testing | PR 1596 changes interfaces, so YAMLs on HF repos need to change | #1801 supplements 1596 (after 1868) |
| unstable-v0.6 | v0.6.0 | #751 CTC decoding & scoring refactorings |

* PR 1868 should be a simple comparison with what's currently on HF
* PR 1801 can be used for tutorials to show how YAMLs need to be updated; here, we keep track of our changes
* PR 1596 is the transformers lib refactoring (this PR), to make use of their AutoConfig, AutoTokenizer, AutoModel, etc.
* PR 751 the legendary v0.6 development - this PR will also need supplementing edits to YAMLs & pretrained interfaces on HF, in a similar tracking fashion as demonstrated with this PR for refactoring the transformers integration

The then "to-come v0.6" will be progressively enhanced on the `unstable-v0.6` branch. When the moment of its merging comes, all YAMLs & interfaces that are on  `hf-interface-testing` are ready to be put on HuggingFace. That in-parallel update procedure implies that each HF repo will have a PR which updates YAML & custom.py interface files with what we kept track of in the `hf-interface-testing` branch.

Before merging this PR on `unstable-v0.6`, please:
1. merge the latest `develop` on the `unstable-v0.6` branch
2. rebase this PR away from & back to the `unstable-v0.6` branch => git tree needs that for updating...

---

Here's how I went for testing this PR.

Preparation steps:
1. Clone & checkout https://github.com/anautsch/speechbrain/tree/hf-integration
2. Please create a fresh Python 3.9 environment (some testinig tools require >= 3.9)
3. Install SpeechBrain from local repo + requirements + `find recipes | grep extra | xargs cat | sort -u | grep -v \# | xargs -I {} pip install {}`

Next => Test if the refactored recipes (SpeechBrain repo only) still work (here, whisper & wav2vec2 only).

4. Run recipe tests by Hparam_file
```shell
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Hparam_file"], filters=[["recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml", "recipes/LibriSpeech/ASR/CTC/hparams/train_hf_whisper_encoder.yaml", "recipes/TIMIT/ASR/transducer/hparams/train_wav2vec.yaml", "recipes/TIMIT/ASR/seq2seq/hparams/train_with_wav2vec2.yaml", "recipes/SLURP/direct/hparams/train_with_wav2vec2.yaml", "recipes/IEMOCAP/emotion_recognition/hparams/train_with_wav2vec2.yaml", "recipes/LibriSpeech/ASR/CTC/hparams/train_sb_wav2vec.yaml", "recipes/LibriSpeech/ASR/CTC/hparams/train_hf_wav2vec.yaml", "recipes/LibriSpeech/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml", "recipes/Switchboard/ASR/CTC/hparams/train_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_fon_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_amh_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_sw_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_dar_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_wol_with_wav2vec.yaml", "recipes/DVoice/ASR/CTC/hparams/train_multi_with_wav2vec.yaml", "recipes/CommonVoice/ASR/seq2seq/hparams/train_en_with_wav2vec.yaml", "recipes/CommonVoice/ASR/seq2seq/hparams/train_fr_with_wav2vec.yaml", "recipes/CommonVoice/ASR/seq2seq/hparams/train_rw_with_wav2vec.yaml", "recipes/CommonVoice/ASR/seq2seq/hparams/train_it_with_wav2vec.yaml", "recipes/CommonVoice/ASR/CTC/hparams/train_en_with_wav2vec.yaml", "recipes/CommonVoice/ASR/CTC/hparams/train_fr_with_wav2vec.yaml", "recipes/CommonVoice/ASR/CTC/hparams/train_de_with_wav2vec.yaml", "recipes/CommonVoice/ASR/CTC/hparams/train_rw_with_wav2vec.yaml", "recipes/CommonVoice/ASR/CTC/hparams/train_it_with_wav2vec.yaml", "recipes/CommonVoice/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml", "recipes/AISHELL-1/ASR/transformer/hparams/train_ASR_transformer_with_wav2vect.yaml", "recipes/AISHELL-1/ASR/CTC/hparams/train_with_wav2vec.yaml", "recipes/timers-and-such/direct/hparams/train_with_wav2vec2.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_ar_hf_whisper.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_mn_hf_whisper.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_hi_hf_whisper.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_sr_hf_whisper.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_fa_hf_whisper.yaml", "recipes/CommonVoice/ASR/transformer/hparams/train_fr_hf_whisper.yaml"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```

_Note: `cat tests/recipes/*.csv | cut -d ',' -f2 | sort -u` will list you the currently recorded recipe datasets available for testing._

5. Prepare testing of YAMLs + custom.py (the to-come HF repo updates/PRs).
```shell
python -c 'from tests.utils.refactoring_checks import init;init(new_interfaces_git="https://github.com/anautsch/speechbrain", new_interfaces_branch="hf-integration")'
```

This clones the specified branch (it's PR 1801) to the nested `tests/tmp/hf_interfaces` folder. There, the `tests/utils/refactoring_checks.py` tool will be able to access updated YAMLs & custom.py files. The same tool gets to-date versions of these files directly from HuggingFace. To compare before and after the refactoring, we take advantage from the local repo installation of SpeechBrain: we can switch between `develop` & PR branches.

6. Run single-file tests for pretrained interfaces
```shell
# Let's revisit the old way to integrate transformers into SpeechBrain
git checkout develop
python -c "from tests.utils.refactoring_checks import gather_expected_results;gather_expected_results()"

# so we can compare it with the proposed way to integrate all of the latest features from the transformers library
git checkout hf-integration
python -c "from tests.utils.refactoring_checks import gather_refactoring_results;gather_refactoring_results()"
```

_Note: A yaml summary file will be created at `tests/tmp/refactoring_results.yaml`._

For the following, please ensure that test partitions of depending recipe datasets are available. As for this example, we assume there's access only available to LibriSpeech. As such, path specifications for other recipe datasets are empty, thus depending pretrained models cannot be tested. The following step aims to reproduce performance metrics using test partitions, as they are reported on the SpeechBrain recipe folders. Therefore, however, pretrained models are used.

7. Prepare your datasets using the recipe, but point their output to the testing structure. Example: LibriSpeech (run on cpu-only; more CPUs, less waiting)
```shell
cd recipes/LibriSpeech/ASR/CTC && python train_with_wav2vec.py hparams/train_hf_wav2vec.yaml --data_folder=/path/to/dataset --output_folder=../../../../tests/tmp/LibriSpeech || cd -
```
Manually cancel after the data preparation finished (when recipe training starts). Repeat for other datasets (each one recipe); check on step (9) for using expected folder names.

8. To avoid that the `recipes` folder cannot be found as a module, we create a symbolic link. (try w/o, you'll see)
```shell
cd tests/utils && ln -s ../../recipes  && cd -
```

9. Run tests with pretrained models on the test partitions of recipe datasets
```shell
git checkout develop
python tests/utils/refactoring_checks.py tests/utils/overrides.yaml --LibriSpeech_data="/path/to/dataset" --CommonVoice_EN_data="" --CommonVoice_FR_data="" --IEMOCAP_data="" --after=False

git checkout hf-integration
python tests/utils/refactoring_checks.py tests/utils/overrides.yaml --LibriSpeech_data="/path/to/dataset" --CommonVoice_EN_data="" --CommonVoice_FR_data="" --IEMOCAP_data="" --after=True
```

_Note: Other refactorings might have expected changes in their testing performance; then, this tool can be used to measure those changes as well._

---

Logs from (4)

```
(1/35) Running test for TIMIT_row_4...
	... 582.89s
(2/35) Running test for TIMIT_row_18...
	=> skipped; took too long – i.e. restart w/o the two TIMIT yamls

(1/33) Running test for LibriSpeech_row_2...
	... 167.02s
(2/33) Running test for LibriSpeech_row_3...
	... 44.51s
(3/33) Running test for LibriSpeech_row_23...
	... 15.72s
(4/33) Running test for LibriSpeech_row_24...
	... 28.28s
(5/33) Running test for LibriSpeech_row_25...
	... 23.78s
(6/33) Running test for DVoice_row_2...
	... 122.35s
(7/33) Running test for DVoice_row_3...
	... 110.42s
(8/33) Running test for DVoice_row_4...
	... 92.57s
(9/33) Running test for DVoice_row_5...
	... 91.19s
(10/33) Running test for DVoice_row_6...
	... 91.76s
(11/33) Running test for DVoice_row_7...
	... 88.60s
(12/33) Running test for AISHELL-1_row_2...
	... 177.38s
(13/33) Running test for AISHELL-1_row_5...
	... 99.93s
(14/33) Running test for timers-and-such_row_6...
	... 67.50s
(15/33) Running test for CommonVoice_row_2...
	... 93.06s
(16/33) Running test for CommonVoice_row_3...
	... 70.77s
(17/33) Running test for CommonVoice_row_4...
	... 70.30s
(18/33) Running test for CommonVoice_row_5...
	... 73.74s
(19/33) Running test for CommonVoice_row_6...
	... 74.12s
(20/33) Running test for CommonVoice_row_12...
	... 104.15s
(21/33) Running test for CommonVoice_row_13...
	... 123.71s
(22/33) Running test for CommonVoice_row_14...
	... 183.87s
(23/33) Running test for CommonVoice_row_15...
	... 144.55s
(24/33) Running test for CommonVoice_row_18...
	... 54.74s
(25/33) Running test for CommonVoice_row_19...
	... 34.37s
(26/33) Running test for CommonVoice_row_20...
	... 33.29s
(27/33) Running test for CommonVoice_row_21...
	... 34.68s
(28/33) Running test for CommonVoice_row_22...
	... 31.71s
(29/33) Running test for CommonVoice_row_23...
	... 40.47s
(30/33) Running test for CommonVoice_row_24...
	... 26.99s
(31/33) Running test for Switchboard_row_2...
	... 33.13s
(32/33) Running test for SLURP_row_4...
	... 25.21s
(33/33) Running test for IEMOCAP_row_2...
	... 20.28s
```

=> Ok, so the refactored recipes are not crashing.

Logs from (6)
```
$ grep same tests/tmp/refactoring_results.yaml
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
  same: true
```

_Note: emotion-recognition-wav2vec2-IEMOCAP entries need manual clean-up._

=> Ok, so for single audios, the refactoring does no harm on inference when using pretrained models.

Logs from (9)
```
Run tests on: asr-wav2vec2-librispeech
speechbrain.core - Beginning experiment!
speechbrain.core - Experiment folder: tests/tmp/LibriSpeech
	repo: asr-wav2vec2-librispeech
	speechbrain.pretrained.EncoderASR
	obj.from_hparams({'source': 'speechbrain/asr-wav2vec2-librispeech', 'savedir': 'pretrained_models/asr-wav2vec2-librispeech', 'run_opts': {'debug': False, 'debug_batches': 2, 'debug_epochs': 2, 'debug_persistently': False, 'device': 'cuda:0', 'data_parallel_backend': False, 'distributed_launch': False, 'distributed_backend': 'nccl', 'find_unused_parameters': False, 'tqdm_colored_bar': False}})
speechbrain.pretrained.fetching - Fetch hyperparams.yaml: Delegating to Huggingface hub, source speechbrain/asr-wav2vec2-librispeech.
speechbrain.pretrained.fetching - Fetch custom.py: Delegating to Huggingface hub, source speechbrain/asr-wav2vec2-librispeech.
Some weights of the model checkpoint at facebook/wav2vec2-large-960h-lv60-self were not used when initializing Wav2Vec2Model: ['lm_head.weight', 'lm_head.bias']
- This IS expected if you are initializing Wav2Vec2Model from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing Wav2Vec2Model from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of Wav2Vec2Model were not initialized from the model checkpoint at facebook/wav2vec2-large-960h-lv60-self and are newly initialized: ['wav2vec2.masked_spec_embed']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
speechbrain.lobes.models.huggingface_wav2vec - speechbrain.lobes.models.huggingface_wav2vec - wav2vec 2.0 is frozen.
speechbrain.pretrained.fetching - Fetch wav2vec2.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/wav2vec2.ckpt.
speechbrain.pretrained.fetching - Fetch asr.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/asr.ckpt.
speechbrain.pretrained.fetching - Fetch tokenizer.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/tokenizer.ckpt.
speechbrain.utils.parameter_transfer - Loading pretrained files for: wav2vec2, asr, tokenizer
Some weights of the model checkpoint at facebook/wav2vec2-large-960h-lv60-self were not used when initializing Wav2Vec2Model: ['lm_head.weight', 'lm_head.bias']
- This IS expected if you are initializing Wav2Vec2Model from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing Wav2Vec2Model from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of Wav2Vec2Model were not initialized from the model checkpoint at facebook/wav2vec2-large-960h-lv60-self and are newly initialized: ['wav2vec2.masked_spec_embed']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
speechbrain.lobes.models.huggingface_wav2vec - speechbrain.lobes.models.huggingface_wav2vec - wav2vec 2.0 is frozen.
speechbrain.dataio.encoder - Load called, but CTCTextEncoder is not empty. Loaded data will overwrite everything. This is normal if there is e.g. an unk label defined at init.
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 328/328 [01:26<00:00,  3.77it/s]
speechbrain.utils.train_logger - [LibriSpeech] - BEFORE: asr-wav2vec2-librispeech, set: test-clean - test CER: 5.00e-01, test WER: 1.90
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 368/368 [01:22<00:00,  4.46it/s]
speechbrain.utils.train_logger - [LibriSpeech] - BEFORE: asr-wav2vec2-librispeech, set: test-other - test CER: 8.83e-01, test WER: 2.95
```

We can compare this with:
https://github.com/speechbrain/speechbrain/tree/develop/recipes/LibriSpeech/ASR/CTC
where for train_hf_wav2vec.yaml a Test Clean WER of 1.90 is reported.

The log for the refactored interfaces:
```
Checking out files: 100% (122/122), done.
Switched to branch 'hf-integration'
Your branch is up to date with 'origin/hf-integration'.

Run tests on: asr-wav2vec2-librispeech
speechbrain.core - Beginning experiment!
speechbrain.core - Experiment folder: tests/tmp/LibriSpeech
	repo: asr-wav2vec2-librispeech
	speechbrain.pretrained.EncoderASR
	obj.from_hparams({'source': 'tests/tmp/hf_interfaces/updates_pretrained_models/asr-wav2vec2-librispeech', 'savedir': 'pretrained_models/asr-wav2vec2-librispeech', 'run_opts': {'debug': False, 'debug_batches': 2, 'debug_epochs': 2, 'debug_persistently': False, 'device': 'cuda:0', 'data_parallel_backend': False, 'distributed_launch': False, 'distributed_backend': 'nccl', 'find_unused_parameters': False, 'tqdm_colored_bar': False}})
speechbrain.pretrained.fetching - Fetch hyperparams.yaml: Linking to local file in tests/tmp/hf_interfaces/updates_pretrained_models/asr-wav2vec2-librispeech/hyperparams.yaml.
speechbrain.pretrained.fetching - Fetch custom.py: Linking to local file in tests/tmp/hf_interfaces/updates_pretrained_models/asr-wav2vec2-librispeech/custom.py.
Some weights of the model checkpoint at facebook/wav2vec2-large-960h-lv60-self were not used when initializing Wav2Vec2Model: ['lm_head.bias', 'lm_head.weight']
- This IS expected if you are initializing Wav2Vec2Model from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing Wav2Vec2Model from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of Wav2Vec2Model were not initialized from the model checkpoint at facebook/wav2vec2-large-960h-lv60-self and are newly initialized: ['wav2vec2.masked_spec_embed']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
speechbrain.lobes.models.transformer.HuggingFace - speechbrain.lobes.models.HuggingFaceTransformer is frozen.
speechbrain.pretrained.fetching - Fetch wav2vec2.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/wav2vec2.ckpt.
speechbrain.pretrained.fetching - Fetch asr.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/asr.ckpt.
speechbrain.pretrained.fetching - Fetch tokenizer.ckpt: Using existing file/symlink in pretrained_models/asr-wav2vec2-librispeech/tokenizer.ckpt.
speechbrain.utils.parameter_transfer - Loading pretrained files for: wav2vec2, asr, tokenizer
Some weights of the model checkpoint at facebook/wav2vec2-large-960h-lv60-self were not used when initializing Wav2Vec2Model: ['lm_head.bias', 'lm_head.weight']
- This IS expected if you are initializing Wav2Vec2Model from the checkpoint of a model trained on another task or with another architecture (e.g. initializing a BertForSequenceClassification model from a BertForPreTraining model).
- This IS NOT expected if you are initializing Wav2Vec2Model from the checkpoint of a model that you expect to be exactly identical (initializing a BertForSequenceClassification model from a BertForSequenceClassification model).
Some weights of Wav2Vec2Model were not initialized from the model checkpoint at facebook/wav2vec2-large-960h-lv60-self and are newly initialized: ['wav2vec2.masked_spec_embed']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
speechbrain.lobes.models.transformer.HuggingFace - speechbrain.lobes.models.HuggingFaceTransformer is frozen.
speechbrain.dataio.encoder - Load called, but CTCTextEncoder is not empty. Loaded data will overwrite everything. This is normal if there is e.g. an unk label defined at init.
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 328/328 [01:30<00:00,  3.61it/s]
speechbrain.utils.train_logger - [LibriSpeech] - AFTER: asr-wav2vec2-librispeech, set: test-clean - test CER: 5.27e-01, test WER: 2.04
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 368/368 [01:25<00:00,  4.30it/s]
speechbrain.utils.train_logger - [LibriSpeech] - AFTER: asr-wav2vec2-librispeech, set: test-other - test CER: 9.21e-01, test WER: 3.15
	before: {'test-clean': {'CER': 0.50, 'WER': 1.90}, 'test-other': {'CER': 0.88, 'WER': 2.95}}
	 after: {'test-clean': {'CER': 0.53, 'WER': 2.04}, 'test-other': {'CER': 0.92, 'WER': 3.15}}
	  same: False
```

=> Well, there's more going on ;-)

It's in the range & with the 751 PR, more refacotrings are coming in—also, we had a hotfixing of wav2vec2 to the end of 2022 & in early 2023, some more edits to whisper (without much performance re-checking). As of internal discussions, at some point the retraining of SpeechBrain models will be necessary.

pplantinga:
Still getting a handle on this PR, but one quick question from the start: many of the changes seem to be related to removing `device` from the arguments to the checkpointer's `.load()` function. Is this a necessary part of this PR or would it make sense to put in a different PR? I would just like to hear the motivation for the change itself and for putting it in this PR, and I can't find a discussion of this point in the comments above.

Adel-Moumen:
Hey @pplantinga, the `device` argument as already been removed in `unstable-v6.0` (see: https://github.com/speechbrain/speechbrain/blob/unstable-v0.6/recipes/LibriSpeech/ASR/seq2seq/train.py#L370) thanks to @lucadellalib's PR (see: https://github.com/speechbrain/speechbrain/pull/1743). 

I think you should not pay attention to this. The PR is not sync anymore to the latest commits in `unstable-v6.0`.

TParcollet:
@mravanelli @Adel-Moumen @pplantinga I have an idea to simplify this PR. What about we take only the 'HuggingFaceTransformer(nn.Module):' class and make it an abstract class. THe idea is to force the user to override a few mandatory functions, like the forward, and the freeze params functions. THis would be MUCH simpler to grasp. There is no way we release a SB v1.0 with huggingface_wav2vec.py in the lobes while it's not even wav2vec2 anymore ...

mravanelli:
I agree on the need for a simplification. Could you show us a code snippet
to figure out how the code can look like with this proposal?

PR #1602, opened 2022-10-13
Parallelize CommonVoice CSV building using PQDM

qanastek:
Hi @TParcollet,

Using the CommonVoice recipes on massive amount of data required a lot of time to build the CSV using `common_voice_prepare.py` (more than 15 minutes in my case and some of my colleagues). Thus, I decide to make it more efficient by parallelizing it using a "thread pool" for Python called [PQDM](https://pqdm.readthedocs.io/en/latest/usage.html). It's generally made the time of building the CSV twice as fast with a busy file system.

I have tested it with the French version of CommonVoice, and it's working perfectly.

Regards,

Yanis LABRAK

### DIFF
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -14,7 +14,9 @@ import re
 import logging
 import torchaudio
 import unicodedata
-from tqdm.contrib import tzip
+
+import multiprocessing
+from pqdm.processes import pqdm
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


Adel-Moumen:
Hello @qanastek, 

Sorry for the delay I didn't know that I was the reviewer of this PR. 

First of all, thank you for contributing to SpeechBrain. 

We have already exchanged on this topic at the lab, but I have some concerns with your PR. First, even though it's faster, it adds another dependency and I'd prefer to have @TParcollet, @mravanelli and @anautsch thoughts on this first (i.e, do we want to support multithreading for data preparation recipes? If so, is `pqdm` the library that we want to use?). 

I do think that this PR could be a new template for future recipes (especially the GigaSpeech which would require the use of multithreading). 

So before reviewing your PR, I'd like to wait for the opinions @TParcollet, @mravanelli and @anautsch thoughts first. 

Thanks again for your PR!

AsuMagic:
> We have already exchanged about this topic at the lab, and you know that I have some concerns with your PR. Even though it's faster, it adds another dependency and I'd prefer to have TParcollet, mravanelli and anautsch thoughts on this first (i.e, do we want to support multithreading for data preparation recipes? If so, is pqdm the library that we want to use?).

@Adel-Moumen should that discussion be moved to its own issue then?

It's an interesting issue. Multithreaded/asynchronous file decoding should be feasible without any external dependencies.

What `process_line` does strikes to me as something that could be done more seamlessly using the data pipeline. I'm not very familiar with it yet, but I am under the impression that recipes rarely make use of data pipeline facilities when it comes to importing datasets, and instead only uses them for computing model-specific stuff.  
Yet the pipeline strikes to me as the best place to shove asynchronous/parallel processing in. Is the data pipeline able to split up work across workers? Is there something else I'm missing?

AsuMagic:
Anecdotally, I made some benchmark experiments based on this step. Due to #1925, I made tests querying the length by loading the whole audio (without normalization) and getting the final tensor length. This is suboptimal, but I don't think there is a better approach with `torchaudio` for now.

My implementation used executors with their `.map` method. I made some interesting findings:

- `ProcessPoolExecutor` generally resulted in better performance and CPU occupancy over `ThreadPoolExecutor`
- `executor.map` with a block size (8 seems to work fine) generally works well, and seems to start up and use much less memory than building a list of futures with `executor.submit` to await on with `as_completed` (to double check)
- `executor.map` works fine with `tqdm` (there's some latency due to chunking, but it's fine)
- There is a benefit to upping process count well beyond the thread count. I found that both in the NFS and local disk case on 16 threads systems, upping the process count to 32 even up to 128 generally yielded benefits with no regressions.

Regarding the last point: I am not sure why that is the case. I did expect spawning more processes than there are threads to benefit performance due to effectively acting as a poor man's asynchronous file IO, possibly batching and reducing seek times for reads, but I didn't expect it to scale that much considering CPU usage was very high at 16 threads already. It would be interesting to figure out the exact reason why that is.

There's also the concern that flooding the filesystem with random reads is maybe sometimes not desirable, so this certainly merits further inspection and probably providing settings. I wonder how disk priorities behave with remote mounts like NFS. I wouldn't want to make sysadmins too unhappy :)

Adel-Moumen:
#1944 merged. No need this PR anymore... Thanks again @qanastek for your time though :-)

PR #1642, opened 2022-10-28
Added german_cleaners

padmalcom:
I added a german cleaner to avoid lowercasing texts.

### DIFF
--- a/speechbrain/utils/text_to_sequence.py
+++ b/speechbrain/utils/text_to_sequence.py
@@ -211,7 +211,7 @@ def basic_cleaners(text):
     text = lowercase(text)
     text = collapse_whitespace(text)
     return text
-	
+
 def german_cleaners(text):
     """Pipeline for German text, that collapses whitespace without transliteration.
     """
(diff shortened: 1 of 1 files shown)


BenoitWang:
Hi @padmalcom thank you for the pr, could you please explain why here you try to avoid lowercasing for German? And in German are abbreviations used? Is this cleaner general for German or for certain datasets? (Excuse that I don't speak German.) Thank you.

padmalcom:
Hi @BenoitWang, sure! In German nouns start with a capital letter, verbs, adverbs etc. don't. E.g. "Zahlen" (numbers) and "zahlen" (to pay) can have different meanings, so cases are important. Abbreviations are used, yes. Since the dataset I use, is generated from mp3s and text is extracted via speech-to-text models and fixed by a punctuation model, I'm pretty sure that it does not contain abbreviations. The cleaner will work for every dataset.

BenoitWang:
Ok I see. Please finish the tests (some simple linter issues). You are also welcome to create a list of german abbreviations if you've got time.

BenoitWang (commented):
Seems good.

padmalcom:
Sorry what ist the issue Here?

PR #1683, opened 2022-11-03
Added workaround for CyclicLR saving

Gastron:
Adds special saver/loader for CyclicLR since they have weakref.WeakMethods, which cannnot be pickled. Fortunately, those are not really state that gets updated during training, but rather they are reconstructed if the new instance is created with the same arguments (i.e. no custom scale function suddenly). Trying to detect this anyway, but only raise a warning - maybe users know what they are doing.


anautsch (approved):
lgtm - thank you for the adhoc fix !

PR #1572, opened 2022-09-09
Fastspeech2

bloodraven66:
@mravanelli , same as fastpitch, need to refactor a bit once fastspeech is merged.

PR #1709, opened 2022-11-16
mask noise for hifigan batch inference

BenoitWang:
Hi @mravanelli, `mel_lens=None` and `hop_len=None` are added to the `decode_batch` function of HifiGan batch inference, a `mask_noise` function is added to mask the noise.


BenoitWang:
done in https://github.com/speechbrain/speechbrain/pull/1572

PR #1717, opened 2022-11-19
[WIP] Whisper finetuning

Adel-Moumen:


speechbrain/lobes/models/huggingface_whisper.py
```diff
@@ -0,0 +1,155 @@
...
+            self.model.train() # we keep it to train to have dropout and LN computed adequaly
+            for param in self.model.parameters():
+                param.requires_grad = False
+        else:
+            self.model.train()
+            if self.freeze_feature_extractor:
+                logger.warning(
+                    "speechbrain.lobes.models.huggingface_whisper - whisper encoder is frozen."
+                )
+                for param in self.model.feature_extractor.parameters():
```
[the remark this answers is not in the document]
TParcollet: Indeed there is a mistake here, I'll change thanks.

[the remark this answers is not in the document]
TParcollet: By the way, we now have a fine-tuning example on libri if you wanna play with it, check the PR ;-)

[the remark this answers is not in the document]
bofenghuang: But there isn't any `feature_extractor` in `WhisperModel`. As said in the warning, it freezes the whisper's encoder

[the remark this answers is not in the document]
bofenghuang: maybe change `freeze_feature_extractor` to `freeze_encoder` ?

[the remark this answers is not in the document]
TParcollet: Well, it's just a naming issue, but you are right, we should rename it properly.

speechbrain/decoders/seq2seq.py
```diff
@@ -224,11 +225,13 @@ def reset_mem(self, batch_size, device):
     def forward_step(self, inp_tokens, memory, enc_states, enc_lens):
         """Performs a step in the implemented beamsearcher."""
         memory = _update_mem(inp_tokens, memory)
-        dec_out = self.module.forward_decoder(
+        # WARNING: the max_decode_ratio need to be under 449 becasue of positinal encoding
+        dec_out, attn = self.module.forward_decoder(
             enc_states, memory
         )  # TODO: switch args
         log_probs = self.softmax(dec_out[:, -1])
-        return log_probs, memory, None
```
TParcollet: Are we adding an argument here?

speechbrain/decoders/seq2seq.py
```diff
@@ -1394,9 +1397,9 @@ def permute_lm_mem(self, memory, index):
     def forward_step(self, inp_tokens, memory, enc_states, enc_lens):
         """Performs a step in the implemented beamsearcher."""
         memory = _update_mem(inp_tokens, memory)
-        dec_out = self.model.forward_decoder(enc_states, memory)
+        dec_out, attn, = self.module.forward_decoder(enc_states, memory)
         log_probs = self.softmax(dec_out[:, -1])
-        return log_probs, memory, None
```
TParcollet: We should make sure that this won't break any model

Adel-Moumen: I just followed what we did for the S2STransformerBeamSearch. The beam search algorithm expects to get the attn weights if provided. In our case, we have attn weights, so it makes sense to return them. One note: I think the return of attn from the decoder should be mandatory (and not optional). 

I did some tests with and without attn in the beam search and did not see any changes. (might be related to how I initialized the beam search. I will investigate later tomorrow).

Adel-Moumen:
LibriSpeech ZeroShot Whisper:
- test-clean: 3.11 WER
- test-other: 7.29 WER

LibriSpeech OneShot Whisper:
- test-clean: 3.46 WER
- test-other: 7.94 WER

The difference might be due to the hyperparameters settings that I used. Furthermore, I took a ckpt at 80% of LibriSpeech and got a test-clean of 2.99 and a test-other of 7.20%. This may imply that the model is doing worse when dealing with long files (it was sorted by ascending order on one GPU).

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
```diff
@@ -0,0 +1,146 @@
...
+test_batch_size: 8
+
+# this value is only used for the searchers as the bos starting token. 
+timestamp_index: 50363 
+eos_index: 50257
+bos_index: 50258
+
+# Decoding parameters
+min_decode_ratio: 0.0
+max_decode_ratio: 0.1
```
Adel-Moumen, line 62: @TParcollet I needed to reduce the max_decode_ratio because of the positional encoding that throws an error when the tokens sequence is superior to a specific value (4XX). Does it seem correct?

speechbrain/decoders/seq2seq.py
```diff
@@ -1400,14 +1400,11 @@ class S2SWhisperBeamSearch(S2SBeamSearcher):
     """
 
     def __init__(
-        self, module, temperature=1.0, temperature_lm=1.0, **kwargs,
+        self, model, temperature=1.0, temperature_lm=1.0, **kwargs,
     ):
         super(S2SWhisperBeamSearch, self).__init__(**kwargs)
 
-        self.model = module[0]
```
csukuangfj: Is there a reason to drop CTC support?

TParcollet: Just a point of view reason: I don't see why one would like to fine-tune using CTC as the Transformer decoder already is pre-trained - i.e. knows how to align properly. + for ST it's not good. If you can think of a reason to keep it, we'll keep it.

Adel-Moumen:
Update on the results: 

LibriSpeech
1-EPOCH Whisper Tiny :
valid WER: 3.47
test-clean WER: 3.09
test-other WER: 8.10

LibriSpeech
1-EPOCH Whisper Tiny.en:
valid WER: 3.22
test-clean WER: 2.61
test-other WER: 7.99

speechbrain/decoders/seq2seq.py
csukuangfj: > I don't see why one would like to fine-tune using CTC as the Transformer decoder already is pre-trained 

I am meaning that we can use CTC loss as an.auxiliary.loss during fine-tuning.

The reason to keep the CTC branch is that we can discard the transformer decoder and use only the CTC branch for decoding. Decoding with the attention decoder is rather slow according to past experience.

If someone does not want to use the CTC branch, he/she can ignore it and uses the transformer decoder for decoding.

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,377 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Define training procedure
+class ASR(sb.Brain):
+    def compute_forward(self, batch, stage):
+        """Forward computations from the waveform batches to the output probabilities."""
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        wavs, wav_lens = wavs.to(self.device), wav_lens.to(self.device)
```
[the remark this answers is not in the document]
Adel-Moumen: Ty I fixed it!

speechbrain/decoders/seq2seq.py
Adel-Moumen: > The reason to keep the CTC branch is that we can discard the transformer decoder and use only the CTC branch for decoding. Decoding with the attention decoder is rather slow according to past experience.

So you mean using only the encoder + a different decoder than the whisper decoder right?

csukuangfj: I am meaning that we only use the transformer encoder + the CTC branch (i.e. ctc_linear + softmax) for decoding.

TParcollet: @Adel-Moumen this makes sense, we should put back CTC xD

csukuangfj: > @Adel-Moumen this makes sense, we should put back CTC xD

Yes, that is only a small change but it provides us an alternative to use CTC for decoding.
I have not found any open-source project that is using CTC loss as an auxiliary loss when fine-tuning the whisper model.

TParcollet: Well, we have a recipe that fine-tune the whisper encoder with CTC only in this PR as well ... but not both :p Let's add it again.

Adel-Moumen: The boss has spoken. I"ll put back the CTC decoding :p

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,373 @@
...
+import os
+import sys
+import torch
+import logging
+import speechbrain as sb
+from speechbrain.utils.distributed import run_on_main
+from speechbrain.utils.data_utils import undo_padding
+from hyperpyyaml import load_hyperpyyaml
+from pathlib import Path
+from transformers.models.whisper.tokenization_whisper import LANGUAGES
```
Adel-Moumen, line 23: this only works when you have git clone the transformers repo... I don't see any other way to fetch the appropriate language token. Any ideas @TParcollet ?

Adel-Moumen: actually, we could copy-paste the LANGUAGES dict directly into the python script...

Moumeneb1: The PR looks great :D ! 
Just a little note: I experimented with whisper a little and found that the decoder can be fairly unpredictable ( because of the noisy data trained on), especially with background sounds. Even with the decoding strategies that are provided in the article, it can still be annoying when whispers persist for entire paragraphs that they have never stated on the wav. Perhaps fine-tuning the encoder and using it as CTC model only can help avoiding that

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
```diff
@@ -0,0 +1,139 @@
...
+save_folder: !ref <output_folder>/save
+train_log: !ref <output_folder>/train_log.txt
+
+# URL for the biggest Fairseq english whisper model.
+whisper_hub: openai/whisper-tiny
+language: english
+
+# Normalize the english inputs with
+# the same normalization done in the paper
+do_normalize: True
```
TParcollet, line 21: Variable name not explicit enough (normalized_transcripts) would be better

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+# Data files
+data_folder: !PLACEHOLDER # e,g./path/to/LibriSpeech
+# If RIRS_NOISES dir exists in /localscratch/xxx_corpus/RIRS_NOISES
```
TParcollet, line 25: remove all comments with respect to RIRS. Rirs are not used here

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
TParcollet: That is weird, you can let it like that but why is it happening?

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
```diff
@@ -0,0 +1,139 @@
...
+auto_mix_prec: False
+sample_rate: 16000
+
+# With data_parallel batch_size is split into N jobs
+# With DDP batch_size is multiplied by N jobs
+batch_size: 12
+test_batch_size: 8
+
+# this value is only used for the searchers as the bos starting token.
+timestamp_index: 50363
```
TParcollet, line 54: Add a comment to specify that this is hardcoded for HF or Whisper official repo

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,357 @@
+#!/usr/bin/env python3
+""" Take the encoder-decoder Whisper and the official tokenizer of whisper to fine-tune it with the NLL.
+
+This recipe can also be used as a zero-shot recipe by removing the training part.
```
TParcollet, line 4: More ref / text on the header (see the CTC whisper header)

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,357 @@
...
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        bos_tokens, bos_tokens_lens = batch.tokens_bos
+
+        # Add augmentation if specified
+        if stage == sb.Stage.TRAIN:
+            if hasattr(self.hparams, "augmentation"):
+                wavs = self.hparams.augmentation(wavs, wav_lens)
+
+        abs_tokens_lens = (bos_tokens_lens * bos_tokens.shape[1]).long()
```
TParcollet, line 41: Add comments to explicit what is happening in the next lines

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,357 @@
...
+            )
+        elif stage == sb.Stage.TEST:
+            self.hparams.train_logger.log_stats(
+                stats_meta={"Epoch loaded": self.hparams.epoch_counter.current},
+                test_stats=stage_stats,
+            )
+            with open(self.hparams.wer_file, "w") as w:
+                self.wer_metric.write_stats(w)
+
+    def init_optimizers(self):
```
TParcollet, line 169: Same here, I am not sure that we need to override this function in this recipe (we only have one optimizer that can be passed as an argument to the Brain class when initialised)

speechbrain/decoders/seq2seq.py
```diff
@@ -196,6 +196,79 @@ def forward(self, enc_states, wav_len):
...
+    This class implements the greedy decoding
+    for Whisper neural nets made by OpenAI in
+    https://cdn.openai.com/papers/whisper.pdf.
+
+    Arguments
+    ---------
+    model : HuggingFaceWhisper
+        The Whisper model.
+    **kwargs
+        see S2SBaseSearcher, arguments are directly passed.
```
TParcollet, line 210: An exemple in the docstring?

speechbrain/decoders/seq2seq.py
```diff
@@ -1313,6 +1386,111 @@ def lm_forward_step(self, inp_tokens, memory):
...
+        if len(module) == 2:
+            self.ctc_fc = module[1]
+
+        self.softmax = torch.nn.LogSoftmax(dim=-1)
+
+        self.temperature = temperature
+        self.temperature_lm = temperature_lm
+
+        self.decoder_input_tokens = None
+        self.language_token = 50259  # default language is english
```
TParcollet, line 1420: Can't we have this as arguments instead? Hard coding it in the init is WOW.

TParcollet (changes requested):
Once the requested changes will be done and results added, the PR will be ready to be merger.

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
Adel-Moumen: done.

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
Adel-Moumen: Done.

recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
Adel-Moumen: Done.

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
Adel-Moumen: Done.

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
Adel-Moumen: Done.

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,357 @@
...
+                target_words = [
+                    self.tokenizer._normalize(text) for text in target_words
+                ]
+
+            self.wer_metric.append(ids, predicted_words, target_words)
+            self.cer_metric.append(ids, predicted_words, target_words)
+
+        return loss
+
+    def fit_batch(self, batch):
```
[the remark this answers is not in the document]
Adel-Moumen: Done.

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
Adel-Moumen: Done.

speechbrain/decoders/seq2seq.py
Adel-Moumen: We can. It was just a default value instead of settings None. The value should be overridden by the setters. I don't fully like the idea of setting it directly in the yaml file as the value will be anyway modified with the setters. We could instead give a default value in args (e.g language_token=None or language_token=50259) instead of hardcoding it in the init. (but it's just a matter of where are we hardcoding the value :p)

recipes/LibriSpeech/ASR/transformer/train_with_whisper.py
```diff
@@ -0,0 +1,357 @@
...
+            "te_splits": hparams["test_splits"],
+            "save_folder": hparams["save_folder"],
+            "merge_lst": hparams["train_splits"],
+            "merge_name": hparams["train_csv"],
+            "skip_prep": hparams["skip_prep"],
+        },
+    )
+
+    # Defining tokenizer and loading it
+    tokenizer = WhisperTokenizer.from_pretrained(hparams["whisper_hub"])
```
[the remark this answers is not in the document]
Adel-Moumen: We can yes.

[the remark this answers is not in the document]
Adel-Moumen: I moved the tokenizer directly into the lobe. The tokenizer will be downloaded only if we are going to use the decoder. Didn't tried it yet but should work.

speechbrain/decoders/seq2seq.py
csukuangfj: > Once the requested changes will be done and results added, the PR will be ready to be merger.

Are there public pre-trained models for this PR?

TParcollet: @csukuangfj we are training models, and we will certainly release them. It is most likely that we will first merge then release (to give code for the HuggingFace fine-tuning challenge). I don't know either if we will provide an inference interface, but maybe ...

TParcollet (approved):
LGTM!

PR #1723, opened 2022-11-24
Fix parameter mismatch at calling fetch()

z-wony:
The fetch() method's parameter has been changed in [pretrained/fetching.py](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/pretrained/fetching.py#L26)
As a result, use_auth_token is moved 6th parameter, from 4th.
So, this commit specifies a parameter name for optional parameters.


anautsch:
Hi @z-wony sorry for getting back to this late; you have a valid point here. 

We had a recent update to this file as of another request. That was more of a quick fix solution. Your PR seems to do a bit more. Please resolve the conflict in a way you see fit.

z-wony:
Thank you for checking.

anautsch:
We failed to include this earlier. Thank you for being responsive, nonetheless.

Issue #1722, opened 2022-11-24
[Bug]: sorting does not works in ASR train


starcell:
### Describe the bug

When training with mult-GPU.
Setting sorting: random and sorting: ascending in hparams/*.yaml file show the same activity. 
In case sorting setted ascending, batch dataset processed randomly in training. 

### Expected behaviour

When I set sorting=ascending, shorter data trained earlier.

### To Reproduce

In yaml file, set as below
sorting: ascending

then ASR train!
It takes the same time when set sorting: random

### Versions

v0.5.13

With v0.5.11, no problem appeared.
I can train with short data at first in v0.5.11.

### Additional context

I tested and come to know there are some probelms in line 1000 of core.py.

1000        with tqdm(
                    train_set,
                    initial=self.step,
                    dynamic_ncols=True,
                    disable=not enable,
                ) as t:
                    for batch in t:

the train_set is sorted befor tqdm, but batch data is not sorted after tqdm().

starcell:
Solution!
I found where is wrong.
The value of 'shuffle' on the line 760 of core.py must be 'Fasle', not 'True'.
dataset, rank=self.rank, shuffle=False, drop_last=drop_last

I wish source code checked and corrected.

AsuMagic:
For the record, this change was introduced by #1518.

AsuMagic:
My understanding of the current code is that `shuffle=True` is passed to the [DistributedSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler) so that each DDP instance receives random subsets of the dataset at each epoch.  

Assuming you are correct, what I'm assuming is happening and may be worth digging into:  
To my understanding, sorting is happening in `dataio_prepare`, before the dataset sampling.  
But the sampler's `shuffle=True` does not only shuffle what subsets DDP instances receives, it also shuffles the order of indices within that sample, which breaks the sorting order.

This is the relevant snippet of PyTorch code:

```python
    def __iter__(self) -> Iterator[T_co]:
        if self.shuffle:
            # deterministically shuffle based on epoch and seed
            g = torch.Generator()
            g.manual_seed(self.seed + self.epoch)
            indices = torch.randperm(len(self.dataset), generator=g).tolist()  # type: ignore[arg-type]
        else:
            indices = list(range(len(self.dataset)))  # type: ignore[arg-type]

        if not self.drop_last:
            # add extra samples to make it evenly divisible
[...]

PR #1730, opened 2022-11-28
fix sorting bug

anautsch:
As described in #1722 , there is an issue with sorting and DDP.

This is to open up a PR. The goal is not fixing all possibilities - the goal is to restore functionality.
A testing is not yet implemented (needs DDP set-up; that might be a different sort of github workflow altogether).

https://github.com/speechbrain/speechbrain/blob/c229dbcde906a850ffcacd7de2c5b02755f2d74e/speechbrain/core.py#L726

defines a shuffle variable that is simply re-used for the flags that were redefined in #1518 ; defaults to `False`, such that sorting is kept if wanted througout.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -739,6 +739,12 @@ class Brain:
 
         # Possibly make a DistributedSampler or a wrapper for some other sampler
         if self.distributed_launch and not isinstance(dataset, IterableDataset):
+            # sort or not
+            if hasattr(self.hparams, "sorting"):
+                shuffle_ddp = self.hparams.sorting == "random"  # False if 'ascending' or 'descending'
+            else:
+                shuffle_ddp = True
+
             drop_last = loader_kwargs.get("drop_last", False)
             # num_replicas arg is equal to world_size
             # and retrieved automatically within
@@ -757,7 +763,7 @@ class Brain:
             elif loader_kwargs.get("batch_sampler") is None:
                 # no sampler and batch-sampler
                 self.train_sampler = DistributedSampler(
-                    dataset, rank=self.rank, shuffle=False, drop_last=drop_last,
+                    dataset, rank=self.rank, shuffle=shuffle_ddp, drop_last=drop_last,
                 )
 
                 # with DistributedSamplerWrapper, one must disable shuffling for dataloader
@@ -767,7 +773,7 @@ class Brain:
                 self.train_sampler = DistributedSamplerWrapper(
                     loader_kwargs.get("batch_sampler", None),
                     rank=self.rank,
-                    shuffle=False,
+                    shuffle=shuffle_ddp,
                 )
                 loader_kwargs["batch_sampler"] = self.train_sampler
         elif self.distributed_launch and isinstance(dataset, IterableDataset):
@@ -1200,11 +1206,18 @@ class Brain:
             for name, module in self.modules.items():
                 if any(p.requires_grad for p in module.parameters()):
                     module = SyncBatchNorm.convert_sync_batchnorm(module)
-                    module = DDP(
-                        module,
-                        device_ids=[self.device],
-                        find_unused_parameters=self.find_unused_parameters,
-                    )
+                    if self.distributed_backend == "gloo":
+                        module = DDP(
+                            module,
+                            device_ids=None,
+                            find_unused_parameters=self.find_unused_parameters,
+                        )
+                    else:
+                        module = DDP(
+                            module,
+                            device_ids=[self.device],
+                            find_unused_parameters=self.find_unused_parameters,
+                        )
                     self.modules[name] = module
         else:
             # data_parallel_backend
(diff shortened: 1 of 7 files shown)


anautsch:
> To what is this variable connected? a yaml parameter?

yeah kinda - but it might be in conflict with some other concept beneath.
maybe the idea is bogus. need to test - let's use this as a playground to get sth working

well, actually, the sorting flag is the flag?

anautsch:
just put it to "False" instead; this simply reverts the other PR that caused the bug w/o needing to rethink recipe & hparam logic.

anautsch:
Added a test and an initial fix outline. The test uses a CPU ddp backend, `gloo`.

I run into these nearby-sortings w/ sorting & ddp:
```
tensor([2.2050, 2.3400, 2.6000, 2.8700, 3.1050, 3.1500, 3.2200, 3.2800, 3.4550,
        3.6500, 3.7000, 4.2650, 4.4701, 4.7600, 4.8350, 4.8900, 5.1600, 6.1450,
        1.7750], dtype=torch.float64)

tensor([6.1450, 5.1600, 4.8900, 4.8350, 4.7600, 4.4701, 4.2650, 3.7000, 3.6500,
        3.4550, 3.2800, 3.2200, 3.1500, 3.1050, 2.8700, 2.6000, 2.3400, 2.2050,
        6.2000], dtype=torch.float64)
```

anautsch:
The test needed to be adjusted to DDP & pytorch's DistributedSampler which has:
```
        drop_last (bool, optional): if ``True``, then the sampler will drop the
            tail of the data to make it evenly divisible across the number of
            replicas. If ``False``, the sampler will add extra indices to make
            the data evenly divisible across the replicas. Default: ``False``.
```

Test works for the proposed fix on DDP & sorting. Bc of the drop_last & replicas, the final elements of the last batch won't be sorted within that batch. Yet, DDP seems to distribute the batches w/ replicas BUT
* for sorted (filtered) datasets, the total number of examples does not change
* for random (shuffling), the number of examples increases, as replicas are included.

The sorting asserts are thus tailored to the dummy data (and a bit lazy with "[:-1]" subset testing across all batches, instead of final batch only).

Ready for review.

speechbrain/core.py
```diff
@@ -1200,11 +1211,18 @@ def _wrap_distributed(self):
             for name, module in self.modules.items():
                 if any(p.requires_grad for p in module.parameters()):
                     module = SyncBatchNorm.convert_sync_batchnorm(module)
-                    module = DDP(
-                        module,
-                        device_ids=[self.device],
-                        find_unused_parameters=self.find_unused_parameters,
-                    )
+                    if self.distributed_backend == "gloo":
```
TParcollet, line 1214: Why has gloo a different behavior now?

speechbrain/utils/distributed.py
```diff
@@ -126,11 +126,12 @@ def ddp_init_group(run_opts):
                 "--distributed_backend=nccl"
             )
         else:
-            if run_opts["local_rank"] + 1 > torch.cuda.device_count():
-                raise ValueError(
-                    "Killing process " + str() + "\n"
-                    "Not enough GPUs available!"
-                )
+            if not run_opts["distributed_backend"] == "gloo":
```
TParcollet, line 129: Again, why is gloo treated like that? :p

TParcollet (commented):
See comments.

speechbrain/core.py
AsuMagic: I think @anautsch might have assumed that `gloo` is only desirable/used in a CPU DDP scenario. [This is the general recommendation,](https://pytorch.org/docs/stable/distributed.html#which-backend-to-use) but using `gloo` for GPU-based DDP appears to be a supported option, even if the recommendation is `nccl`.

anautsch: `gloo` is the only way for CPU to do DDP - i.e. what it wants, must be there for testing when GPUs are not there (github workflows).

`gloo` has API differences to the other two, as it uses here one CPU node only, and not several GPU nodes. Otherwise, pytorch throws an error - and that's it then.

speechbrain/utils/distributed.py
anautsch: the way it is implemented for GPU is fully incompatible with GPU, in fact: gloo will crash when run in cpu only.

This pytorch example for gloo is incompatible with SB, as it was before the contributed fix
https://pytorch.org/tutorials/intermediate/dist_tuto.html

anautsch: for this part, if there is only CPU, ofc there are not enough GPUs ;)

speechbrain/core.py
anautsch: btw - I did not try to make `mpi` available, compare
https://pytorch.org/docs/stable/distributed.html

anautsch: `gloo` cannot do GPU !!!

anautsch: a DDP integration testing -> not possible w/o GPUs
no idea how to make that workable within github workflows ;)
could be done offline & before another release is published - but then, what should the test case look like?

TParcollet (approved):
LGTM

Issue #1678, opened 2022-10-31
[Bug]: `foreign_class` not working with auth token


liaspas:
### Describe the bug

Hello,
I am trying to load a custom encoder from a private HuggingFace model with `foreign_class` but am faced with the following issues.
1. It seems `use_auth_token` is not properly passed here:
https://github.com/speechbrain/speechbrain/blob/64c8fa951bdade7e28c81b37452331c61af7b548/speechbrain/pretrained/interfaces.py#L91-L92
That is because `use_auth_token` is the sixth parameter in `fetch` and as such should be called like:
   ```
   fetch(hparams_file, source, savedir, use_auth_token=use_auth_token)
   ```
   By changing that I was able to download the custom encoder.py and hyperparams.yaml

2. Authentication fails again while trying to fetch the model checkpoints during `collect_files` of the Pretrainer:
https://github.com/speechbrain/speechbrain/blob/64c8fa951bdade7e28c81b37452331c61af7b548/speechbrain/utils/parameter_transfer.py#L202-L204
Where there seems to be no option to pass the auth token. 

### Expected behaviour

Passing the HuggingFace access token in `foreign_class(..., use_auth_token={token})` should authenticate and download the required files.

### Versions

speechbrain==0.5.13

PR #1734, opened 2022-12-01
Fix pretrained  / fetch calls

anautsch:
Fixing #1678.

The same issue popped up in other PRs that are currently under development; some hidden merge conflicts are ahead ;-)
This PR covers fully the existing fetch parameters and puts forth their defaults even - if one likes to override later on.


anautsch:
Closed here, it should have had only one commit but got mumbled up with commits for another PR.

PR #1735, opened 2022-12-01
explicit fetch args & download-only option

### LINKED ISSUE #1678: [Bug]: `foreign_class` not working with auth token

anautsch:
Cleaner attempt to fix #1678 - includes a download-only option for pretrained models

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -7,6 +7,7 @@ Authors:
  * Mirco Ravanelli 2020
  * Titouan Parcollet 2021
  * Abdel Heba 2021
+ * Andreas Nautsch 2022
 """
 import logging
 import hashlib
@@ -41,6 +42,7 @@ def foreign_class(
     overrides={},
     savedir=None,
     use_auth_token=False,
+    download_only=False,
     **kwargs,
 ):
     """Fetch and load an interface from an outside source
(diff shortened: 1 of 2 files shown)


Adel-Moumen (approved):
LGTM! I really like the idea of the download_only.

PR #1737, opened 2022-12-01
remove discussions references

Adel-Moumen:


README.md
```diff
@@ -214,7 +214,7 @@ The results will be saved in the `output_folder` specified in the yaml file. The
 
 # SpeechBrain Roadmap
 
-As a community-based and open-source project, SpeechBrain needs the help of its community to grow in the right direction. Opening the roadmap to our users enables the toolkit to benefit from new ideas, new research axes, or even new technologies. The roadmap, available on our [Discourse](https://speechbrain.discourse.group/t/speechbrain-a-community-roadmap/179) lists all the changes and updates that need to be done in the current version of SpeechBrain. Users are more than welcome to propose new items via new Discourse topics!
+As a community-based and open-source project, SpeechBrain needs the help of its community to grow in the right direction. Opening the roadmap to our users enables the toolkit to benefit from new ideas, new research axes, or even new technologies. The roadmap, will be available on our [Discussions](https://github.com/speechbrain/speechbrain/discussions/categories/announcements) and will lists all the changes and updates that need to be done in the current version of SpeechBrain. Users are more than welcome to propose new items via new Discussions topics!
```
AsuMagic, line 217:
```suggestion
As a community-based and open-source project, SpeechBrain needs the help of its community to grow in the right direction. Opening the roadmap to our users enables the toolkit to benefit from new ideas, new research axes, or even new technologies. The roadmap will be available in our [GitHub Discussions](https://github.com/speechbrain/speechbrain/discussions/categories/announcements) and will list all the changes and updates that need to be done in the current version of SpeechBrain. Users are more than welcome to propose new items via new Discussions topics!
```

Suggesting here:
- Removing the comma that flows awkwardly
- Fixing "available on" => "available in" (AFAIK that is the correct English form)
- Clarifying "GitHub Discussions" rather than just "Discussions". Probably not worth clarifying on the second mention, though
- Fixing `will lists` -> `will list`

AsuMagic (changes requested):
Those are the only two repo mentions to Discourse I could find back when I checked so that should be good.

README.md
Adel-Moumen: good catch :-)

AsuMagic (approved):

PR #1738, opened 2022-12-01
[WIP] Symbolic tracing for sb.nnet layers

fpaissan:
Starting a recipe to test and adapt speechbrain layers for [symbolic tracing](https://pytorch.org/docs/stable/fx.html).

### DIFF
--- a/speechbrain/nnet/RNN.py
+++ b/speechbrain/nnet/RNN.py
@@ -102,6 +102,7 @@ class RNN(torch.nn.Module):
         dropout=0.0,
         re_init=True,
         bidirectional=False,
+        padded_sequence_eval=True
     ):
         super().__init__()
         self.reshape = False
@@ -150,7 +151,7 @@ class RNN(torch.nn.Module):
         self.rnn.flatten_parameters()
 
         # Pack sequence for proper RNN handling of padding
-        if lengths is not None:
+        if lengths is not None and not self.training and self.padded_sequence_eval:
             x = pack_padded_sequence(x, lengths)
 
         # Support custom initial state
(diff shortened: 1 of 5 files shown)


speechbrain/nnet/RNN.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         dropout=0.0,
         re_init=True,
         bidirectional=False,
+        padded_sequence_eval=True,
```
mravanelli, line 105: This new variable is not described in the docstring

speechbrain/nnet/RNN.py
```diff
@@ -150,7 +151,11 @@ def forward(self, x, hx=None, lengths=None):
         self.rnn.flatten_parameters()
 
         # Pack sequence for proper RNN handling of padding
-        if lengths is not None:
+        if (
+            lengths is not None
+            and not self.training
```
mravanelli, line 156: I'm not sure about this line. This seems to lead to something different from the previous code as we call   `pack_padded_sequence` only when we are in eval modality.

fpaissan: you were right, I fixed it in [2e44f97](https://github.com/speechbrain/speechbrain/pull/1738/commits/2e44f978f23b76e2b939b00c2ff1975436708865)

speechbrain/nnet/RNN.py
fpaissan: Now this is actually not needing as parameter for the init, as we can use this as "shadow" flag to call inside the quantisation APIs..

Adel-Moumen:
Hello @fpaissan, 

What is the current state of this PR? Do you think it could be merged before August 2023? Thanks.

fpaissan:
> Hello @fpaissan,
> 
> What is the current state of this PR? Do you think it could be merged before August 2023? Thanks.

Sure!

Adel-Moumen:
Hello @fpaissan, sorry for the ping again. Do you have any news on this PR? CC: @mravanelli

PR #1739, opened 2022-12-03
[FIX] Flush gradients and save memory for validation.

MartinKocour:
When `fit_train` is done we should call optimizer.zero_grad(set_to_none=True) , which should force to clear the gradients from memory. In the current version of the code, the `optimizer.zero_grad()`  is the last call, which however store zeros in memory.
This minor change should allow to use bigger batch during `_fit_valid`  or `evaluate` and thus speed-it up. This is especially useful, when you use large models like SepFormer with significant memory footprint.

More details can be found in [this PyTorch Tuning Guide](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#use-parameter-grad-none-instead-of-model-zero-grad-or-optimizer-zero-grad).

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -822,7 +822,7 @@ class Brain:
         if ``set_to_none=False`` (default) or to None otherwise.
 
         Setting gradients to None should save the memory, e.g.
-        during ``evaluate()`` and thus larger batch might be used
+        during ``evaluate()`` and thus larger batch might be used.
         """
         if hasattr(self, "optimizer"):
             self.optimizer.zero_grad(set_to_none)
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -818,8 +818,11 @@ class Brain:
                 self.checkpointer.add_recoverable("optimizer", self.optimizer)
 
     def zero_grad(self, set_to_none=False):
-        """Sets the gradients of all optimized `torch.Tensor`s to zero
-        if `set_to_none=False` (default) or to `None` otherwise.
+        """Sets the gradients of all optimized ``torch.Tensor``s to zero
+        if ``set_to_none=False`` (default) or to None otherwise.
+
+        Setting gradients to None should save the memory, e.g.
+        during ``evaluate()`` and thus larger batch might be used
         """
         if hasattr(self, "optimizer"):
             self.optimizer.zero_grad(set_to_none)
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -817,6 +817,10 @@ class Brain:
             if self.checkpointer is not None:
                 self.checkpointer.add_recoverable("optimizer", self.optimizer)
 
+    def zero_grad(self, set_to_none=False):
+        if hasattr(self, "optimizer"):
+            self.optimizer.zero_grad(set_to_none)
+
     def on_evaluate_start(self, max_key=None, min_key=None):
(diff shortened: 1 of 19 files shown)


MartinKocour:
I also spotted that gradients are not accumulated correctly, so I fixed that too.

mravanelli:
Thank you for this PR @MartinKocour. I think it can be useful. One thing I'm not sure about is why calling the function you defined (`zero_grad`) in `fit_batch` as well. As you can see, in `core.py` you still call the original function  `self.optimizer.zero_grad()`, but you modified the recipes files such that they call `zero_grad` in their `fit_batch `. I'm not sure we need to delete the gradient buffers at the end of each batch (this might slow down?), but likely we really need to do it at the end of a training epoch only, right?

MartinKocour:
Mirco @mravanelli, in recipes I am not deleting anything.  My `zero_grad(set_to_none=False)` has the same behaviour like original `torch.optimizer.zero_grad()`, i.e. by default it does not delete the gradients from memory, but instead it just zeros the gradients when `set_to_none=False` (default).

But you are right, I don't have to call self.zero_grad() in the recipes. I just didn't want to duplicate same lines of code. But it is not necessary, I can remove it. But the behaviour with `self.zero_grad()` is the same like with original code.

In `core.py` I did not use it since it is just single line of code + calling `optimizer.zero_grad()` is faster than calling `Brain.zero_grad()` since it contains additional `if` statement.

speechbrain/core.py
```diff
@@ -878,6 +887,7 @@ def fit_batch(self, batch):
                 if self.check_gradients(loss):
                     self.scaler.step(self.optimizer)
                 self.scaler.update()
+                self.optimizer.zero_grad()
```
pplantinga, line 890: While I appreciate that this is technically faster than `self.zero_grad()`, I think it would still make more sense to me to call `self.zero_grad()` here. Its more consistent and if for some reason someone wanted to override this function it would be better to call it here as well.

pplantinga (commented):
Nice change! I especially appreciate saving space for evaluation and fixing the gradient accumulation bug. Quick question, have you verified that this does in fact save space for evaluation and there's nothing else (e.g. manual garbage collection) that is needed?

speechbrain/core.py
MartinKocour: I agree, but what about the recipes, which uses custom optimisers and thus has custom `fit_batch`? Should I also call `self.zero_grad()` there?

MartinKocour: > Nice change! I especially appreciate saving space for evaluation and fixing the gradient accumulation bug. Quick question, have you verified that this does in fact save space for evaluation and there's nothing else (e.g. manual garbage collection) that is needed?

This is what I am evaluating this week, that is why the PR is marked as WIP

pplantinga: I don't think we need to add `self.zero_grad()` for recipes with custom `fit_batch`. While it probably wouldn't hurt anything, it doesn't seem a necessary change to me, recipes will continue to work as before. If in the future there is some change to `self.zero_grad()`, the recipes could be updated at that time.

MartinKocour: Okay, I can confirm that calling `self.zero_grad(set_to_none=True)` will free the memory. So I think you can merge it.

MartinKocour:
Ready to be merged

pplantinga (approved):
Nice! LGTM

PR #1743, opened 2022-12-04
Avoid loading checkpoint parameters on the target device

lucadellalib:
When 'map_location=device', the checkpoint parameters are loaded on the CPU first (see docs of torch.load) and then moved to the target device. This means that, before copying the checkpoint parameters into the model, we have 2 full independent copies of the model parameters on the target device (first copy are the parameters in the model, second copy are the parameters being recovered from the checkpoint). If the model is on the CPU, we cannot avoid the waste of memory. However, if the model is on a device different from CPU (e.g. "cuda"), we can avoid moving the loaded checkpoint parameters on the device and hence avoid wasting the device memory and potentially having an out-of-memory error when dealing with huge models. Since 'obj.load_state_dict' copies the loaded parameters into the model/optimizer/scheduler in-place one by one, it automatically takes care of moving them to the model's device, even if they are on the CPU.

### DIFF
--- a/recipes/AISHELL-1/ASR/seq2seq/train.py
+++ b/recipes/AISHELL-1/ASR/seq2seq/train.py
@@ -328,7 +328,7 @@ if __name__ == "__main__":
 
     # We download and pretrain the tokenizer
     run_on_main(hparams["pretrainer"].collect_files)
-    hparams["pretrainer"].load_collected(device=run_opts["device"])
+    hparams["pretrainer"].load_collected()
 
     # Trainer initialization
     asr_brain = ASR(
--- a/recipes/AISHELL-1/ASR/transformer/train.py
+++ b/recipes/AISHELL-1/ASR/transformer/train.py
@@ -268,7 +268,7 @@ class ASR(sb.core.Brain):
             max_key=max_key, min_key=min_key
         )
         ckpt = sb.utils.checkpoints.average_checkpoints(
-            ckpts, recoverable_name="model", device=self.device
+            ckpts, recoverable_name="model",
         )
 
         self.hparams.model.load_state_dict(ckpt, strict=True)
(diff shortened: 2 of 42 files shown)


mravanelli:
@Gastron could you please take a look?

Gastron:
This is a good idea, I think we should go through with this change. Like noted, for example torch.load [documentation](https://pytorch.org/docs/stable/generated/torch.load.html) suggests this.

However, the `device` argument of the loading code becomes unnecessary with this approach. It should be removed - unfortunately it gets used (again, unnecessarily) in many places, e.g. https://github.com/speechbrain/speechbrain/blob/46be2d16bdb222ad06821b804c5f61d17a5431f4/speechbrain/core.py#L827

@lucadellalib would you be willing to go over the codebase and remove the unnecessary `device` argument? It should be removed from the `torch_recovery`, `torch_parameter_transfer`, as well as `mark_as_loader `and `mark_as_transfer`. And then it gets used in calls to `recover_if_possible` and possibly `load_checkpoint`? Note that this will take some time.

lucadellalib:
@Gastron I think it would be better to leave it as it is to minimize the number of changes (easier to debug in case of problems, lower risk of breaking other components, etc.) and especially for backward compatibility. Some users might be using those functions and suddenly find their code not working anymore because they are passing an unexpected argument - `device`.

mravanelli:
What about adding the change in the unstable branch? This way it will be
available for the next major version (where some interface changes are
planned).

Gastron:
I think it would be better to take the whole step of removing the device argument. However, I agree with Mirco that this is then a breaking change, meaning we should only add it during the next major version.

Of course we understand if @lucadellalib doesn't want to take the time to remove the argument from so many places, or how do you feel Luca? Perhaps someone else should contribute that part?

mravanelli:
Hi Luca,
feel free to move on with it and ask a PR on the "unstable" branch (the one
for the new version of speechbrain).
Note that we will soon merge the recipe testing PR that allows us to test
every single recipe.

Best,

Mirco

lucadellalib:
@Gastron I removed the unnecessary `device` argument from all the places as discussed, please take a look.

Gastron:
Great work! I browsed all the changed, looks good to me. This touches a lot of recipes, is it ok to merge this now, before the recipe testing PR @anautsch ?

anautsch:
lgtm!

@Gastron we'll see what breaks & fix it - but this PR's changes seem complementary. Thank you, @lucadellalib !

PR #1744, opened 2022-12-04
loss must be avg when BS>1 when calling evaluate_batch()

sangeet2020:
Hi @ycemsubakan ,

I initiate this PR to fix a bug in the `WHAMandWHAMR` recipe. I realized that the current recipe works fine on batch_size as 1. However, when increased, losses are not averaged when `evaluate_batch` is called. 
Basically, it makes sense to write
```python
loss = self.compute_objectives(predictions, targets)
loss = torch.mean(loss)   # Avg losses in a batch
```

Thank You

### DIFF
--- a/recipes/WHAMandWHAMR/enhancement/train.py
+++ b/recipes/WHAMandWHAMR/enhancement/train.py
@@ -233,8 +233,7 @@ class Separation(sb.Brain):
 
         with torch.no_grad():
             predictions, targets = self.compute_forward(mixture, targets, stage)
-            loss = self.compute_objectives(predictions, targets)
-            loss = torch.mean(loss)
+            loss = self.compute_objectives(predictions, targets).mean()
 
         if stage != sb.Stage.TRAIN:
             self.pesq_metric.append(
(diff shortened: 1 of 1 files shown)


mravanelli:
@ycemsubakan could you please take a look?

ycemsubakan:
@mravanelli @sangeet2020 I will try to do it this week.

MartinKocour:
@sangeet2020 take a look into mine LibriCSS recipe. IMHO, it should be done inside compute_objective
LibriCSS recipe is here:
https://github.com/MartinKocour/speechbrain/blob/49ab1668c8e8952cf3f33efa3d8d1f15cd3f42a4/recipes/LibriCSS/separation/train.py#L139

ycemsubakan:
@MartinKocour , thanks for pointing this out. I 'll take a look tonight.

sangeet2020:
> @sangeet2020 take a look into mine LibriCSS recipe. IMHO, it should be done inside compute_objective LibriCSS recipe is here: https://github.com/MartinKocour/speechbrain/blob/49ab1668c8e8952cf3f33efa3d8d1f15cd3f42a4/recipes/LibriCSS/separation/train.py#L139

yes, this is a suitable choice. I will modify it and re-commit. thanks @MartinKocour

ycemsubakan:
Hmm... we need make sure it works with both frequency domain and time.. I am checking now..

ycemsubakan:
I pushed a small cosmetic update. @sangeet2020 , are you planning to have more updates? If yes, we can keep this PR open. If not, we can merge.

ycemsubakan (approved):

ycemsubakan:
> Hi Cem, I dont plan any more updates on this PR, I think we are safe to close and merge it.

Alright, thanks for the PR @sangeet2020!

PR #1751, opened 2022-12-07
Enabling the retrieval of whisper's hidden states

Hguimaraes:
Hello guys,

I'm trying to extend the Whisper class to enable the retrieval of the Hidden States. I think it is handy for people working with speech representation learning. It is a small change. Also, I tried to make it compatible with the previous version, so it should not break any examples or codes that are already using this class.

Best regards,
Heitor

### DIFF
--- a/speechbrain/lobes/models/huggingface_whisper.py
+++ b/speechbrain/lobes/models/huggingface_whisper.py
@@ -65,6 +65,7 @@ class HuggingFaceWhisper(nn.Module):
         freeze=False,
         freeze_encoder=False,
         output_attentions=True,
+        output_hidden_states=False,
     ):
         super().__init__()
         self.sampling_rate = sampling_rate
@@ -72,6 +73,7 @@ class HuggingFaceWhisper(nn.Module):
         self.freeze = freeze
         self.freeze_encoder = freeze_encoder
         self.output_attentions = output_attentions
+        self.output_hidden_states = output_hidden_states
 
         self.tokenizer = None
         # Download the tokenizer only if we are going to use the Decoder.
@@ -131,18 +133,29 @@ class HuggingFaceWhisper(nn.Module):
                 out_encoder = self.forward_encoder(wav)
                 if self.encoder_only:
                     return out_encoder
-                logits, attn = self.forward_decoder(
-                    out_encoder, decoder_input_ids
-                )
+
+                if self.output_hidden_states:
+                    logits, attn = self.forward_decoder(
+                        out_encoder[-1], decoder_input_ids
+                    )
+                else:
+                    logits, attn = self.forward_decoder(
+                        out_encoder, decoder_input_ids
+                    )
                 return out_encoder, logits, attn
         else:
             if self.encoder_only:
                 return self.forward_encoder(wav)
             else:
                 out_encoder = self.forward_encoder(wav)
-                logits, attn = self.forward_decoder(
-                    out_encoder, decoder_input_ids
-                )
+                if self.output_hidden_states:
+                    logits, attn = self.forward_decoder(
+                        out_encoder[-1], decoder_input_ids
+                    )
+                else:
+                    logits, attn = self.forward_decoder(
+                        out_encoder, decoder_input_ids
+                    )
                 return out_encoder, logits, attn
 
     def forward_encoder(self, wav):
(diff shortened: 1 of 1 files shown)


speechbrain/lobes/models/huggingface_whisper.py
```diff
@@ -65,13 +65,15 @@ def __init__(
         freeze=False,
         freeze_encoder=False,
         output_attentions=True,
+        output_hidden_states=False,
```
Adel-Moumen, line 68: I'd prefer to see "output_all_hiddens" instead of "output_hidden_states" as we already have something really similar to what you did with Wav2vec2 and we named it "output_all_hiddens" see: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/huggingface_wav2vec.py

Adel-Moumen (changes requested):
Hello, 

Many thanks for the PR! 

I left one comment on the naming of the variable + docstring. After that, I think everything is LGTM.

speechbrain/lobes/models/huggingface_whisper.py
Hguimaraes: Thank you for your answer! I pushed the modifications. I changed the variable name, inserted it on the docstring, and also stack the tensors instead of returning a tuple... Now it is more similar to the wav2vec2 class.

speechbrain/lobes/models/huggingface_whisper.py
```diff
@@ -131,18 +138,29 @@ def forward(self, wav, decoder_input_ids=None):
                 out_encoder = self.forward_encoder(wav)
                 if self.encoder_only:
                     return out_encoder
-                logits, attn = self.forward_decoder(
-                    out_encoder, decoder_input_ids
-                )
+
+                if self.output_all_hiddens:
+                    logits, attn = self.forward_decoder(
+                        out_encoder[-1, ...], decoder_input_ids
```
Adel-Moumen, line 144: why is there now `out_encoder[-1, ...]` instead of `out_encoder[-1]` ?

speechbrain/lobes/models/huggingface_whisper.py
```diff
@@ -155,10 +173,16 @@ def forward_encoder(self, wav):
...
-                return self.model.encoder(mel).last_hidden_state
+                return self._get_encoder_states(wav)
+        else:
+            return self._get_encoder_states(wav)
+
+    def _get_encoder_states(self, wav):
+        mel = self._get_mel(wav)
+        if self.output_all_hiddens:
+            states = self.model.encoder(mel, output_hidden_states=True)
+            return torch.stack(states.hidden_states)
```
Adel-Moumen, line 184: is it the same to `torch.stack(list(states.hidden_states), dim=0)` (taken from wav2vec2)?

Hguimaraes: Yes! I think it is not necessary to take this extra step to first convert to a list, we can stack the tuples directly.

speechbrain/lobes/models/huggingface_whisper.py
Hguimaraes: Because now I'm not returning a tuple of tensors as before, I'm returning the stacked tensor (num_layers, B, T, C) [As I answered you in the other comment]

Hguimaraes:
Hi @Adel-Moumen,

Please let me know if there is anything else I could do on this PR. I think I have answered the previous questions and the interface is similar to the wav2vec2 class.

Best regards,

Adel-Moumen:
> Hi @Adel-Moumen,
> 
> Please let me know if there is anything else I could do on this PR. I think I have answered the previous questions and the interface is similar to the wav2vec2 class.
> 
> Best regards,

Hello @Hguimaraes, 

I will finish the review this week :-)

speechbrain/lobes/models/huggingface_whisper.py
```diff
@@ -155,10 +173,16 @@ def forward_encoder(self, wav):
 
         if self.freeze_encoder:
             with torch.no_grad():
-                mel = self._get_mel(wav)
-                return self.model.encoder(mel).last_hidden_state
+                return self._get_encoder_states(wav)
+        else:
+            return self._get_encoder_states(wav)
+
+    def _get_encoder_states(self, wav):
```
Adel-Moumen, line 180: please add some doc string on this function

speechbrain/lobes/models/huggingface_whisper.py
Adel-Moumen, line 144: From my understanding, `out_encoder[-1, ...]` is exactly the same as `out_encoder[-1]`, so why adding "..." ?

Adel-Moumen (changes requested):
Hello, 

Thanks again for your great work! Please take a look at my latest comments :-) 

Btw, I wish you a great end of the year :-)

Hguimaraes:
Hi @Adel-Moumen,

Implemented the changes. If there is anything else, just let me know, and I can update it.
I hope you have a great end of the year as well!

Best,

Adel-Moumen (approved):
Everything sounds good! Many thanks for the PR! :-)

Issue #1164, opened 2021-11-25
E   TypeError: save() missing 1 required positional argument: 'sample_rate'


Timodhau:
Hello,

Getting error in using 

`enhancer = SpectralMaskEnhancement.from_hparams(
    source="speechbrain/mtl-mimic-voicebank",
    savedir='stream_folder/',
    run_opts={"device":"cuda"})

enhanced = enhancer.enhance_file("stream_folder/fr_yourname.wav", output="stream_folder/test.wav)`

test_speechbrain.py:None (test_speechbrain.py)
test_speechbrain.py:13: in <module>
    enhanced = enhancer.enhance_file("stream_folder/fr_yourname.wav",
/home/tim/env/python3.8/lib/python3.8/site-packages/speechbrain/pretrained/interfaces.py:1929: in enhance_file
    torchaudio.save(output_filename, enhanced, channels_first=False)
E   TypeError: save() missing 1 required positional argument: 'sample_rate'

probably an update on torchaudio.save you need to handle.

Regards

TParcollet:
@pplantinga ?

pplantinga:
I guess this error no longer occurs.

Ca-ressemble-a-du-fake:
Hi,

I've installed SpeechBrain today from the git repos following the [guide](https://github.com/speechbrain/speechbrain#install-with-github) and ran the [example](https://github.com/speechbrain/speechbrain/blob/e498b87d6d7a99dc9989bf943b79eda9b4e284e7/speechbrain/pretrained/interfaces.py#L2166) shown in code.

Just to let you know the issue still occurs.

So I modified the code to be able to pass in the sample_rate :
`torchaudio.save(output_filename, enhanced, sample_rate, channels_first=False)`

But the saved file is 59 bytes (the input noisy wav file was 2.9MB).

I tried to save the `noisy` variable and the same occured : output file was 59 bytes.

AsuMagic:
I can confirm this successfully reproduces the issue:

```python
from speechbrain.pretrained import SpectralMaskEnhancement
enhancer = SpectralMaskEnhancement.from_hparams(
    source="speechbrain/metricgan-plus-voicebank"
)
enhanced = enhancer.enhance_file(
    "speechbrain/metricgan-plus-voicebank/example.wav",
    output_filename="test.wav"
)
```

There are two issues:
1. `sample_rate` is never passed to `torchaudio.save` (I suspect file output was actually never tested)
[...]

PR #1754, opened 2022-12-09
Update interfaces.py

yingshaoxo:
Fix a bug:

```
        if output_filename is not None:
>           torchaudio.save(output_filename, enhanced, channels_first=False)
E           TypeError: save() missing 1 required positional argument: 'sample_rate'

/opt/homebrew/lib/python3.10/site-packages/speechbrain/pretrained/interfaces.py:2498: TypeError
```

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -2221,7 +2221,7 @@ class SpectralMaskEnhancement(Pretrained):
         # Return resynthesized waveforms
         return self.hparams.resynth(torch.expm1(enhanced), noisy)
 
-    def enhance_file(self, filename, output_filename=None):
+    def enhance_file(self, filename, output_filename=None, sample_rate=16000):
         """Enhance a wav file.
 
         Arguments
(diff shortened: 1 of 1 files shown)


speechbrain/pretrained/interfaces.py
```diff
@@ -2221,7 +2221,7 @@ def enhance_batch(self, noisy, lengths=None):
         # Return resynthesized waveforms
         return self.hparams.resynth(torch.expm1(enhanced), noisy)
 
-    def enhance_file(self, filename, output_filename=None):
+    def enhance_file(self, filename, output_filename=None, sample_rate=16000):
```
AsuMagic, line 2224: I don't think taking a `sample_rate` parameter here is a good idea. It is confusing because it must be made to match the input sample rate, but that is already known at `load_audio` time. Mismatching the sample rate will only silently produce an output file with an incorrect sample rate.

AsuMagic (changes requested):
Related to #1164 - it seems like a bunch more classes are affected, so this probably needs more thought, and should fix all classes at once.

speechbrain/pretrained/interfaces.py
yingshaoxo: OK, that's fine.

I think the community will fix it at some point.

PR #1767, opened 2022-12-15
threshold SpeakerRecognition

joshipunitram:
The modification in the threshold is very important to ensure the accuracy for speaker verfication through SpeakerRecognition class.

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -1053,7 +1053,7 @@ class SpeakerRecognition(EncoderClassifier):
         self.similarity = torch.nn.CosineSimilarity(dim=-1, eps=1e-6)
 
     def verify_batch(
-        self, wavs1, wavs2, wav1_lens=None, wav2_lens=None, threshold=0.25
+        self, wavs1, wavs2, wav1_lens=None, wav2_lens=None, threshold=0.5
     ):
         """Performs speaker verification with cosine distance.
 


joshipunitram:
The modification in the threshold is very important to ensure the accuracy for speaker verfication through SpeakerRecognition class. In the former code the threshold is defined to be 0.25, which gives an erroneous result. Redefining it to 0.5 gives quite better results.

AsuMagic:
IMO if this value is to be changed it should be done very very carefully, especially as this is defined globally rather than at model level.  
Given the applications of speaker recognition, I think it may be more worthwhile to add a big fat warning to tell users that they should carefully evaluate the threshold for their particular model *and data*.

I am using `speechbrain/spkrec-ecapa-voxceleb` for a university assignment, using a subset of VoxCeleb for evaluation. Coincidentally, I had been testing threshold values. Unless I got something wrong, some of my findings are as follows (where "seuil" stands for "threshold"):

false acceptance/false reject vs threshold

where

```python
    false_reject = false_negative / (true_positive + false_negative) # FR
    false_accept = false_positive / (false_positive + true_negative) # FA
```

As you can see, the value of `0.25` is much, _much_ closer to the EER (i.e. where FA/FR intersect) than to `0.5`. That is not to say that a higher threshold might not be desirable for applications with much stronger FR requirements, but I don't think the default should be changed.

asumagic:
Closing for the reason I provided, IMO the fix would be better documentation

PR #1781, opened 2023-01-03
Added fix to use DDP with hifi_gan training on ljspeech

padmalcom:
Function call in __main__ to `ddp_init_group()` was missing in train.py for hifi_gan. This MR adds this line. Without it, creating the required json files for training fails when using ddp.


anautsch (approved):
lgtm (thank you!!)

PR #1782, opened 2023-01-03
Conformer Transducer Librispeech (Contribution from Samsung AI Cambridge)

TParcollet:
This PR refactors our transducer recipe for Librispeech. In practice, we will drop the current CRDNN that hasn't even been training on the full set with a better-performing conformer_transducer.

- [X] Develop a first working multitask (CTC only for now) conformer transducer. This is with RNNLM. @TParcollet 
- [X] Refactor the recipe to be simpler. @TParcollet 
- [X] Go below 3% of WER. Now at 2.8% with BS on test-clean with RNNLM.

**Todo in another PR**
- Refactor entirely transducer BS to match our new interface @Adel-Moumen.
- Turn this recipe into something that works with TransformerLM and match ESPnet 2.4%.
- Transpose all this to all transducer recipes (CommonVoice I believe).


recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -213,16 +219,17 @@ dec: !new:speechbrain.nnet.RNN.LSTM
...
-   input_size: !ref <dec_dim>
-   n_neurons: !ref <joint_dim>
+   input_size: !ref <joint_dim>
+   n_neurons: !ref <output_neurons>
    bias: False
 
 # For MTL
-#ce_cost: !name:speechbrain.nnet.losses.nll_loss
-#   label_smoothing: 0.1
+ce_cost: !name:speechbrain.nnet.losses.nll_loss
```
AsuMagic, line 232: Note that `ce_weight` is `0.0` so this isn't effective right now (stating just in case).

Is it possible that applying label smoothing over the auxiliary loss here would help the other losses? I would think not but I am wondering.

recipes/LibriSpeech/ASR/transducer/train.py
```diff
@@ -315,7 +327,12 @@ def dataio_prepare(hparams):
     @sb.utils.data_pipeline.takes("wav")
     @sb.utils.data_pipeline.provides("sig")
     def audio_pipeline(wav):
-        sig = sb.dataio.dataio.read_audio(wav)
+        if hparams["speed_perturb"]:
+            sig = sb.dataio.dataio.read_audio(wav)
+
+            sig = hparams["speed_perturb"](sig.unsqueeze(0)).squeeze(0)
```
AsuMagic, line 353: FWIW you mentioned slow training at some point so I am wondering if it could be related to the speed perturbation being performed here. Could be worth investigating (wondering if it's done on CPU or GPU, what happens wrt. DDP, wrt. PyTorch threading because of workers, etc.)

AsuMagic: Personally, I didn't notice a performance degradation performing this in the model's `forward` when in train mode.

AsuMagic (dismissed):
Made some general remarks and pointed out at one issue in the hparams. Other than that, LGTM, and I will be retraining a model shortly.

recipes/LibriSpeech/ASR/transducer/train.py
AsuMagic: I would need to double check, but moving it to the forward seems to have made the end-of-epoch validation step much faster..

Adel-Moumen:
Hey @TParcollet, any news please? Thanks. :-)

AsuMagic:
Taking the liberty of pushing my current code (without streaming stuff, only the stuff that gets me to 2.8) but bear in mind I'm currently trying to fight git for a bit.

mravanelli:
Sure, ping me wheb you are ready for a review.

AsuMagic:
I have yet to upload the checkpoint + edit the README with new results, I will do this tomorrow. The code side of things is ready if you prefer not to delay reviewing for it. With my changes, using gradient averaging, we reach a greedy-search `test-clean` WER of **2.84%** as roughly was the goal.

There is one issue I had to work around: In `TransformerASR`, `custom_tgt_module` is initialized regardless of if transformer decoding is used at all, which is not the case here.  
This is a problem, because it broke DDP for me as it discovered an unused parameter. It also adds a useless parameter to trained model checkpoints (~0.5M).  
The `Transformer` class (which `TransformerASR` extends) properly gates that initialization behind a `if num_decoder_layers > 0:` check. The problem is that simply adding this check will not cut it as it will break every existing trained model, since trained models will include keys that will not exist (as the `custom_tgt_module` will not have been initialized).  
I am not sure what the best approach to fix the issue is, so for now as to avoid delaying things, I reintroduced the useless parameter to my checkpoint after finishing training so that this code can be left unchanged.

mravanelli:
@TParcollet, @AsuMagic  thank you for this PR!

- I did some tests and everything LGTM. I ran the recipe in debug model and it looks like the code is runnable end-to-end. I couldn't test the DDP part as I could do my test on a single GPU (with reduced batch size). 

Regarding DDP, here are my questions:
- @TParcollet, did you observe the problem mentioned by @AsuMagic ?
- @AsuMagic, do you see a way to fix this issue in a backward-compatible way? Maybe we can write a custom load_state_dict that loads the optional parameters only if they are available in the checkpoint.

TParcollet:
@AsuMagic what about gating the custom_tgt_module as well if num_decoder_layers == 0?

AsuMagic:
> @AsuMagic what about gating the custom_tgt_module as well if num_decoder_layers == 0?

What do you mean precisely? Only instantiating the `custom_tgt_module` if `num_decoder_layers > 0` is indeed possible but it will break existing trained models that have no decoder layers without further changes. Quite a few recipes seem to use `num_decoder_layers: 0`.

> Maybe we can write a custom load_state_dict that loads the optional parameters only if they are available in the checkpoint.

Possibly, but I am not sure what is the best way to do it. If the caller code (i.e. not `TransformerASR` itself) has to be modified then it would be a breaking change. Ideally, the workaround would be at `TransformerASR` level... But I don't know if that's possible.

AsuMagic:
Updated README with all results + uploaded the checkpoint files. It took me a bit because I had to rerun evaluation.

The current checkpoint includes the useless `custom_tgt_module` at the moment.

AsuMagic:
Since the Dropbox PR was merged, this now conflicts. Also I forgot that the `LibriSpeech.csv` recipe file should be updated with a link.

But let's first decide what we should do regarding the `custom_tgt_module` as a checkpoint reupload may be needed. The current checkpoint contains the unnecessary parameter (this will still work, and is a workaround for now).

Other than that, it should be all good AFAIK.

TParcollet:
If you can think of a quick and easy way of fixing the incompatibility that will arise with existing checkpoints and the removal of custom tgt, then we fix it. Otherwise we move forward

mravanelli:
Thank you @AsuMagic,
- It looks like the log folder is still in gdrive, while we need to switch to Dropbox. Do you want me to upload to the speechbrain dropbox and change the link?
- It looks like the DDP issue is important, but might be fixed in a new PR (maybe for the new major version of SpeechBrain). What do you suggest to do @TParcollet?

AsuMagic:
> It looks like the log folder is still in gdrive, while we need to switch to Dropbox. Do you want me to upload to the speechbrain dropbox and change the link?

Sure @mravanelli !

mravanelli (approved):

PR #1800, opened 2023-01-13
Hf yaml cleanup

anautsch:
This is not supposed to go anywhere near the `develop` branch.

Purpose: tracking HF yaml/interface specifications & their testing.

When I created this branch, I simply checked out from develop - that was a bit too much. All that is really needed here from the develop branch are the linters.
* [x] drop all files unrelated to HF repo yaml/interfaces
* [x] patch existing test descriptors to their latest definition, compare #1623 

I hope making a PR here isn't overkill, it's just so there's awareness of what's going on & some participation in how this branch can be shaped for testing automation when refactorings touch upon HF repos (such as wav2vec2 fixes & refactoring SB's transformers interface).


anautsch:
If this is too odd, we might also drop that branch on the speechbrain repo & create a new branch for this purpose that does not start from a speechbrain/develop copy (and retire this branch).

anautsch:
Nvmd - closing here. I created this orphan branch instead: https://github.com/speechbrain/speechbrain/tree/hf-interface-testing

PR #1808, opened 2023-01-19
Fix comparison order of train loss

z-wony:
If batch size > 1 and all of the losses are less than threshold,
RuntimeError raised from loss comparison block.


z-wony:
For example,
if loss is [ -10.1, -11.1 ] and threshold is -10.0
-> `if loss < self.hparams.loss_upper_lim` block raise error.

RuntimeError: Boolean value of Tensor with no values is ambiguous

TParcollet:
@pplantinga could you have a look please?

pplantinga (approved):
Yes, this corrects the order of comparison. LGTM!

PR #1838, opened 2023-02-16
edits for refactoring check tool

anautsch:
Some minor patches (need became obvious during a tool review with @Adel-Moumen).


Adel-Moumen (approved):
LGTM

PR #1841, opened 2023-02-17
Fix InputNormalization device

lucadellalib:
@anautsch

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -1115,7 +1115,7 @@ class InputNormalization(torch.nn.Module):
 
                     self.count = self.count + 1
 
-                x = (x - self.glob_mean.data) / (self.glob_std.data)
+                x = (x - self.glob_mean.data.to(x)) / (self.glob_std.data.to(x))
 
         return x
 


anautsch:
Hi @lucadellalib as you mentioned offline, this happens during evaluation. In the logs I saw that the error occured right away after loading a checkpoint (to a line which had before the device information and is now changed to simplify things).

Passing a `.to()` for each batch, or each audio file during eval with bs=1, might be overheads. The hparam files usually consider the normalisation as part of the module, so they get put on the correct device automatically (using the `.to()` function once). It's though in line with your previous PR #1743 which has a strong opening point. Out-of-memory is another downside what needs to be considered.

cc @Adel-Moumen please take a quick look at https://github.com/speechbrain/speechbrain/pull/1743#issue-1475131881 – what's your take?

@lucadellalib is the `.to()` here necessary, also?
`x = (x - current_mean.data) / (current_std.data)`
https://github.com/lucadellalib/speechbrain/blob/db23b59a7327e78f0442689d24f1e456bec7ba52/speechbrain/processing/features.py#L1090

The recipe test for TIMIT_row_2 worked out well (some run instances take the same time as with the current develop version).

Adel-Moumen:
@lucadellalib are you getting a device error when you are only testing a model, i.e., no training? Because I faced this issue many times and it was because I was testing a model that wasn't trained first...

anautsch:
This PR is a follow-up to his earlier one—the recipe tests fail on our unstable branch, here's a fix. While testing it (it works) I started to wonder though if this way of putting the `.to()` is not causing too much device swapping. But hey, it's just a few floats (input norm stat values). In comparison to avoidable out-of-memory cases it's an advantage. Yet, I'd like to have a second opinion on this strategy.

Adel-Moumen:
I think the `.to()` is necessary (e.g., the issue that I have encounter on testing first). About overheads I don't know. But what I know is that without `.to()` when computing `x` you can encounter some problems when doing zero shot.

lucadellalib:
@anautsch, after the first batch `self.glob_mean` should be already on the correct device due to this line:
```
self.glob_mean = (1 - self.weight) * self.glob_mean.to(
    current_mean
) + self.weight * current_mean
```
Therefore, from the second batch on `to` should become a no-op (it just returns `self`, https://pytorch.org/docs/stable/generated/torch.Tensor.to.html).
Regarding `x = (x - current_mean.data) / (current_std.data)`, `current_mean` and `current_std` are computed based on current and previous `x` so they already on the same device as `x` (assuming that all the batches are on the same device, otherwise it would fail already at the previous line `current_mean = torch.mean(torch.stack(current_means), dim=0)`).

anautsch:
lgtm

Issue #1613, opened 2022-10-21
[Bug]:  causal=True in speechbrain.lobes.models.conv_tasnet.MaskNet results in an error


ahmad-aloradi:
### Description

Hi everyone, 

I am running into a seemingly simple bug when trying to run the causal version of [speechbrain.lobes.models.conv_tasnet.MaskNet](https://speechbrain.readthedocs.io/en/latest/API/speechbrain.lobes.models.conv_tasnet.html#speechbrain.lobes.models.conv_tasnet.MaskNet). Simply setting the 'causal' flag to True triggers this error: TypeError: bad operand type for unary -: 'str'. 

### Expected behaviour

I expect that simply changing the flag allows switching between the causal and acausal implementations. 

### To Reproduce

Here is a minimal example:

```
from speechbrain.lobes.models.conv_tasnet import MaskNet

hprams = {'N': 64, 'B': 32, 'H': 64, 'P': 3, 'X': 6, 'R': 3, 'C': 1}

acausal_net = MaskNet(**hparams, causal=False)
# -> defines network 

causal_net = MaskNet(**hparams, causal=True)

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ALABSAD/aloradad/anaconda3/envs/se/lib/python3.8/site-packages/speechbrain/lobes/models/conv_tasnet.py", line 238, in __init__
    self.temporal_conv_net = TemporalBlocksSequential(
  File "/home/ALABSAD/aloradad/anaconda3/envs/se/lib/python3.8/site-packages/speechbrain/lobes/models/conv_tasnet.py", line 160, in __init__
    self.append(
  File "/home/ALABSAD/aloradad/anaconda3/envs/se/lib/python3.8/site-packages/speechbrain/nnet/containers.py", line 112, in append
    layer = layer(*args, input_shape=input_shape, **kwargs)
  File "/home/ALABSAD/aloradad/anaconda3/envs/se/lib/python3.8/site-packages/speechbrain/lobes/models/conv_tasnet.py", line 346, in __init__
    self.layers.append(
  File "/home/ALABSAD/aloradad/anaconda3/envs/se/lib/python3.8/site-packages/speechbrain/nnet/containers.py", line 112, in append
    layer = layer(*args, input_shape=input_shape, **kwargs)
[...]

PR #1851, opened 2023-02-20
fixing the convtasnet causal=True bug

ycemsubakan:
This PR aims to fix the bug pointed out in https://github.com/speechbrain/speechbrain/issues/1613. 

I had to modify `Conv1d` class in `CNN.py`, but the default behavior is not changed. @Adel-Moumen please double check that everything is fine.


Adel-Moumen (approved):
LGTM!

Thanks @ycemsubakan, that solves the problem in question :-)

Issue #1696, opened 2022-11-10
[Bug]: The output_neurons dimension isn't consistent with SentencePiece vocabulary size when training a model using a character based tokenizer 


qanastek:
### Describe the bug

Dear maintainers,

My coworkers and I were confronted multiple time to a bug caused by an inconsistency between the dimension of the `output_neurons` and the vocabulary size of the SentencePiece character based tokenizer.

Let me explains by taking for example the recipe for `Wav2Vec + CTC` for CommonVoice.

If we are training an ASR model by setting up the tokenizer to produce a maximum of 100 characters by changing the parameter `model_type` to `char` and `vocab_size` to `100` in the SpeechBrain implementation of SentencePiece.

We will obtain an `NaN` loss after a few epochs if the real number of characters present in the training subset is higher than the `output_neurons` since the implementation of the SentencePiece present in SpeechBrain exclude the parameter `--vocab_size=<output_neurons>` in the case of the characters based model at the line 306. 

https://github.com/speechbrain/speechbrain/blob/11c70de96b3cc33ccd31fad7e470297e5c0ca69f/speechbrain/tokenizers/SentencePiece.py#L279-L314

If I change in this recipe, the `output_neurons` to very low number like `15` and the `model_type` to `char`. I will technically train a model with only 15 outputs neurons, but my tokenizer will encode and give me as inputs the documents with `68` characters. Which will progressively conduct to a `NaN` loss since the model try to produce unknown characters.

### Expected behaviour

I expect the character based tokenizer to limit the number of characters to the given `vocab_size`.

### To Reproduce

1. Take the Wav2Vec + CTC recipe for CommonVoice [here](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/CTC/hparams/train_fr_with_wav2vec.yaml)
2. Just replace `output_neurons` by a small number like `15` and `token_type` to `char`.
3. Enjoy the `NaN` loss after 7 to 10 epochs.

### Versions

Up to date with the `default` branch.

### Relevant log output

```shell
Image below in "Additional context".
```

### Additional context

![image (6)](https://user-images.githubusercontent.com/19389475/201071743-2f7eb523-cb36-4c45-aef3-1ae02bfd9196.png)

TParcollet:
[...]

Issue #1589, opened 2022-10-03
Incorrect log_probs dim causes a confusing error about NaN losses on CTC loss/search


AsuMagic:
## Summary

Accidentally undersizing the output vector for a CTC-based model (and possibly others) may produce a confusing NaN loss error after a couple of batches, which is not helpful in finding the root cause.  
PyTorch's `detect_anomaly` only points as NaNs being found in `log_softmax`, which does not help either.

Ideally, SB could instead detect this issue and produce a more helpful error/warning/hint when this occurs.

## Context

(This is mostly skippable, but should help making the issue more googleable and to understand the developer experience issue)

While building an ASR recipe based on the LibriSpeech wav2vec2-based CTC recipe, I have encountered a surprising issue where my training would crash due to a NaN loss just after a couple of batches.  
After some shotgun debugging, I found out that I had completely overlooked **the `output_neurons` hyperparameter, which was set way too low for my dataset**, which contains many more output tokens than LibriSpeech. This apparently didn't result in any meaningful warning or error.  
Incidentally, it seems that a colleague has run into the exact same issue, but that has not been confirmed yet.

## Possible fix

My suggestion is that all relevant loss functions in `nnet.losses` (CTC loss in my case) could include sanity checks to ensure that the prediction tensor is large enough to accept all possible targets. The same could be applied to `ctc_greedy_decode` (which is only used at inference time), and possibly to other functions in SB that could exhibit the same issue that are to be investigated.

That being said, I have some concerns around this but I don't have enough knowledge right now to know whether they're an issue in practice, or to know what the best course of action is:
- The parameters to `ctc_loss` do not seem to help knowing how many tokens should exist. It could be possible to check for the `max` of `targets` but that's a little silly and means we don't catch the issue as early as possible.
- I am not sure whether adding such a check could cause unintended side effects. I cannot think of one right now, though.

TParcollet:
The real issue is that in SB, we do not require the user to identify a layer as an output layer. A quick solution, imho, would be to add a really visible log output explicitons the size of the vocabulary. This error is very very very common among our community. 

An other possibility would be to add a check in our SentencePiece interface. We could ask for a mandatory argument "vocab_size" and check if what we produce actually is of this size. Imho this would be even better.

Adel-Moumen:
@AsuMagic are you willing to do a PR with the solution mentioned by @TParcolled, please? That would be awesome! Thanks  :-)

AsuMagic:
[...]

PR #1852, opened 2023-02-20
Allow specifying expected encoder category count

AsuMagic:
This introduces new user methods to `CategoricalEncoder`: `expect_len` and `ignore_len`. When neither of these are called, a `UserWarning` will be emitted once, informing the user that it is recommended to explicitly provide the category count they expect.

`expect_len` can be used to specify the expected final length of the encoder. When a mismatch is observed between that value and the real observed `len()` of the encoder during encoding or decoding, an exception is raised.

Solves #1589 and #1696 as `CTCTextEncoder` is a subclass of `CategoricalEncoder` (through `TextEncoder`).

TODOs:

- [x] Implement tests for the feature (alongside other tests in `test_categorical_encoder`)
- [x] ~~Modify existing recipes to eliminate warnings~~ See comments
- [x] Modify existing tests to eliminate warnings
- [x] Modify existing doctests to eliminate warnings
- [x] Add examples/doctests to the docs

### DIFF
--- a/speechbrain/dataio/encoder.py
+++ b/speechbrain/dataio/encoder.py
@@ -69,6 +69,7 @@ class CategoricalEncoder:
     >>> from speechbrain.dataio.dataset import DynamicItemDataset
     >>> dataset = [[x+1, x+2] for x in range(20)]
     >>> encoder = CategoricalEncoder()
+    >>> encoder.ignore_len()
     >>> encoder.update_from_iterable(dataset, sequence_input=True)
     >>> assert len(encoder) == 21 # there are only 21 unique elements 1-21
 
@@ -636,6 +637,7 @@ class CategoricalEncoder:
         >>> # So the first time you run the experiment, the encoding is created.
         >>> # However, later, the encoding exists:
         >>> encoder = CategoricalEncoder()
+        >>> encoder.expect_len(4)
         >>> if not encoder.load_if_possible(encoding_file):
         ...     assert False  # We won't get here!
         >>> encoder.decode_ndim(range(4))
@@ -677,7 +679,18 @@ class CategoricalEncoder:
             The expected final category count, i.e. `len(encoder)`.
 
         Example
-        -------"""
+        -------
+        >>> encoder = CategoricalEncoder()
+        >>> encoder.update_from_iterable("abcd")
+        >>> encoder.expect_len(3)
+        >>> encoder.encode_label("a")
+        Traceback (most recent call last):
+          ...
+        RuntimeError: .expect_len(3) was called, but 4 categories found
+        >>> encoder.expect_len(4)
+        >>> encoder.encode_label("a")
+        0
+        """
         # TODO
         self.expected_len = expected_len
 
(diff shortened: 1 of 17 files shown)


AsuMagic:
Note that, in this current state, the expected category count is not saved (because it often gets loaded when used in training anyway).

When using saved models, e.g. `classifier = EncoderClassifier.from_hparams(source="speechbrain/spkrec-ecapa-voxceleb")`, this means that a warning may be emitted because of `classifier.hparams.label_encoder` having an unspecified length.  
This also means inhibiting the warning looks a bit clunky (e.g. `classifier.hparams.label_encoder.ignore_len()`/`expect_len()`)

I am not sure if anything can really be done about this while preserving the desired properties. It's only a warning at worst, though, which is not very significant.

AsuMagic:
Following a discussion with @anautsch, I am thinking this could be reviewed & merged despite recipes not being updated. Feel free to share thoughts if needed.

Some recipe tests seem to be failing, which I've been told is to be fixed in `0.6`, so it would probably be best to wait for these to be all green before updating them all to remove the newly added warning (so that work would be postponed to a future PR).

This is a legacy-preserving change because no interfaces have been modified. Further, this is **not** a breaking change, so existing recipes are still functional and will not fail. A warning will merely be shown, indicating that it is recommended to specify the expected categorical encoder size. Said warning also indicates the category count (i.e. vocabulary size often enough) as observed at the time of the warning.

anautsch:
We could start with SpeechBrain linters, bc the underlying idea seems like one ;-)
No clue how much effort that is, though.

lmk when it's ready for review.

(side note: I get a bit 'sensitive'/alert when I read «force-pushed»)

@TParcollet @Adel-Moumen  how about throwing straight errors instead of warnings?

AsuMagic:
> lmk when it's ready for review.

@anautsch it is as of now. Existing tests (regular+doctest) have been modified to adapt to the recommended check and do not raise warnings anymore. A basic test was added to check that `ensure_len` is functioning properly.

> (side note: I get a bit 'sensitive'/alert when I read «force-pushed»)

I probably shouldn't have, this was to keep commits tidy as I had forgotten to cover one test :P

speechbrain/dataio/encoder.py
```diff
@@ -655,6 +662,74 @@ def load_if_possible(self, path, end_of_epoch=False, device=None):
...
+        >>> encoder.expect_len(3)
+        >>> encoder.encode_label("a")
+        Traceback (most recent call last):
+          ...
+        RuntimeError: .expect_len(3) was called, but 4 categories found
+        >>> encoder.expect_len(4)
+        >>> encoder.encode_label("a")
+        0
+        """
+        # TODO
```
Adel-Moumen, line 694: ?

AsuMagic: I think this is just a forgotten comment which can be removed.
```suggestion
```

Adel-Moumen (approved):
LGTM! 

Tried it and run the recipe/integration tests. Everything works as expected.

PR #1861, opened 2023-02-24
Rewrite multi-GPU documentation

### LINKED ISSUE #1860: [Feature Request]: multi-GPU docs should be improved/updated
### 🚀The feature

See:

- https://speechbrain.readthedocs.io/en/latest/multigpu.html
- https://github.com/speechbrain/speechbrain/blob/develop/docs/multigpu.md

There are several issues:

- SpeechBrain recommends against DataParallel, but none of that is mentioned in this document (it instead seems like a recommended option, if not more than DDP).
- It is not immediately clear _what_ one needs to ensure to make their recipe DDP-safe.
- "For further information, please see our multi-gpu tutorial: amazing multi-gpu tutorial" -> either the tutorial does not exist/is the page itself (I don't recall seeing one?) or the link was forgotten.
- The page mentions that when issues are being encountered with `nccl`, the `gloo` backend can be tested instead. ~~However, PyTorch docs seem to indicate that `gloo` simply cannot be used with CUDA builds of PyTorch.~~ It is only recommended for CPU training. Hence, that recommendation seems unnecessary.
- ... Hence providing the backend in the commandline seems unnecessary ~~(?)~~
- "The future version of pytorch will improve the stability of DDP." -> I am not sure what this refers to, but this is most likely outdated?
- The page could be laid out slightly better for navigability - in particular, "Slurm" could be put into its own section, the single-node multi-GPU DDP case could be put in a separate subsection from multi-node multi-GPU, etc.

Some points could benefit from more details.

### Solution outline

The above issues should be addressed.

### Additional context

I could work on making a PR to update the document so I assigned myself.

AsuMagic:
Closes #1860. The changes to the document are quite in-depth and there is new added info, so this PR needs to be fact-checked aside of proofreading.

The resulting compiled document looks as intended.

## Overview of the changes

- The document was generally rewritten to be better laid out and is significantly more detailed.
- DP is explicitly recommended against and is noted as deprecated, which seems to be the general stance of SB core contributors. (I made it go as far as saying SB does not support it anymore, considering we generally always tell DP users to use DDP instead when they face DP-related bugs)
- I tried to make the phrasing much more explicit in what DDP enables and how multi-GPU training works.
- "Writing DDP-safe code" gets its own, more detailed subsection.

## Additional notes

The Colab tutorial (which was brought up in the issue [here](https://github.com/speechbrain/speechbrain/issues/1860#issuecomment-1442031524)) is not linked in the document as of this PR. It should probably be updated to the contents of this PR, but as unless it actually gets any different content from the documentation, it probably doesn't need to be linked from the docs.

@anautsch (in DMs) mentioned [DDP tests](https://github.com/speechbrain/speechbrain/blob/develop/tests/integration/sampling/example_sorting.py) and some changes from #1817 that indirectly affect DDP. This PR mentions neither at the time, but perhaps someone more qualified than me could enhance the DDP-safety subsection further.

### DIFF
--- a/docs/multigpu.md
+++ b/docs/multigpu.md
@@ -36,7 +36,7 @@ python -m torch.distributed.launch --nproc_per_node=4 experiment.py hyperparams.
 
 _This covers the case where you want to split training across **multiple machines** on a network, with any amount of GPUs per machine._
 
-Note that using DDP across multiple machines introduces a **communication overhead** that might slow down training significantly, sometimes more than if you were to train on a single node! This largely depends on the network speed between the nodes.  
+Note that using DDP across multiple machines introduces a **communication overhead** that might slow down training significantly, sometimes more than if you were to train on a single node! This largely depends on the network speed between the nodes.
 Make sure you are actually observing any benefits from distributing the work across machines.
 
 While DDP is more efficient than `DataParallel`, it is somewhat prone to exhibit unexpected bugs. DDP is quite server-dependent, so some setups may face issues. If you are encountering problems, make sure PyTorch is well up to date.
--- a/docs/multigpu.md
+++ b/docs/multigpu.md
@@ -43,18 +43,29 @@ While DDP is more efficient than `DataParallel`, it is somewhat prone to exhibit
 
 #### Basics & manual multi-node setup
 
-Let's start with a simple example where a user is able to connect to each node directly. If we want to run 2 GPUs on 2 different nodes (for a total of 4 GPUs), we must do:
+Let's start with a simple example where a user is able to connect to each node directly. Consider that we have 2 nodes with 2 GPUs each (for a total of 4 GPUs).
+
+We use `torch.distributed.launch` once on each machine, with the following parameters:
+
+- `--nproc_per_node=2` means we will spawn 2 processes per node, which equates to 2 GPUs per nodes.
+- `--nnodes=2` means we will be using two nodes in total.
+- `--node_rank=0` and `--node_rank=1` refer to the rank/"index" we are attributing to the node/machine.
+- `--master_addr`/`--master_port` define the IP address and the port of the "master" machine. In this case, we're arbitrarily choosing the first machine to be the "master" of everyone else (the 2nd machine in our case). Note that `5555` might be taken by a different process if you are unlucky, so you may need to choose a different free port.
+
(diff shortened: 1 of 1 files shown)


AsuMagic:
Ugh, pre-commit checks fail to intended trailing whitespaces in the markdown. I'll remove them for now.

docs/multigpu.md
```diff
@@ -1,62 +1,83 @@
...
+Note that using DDP across multiple machines introduces a **communication overhead** that might slow down training significantly, sometimes more than if you were to train on a single node! This largely depends on the network speed between the nodes.
+Make sure you are actually observing any benefits from distributing the work across machines.
+
+While DDP is more efficient than `DataParallel`, it is somewhat prone to exhibit unexpected bugs. DDP is quite server-dependent, so some setups may face issues. If you are encountering problems, make sure PyTorch is well up to date.
 
-Running DDP over multiple servers (nodes) is quite system dependent. Let's start with a simple example where a user is able to connect to each node directly. If we want to run 2 GPUs on 2 different nodes (i.e total of 4 GPUs), we must do:
+#### Basics & manual multi-node setup
 
-```shell
+Let's start with a simple example where a user is able to connect to each node directly. If we want to run 2 GPUs on 2 different nodes (for a total of 4 GPUs), we must do:
```
Adel-Moumen, line 46: I think it would be great to remind the user that `--nproc_per_node=2` will expect 2 GPUs, hence, having 2 nodes will lead to 4 GPUs. I mean, you already said it but I think we should be more specific on that

docs/multigpu.md
```diff
@@ -1,62 +1,83 @@
...
+
+```bash
 # Machine 1
 cd recipes/<dataset>/<task>/
-python -m torch.distributed.launch --nproc_per_node=2 --nnodes=2 --node_rank=0 --master_addr machine_1_adress --master_port 5555 experiment.py hyperparams.yaml --distributed_launch --distributed_backend='nccl'
+python -m torch.distributed.launch --nproc_per_node=2 --nnodes=2 --node_rank=0 --master_addr machine_1_adress --master_port 5555 experiment.py hyperparams.yaml --distributed_launch
```
Adel-Moumen, line 51: Maybe we should also add a line on the `master_port` by saying "blblblbl you need to have a different address if launching multiple .sh on a same machine with multiple GPUs to avoid issues`

Adel-Moumen:
Thanks @AsuMagic!

Adel-Moumen (approved):
LGTM! I really like your changes. I think it will help a lot of ppl! :-)

PR #1864, opened 2023-02-25
Bayesian ASR

lucadellalib:
Add Bayesian transformer for LibriSpeech following https://arxiv.org/abs/2301.11276


lucadellalib:
@pplantinga a draft PR for BayesSpeech, if you have time take a look

recipes/LibriSpeech/ASR/transformer/train_bayesspeech.py
```diff
@@ -0,0 +1,565 @@
...
+            if should_step:
+                if self.check_gradients(loss):
+                    self.optimizer.step()
+                self.optimizer.zero_grad()
+                self.optimizer_step += 1
+
+                # anneal lr every update
+                self.hparams.noam_annealing(self.optimizer)
+
+        return loss.detach().cpu()
```
pplantinga, line 190: I'm curious, is the only reason for reproducing the `fit_batch()` here in order to add the `noam_annealing()`? This could better be done with the new function `on_fit_batch_end()`. If this code was inherited from another recipe, that one should eventually get updated too.

recipes/LibriSpeech/ASR/transformer/train_bayesspeech.py
```diff
@@ -0,0 +1,565 @@
...
+> python train_bayesspeech.py hparams/transformer_bayesspeech.yaml
+
+With the default hyperparameters, the system employs a convolutional frontend and a transformer.
+The decoder is based on a Transformer decoder. Beamsearch coupled with a Transformer
+language model is used  on the top of decoder probabilities.
+
+The neural network is trained on both CTC and negative-log likelihood
+targets and sub-word units estimated with Byte Pairwise Encoding (BPE)
+are used as basic recognition tokens. Training is performed on the full
+LibriSpeech dataset (960 h).
```
pplantinga, line 17: Should this add something about KL divergence?

recipes/LibriSpeech/ASR/transformer/README.md
```diff
@@ -12,10 +12,10 @@ python train.py hparams/transformer.yaml
 installed in your environment (see extra-requirements.txt)**
 # Results
 
-| Release | hyperparams file | Test Clean WER | HuggingFace link | Model link | GPUs |
-|:-------------:|:---------------------------:| :-----:| :-----:| :-----:| :--------:|
+| Release  | hyperparams file | Test Clean WER | HuggingFace link | Model link | GPUs |
+|:--------:|:---------------------------:| :-----:| :-----:| :-----:| :--------:|
 | 24-03-22 | transformer.yaml | 2.26 | [HuggingFace](https://huggingface.co/speechbrain/asr-transformer-transformerlm-librispeech) | [GoogleDrive](https://drive.google.com/drive/folders/1sM3_PksmGQZMxXPibp7W7mQfPXFdHqc5?usp=sharing) | 1xA100 40GB |
-| 06-12-23 | train_hf_whisper.yaml | 3.60 | Not Avail. | Not Avail. | 1xA100 40GB |
+| 06-12-22 | train_hf_whisper.yaml | 3.60 | Not Avail. | Not Avail. | 1xA100 40GB |
```
pplantinga, line 18: This seems unrelated to bayesian asr. Also, are there results available for bayesian ASR?

pplantinga (commented):
This looks like a nice add to the toolkit, thank you! I'm curious, were you able to verify any of the claims in 2301.11276, such as faster training time or better-calibrated posterior?

Other than some sort of results showing the benefit of this recipe, the only thing remaining is for someone to try it and ensure it works (I haven't had time for this yet).

recipes/LibriSpeech/ASR/transformer/train_bayesspeech.py
lucadellalib: It was indeed copied from the `train.py` script in the same directory. I guess the reason is that we want to anneal the learning rate at each step, but only if the `should_step` flag is `True`.

lucadellalib: In theory it should help neither with training time nor convergence speed, since wall-clock time for each epoch is definitely larger (2x parameters for each Bayesian layer, requires sampling for each forward pass, etc.) and we are over-regularizing (LibriSpeech has plenty of data while Bayesian neural networks are good in low data regimes).

lucadellalib: I trained both Bayesian transformer and standard one for 20 epochs with same batch size, this is the log:
[bayesspeech_vs_transformer.txt](https://github.com/speechbrain/speechbrain/files/10974121/bayesspeech_vs_transformer.txt)

pplantinga: That's a neat chart, looks like Bayes Speech has better validation accuracy until ~epoch 7 or 8. I wonder how this might work alongside hyperparameter search or neural architecture search to reduce computational budget... Food for thought anyway!

pplantinga (approved):
I ran librispeech-100 and got an accuracy of 83.3% and WER of 17.94 on dev-clean. I assume  hparams more tuned to this subset (longer training?) would result in a better score but it seems like its working at least. Any extra feedback before merging @TParcollet @mravanelli ?

Adel-Moumen:
Would it be possible to have access to a checkpoint (in the google drive) as well as a HuggingFace Hub model? We should make sure that training for longer (i.e., more epochs on full LS 960h) is not giving wrong results.

Adel-Moumen:
Hello @lucadellalib, any news regarding this PR please? Thanks :-)

lucadellalib:
@Adel-Moumen https://www.dropbox.com/sh/k8t08hy8j12968s/AACIzS83iBy_hFamluE-QsyWa?dl=0

PR #1872, opened 2023-03-08
minor fixes for recipe testing

anautsch:
As @Adel-Moumen pointed out in the 751 PR, there are some open tasks with the recipe testing. Here are some fixes. Added a missing extra requirements also.

The REAL-M recipe fix touches on its inner logic. Two nested flags are used to create a specfic field, if both are true - in this testing example, however, only one of them is and so, the recipe test fails for a later component expecting that specific field which never would be created, since one of the flags is False. That's addressed now with adding the least changes to the recipe.

Please merge first to develop & then fetch/merge the develop into pending PRs (+unstable branch).

cc @TParcollet


speechbrain/lobes/models/huggingface_wav2vec.py
```diff
@@ -457,7 +457,7 @@ def forward(self, wav, wav_lens):
         torch_mask_time_indices = torch.tensor(
             mask_time_indices, device=wav.device, dtype=torch.long,
         )
-        padding_mask = self.make_masks(wav, wav_len=wav_lens)
+        padding_mask = self.make_padding_masks(wav, wav_len=wav_lens)
```
Adel-Moumen, line 460: why? It seems to be the same for me.. not?

anautsch: different classes but the same function has a different name in each class

Adel-Moumen (approved):
LGTM!

Issue #1749, opened 2022-12-07
[Feature Request]: isort for sorting import statements


lucadellalib:
### 🚀The feature

A standard good practice in Python is to group import statements in 3 sections separated by one blank line (see [https://peps.python.org/pep-0008/#imports](https://peps.python.org/pep-0008/#imports)):
1. Standard library imports
2. Third party imports
3. Local library imports
Furthermore, imports within each group should be sorted in lexicographically, with `import package`-style statements followed by `from package import ...`-style statements. These are the conventions followed by most top Python repositories (e.g. Pandas, TensorFlow, etc.). We should do the same in SpeechBrain to improve readability.

### Solution outline

`isort` ([https://github.com/PyCQA/isort](https://github.com/PyCQA/isort)) is a tool similar to `black` that automatically sorts import statements in all Python files. Just like `black` it can be customized and run as a `pre-commit` hook. Integration is straightforward and usually doesn't cause any compatibility issue. We should consider adding it to SpeechBrain.

anautsch:
I like the idea, thanks for suggesting @lucadellalib ! 

There will be cases where manual sorting is needed though (@Adel-Moumen we discussed on the beamsearch refactoring). Referencing #1353 here, as we should put this one in too when we polish up our housekeeping for the next version.

PR #1873, opened 2023-03-08
Linters updates (housekeeping)

anautsch:
This PR accompanies #1353.

When running
```shell
pre-commit run --all-files
```
in a fresh environment (e.g. Pyhon 3.8), the log will look like this:
```python
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/PyCQA/flake8.
[INFO] Initializing environment for https://github.com/adrienverge/yamllint.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/PyCQA/flake8.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/adrienverge/yamllint.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
fix requirements.txt.....................................................Failed
- hook id: requirements-txt-fixer
- exit code: 1
- files were modified by this hook

Sorting docs/docs-requirements.txt

mixed line ending........................................................Passed
check for added large files..............................................Passed
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted recipes/AISHELL-1/Tokenizer/pretrained.py
...
reformatted tests/utils/recipe_tests.py

All done! ✨ 🍰 ✨
264 files reformatted, 115 files left unchanged.

flake8...................................................................Failed
- hook id: flake8
- exit code: 1

recipes/Aishell1Mix/separation/scripts/create_aishell1mix_from_metadata.py:313:27: F541 f-string is missing placeholders
...
speechbrain/lobes/models/HifiGAN.py:650:13: E741 ambiguous variable name 'l'

yamllint.................................................................Passed
```

As this line indicates
```python
fix requirements.txt.....................................................Failed
```
the linters checks for this PR itself & its direct changes is not the goal at its opening. This PR is not to loose sight of that requirements will need updating when it comes to v0.6 (at some later point; hence its merging to the `unstable-v0.6` branch).

When merging this one alongside with cspell, the latest versions need to be agreed upon. As one can readily identify from this PR's file changes, there are quite some version gaps which can be recovered ;-)

Recommendation to keep the `unstable-v0.6` branch as clean as possible:
* get everything that needs to be on unstable-v0.6 on it
* cspell PR: move to latest version & check proposed file changes
* this PR: final reformattings -> please feel free to add into this PR branch directly

My gut feeling is that this will happen after April.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.5.0  # Use the ref you want to point at
+    rev: v4.4.0  # Use the ref you want to point at
     hooks:
       - id: trailing-whitespace
         types: [file, text]
@@ -14,18 +14,17 @@ repos:
         args: [--maxkb=1024]
 
   - repo: https://github.com/psf/black
-    rev: 19.10b0
+    rev: 23.1.0
     hooks:
       - id: black
         types: [python]
-        additional_dependencies: ['click==8.0.4']
   - repo: https://github.com/PyCQA/flake8
-    rev: 3.7.9
+    rev: 6.0.0
     hooks:
       - id: flake8
         types: [python]
 
   - repo: https://github.com/adrienverge/yamllint
-    rev: v1.23.0
+    rev: v1.29.0
     hooks:
       - id: yamllint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,10 +18,17 @@ repos:
     hooks:
       - id: black
         types: [python]
+
   - repo: https://github.com/PyCQA/flake8
     rev: 6.0.0
     hooks:
       - id: flake8
+        types: [python]
+
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.12.0
+    hooks:
+      - id: isort
         types: [python]
 
   - repo: https://github.com/adrienverge/yamllint
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,5 +1,6 @@
 black==23.1.0
 flake8==6.0.0
+isort==5.12.0
 pycodestyle==2.10.0
 pytest==7.2.2
 yamllint==1.29.0
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,6 +1,5 @@
-black==19.10b0
-click==8.0.4
-flake8==3.7.9
-pycodestyle==2.5.0
-pytest==5.4.1
-yamllint==1.23.0
+black==23.1.0
+flake8==6.0.0
+pycodestyle==2.10.0
+pytest==7.2.2
+yamllint==1.29.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
 -r lint-requirements.txt
+cchardet
+chardet
 huggingface_hub>=0.12.1
 hyperpyyaml>=1.1.0
 joblib>=1.2.0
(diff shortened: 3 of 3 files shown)


anautsch:
Interesting... from the [failing] Pyton 3.7 log
```python
ERROR: Ignored the following versions that require a different python version: 6.0.0 Requires-Python >=3.8.1
```

i.e. if SpeechBrain opts for latest dependencies when migrating to its v0.6, then Python 3.7 support might need to be dropped, also.

Note: the pre-commits/linters fail as expected. (Their versions are updated here, but the code base remained unaltered.)

anautsch:
We discussed Python versions at some earlier point with @AsuMagic - revisiting the end-of-life cycles:
https://devguide.python.org/versions/

and PyTorch's 2/1/2023 notice:
https://pytorch.org/blog/deprecation-cuda-python-support/

The table provided by PyTorch seems reasonable. SpeechBrain might limit this further to 1.13 & 2+ support for PyTorch. Python 3.10 has end-of-life 2026-10. Yet, PyTorch 2+ also supports Python 3.11. I don't see SB being able to maintain multiple branches, i.e., a decision will have to be made.

AsuMagic:
> I don't see SB being able to maintain multiple branches, i.e., a decision will have to be made.

What would the multiple branches cover? Shouldn't PyTorch 2.0 should be largely backwards compatible with 1.13? And would there be any issue in requiring Python 3.8+ and PyTorch 1.13+?

anautsch:
Hey @AsuMagic !

> > I don't see SB being able to maintain multiple branches, i.e., a decision will have to be made.
> 
> What would the multiple branches cover? 

I was wondering if 
  (a) older python/dependency versions would require slight code changes => multiple `develop` branches could be a consequence
  (b) newer python versions are becoming the norm (e.g. 3.8+ over 3.7), some new coding elements will be introduced that over time are not backward compatible (since they were introduced in 3.8 or in 3.9 or ...), and as such, multiple `develop` branches could be again a consequence.

Hence, should SpeechBrain drop python 3.7 support when moving to the next major version?

> Shouldn't PyTorch 2.0 should be largely backwards compatible with 1.13? 

That's a part of the problem space, but outside of the above concern for what SB should be "officially" supporting. This is about the Python versions, not about PyTorch. The new PyTorch version just makes it more urgent or puts the topic directly on the table.

Earlier versions than PyTorch 1.13 had also their coding styles and subpackages changing (especially their profiler had quite some changes over the last 3.10-3.13 minor versions). Offering support for all would require to be compatible with all different subpackage structures etc. - a clear cut would make things easier on our side to maintain SB.

> And would there be any issue in requiring Python 3.8+ and PyTorch 1.13+?

Not at all. That's what I'd argue for.

Some users might have their envs set - so changing Python version requirement with a minor SB version release might be a bit harsh ... with a new SB major version, moving on with updated dependencies would be another level.

AsuMagic:
> Hence, should SpeechBrain drop python 3.7 support when moving to the next major version?

IMO that would be okay. The entire ecosystem is eventually going to drop 3.7 anyway, if PyTorch can afford to do it I would not worry about it too much.

anautsch:
added `isort`, see #1749

PR #1874, opened 2023-03-09
add preprocess in speechbrain/vad-crdnn-libriparty

HildaGuAppen:
make compatible with multi-channel audio file, or different sample rate.

> As the system expects input recordings sampled at 16kHz (single channel). If your signal has a different sample rate, resample it (e.g., using torchaudio or sox) before using the interface. 

1. Add a parameter `use_channel` to choose which channel to use.
2. If origin audio samples unequals 16kHz, resample by kaiser window.
3. Save the single channel signal to local, continue.


henry-wang-ap:
update

PR #1876, opened 2023-03-10
adding pretrainer in hyperparameter file LJSPEECH/TTS recipe

imen185:
Adding pretrainer in the hyperparameter file for both Tacotron2 and the vocoder in order to be able to do the inference with a local trained model since for the TTS task the best way to evaluate a model is listening.


BenoitWang:
Hi @imen185 , usually, we do not put the pretrainer in the training scripts. It is only added in the yaml file for inference or when a pre-trained model is needed, which is not the case for this recipe.

imen185:
OK, I get it, thank you. I'll close the PR.

PR #1882, opened 2023-03-14
Fix Whisper `avoid_if_longer_than` never used

Adel-Moumen:


fpaissan (approved):
Tested. Seems okay now

Issue #1883, opened 2023-03-15
[Bug]: pad_right_to in data_utils.py does not check ndim correctly


anthony-wss:
### Describe the bug

Hi,
When I am training the voxlingua107 recipe with my own data, I found out that in line 418 of `speechbrain/utils/data_utils.py`:

```python
if not (
    any(
        [tensors[i].ndim == tensors[0].ndim for i in range(1, len(tensors))]
    )
):
    raise IndexError("All tensors must have same number of dimensions")
```

I suppose that this code should raise an error when the `ndim` of the tensor in `tensors` list is not the same, but it should be `all` instead of `any` in order to go into the if statement if some tensors have wrong `ndim`?

Thanks!

### Expected behaviour

Raise an exception when the dimension of `tensors` is not the same.

PR #1884, opened 2023-03-15
fix batch_to_right

anthony-wss:
#1883 

This is the PR related to the issue "[Bug]: pad_right_to in data_utils.py does not check ndim correctly".

Thanks!


TParcollet:
@Gastron Can you take a look?

Gastron:
Yes this looks like a straight-forward mistake. Weird that it wasn't caught before. Thank you for fixing this.

Gastron (approved):

Issue #1886, opened 2023-03-15
[Bug]: torch version check in STFT


grazder:
### Describe the bug

Here is check:

https://github.com/speechbrain/speechbrain/blob/beec862238c4c2ccfcb20d9ca592b671673ee6af/speechbrain/processing/features.py#L148

I don't understand the meaning of this because you have `torch>=1.9` in the requirements

https://github.com/speechbrain/speechbrain/blob/beec862238c4c2ccfcb20d9ca592b671673ee6af/setup.py#L40

So this if seems useless

### Expected behaviour

I expected that there would be no such check, because defined to support `torch>=1.9.0`

### To Reproduce

Plus it is became a problem for scripting Fbank (my local error):
```
torch.jit.frontend.UnsupportedNodeError: try blocks aren't supported:
  File "/opt/conda/lib/python3.9/site-packages/packaging/version.py", line 48
    a valid PEP 440 version or a legacy version.
    """
    try:
    ~~~ <--- HERE
        return Version(version)
    except InvalidVersion:
```
because version is using try-catch

and error in colab:
```
import torch
from speechbrain.lobes.features import Fbank

fbank = Fbank(sample_rate=8000, n_fft=512, n_mels=64, left_frames=0, right_frames=0, deltas=False)
torch.jit.script(fbank)
```
```
RuntimeError: 
'__torch__.packaging.version.Version (of Python compilation unit at: 0x6dd2a30)' object has no attribute or method '_regex'. '_regex' is defined as a class attribute which currently is not supported. Consider converting this to an instance attribute.:
  File "/usr/local/lib/python3.9/dist-packages/packaging/version.py", line 195
    
        # Validate the version and parse it into pieces
        match = self._regex.search(version)
                ~~~~~~~~~~~ <--- HERE
        if not match:
            raise InvalidVersion(f"Invalid version: '{version}'")
'Version.__init__' is being compiled since it was called from '__torch__.packaging.version.Version'
  File "/usr/local/lib/python3.9/dist-packages/packaging/version.py", line 43
def parse(version: str) -> "Version":
                           ~~~~~~~~ <--- HERE
    """Parse the given version string.
'__torch__.packaging.version.Version' is being compiled since it was called from 'parse'
  File "/usr/local/lib/python3.9/dist-packages/packaging/version.py", line 43
def parse(version: str) -> "Version":
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    """Parse the given version string.
[...]

PR #1890, opened 2023-03-20
Fix for #1886

anautsch:
As pointed in #1886 it's an out-dated check.
https://github.com/speechbrain/speechbrain/blob/beec862238c4c2ccfcb20d9ca592b671673ee6af/speechbrain/processing/features.py#L148-L173

Dropped the if branch of the statement. 
Note: the then branch adds `return_complex=False,` to the parameters.

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -36,7 +36,6 @@ Authors
 import math
 import torch
 import logging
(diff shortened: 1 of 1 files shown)


Adel-Moumen (approved):
LGTM

Issue #1881, opened 2023-03-13
[Bug]: dynamic batching not behaving as expected with max_batch_ex


TParcollet:
### Describe the bug

When using dynamic batching and setting max_batch_ex to any number, we can still see batches created above this limit. E.g. max_batch_ex=64; the first bucket boundary will be a batch size of 80; when printing in compute forward we will effectively see a batch of 80.

### Expected behaviour

max_batch_ex, as explained in the documentation, should limit the number of samples per batch. Hence; if max_batch_ex=64, then the maximal batch size should be 64.

### To Reproduce

Any recipe with dynamic batching on

anautsch:
Might be a confusion about wording issues in the log message. Please let's check.

I went for this recipe:
```bash
recipes/LibriSpeech/ASR/transformer$ python train.py hparams/transformer.yaml --data_folder=
```

which has:
https://github.com/speechbrain/speechbrain/blob/beec862238c4c2ccfcb20d9ca592b671673ee6af/recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml#L61-L73

where `max_batch_len` is used for TRAIN & `max_batch_len_val` for VALID.

This is how the log spawns:
https://github.com/speechbrain/speechbrain/blob/beec862238c4c2ccfcb20d9ca592b671673ee6af/speechbrain/dataio/sampler.py#L633-L647

Let's look at the log for the latter (max ex: 100):
```
Generating dynamic batches
Bucket 0 with boundary 0.0-0.6 and batch_size 173: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
Bucket 1 with boundary 0.6-0.7 and batch_size 135: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
Bucket 2 with boundary 0.7-0.9 and batch_size 115: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
Bucket 3 with boundary 0.9-1.0 and batch_size 102: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
Bucket 4 with boundary 1.0-1.1 and batch_size 93: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
Bucket 5 with boundary 1.1-1.2 and batch_size 86: Num Examples 0.0, Num Full Batches 0.000, Pad Factor 0.000.
[...]

PR #1891, opened 2023-03-22
Fix for #1881

anautsch:
As it looks like, none of the recipes which referred to the dynamic batch sampler were able to make use of all parameters specified in the recipe's hparams file. The reason is that the depending class is not instantiated through hyperpyyaml but during the recipe.

Furthermore, "length" as a variable name seems inconclusive as there were two args (length as duration & length as number of examples). As a result, the `max_batch_len` can be confused with `max_batch_ex`. The former term has been renamed here to make clear, it is about the duration. This change breaks the interface of the dynamic batch sampler, which is relevant only during training, for only few recipes (five of them have dynamic batching on). If that's an issue -> we can put this PR also on the unstable-v0.6 branch.

To test, simply run:
```
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Hparam_file"], filters=[["recipes/TIMIT/ASR/seq2seq/hparams/train.yaml", "recipes/Switchboard/ASR/seq2seq/hparams/train_BPE_2000.yaml", "recipes/LibriSpeech/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml", "recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml", "recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_5000.yaml", "recipes/LibriSpeech/ASR/seq2seq/hparams/train_BPE_1000.yaml", "recipes/CommonVoice/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml", "recipes/AISHELL-1/ASR/transformer/hparams/train_ASR_transformer_with_wav2vect.yaml", "recipes/AISHELL-1/ASR/transformer/hparams/train_ASR_transformer.yaml", "recipes/AISHELL-1/ASR/seq2seq/hparams/train.yaml", "recipes/AISHELL-1/ASR/CTC/hparams/train_with_wav2vec.yaml", "recipes/Switchboard/ASR/CTC/hparams/train_with_wav2vec.yaml", "recipes/Switchboard/Tokenizer/hparams/2K_unigram_subword_bpe.yaml", "recipes/Switchboard/LM/hparams/transformer.yaml"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```
This includes extra test cases for one Switchboard recipe, which are dependency test cases (need to run before the one depending on this PR).

### DIFF
--- a/recipes/AISHELL-1/ASR/CTC/hparams/train_with_wav2vec.yaml
+++ b/recipes/AISHELL-1/ASR/CTC/hparams/train_with_wav2vec.yaml
@@ -43,11 +43,11 @@ test_batch_size: 4
 
 dynamic_batching: False
 dynamic_batch_sampler:
-   feats_hop_size: 0.01
-   max_batch_len: 15 # in terms of "duration" in annotations by default, second here
-   left_bucket_len: 200 # old implementation attributs
-   multiplier: 1.1 # old implementation attributs
-   shuffle_ex: False # if true re-creates batches at each epoch shuffling examples.
+   # feats_hop_size: 0.01
+   max_duration_per_batch: 15 # in terms of "duration" in annotations by default, second here
+   # left_bucket_len: 200 # old implementation attributs
+   # multiplier: 1.1 # old implementation attributs
+   shuffle: False # if true re-creates batches at each epoch shuffling examples.
    num_buckets: 10 # floor(log(max_batch_len/left_bucket_len, multiplier)) + 1
    batch_ordering: ascending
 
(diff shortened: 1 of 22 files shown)


anautsch:
cc @TParcollet on #1881

TParcollet:
@anautsch Let's put this on unstable for now. Can we simply delete what has been commented out?

anautsch:
@TParcollet unstable might need to fetch develop first. Just changed base. It's a lot.

Yes, commented out details can be deleted.

Adel-Moumen:
Closed in favour of #2170

PR #1898, opened 2023-03-24
fix core.py bf16

Adel-Moumen:


TParcollet (approved):
LGTM necessary as enabled still evaluates bfloat16 on GPU that can't do it... leading to errors

PR #1899, opened 2023-03-24
S2SGreedySearcher : Do not continue decoding when EOS token was generated for all samples from a batch

Jeronymous:
This can speed up things when validating a seq2seq model

### DIFF
--- a/speechbrain/decoders/seq2seq.py
+++ b/speechbrain/decoders/seq2seq.py
@@ -188,12 +188,16 @@ class S2SGreedySearcher(S2SBaseSearcher):
             0, max_decode_steps
         )
 
+        has_ended = enc_states.new_zeros(batch_size).bool()
         for t in range(max_decode_steps):
             log_probs, memory, _ = self.forward_step(
                 inp_tokens, memory, enc_states, enc_lens
             )
             log_probs_lst.append(log_probs)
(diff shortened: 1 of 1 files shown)


TParcollet:
@Adel-Moumen what do you think? It could be added to the minor?

Adel-Moumen (approved):
LGTM! Thanks for spotting/solving the bug!

TParcollet:
This PR will be shipped in the new release today, this was perfect timing ahah.

mravanelli:
Did you test that everything is running and providing the same performance
as before (at least in one case like librispeech)?

Adel-Moumen:
Yes @mravanelli, I conducted the experiment using Whisper on LibriSpeech.

I was previously aware of this bug and had attempted a somewhat similar approach that partially addressed the issue. The current fix fully resolves the problem by monitoring the batches that are reaching or have reached the eos.

mravanelli:
Perfect,
we should inherit this fix into the beamsearch refactor for the major
version.

Adel-Moumen:
Yep. 

Just so everyone knows about this change:

We went from 41 mins and 4 seconds with Whisper Tiny on LibriSpeech test-clean to 2 mins and 36 seconds. That's like a ~20x speed boost. Both of them led to the same WER of 7.50.

PR #1901, opened 2023-03-25
Whisper docstring example

Adel-Moumen:


anautsch:
Hi @Adel-Moumen why the « doctest: +SKIP » ?

anautsch (changes requested):
On skipped test. In an ealier PR, we dropped a doctest for it required jit & compilation (taking additional time). This way might be good to keep record of how to test what, even if tests are skipped on a regular basis - just so one knows how to start any debugging.

This one downloads model data of 12+ GB and requires a node to have that available as RAM. This is not the case for github workflows. It's better skipped for what is needed.

There's an error
```
>>> asr_model.transcribe_file("samples/audio_samples/example_fr.wav")
  File "huggingface_hub/utils/_errors.py", line 264, in hf_raise_for_status
    response.raise_for_status()
  File "requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/samples/audio_samples/resolve/main/example_fr.wav
```

This one works better (but depends on mp3 backend):
```
asr_model.transcribe_file("speechbrain/asr-whisper-large-v2-commonvoice-fr/example-fr.mp3")
```

The other issue is, we don't have a FR test sample in the repo:
```
>>> asr_model.transcribe_file("tests/samples/single-mic/example1.wav")
[['le', 'berchecanneu', 'slid', 'on', 'the', 'smooth', 'planks']]
```

Adel-Moumen:
> On skipped test. In an ealier PR, we dropped a doctest for it required jit & compilation (taking additional time). This way might be good to keep record of how to test what, even if tests are skipped on a regular basis - just so one knows how to start any debugging.
> 
> This one downloads model data of 12+ GB and requires a node to have that available as RAM. This is not the case for github workflows. It's better skipped for what is needed.
> 
> There's an error
> 
> ```
> >>> asr_model.transcribe_file("samples/audio_samples/example_fr.wav")
> Traceback (most recent call last):
>   File "/users/anautsch/.local/lib/python3.7/site-packages/huggingface_hub/utils/_errors.py", line 264, in hf_raise_for_status
>     response.raise_for_status()
>   File "/users/anautsch/.conda/envs/pY/lib/python3.7/site-packages/requests/models.py", line 1021, in raise_for_status
>     raise HTTPError(http_error_msg, response=self)
> requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/samples/audio_samples/resolve/main/example_fr.wav
> ```
> 
> This one works better (but depends on mp3 backend):
> 
> ```
> asr_model.transcribe_file("speechbrain/asr-whisper-large-v2-commonvoice-fr/example-fr.mp3")
> ```
> 
> The other issue is, we don't have a FR test sample in the repo:
> 
> ```
> >>> asr_model.transcribe_file("tests/samples/single-mic/example1.wav")
> [['le', 'berchecanneu', 'slid', 'on', 'the', 'smooth', 'planks']]
> ```

Hi, ty for the feedbacks.

We can switch to the mp3 file but as you said, it depends on the mp3 backend... 

I can add the example-fr.mp3 file in the test folder so that we can use it. What do you think?

Adel-Moumen:
The last commit switched to the mp3 file.

PR #1903, opened 2023-03-27
Fix unnecessary warning in train logger

pplantinga:
This code uses a deprecated call to `logger.warn` which is superseded by `logger.warning`


anautsch (approved):
lgtm

PR #1929, opened 2023-04-07
Fix broken CUDA check in TransducerLoss forward

### LINKED ISSUE #1928: [Bug]: Error while running the TransducerLoss module
### Describe the bug

Originally reported by #1867

`ValueError: Found inputs tensors to be on [device(type='cuda', index=0), device(type='cuda', index=0), device(type='cuda', index=0), device(type='cuda', index=0)] while needed to be on a 'cuda' device to use the transducer loss.`

### Expected behaviour

No error should occur.

### To Reproduce

The `TransducerLoss` doctest is sufficient to reproduce the issue.

```python
from speechbrain.nnet.loss.transducer_loss import TransducerLoss
import torch
loss = TransducerLoss(blank=0)
logits = torch.randn((1,2,3,5)).cuda().requires_grad_()
labels = torch.Tensor([[1,2]]).cuda().int()
act_length = torch.Tensor([2]).cuda().int()
label_length = torch.Tensor([2]).cuda().int()
l = loss(logits, labels, act_length, label_length)
```

### Versions

The offending line exists as of current `develop`.

### Relevant log output

_No response_

### Additional context

The offending line of code is the following: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/nnet/loss/transducer_loss.py#L343 (as of issue making time)

`git blame` points to https://github.com/speechbrain/speechbrain/commit/2c491a4662b1d7bb03f71c38a8f82d9faa766f33 cc @Adel-Moumen. I can look into it and make a PR.

The check is incorrect because `device` is not a `str` but some PyTorch object.

AsuMagic:
The previous check was incorrect because .device does not return a `str` but a PyTorch device object, hence the check was incorrect.

Fixes #1928; the provided repro now successfully runs on GPU and changing any of the tensors to CPU correctly errors out.

### DIFF
--- a/speechbrain/nnet/loss/transducer_loss.py
+++ b/speechbrain/nnet/loss/transducer_loss.py
@@ -340,7 +340,7 @@ class TransducerLoss(Module):
     def forward(self, logits, labels, T, U):
         """Computes the transducer loss."""
         # Transducer.apply function take log_probs tensor.
-        if logits.device == labels.device == T.device == U.device == "cuda":
+        if all(t.is_cuda for t in (logits, labels, T, U)):
             log_probs = logits.log_softmax(-1)
             return self.loss(
                 log_probs, labels, T, U, self.blank, self.reduction


Adel-Moumen (approved):
LGTM

PR #1934, opened 2023-04-11
Add `read_audio_info`, fix improper MP3 metadata read on CommonVoice/DVoice

### LINKED ISSUE #1925: [Bug]: CommonVoice sorting is likely broken with certain torchaudio versions
### Describe the bug

CommonVoice is a dataset in MP3 format. It seems that a recent `torchaudio` version (linked issue reports `0.12`, but my `0.11` environment seems to reproduce) more or less intentionally broke querying the length of MP3 files when using `torchaudio.info(path).num_frames`, as it seems to always return 0, see <https://github.com/pytorch/audio/issues/2524>.

It seems that a PR was merged at some point to solve that issue but the minimum torchaudio version we require is probably still affected.

SpeechBrain uses this to generated the CSVs. As a result, they contain a `0.0` duration column.

`train.py` uses that column to sort data.

### Expected behaviour

`duration` should never be zero.

### To Reproduce

Run any preprocessing step in a CommonVoice recipe, then inspect `results/<seed>/save/*.csv`.

### Versions

_No response_

### Relevant log output

Example contents of CSV:

```
ID,duration,wav,spk_id,wrd
common_voice_fr_19642221,0.0,/users/sdelangen/CommonVoice8.0/fr/clips/common_voice_fr_19642221.mp3,4e64587f8dd935d70e50607fe60f73b5badcd6e967db2c331a6c92ca00e9d1ddde6ff2385cd1a90c6d5ceea960e708248914dcb6cdc69fe6ea7edd0208a3ff5e,C EST LE PLUS VASTE ET LE PLUS PEUPLÉ DE LA COMMUNE
common_voice_fr_19642222,0.0,/users/sdelangen/CommonVoice8.0/fr/clips/common_voice_fr_19642222.mp3,4e64587f8dd935d70e50607fe60f73b5badcd6e967db2c331a6c92ca00e9d1ddde6ff2385cd1a90c6d5ceea960e708248914dcb6cdc69fe6ea7edd0208a3ff5e,CES SEPT QUATUORS ONT ÉTÉ ÉDITÉS PAR MARC LOMBARD
common_voice_fr_19642223,0.0,/users/sdelangen/CommonVoice8.0/fr/clips/common_voice_fr_19642223.mp3,
```

### Additional context

_No response_

AsuMagic:
Fixes #1925.

This PR introduces a `dataio` function `read_audio_info` which, for the most part, behaves identically to `torchaudio.info`. However, because certain `torchaudio` versions have a "bug" which makes it so that `num_frames` is not populated with MP3 (which was solved in very recent versions), this function additionally decodes the file to obtain the frame count.  
This is slow, but this is what newer torchaudio does anyway(? might be ffmpeg now), as there doesn't seem to be a perfectly consistent way to achieve this with MP3.

Data preparation was tested over for CommonVoice, but not for DVoice, which also uses MP3. However, because these steps are very similar, there should be no reason for it to fail.

A test was also added to verify whether it appears to function properly (and also does so for mp3+ogg+wav for good measure).

Recent versions of torchaudio offload MP3 handling to ffmpeg, so this PR also adds a ffmpeg install step for the CI (it does not touch requirements files and such, as it is often system provided). This could be split to a different PR if required.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -24,6 +24,9 @@ jobs:
             - name: Install libsndfile
               run: |
                   sudo apt-get install -y libsndfile1
+            - name: Install ffmpeg
+              run: |
+                  sudo apt-get install -y ffmpeg
             - name: Display Python version
               run: python -c "import sys; print(sys.version)"
             - name: Full dependencies
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -18,6 +18,7 @@ import unicodedata
 import functools
 
 from speechbrain.utils.parallel import parallel_map
+from speechbrain.dataio.dataio import read_audio_info
 
 logger = logging.getLogger(__name__)
 
@@ -180,7 +181,7 @@ def process_line(line, data_folder, language, accented_letters):
 
     # Reading the signal (to retrieve duration in seconds)
     if os.path.isfile(mp3_path):
-        info = torchaudio.info(mp3_path)
+        info = read_audio_info(mp3_path)
     else:
         msg = "\tError loading: %s" % (str(len(file_name)))
         logger.info(msg)
--- a/recipes/DVoice/dvoice_prepare.py
+++ b/recipes/DVoice/dvoice_prepare.py
@@ -19,6 +19,7 @@ import pandas as pd
 from tqdm import tqdm
 import numpy as np
 import glob
+from speechbrain.dataio.dataio import read_audio_info
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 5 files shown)


AsuMagic:
The CI error seems to hint that ffmpeg is not available for the `torchaudio` version in use, judging from where this error originates from: <https://pytorch.org/audio/stable/_modules/torchaudio/backend/sox_io_backend.html>.

I will look into it and add fix CI within the PR if I'm correct.

Adel-Moumen:
Hi @AsuMagic,

Many thanks for this PR! 

@Gastron could you please have a look? Thanks!

Gastron:
I can take a look, but as I understand, let's work on #1944 first, and then this.

Gastron (approved):
Ok now that #1944 was merged I looked at these changes. The read_audio_info is great, let's merge this after the merge conflicts (from #1944) have been resolved.

AsuMagic:
Rebased against `develop`, contents are identical other than `read_audio_info` being moved around in CV prepare. I tested it again, it still works.

PR #1210, opened 2021-12-24
add CTC recipe GigaSpeech

aheba:
Hi, this PR for adding GigaSpeech (https://github.com/SpeechColab/GigaSpeech) as discussed in #1198.

the database is using OPUS format instead of Wav. I add an optional skip_opus2wav_conversion for allowing the conversion (using FFMPEG)..
2- Each audio file have avg of 20min.. our on fly loading time with `torchaudioload(..start_time, num_frame)` is very huge... one solution may consist into split the 20min audio to X utterance_audio...

what do you think @TParcollet , @mravanelli

PR #1944, opened 2023-04-17
Add `parallel_map`, parallelize CommonVoice data preparation

AsuMagic:
Supersedes #1602. This adds a `speechbrain.utils.parallel` import which provides `parallel_map`. Its main goal is to allow transforming elements from a list using a function, in a way that is executed in parallel. Elements are guaranteed to be returned in order, which makes this generally fairly usable. By default, a progress bar is shown.

By default, a process pool is used. I found that for certain tasks, `ThreadPoolExecutor` has unacceptable overhead, even when using its `.map` method. When the GIL is an issue, process pools appear to simply perform much better.

This does not require adding a new dependency to SpeechBrain and generally differs in some ways compared to most alternatives, in ways that seem somewhat beneficial. For one, `parallel_map` uses a (configurable) queue for dispatching jobs from the main thread. Furthermore, it will not attempt consuming the entire input stream at once, like `.map` would. It will also generate the output in-order, and not all at once.  
The hope here is to keep it flexible and viable for a wider range of usecases, including cases where it would not be acceptable to have the entire input or output live in memory at once.

The main design rationale is to:
1. Fill a queue (n=128 by default) of futures which are invoked jobs in the process pool, where every job operates over a chunk of the input.
2. Attach a conditional variable signal to the future to wake up the main scheduler thread.
3. When the main thread is woken up by the condition variable, re-enqueue as many jobs as were finished (in order to maintain the queue size) -- as long as there is enough input data.
4. If the first future in the queue is ready, get its result and yield it to the calling function. Keep going as long as we have ready futures in the front. This ensures in-order yielding even if the chunks are processed out-of-order across processes.
5. Loop back to step 2 until there isn't any future left to yield.

As it stands, this is a single function and is not integrated into things like the dynamic item dataset. I was discussing this in #1602 but I am not certain whether this would be necessary for data preparation.

This was tested with a real recipe, CommonVoice, for which the new data preparation steps are included in this PR. The observed performance improvement appears to scale with core count (in a case where the data preparation step was CPU bound). Will conflict with #1934, which will need to be fixed once either of the PRs are merged.

### DIFF
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -271,7 +271,10 @@ def create_csv(
         accented_letters=accented_letters,
     )
 
-    with open(csv_file, mode="w", encoding="utf-8") as csv_f:
+    # Stream into a .tmp file, and rename it to the real path at the end.
+    csv_file_tmp = csv_file + ".tmp"
+
+    with open(csv_file_tmp, mode="w", encoding="utf-8") as csv_f:
         csv_writer = csv.writer(
             csv_f, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL
         )
@@ -292,6 +295,8 @@ def create_csv(
                     row.words,
                 ]
             )
+
+    os.replace(csv_file_tmp, csv_file)
 
     # Final prints
     msg = "%s successfully created!" % (csv_file)
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -268,7 +268,7 @@ def create_csv(
         process_line,
         data_folder=data_folder,
         language=language,
-        accented_letters=accented_letters
+        accented_letters=accented_letters,
     )
 
     with open(csv_file, mode="w", encoding="utf-8") as csv_f:
@@ -283,7 +283,15 @@ def create_csv(
                 continue
 
             total_duration += row.duration
-            csv_writer.writerow([row.snt_id, str(row.duration), row.mp3_path, row.spk_id, row.words])
+            csv_writer.writerow(
+                [
+                    row.snt_id,
+                    str(row.duration),
+                    row.mp3_path,
+                    row.spk_id,
+                    row.words,
+                ]
+            )
 
     # Final prints
     msg = "%s successfully created!" % (csv_file)
--- a/recipes/CommonVoice/common_voice_prepare.py
+++ b/recipes/CommonVoice/common_voice_prepare.py
@@ -8,13 +8,16 @@ Luca Della Libera 2022
 Pooneh Mousavi 2022
 """
 
+from dataclasses import dataclass
 import os
 import csv
 import re
 import logging
 import torchaudio
 import unicodedata
-from tqdm.contrib import tzip
+import functools
+
+from speechbrain.utils.parallel import parallel_map
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Adel-Moumen:
Hey @AsuMagic,

This look very promising and awesome. Really neat! 

Do you have some numbers to share (e.g., CV preprocessing time taken with and without multithreading?)? 
Ty :-)

AsuMagic:
Quick test on CV 8.0 and a PyTorch 2.0 environment on a cluster node (host machine is a Xeon Silver 4210R). Files are local to the node I tested it on and either the drive is fast enough or the files are sitting in page cache. Very unscientific test, but:

PR code but with `tqdm(map())` (equivalent to previous code): `  2%|▏         | 6620/430464 [00:34<37:20, 189.15it/s]`
PR code with 24 processes (24t allocated): ` 47%|████▋     | 200568/430464 [01:24<01:32, 2473.53it/s]`  
PR code with 32 processes (32t allocated): ` 12%|█▏        | 50704/430464 [00:17<02:02, 3100.79it/s]`

Here the scaling is not linear with the number of threads, but that is likely due to hyperthreading.  
CPU usage is essentially maxxed out across the allocated threads.

Adel-Moumen:
Ok, really really cool. 

I suppose this new feature will help a lot regarding this #1210

AsuMagic:
> I suppose this new feature will help a lot regarding #1210

Depends. If the file loading is done at batch preparation time, then increasing the worker count would be sufficient as this is already done over multiple processes. If data preparation is slow then this function could help.

FWIW, from the linked issue:

>  2- Each audio file have avg of 20min.. our on fly loading time with torchaudioload(..start_time, num_frame) is very huge... one solution may consist into split the 20min audio to X utterance_audio...

This should not be the case, but I also observed that `torchaudio.load` with an offset would unnecessarily load the entire file and then slice it, which is atrocious for performance (at least the `sox` backend did, and it might depend on torchaudio versions as well). I honestly don't think Opus decoding should be a bottleneck.

Gastron (approved):
This is a nice general purpose tool to have in the utilities and the recipe demonstrate a use case, so I support its inclusion in SpeechBrain.

I haven't gone over the changes to the CommonVoice data prep but I trust that they work the same as before, just in parallel.

AsuMagic:
> I haven't gone over the changes to the CommonVoice data prep but I trust that they work the same as before, just in parallel.

Yes, the diff for `train.csv` (at least) is empty. Do note that it now writes to the `.csv` directly instead of waiting for all results in a list (mostly because it was simpler that way).

I'll fix the two issues mentioned above.

Adel-Moumen:
> Yes, the diff for train.csv (at least) is empty. Do note that it now writes to the .csv directly instead of waiting for all results in a list (mostly because it was simpler that way).

Hmmm it may creates an issue with how we detect that a csv is "valid". We do not check if the contents of a csv is the one expected rather we check if the csv exist or not which means that if for some reasons the user kill the process during the data-processing step, when relaunching no errors will be thrown while it should have been the case.

You can take a look at the fn 'skip' in the data prepare file.

AsuMagic:
> Hmmm it may creates an issue with how we detect that a csv is "valid". We do not check if the contents of a csv is the one expected rather we check if the csv exist or not which means that if for some reasons the user kill the process during the data-processing step, when relaunching no errors will be thrown while it should have been the case.

OK, I thought about this for a bit and I'm thinking that streaming into a `.tmp` file then using `os.replace` to rename the temporary file to the final file should be the most robust way to go in general (and it is how a lot of software override files anyway). Does that seem good?

Adel-Moumen:
> OK, I thought about this for a bit and I'm thinking that streaming into a .tmp file then using os.replace to rename the temporary file to the final file should be the most robust way to go in general (and it is how a lot of software override files anyway). Does that seem good?

It sounds good to me. What do you think @Gastron ?

AsuMagic:
FWIW I have gone ahead and implemented the rename solution and tested it with the CV recipe. If there's a better option it can still be changed.

TParcollet (approved):
Went from 1H to 2min on a slow NFS, massive. LGTM

PR #1947, opened 2023-04-19
Add torch compilation features (torch.compile, and more!)

AsuMagic:
This is a very WIP proof-of-concept for compiling specific modules in a model through a `compile_module_keys` hyperparameter, which functions similarly to `compile_jit_keys` but uses PyTorch 2.0's `torch.compile` instead.

**This is not intended to be reviewed at the moment,** but people can experiment with it.

A few observations and thoughts from the experiment:

- If you're going to experiment around `torch.compile`, use the nightlies, as they have some common bugs fixed.
- Dynamic shape support is still janky (most likely worsened because of training).
- A mere `compile_module_keys` might not be ideal. In some cases, the user or the compiled module may need specific compile parameters to work, or might otherwise benefit from tuning.
- As of right now, the compiler just seems to dislike sequences changing lengths every batch, and will either trigger constant recompiles or fail.
- The `compile_module_keys` model might be suboptimal. _Ideally_, in the future, we would simply be able to compile the entire thing and it would Just Work. In the mean time, compiling individual modules may prove beneficial, but when the recipes defines a bunch of small modules (e.g. mere linear layers) we _might_ be missing some significant fusing benefit.


Adel-Moumen:
Hey @mravanelli,

Do you know why in `core.py` we named our modules `self.modules` (https://github.com/speechbrain/speechbrain/blob/723e38e96c8b25fc49b73615d68768a2006a9e68/speechbrain/core.py#L599) while in `interfaces.py` It is called `self.mods` (https://github.com/speechbrain/speechbrain/blob/723e38e96c8b25fc49b73615d68768a2006a9e68/speechbrain/pretrained/interfaces.py#L222) ? If there's is no technical reason, I believe we should change it to match the one in `core.py`... 

Overall, regarding the PR, `torch.compile` is working. However, it is quite architecture specific and does not work everywhere. In fact, I was unable to make it faster than without any compilation with an RTX 3090. I believe this is due to two reasons: the first one being torch.compile is still quite "new" and the second one is that I need to use a better gpu (e.g., A100) to completely leverage the performances of torch.compile.

The PR is introducing four new flags:
`--compile_module_keys`, `--compile_mode`, `--compile_using_fullgraph`, `--compile_using_dynamic_shape_tracing`. The first flag is similar to `--jit_module_keys` while the three left are specific to `torch.compile`. 

Note: As mentioned, `torch.jit.script` is still in SpeechBrain and now will coexist with `torch.compile`.

pplantinga:
> Do you know why in `core.py` we named our modules `self.modules` while in `interfaces.py` It is called `self.mods`

I think this is due to the fact that `self.modules` is already used in `torch.nn.Module` so is not allowed in sub-classes. `Brain` class is not a `torch.nn.Module` whereas `Pretrained` class is a `torch.nn.Module`.

Not sure why `Pretrained` was set up this way but that's the likely reason.

mravanelli:
I added a flag `--compile `and `--jit `that, when enabled, will compile all the models in modules. That's pretty convenient because you can potentially run `python train.py hparams.yam --compile` without necessary modifying the yaml file if the user wants to compile all the modules. Please @Adel-Moumen  or @pplantinga, review this change.

Overall, I tried the compilation flags and I agree with what @Adel-Moumen  observed. It looks like an immature feature and I struggled a lot to have the model compiling. We can support this feature anyway with the hope that things will rapidly improve at the Pytorch level.

My next step is to add information about compilation in the documentation and maybe set up a short tutorial on this topic.

speechbrain/core.py
```diff
@@ -1345,16 +1410,59 @@ def _save_intra_epoch_ckpt(self):
...
+                compile_module_keys = set(self.compile_module_keys)
 
-        for name in self.jit_module_keys:
+        # Modules to compile with jit
+        jit_module_keys = set()
+        if self.jit:
+            if self.jit_module_keys is None:
+                jit_module_keys = set(self.modules)
+            else:
+                jit_module_keys = set(self.jit_module_keys)
```
Adel-Moumen, line 1436: I think we should log this behaviour to the user i.e. if `--jit` and `--jit_module_keys` are enabled then we should warn the user that we will only use the `--jit_module_keys`.

mravanelli (approved):

Adel-Moumen (approved):
LGTM @mravanelli. 

I just improved the warning message / added some small changes to the compilation.md.

mravanelli:
Everything LGTM. Thank you @AsuMagic and @Adel-Moumen for working on this PR.

Issue #1965, opened 2023-05-02
[Bug]: Can't pickle an EncoderDecoderASR because of functools._lru_cache


lucilemg:
### Describe the bug

When trying to export a package with a pickled EncoderDecoderASR object, the pickle operation does not work. From this [similar issue](https://stackoverflow.com/questions/54363043/pickling-lru-cached-function-on-object) (non related to Speechbrain) seems to come from the pickle module that has issues when dealing with decorators. Although the issue comes from the pickle module, maybe there's a way around to be able to pickle a model ?

The issue is caused by a decorator in speechbrain/dataio/preprocess.py in the AudioNormalizer class. When initiating an EncoderDecoderASR object (and thus a Pretrained class), an AudioNormalizer is initialized (see [code](https://github.com/speechbrain/speechbrain/blob/0423bda83237070ad219d88b834d480f58cbe53e/speechbrain/pretrained/interfaces.py#L225)). In its `__init__` function (see [code](https://github.com/speechbrain/speechbrain/blob/0423bda83237070ad219d88b834d480f58cbe53e/speechbrain/dataio/preprocess.py#L44)), it creates `self._cached_resample` with `functools.lru_cache` for the Resample class, which causes the error below.

### Expected behaviour

The package should be created with the pickled model inside.

### To Reproduce

```
from torch import package
from speechbrain.pretrained import EncoderDecoderASR

if __name__ == "__main__":
    asr_model = EncoderDecoderASR.from_hparams(
    	source="speechbrain/asr-transformer-switchboard",
    )

    with package.PackageExporter("test_pickle.pt", debug=True) as exp:
        exp.extern("speechbrain.**")
        exp.intern("__main__")
        exp.save_pickle("binary", "test_pickle.pkl", asr_model)
```

### Versions

speechbrain 0.5.14
torch 2.0.0
torchaudio 2.0.1

### Relevant log output

```shell
Traceback (most recent call last):
  File "/home/lucile/speech_repo/speech-lab/modelling/use_cases/test_pickle_issue.py", line 15, in <module>
    exp.save_pickle("binary", "test_pickle.pkl", asr_model)
  File "/home/lucile/speech_repo/speech-lab/modelling/.venv/lib/python3.9/site-packages/torch/package/package_exporter.py", line 631, in save_pickle
    pickler.dump(obj)
[...]

PR #1966, opened 2023-05-02
Pickle issue

lucilemg:
# Contribution in a nutshell
Hey, 
This PR has for objective to find a solution to [this issue](https://github.com/speechbrain/speechbrain/issues/1965) when trying to export a package with a pickled EncoderDecoderASR object.

# Scope
* [x] Hopefully fixes [[Bug]: Can't pickle an EncoderDecoderASR because of functools._lru_cache](https://github.com/speechbrain/speechbrain/issues/1965)

# Pre-review
* [x] exhausted patience before clicking « Ready for review » in the merge box 🍄

---

### DIFF
--- a/speechbrain/processing/speech_augmentation.py
+++ b/speechbrain/processing/speech_augmentation.py
@@ -16,7 +16,7 @@ Authors
 import math
 import torch
 import torch.nn.functional as F
-from methodtools import lru_cache
+from functools import lru_cache
 from speechbrain.dataio.legacy import ExtendedCSVDataset
 from speechbrain.dataio.dataloader import make_dataloader
 from speechbrain.processing.signal_processing import (
--- a/speechbrain/processing/speech_augmentation.py
+++ b/speechbrain/processing/speech_augmentation.py
@@ -16,6 +16,7 @@ Authors
 import math
 import torch
 import torch.nn.functional as F
+from methodtools import lru_cache
 from speechbrain.dataio.legacy import ExtendedCSVDataset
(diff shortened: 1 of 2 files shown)


Gastron (changes requested):
Thanks for taking the time to try and solve this and still keep some cache in use. Unfortunately your implementation changes what keys the cache would work on, see comment for specific line.

Looking back, I think the LRU cache was a premature optimisation. I suggest a simpler version which doesn't evict any results from the cache - then we can just use a dict. I think it is wise to keep the Resample object around, not to waste time creating one every time this gets called.

speechbrain/processing/speech_augmentation.py
```diff
@@ -566,6 +567,7 @@ def _compute_strides(self):
         self.conv_stride = input_samples_in_unit
         self.conv_transpose_stride = self.output_samples
 
+    @lru_cache(maxsize=12)
```
Gastron, line 570:
```suggestion
```

speechbrain/processing/speech_augmentation.py
```diff
@@ -16,6 +16,7 @@
 import math
 import torch
 import torch.nn.functional as F
+from functools import lru_cache
```
Gastron, line 19:
```suggestion
```

Gastron (changes requested):
Oh and remove the LRU cache from Resample too.

Gastron (approved):

Gastron:
Thanks @lucilemg! LGTM.

PR #1978, opened 2023-05-15
Transducer Beam Search Implemented original algorithm from paper 

### LINKED ISSUE #1969: [Bug]: Transducer beamsearch decoder can never finish
### Describe the bug

I was looking into customizing the transducer decoder beamsearch to support multitask as in this [paper](https://arxiv.org/pdf/2204.00558.pdf) when i realized that in the case of degenerate output from the beamsearch, it may never finish.
This behavior can also be seen in the doctest of the decoder/transducer.py file. (If you run it without training, it is stuck in an infinite loop) 
This may be because the implementation in speechbrain is slightly different from the one in the original [paper](https://arxiv.org/pdf/1211.3711.pdf) by Alex Graves. 

Screenshot from 2023-05-03 13-08-09

In the original paper, y*+blank will always be added to B. Whereas in the speechbrain implementation, it will only be added if the blank token is in the topK probabilities. Meaning that if an output degenerate on a bigram for exemple, the blank index may never be in topK probabilities, the beam in A will never be added to B, and it will never finish. 
Screenshot from 2023-05-03 13-08-32
Maybe adding a safeguard against degenerate sentence can be done ? Based on the number of token generated in one step, for example. 

There are ways to bypass this kind of problem simply by training the model more or maybe by using a bigger beam, but I figured I had let you know of this behavior since it can be significantly different from the expected one. 

### Expected behaviour

The beamsearch should end even if it returns a degenerate output.

### To Reproduce

_No response_

### Versions

_No response_

### Relevant log output

_No response_

### Additional context

_No response_

Pupiera:
Implemented original algorithm from paper  : https://arxiv.org/pdf/1911.01629.pdf

Resolves #1969


mravanelli:
@Adel-Moumen, what should we do with this PR?

PR #1979, opened 2023-05-15
fix bug: eos_threshold in S2SBeamSearcher

My-captain:
In my experiments, logits of <EOS> is zero when the decoding step should stop.
Absolutely, `(self.eos_threshold * max_probs)` is zero too. So the `_check_eos_threshold` return False, and the heuristic search algorithm from [Sequence-to-Sequence Speech Recognition with Time-Depth Separable Convolutions](https://arxiv.org/pdf/1904.02619.pdf) will fail.

Considering the efficiency of beam search, changing `>` to `>=` is the simplest way to fix it.

# Contribution in a nutshell
Hey, this could help our community 🌱

# Scope
* [ ] I want to get done ...
* [ ] ... and hope to also achieve ...

# Pre-review
* [ ] (if applicable) add an `extra_requirements.txt` file
* [ ] (if applicable) add database preparation scripts & use symlinks for nested folders (to the level of task READMEs)
* [ ] (if applicable) add a recipe test entry in the depending CSV file under: tests/recipes
* [ ] create a fresh testing environment (install SpeechBrain from cloned repo branch of this PR)
* [ ] (if applicable) run a recipe test for each yaml/your recipe dataset
* [ ] check function comments: are there docstrings w/ arguments & returns? If you're not the verbose type, put a comment every three lines of code (better: every line)
* [ ] use CI locally: `pre-commit run -a` to check linters; run `pytest tests/consistency`
* [ ] (optional) run `tests/.run-doctests.sh` & `tests/.run-unittests.sh`

---

### DIFF
--- a/speechbrain/decoders/seq2seq.py
+++ b/speechbrain/decoders/seq2seq.py
@@ -578,7 +578,7 @@ class S2SBeamSearcher(S2SBaseSearcher):
         """
         max_probs, _ = torch.max(log_probs, dim=-1)
         eos_probs = log_probs[:, self.eos_index]
-        cond = eos_probs > (self.eos_threshold * max_probs)
+        cond = eos_probs >= (self.eos_threshold * max_probs)
         return cond
 
     def _update_hyp_and_scores(


Adel-Moumen:
Hi @My-captain, 

Thanks for this PR. However, I believe we should not change the operator from `>` to `>=`. Indeed, in the official paper, they are using a hard constraint, and in all of the other implementations of this heuristic, people are also using the hard constraint (see: espnet). I believe that in SpeechBrain we should always match as strictly as possible the official implementation of a research work. Thus, I prefer to close this PR. 

Thanks again for your work.

PR #1982, opened 2023-05-17
Fix pyctcdecode and add padding example

Adel-Moumen:
This PR add an example of how to change the default padding value in SpeechBrain. Furthermore, this PR fix a tiny bug in the ASR CTC LibriSpeech recipe where the greedy search was performed alongside the beam search when using PyCTCDecode which is unwanted.


recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml
```diff
@@ -84,12 +89,21 @@ train_dataloader_opts:
     batch_size: !ref <batch_size>
     shuffle: True
     num_workers: !ref <num_workers>
+    collate_fn: !name:speechbrain.dataio.batch.PaddedBatch
+        padding_kwargs:
+            value: !ref <pad_index>
 
 valid_dataloader_opts:
     batch_size: 1
```
Adel-Moumen, line 97: @TParcollet do you know why we have a batch size of 1 when doing eval/test ?

AsuMagic (approved):

PR #1992, opened 2023-05-23
Implementation of the Branchformer (code from Samsung AI Cambridge)

TParcollet:
# Contribution in a nutshell

This is the implementation of the Branchformer coming from _"Branchformer: Parallel MLP-Attention Architectures
to Capture Local and Global Context for Speech Recognition and Understanding"_ from Shinji's team. The code is given by Samsung AI Cambridge. 

Result-wise, we are at 2.2 on test-clean against 2.0 from ESPnet. I suspect that further hyperparameter tuning must be done, but at the same time, it honestly is pointless on this dataset. Our LM might also be weaker. 

@Adel-Moumen please run the models on 4 GPUs with DDP (compute canada?) and see if you are able to reach the same numbers (conformer_large and branchformer_large). Also, could you upload the models to the dropbox? I can't share logs from the Samsung training center. (Having a model on HF would be great as well I believe, maybe the one of the branchformer)
@Gastron please have a look at the function removing bias and LN from weight decay in the core function, as this may basically affect ALL existing recipes (hopefully in good). 

# Pre-review
* [x] Add README results.
* [x] add a recipe test entry in the depending CSV file under: tests/recipes


speechbrain/utils/optimizers.py
```diff
@@ -0,0 +1,23 @@
+"""Implements functions to avoid optimizing certain parameters
+
+Authors
+ * Titouan Parcollet 2023
+"""
+
+
+def rm_weight_decay_bias_and_norm_params(
```
Gastron, line 8: This function name suggests that it targets bias and norm parameters, and I suppose it most often ends up doing that, but the criterion is `len(param.shape) == 1`. Are you sure that is robust - will this sometimes catch other parameters? If I am thinking this out correctly, won't this essentially catch all vectors?

Is this something we always want to do for bias and norm params, I am not aware of the theory behind this?

speechbrain/core.py
```diff
@@ -867,8 +868,17 @@ class being passed at initialization that takes only a list
 
         Override this class if there are multiple optimizers.
         """
+
+        # We remove biases and normalization parameters from weight_decay
+        all_params = self.modules.parameters()
+        if hasattr(self, "hparams"):
+            if hasattr(self.hparams, "weight_decay"):
```
Gastron, line 875: This behaviour is a little too magical in my opinion. We don't rely on a specific key in hparams anywhere. Instead, this should probably be controlled by a run_opts flag.

speechbrain/utils/optimizers.py
TParcollet: You are right, it will catch all parameters that are just vectors. But in most case, we may want to avoid pushing such parameters to 0. For BN and LN layers, it is detrimental to the results to do it. For biases, it simply is useless. ESPnet and K2 do this for instance. The len==1 is a global extension because I don't see any reason why someone would have scalar learnable parameters being pushed to 1. Note: This is only triggered if the yaml explicitly contains a self.weight_decay parameter .

speechbrain/core.py
TParcollet: I will think about it. Applying weight decay to LN and BN plainly is wrong. Simply purely wrong. Hence, no magic in forcing things to be right. And the problem here is that I need the weight_decay to be set in the hparams to access its value... a run_opts would not give me the value (no way that we put the weight decay value as a run_opt given that this is linked to the optimiser)

TParcollet: You are write, however, in the sense that if someone use weight decay in the optimiser but not a hparams.weight_decay params, then it won't be triggered. I think I will instead directly check the optimiser.

TParcollet: @Gastron have a look at the latest version I added your comments (check name as well and remove the magic)

Gastron (changes requested):
I am reviewing the part where vectors are excluded from weight_decay.

Thanks for responding to my comments @TParcollet. I now think you are right about this being desirable behaviour, not just for norms and biases, but probably all vectors.

However, I still believe the implementation does things a little too automagically.

Here's a suggestion: 
1. Rename the behaviour to be explicit about what it does: `remove_vector_weight_decay`.
2. Make this behaviour conditional to a run_opts flag instead of looking for weight_decay in optimizers.

Implementation outline for 2:
In `core.py`, function `parse_arguments`, add
```python
parser.add_argument(
        "--remove_vector_weight_decay",
        default=None,
        action="store_true",
        help="Make vectors (e.g. norms and biases) a separate parameter group without weight_decay.",
    )
```
And then in the Brain `__init__`, add a corresponding default:
```python
run_opt_defaults = {
...
remove_vector_weight_decay=False
...
}
```
Finally in `init_optimizers`:
```python
        ...
        if self.remove_vector_weight_decay:
                all_params = rm_weight_decay_bias_and_norm_params(
                    self.modules, self.hparams.weight_decay
                )
```

For the reproducibilty of results, I really recommend making the default False.

recipes/LibriSpeech/ASR/transformer/hparams/branchformer_large.yaml
```diff
@@ -0,0 +1,291 @@
...
+# Tokens: unigram
+# losses: CTC + KLdiv (Label Smoothing loss)
+# Training: Librispeech 960h
+# Authors:  Titouan Parcollet
+# ############################################################################
+# Seed needs to be set at top of yaml, before objects with parameters are made
+
+seed: 7775
+__set_seed: !apply:torch.manual_seed [!ref <seed>]
+output_folder: !ref results/conformer_small/<seed>
```
Adel-Moumen, line 14: rename it to branchformer_large

TParcollet: @Adel-Moumen I agree with the latest comment of @Gastron could you do it and merge please?

TParcollet:
@Adel-Moumen I agree with the latest comment of @Gastron could you do it and merge please?

Adel-Moumen (approved):
LGTM. As mentioned by Titouan, I fixed the issue raised by @Gastron as well as some tiny typos.

speechbrain/utils/optimizers.py
Gastron, line 8:
```suggestion
def rm_vector_weight_decay(
```

speechbrain/utils/optimizers.py
```diff
@@ -0,0 +1,23 @@
...
+        if not param.requires_grad:
+            continue
+        if len(param.shape) == 1 and ("bias" in name or "norm" in name):
+            no_decay.append(param)
+        else:
+            decay.append(param)
+    return [
+        {"params": no_decay, "weight_decay": 0.0},
+        {"params": decay, "weight_decay": weight_decay},
+    ]
```
Gastron, line 23:
```suggestion
def rm_vector_weight_decay(
    modules,
):
    decay = []
    no_decay = []
    for name, param in modules.named_parameters():
        if not param.requires_grad:
            continue
        if len(param.shape) == 1:
            no_decay.append(param)
        else:
            decay.append(param)
    return [
        {"params": no_decay, "weight_decay": 0.0},
        {"params": decay},
    ]
```

speechbrain/core.py
```diff
@@ -867,8 +875,16 @@ class being passed at initialization that takes only a list
...
         """
+
+        all_params = self.modules.parameters()
+
         if self.opt_class is not None:
-            self.optimizer = self.opt_class(self.modules.parameters())
+            if self.remove_vector_weight_decay:
+                all_params = rm_weight_decay_bias_and_norm_params(
+                    self.modules, self.hparams.weight_decay
+                )
```
Gastron, line 885:
```suggestion
            if self.remove_vector_weight_decay:
                all_params = rm_vector_weight_decay(self.modules)
```

speechbrain/core.py
```diff
@@ -34,6 +34,7 @@
 from torch.nn.parallel import DistributedDataParallel as DDP
 from hyperpyyaml import resolve_references
 from speechbrain.utils.distributed import run_on_main
+from speechbrain.utils.optimizers import rm_weight_decay_bias_and_norm_params
```
Gastron, line 37:
```suggestion
from speechbrain.utils.optimizers import rm_vector_weight_decay
```

Gastron (changes requested):
Continuing on the vector weight decay part.

Sorry, only now I realized that the implementation also used to hardcode `self.hparams.weight_decay` as the parameter we looked for. The nice thing is that we can just leave that up to the default parameters given to the optimizer, so we don't need to pass the weight_decay at all.

Also, I made a change which removes all vectors from weight decay. I think we agreed earlier that that is pretty sensible. However, if you want to keep the additional check `("bias" in name or "norm" in name)`, then I would actually prefer the older name (`rm_weight_decay_bias_and_norm_params`).

Adel-Moumen:
@TParcollet could you please pull the latest commits from the dev branch? Ty. 

Thanks @Gastron for your changes. I added the commits.

speechbrain/utils/optimizers.py
Gastron: This function also needs a docsring. I suggest something like:

```
"""Put vectors in a parameter group without weight decay

Takes in a list of modules and separates their parameters into two parameter groups,
which can be passed to a PyTorch Optimizer class. Vector parameters get weight_decay overridden to zero.
This is particularly useful for biases and norms, which we expect to deviate from zero. Other vectors as parameters are also likely not meant to be pushed toward zero.

Arguments
----------
modules : torch.ModuleList, torch.Module
    Torch modules to operate on

Returns
-------
list
    The parameter groups in the Pytorch Optimizer specification format.
```

Adel-Moumen: Yep. I'm currently working on it as well as fixing the pre-commits.

Gastron (approved):
At this point, the vector parameter groups stuff looks good to me.

Adel-Moumen:
I'm currently training a conformer large / branchformer large models. When the training will be finished, I will create a new PR with the links of the new HF/dropbox models.

Issue #1803, opened 2023-01-15
[Bug]: CTC and w2v2 with fp16 is going NaN 


TParcollet:
### Describe the bug

When fine-tuning an ASR model based on CTC only and a w2v2 from HF (for instance CommonVoice/ASR/CTC), the loss will explode to NaN in few batches when fp16 is activated.

I suspect that this is due to the CTC loss and log_softmax being done in fp16 while they should be computed in fp32.

Don't have time to fix this now. 

### Expected behaviour

Training without NaN

Adel-Moumen:
I was able to reproduce the bug. 

I started digging into the problem and found that GradScaler was the source of NaNs. 

I will get back to you when I find the reason/solution to fix the problem. :-)

JuanPZuluaga:
@TParcollet did you solve this issue? Or simply moving out the loss computation from autocast (as now in `core.py` see below) solved the issue? 

https://github.com/speechbrain/speechbrain/blob/891318f5950c337bb951912bf64bd5973af7c908/speechbrain/core.py#L937

Adel-Moumen:
Closing this issue as #2010 is merged in develop.

Issue #1517, opened 2022-07-26
Can NOT speed up with DDP


poor1017:
Hi

I can not speed up training with DDP for `recipes/WHAMandWHAMR/enhancement/train.py`. My DDP command line is:

```
python3 -m torch.distributed.launch --nproc_per_node=4 --master_port 6666 train.py hparams/sepformer-whamr-16k-DM.yaml --data_folder /workspace/cpfs-data/Workstations/speechbrain/recipes/WHAMandWHAMR/data/whamr --base_folder_dm /workspace/cpfs-data/Data/wham/wsj0/si_tr_s/ --rir_path rir_wavs --distributed_launch --distributed_backend='nccl'
```

Compare to:

```
python3 train.py hparams/sepformer-whamr-16k-DM.yaml --data_folder /workspace/cpfs-data/Workstations/speechbrain/recipes/WHAMandWHAMR/data/whamr --base_folder_dm /workspace/cpfs-data/Data/wham/wsj0/si_tr_s/ --rir_path rir_wavs
```

DDP mode cost almost the same time.

Could you help to check this issue?

Thank you!

poor1017:
Hi @Samuele Cornell, @Cem Subakan
I found the reason that makes DDP do not work:
[https://github.com/speechbrain/speechbrain/blob/develop/recipes/WHAMandWHAMR/enhancement/dynamic_mixing.py#L216](url)

Can you tell me the reason you wrap `train_data` with `torch.utils.data.DataLoader`?

Thank you

popcornell:
Hi, your tag did not work so I ve missed this issue sorry, thanks to @anautsch that pinged me. 
Does it happens only with Dynamic Mixing ? Or also with the plain recipe ?

ycemsubakan:
Also, @poor1017 , do you see the same behavior with WSJ0Mix recipe also?

poor1017:
@popcornell Only with Dynamic Mixing, because with DM, you use torch.utils.data.DataLoader to cover sb.dataio.dataset.DynamicItemDataset
@ycemsubakan I have not tried WSJ0Mix.

mravanelli:
Samuele, could you please take a look at it?

Adel-Moumen:
Hello, I believe that this issue is related to #1802 and should be solved in #2010.

Adel-Moumen:
Closing this issue as #2010 is merged in develop.

Issue #1496, opened 2022-07-05
Brain's check_gradients seems to not do what name implies?


RuABraun:
Was trying to debug something of mine and noticed two things:

1. [check_gradients](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/core.py#L888) does not actually check the gradients, it checks the loss [which does not have to be NaN if the gradients will be](https://discuss.pytorch.org/t/loss-is-not-nan-but-the-gradients-are/21224)

2. the [GradScaler's step will use unscale (which checks for nonfinite gradients) to skip an optimization step if any of the gradients are null](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler.step), so the faulty behaviour of `check_gradients` works out because it's actually already done by GradScaler

So it seems to me that at minimum `check_gradients` should be renamed to check loss or something like that, because it's confusing to read the code and see it doesn't match the function name and docstring, and then one wonders if there's a bug.

pplantinga:
I agree that a different name could be more informative. Something like `check_loss_isfinite` or similar.

Adel-Moumen:
Hi guys, 

What should we do about this? Should we create a PR to change the naming of `check_gradients`?

Adel-Moumen:
Closing this issue as #2010 is merged in develop.

Issue #1802, opened 2023-01-13
[Bug]: DDP is going blblblblblblblbl


TParcollet:
### Describe the bug

On certain recipe, and with certain models, DDP with X GPUs is slower than mono GPU

### Expected behaviour

Should be (X - loss factor) times faster with X the number of GPUs.

### To Reproduce

Simply run the CommonVoice/ASR/CTC/train_with_wav2vec.py with and without DDP. The expected training time will be longer with DDP.

HOWEVER, if we add self.no_sync() to the fit_batch around compute_forward(): then it works as intended. @RuABraun worked a bit with this no_sync() stuff, maybe he could help? I don't understand why models would be syncing during the forward though ...

AsuMagic:
Adel and I investigated the issue a little and it seems that the problem boils down to the fact that we also need to include the **forward** pass (not just the backward pass) in the `no_sync` context manager.

So only at the last forward/backward should the gradient be synchronized. If you omit the `no_sync` context manager for the **forward**, PyTorch seems to call into some code, `prepare_for_backward`, which presumably induces a bunch of synchronization (either immediately or later).

The PyTorch documentation says so and a [HuggingFace Accelerate tutorial](https://huggingface.co/docs/accelerate/concept_guides/gradient_synchronization) mention the same thing. The performance degradation HF notes is similar to what you are observing (i.e. worse than single-GPU).

So what we are doing in SB right now is incorrect. It shouldn't cause issues other than unnecessary slowdowns, though. Considering the scope of #2010 which might be very related, maybe it should be implemented there? @Adel-Moumen

RuABraun:
Interesting catch! Didn't realize this and the torch docs didn't have that warning at the time.

Adel-Moumen:
Fixed in SB 1.0 :p (see changes in speechbrain.core.py and more precisely the fit_batch function).

PR #2010, opened 2023-05-31
Refactoring of the 'fit_batch' function

Adel-Moumen:
This PR aims at solving the fp16 nan issue with CTC common voice wav2vec #1803.

The problem was coming from the loss. basically, we were not asking the accumulation of the gradients when the loss was NaN which led to NaN gradients. It is based on the pr #2018. 

I also added bfloat16 support. 

This pr change the name `check_gradients` to `check_loss_isfinite` (#1496).

WIP results: 
FP16 A100 80Gb - CommonVoice FR - Wav2vec2 
100%|██████████████████████████████████████████████████████████████████████████| 17829/17829 [1:35:12<00:00,  3.12it/s, train_loss=0.367]
100%|████████████████████████████████████████████████████████████████████████████████████████████████| 3978/3978 [05:22<00:00, 12.32it/s]
speechbrain.utils.train_logger - epoch: 1, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 3.67e-01 - valid loss: 1.50e-01, valid
CER: 3.89, valid WER: 12.58

Related issues:
#1517  #1803 #1496 #2085 

Related code:
ESPNet way to handle NaNs: https://github.com/espnet/espnet/blob/5d0758e2a7063b82d1f10a8ac2de98eb6cf8a352/espnet/asr/pytorch_backend/asr.py#L301 -> if a bwd leads to NaN gradients, they still continue to accumulate. There is no such thing as "valid_step". This is a simpler approach but it could lead to "wrong" accumulation steps. If grad_accu is set to 3, and during training there's one bwd which leads to NaN then it means there are going to step the opt with only 2 bwd of grads not 3 as expected. 

Furthermore, there are not removing the loss from the autocast part. And also, they are using the "old" way of handling autocast not the new one with GradScaler.

NeMO way to handling NaNs/Infs = https://github.com/NVIDIA/NeMo/blob/1c10b3677040cde4785ea73f7d9ec887578885c7/nemo/collections/asr/models/asr_model.py#L133 basically there are setting the value of the grad to be zero. 

[PyTorch Discussion](https://discuss.pytorch.org/t/ctc-loss-ctc-loss-not-support-float16/148800) which indicates that the loss should be excluded from the fp16 region.


Adel-Moumen:
```python
    def check_loss_isfinite(self, loss):
        """Check if loss is finite and that gradients are not too large.
        Automatically clips large gradients.

        Arguments
        ---------
        loss : tensor
            The loss tensor after ``backward()`` has been called but
            before the optimizers ``step()``.

        Returns
        -------
        bool
            Whether or not the optimizer step should be carried out.
        """
        if not torch.isfinite(loss):
            self.nonfinite_count += 1

            # Print helpful debug info
            logger.warning(f"Loss is {loss}.")
            for p in self.modules.parameters():
                if not torch.isfinite(p).all():
                    logger.warning("Parameter is not finite: " + str(p))

            # Check if patience is exhausted
            if self.nonfinite_count > self.nonfinite_patience:
                raise ValueError(
                    "Loss is not finite and patience is exhausted. "
                    "To debug, wrap `fit()` with "
                    "autograd's `detect_anomaly()`, e.g.\n\nwith "
                    "torch.autograd.detect_anomaly():\n\tbrain.fit(...)"
                )
            else:
                logger.warning(
                    "Patience not yet exhausted, ignoring this batch."
                )
                return False

        if self.max_grad_norm > 0.0:
            torch.nn.utils.clip_grad_norm_(
                (p for p in self.modules.parameters()), self.max_grad_norm
            )

        return True
```

There's something that I don't understand. This function is doing too much. First, it checks if the loss is not nan, then checks the  gradients, and finally perform gradient clipping. However, with the latest PR #2018, we now use this function to check whether the loss is finite, if it's the case then we backpropagate. However, doing so is skipping the clipping / check gradients part because we are are adding gradients AFTER having clipped/checked grads...  Furthermore, I saw many recipes that were using `check_loss_isfinite` to just check/clip grads (and not to check the loss). 

I propose one solution that would be to 1. re-rename `check_loss_isfinite` to `check_gradients` (contrary to #1496), and 2. remove the part "if not torch.isfinite(loss)" in a new function such as: 

```python
def check_loss_isfinite(self, loss):
    """Check if loss is finite and that gradients are not too large.
    Automatically clips large gradients.

    Arguments
    ---------
    loss : tensor
        The loss tensor after ``backward()`` has been called but
        before the optimizers ``step()``.

    Returns
    -------
    bool
        Whether or not the optimizer step should be carried out.
    """
    if not torch.isfinite(loss):
        self.nonfinite_count += 1
        # Check if patience is exhausted
        if self.nonfinite_count > self.nonfinite_patience:
            raise ValueError(
                "Loss is not finite and patience is exhausted. "
                "To debug, wrap `fit()` with "
                "autograd's `detect_anomaly()`, e.g.\n\nwith "
                "torch.autograd.detect_anomaly():\n\tbrain.fit(...)"
            )
        else:
            logger.warning(
                "Patience not yet exhausted, ignoring this batch."
            )
        return False
    else:
        return True


def check_gradients(self):
    """Check that gradients are finite and not too large.
    Automatically clips large gradients.
    """
    for p in self.modules.parameters():
        if not torch.isfinite(p).all():
            logger.warning("Parameter is not finite: " + str(p))

    if self.max_grad_norm > 0.0:
        torch.nn.utils.clip_grad_norm_(
            (p for p in self.modules.parameters()), self.max_grad_norm
        )
``` 

The idea is to be more versatile. 

What's your view on that please @TParcollet @AsuMagic (@pplantinga @RuABraun as you were in #1496) ?

AsuMagic:
So, we discussed this a little at the lab, and the tl;dr of the steps above seem correct. There are still a bunch of implementation details to figure out, and we determined a few things. Mostly dumping what we talked about and what I think about it.

For context, here, when referring to "stepping", we mean running the optimizer step function after we accumulated as many gradients as we desired (if using that).

So basically, to reiterate, the steps while backwarding/accumulating gradients could become (more details later):

- If using DDP, disable gradient sync for all but the last time we are accumulating gradients before a step
- Run the forward pass
- Check for non-finite loss: If bad, skip the batch
- Run the backward pass
- Check for non-finite gradients: If bad, flag the step as broken
- Clip gradients

Then, at stepping time:

- Check if the step is flagged as broken, if so, zero the gradient and early quit
- Step the optimizer
- Update the AMP scaler
- Zero the gradient

### Handling non-finite losses and gradients

- Our current code to detect non-finite gradients is incorrect. It checks for the model parameters, not for their gradients. At this point, the parameters have already been "corrupt" (driven to NaN/+-infinity), so we can't recover that.
    - It might still be useful to track non-finite parameters, to be fair?
    - The check also only occurs if the loss was detected to be non-finite so it does not really do what it's advertised for.
- We probably should use a counter for successful gradient accumulations (mostly because of the next point).
- If we detect a non-finite loss, then we should skip the batch (but not the step). At this point, we have not run the backward pass yet, so the gradient is not corrupt. We can carry on with our step and not backpropagate the corrupt loss.
- The patience mechanism would only be applied to non-finite *loss* detection, unless we believe it makes sense to apply it to non-finite gradients too.
- If we detect non-finite gradients, then our gradient is "corrupt", which we cannot undo. Thus, we mark the step as corrupt, and only `.zero_grad()` will be called at stepping time.
    - Yes, it is suboptimal as we could technically discard the step immediately. It is however not a big deal and keep things simpler when we think about DDP, etc.
- There is _some_ overlap with what AMP scalers do. Presumably, non-finite gradients would already be skipping the step. This is not the case for fp32, where we use no scaler, so it still makes sense to handle in SB.

### Gradient clipping

- We can apply it either after every batch, or we can apply it when stepping. It can make sense to be done after every batch since it reduces the odds that gradients explode to infinity, in particular with fp16.
- **@Adel-Moumen : There could be a good reason for doing it at step time:** Gradient clipping requires us to unscale the gradient when using AMP. I don't know how much of a problem in practice it would be, would need to check

### API design

- Lots of recipes have a lot of duplicated code so we should be careful about API design so that it is modular enough and easy to override when _really_ needed (which should be rare). So we should study the recipes to make sure we come with a good design.
    - This is so easy to get wrong we should ensure that it is difficult to use things incorrectly. For the most part, that is solved by making the API flexible and natural enough.
    - Example: With wav2vec models we might have split optimizers. In our new stock stepping functions, it would be trivial and clean to handle this usecase by taking in a list of optimizers.
- We make a clear cut between accumulating gradients, and stepping the optimizer. This would split most of the backwards logic to two relatively self-contained functions.
- Most of the new functionality would be included as Brain methods.
- Be careful about naming, IMO:
    - A function named `check_`something should _not_ include patience logic inside, unless explicitly stated in the function name (merely "checking" feels like it should not carry any state)
- As stated in #1802, our `no_sync` DDP logic is broken, resulting in extraneous synchronization. This may influence API design here if we want to end up with something clean.

For example the recipe code for the dual-optimizer wav2vec code could roughly look like:

```python
active_optimizers = [self.adam_optimizer]
if not self.hparams.wav2vec2_frozen:
    active_optimizers.append(self.wav2vec_optimizer)

#...

if self.safe_backward(loss):
    if self.should_step():
        self.step_optimizer(active_optimizers)
```

Also OFC it might make sense to check what other toolkits do.

pplantinga:
This seems well thought out, and I agree with most of the points. I wonder if we could add a basic test with a ridiculously high learning rate to see if it works as expected.

Adel-Moumen:
Gentle ping @TParcollet, what do you think about this https://github.com/speechbrain/speechbrain/pull/2010#issuecomment-1600926440 ? 

Btw, @AsuMagic, I'd like to have something even more generic than what you are proposing. I prefer to override a function for example `step_optimizer(self)` so that we don't have to rewrite each time the `fit_batch(self, batch)`. So we would have the generic code in the core.py, and then we would only override specific functions such as `step_optimiser`.

AsuMagic:
> Gentle ping @TParcollet, what do you think about this [#2010 (comment)](https://github.com/speechbrain/speechbrain/pull/2010#issuecomment-1600926440) ?
> 
> Btw, @AsuMagic, I'd like to have something even more generic than what you are proposing. I prefer to override a function for example `step_optimizer(self)` so that we don't have to rewrite each time the `fit_batch(self, batch)`. So we would have the generic code in the core.py, and then we would only override specific functions such as `step_optimiser`.

Could make sense, I'm not sure how many recipes would _not_ need a custom `fit_batch` though. Seems like a bunch of them could, and that possibly would fix a few bugs/automatically add things like AMP support...

pplantinga:
Hard agree that we should add the support needed to avoid overriding fit_batch where possible.

Adel-Moumen:
> > Gentle ping @TParcollet, what do you think about this [#2010 (comment)](https://github.com/speechbrain/speechbrain/pull/2010#issuecomment-1600926440) ?
> > Btw, @AsuMagic, I'd like to have something even more generic than what you are proposing. I prefer to override a function for example `step_optimizer(self)` so that we don't have to rewrite each time the `fit_batch(self, batch)`. So we would have the generic code in the core.py, and then we would only override specific functions such as `step_optimiser`.
> 
> Could make sense, I'm not sure how many recipes would _not_ need a custom `fit_batch` though. Seems like a bunch of them could, and that possibly would fix a few bugs/automatically add things like AMP support...


As you can see in the pr I started to change a bunch of recipes, and basically what I saw is that A LOT of recipes are actually doing the exact same things over and over. I really think that's its possible to have something generic. The only recipes that will have to change something are the one with multiple optimisers. (But we will have to look more closely)

AsuMagic:
This is the mockup/prototype document I wrote that we discussed in the core meeting: https://gist.github.com/AsuMagic/8faaf3c15b24e845ab1767d06f1eb2d2

TParcollet (approved):
I like the changes.

pplantinga:
Any progress on this?

Adel-Moumen:
> Any progress on this?

Hello, no progress so far. I will work on this after having finish CTC and Rescoring PRs.

Adel-Moumen:
update results: 

RTX 3090
> DDP Fp32 - 10%|███████████▎                                                                                                          | 1709/17829 [12:30<1:57:55,  2.28it/s, train_loss=5.64]
DDP Fp16 -  11%|████████████▋                                                                                                         | 1924/17829 [12:30<1:43:26,  2.56it/s, train_loss=2.45]
No DDP Fp32  7%|███████▋                                                                                                              | 2341/35658 [12:30<2:58:01,  3.12it/s, train_loss=4.35]
No DDP Fp16   9%|██████████▊                                                                                                           | 3270/35658 [12:30<2:11:16,  4.11it/s, train_loss=1.81]

A100 80 Gb 
> Fp32  8%|████████▉                                                                                                             | 2688/35658 [12:32<2:33:50,  3.57it/s, train_loss=1.28]
Fp16  13%|███████████████▍                                                                                                      | 4680/35658 [17:19<1:55:32,  4.47it/s, train_loss=1.41]
Bf16  11%|████████████▋                                                                                                         | 3842/35658 [12:47<1:53:41,  4.66it/s, train_loss=2.82]

We can clearly see that now the instabilities issues are tackled down. Now, you can train a CommonVoice CTC ASR recipe with Fp16 without any problems. 

The issue was located at the level of GradScaler unscale_ and gradient clipping. GradScaler was returning some NaNs gradient (which is expected), and when fed to the gradient clipping it was leading to NaN issues. Setting the gradient to zero is effectively solving the issue. 

Furthermore, now there is no more `auto_mix_prec` or `bfloat16_mix_prec` flags. There are merged in favour of a new flag called `autocast` which expect a string. By default this is `fp32`. It is linked to the `AutocastConfig` obj (https://github.com/speechbrain/speechbrain/blob/15a9e824cfaec4d26f7548a9de6bd7d302f62a52/speechbrain/core.py#L57). By default, if `fp32` is given, GradScaler is set to `False` (enabled=False). So no need to modify anything in the code structure. If `fp16` or `bf16` is given then GradScaler enabled=True. 

Regarding DDP, now everything is wrapped inside of `self.no_sync(not should_step)` to comply with PyTorch documentation. 

Finally, I am introducing a new function called `freeze_optimizers` which is returning the current active optimisers. The idea is to override this function if one wants to freeze some optimisers. The function takes in input the dict of current optimisers and return the active optimisers. 

in summary, what is happening now is the following: 
1. Fp16 is working. 
2. if a gradient is NaN or Inf we replace it to zero 
3. if the loss function is not finite there's still the patience mechanism being active
4. One flag for all: `autocast` 
5. DDP is fixed as everything is wrapped inside of the no_sync region. 
6. remove the counter `valid_step`. Indeed, I found that no toolkit is really looking at the valid steps (e.g., ESPNet, NeMO...). Using `valid_step` is adding some extra complexity to the code, and the case where a loss is NaN/Inf is not really common. It means we can accept that some optimisers step are done with a number of losses < grad_accumulation_factor. 

What is remaining ? 
1. Change all ASR recipes to match the new interface ; I propose to only cover Aishell / CommonVoice / LibriSpeech as a first step. We can latter (in another PR) change all the other recipes (this PR is not backward incompatible). 
2. Clean everything (i.e., add docstring for instance) 
3. Get feedbacks from the core team CC: @pplantinga @TParcollet @mravanelli @Gastron

pplantinga:
>  If `fp16` or `bf16` is given then GradScaler `enabled=True`.

One minor point here: the scaler should not be needed at all if the autocast target is `bf16`. The exponent has the same number of bits so there's no risk of over/underflow.

speechbrain/core.py
```diff
@@ -1050,100 +1069,45 @@ def fit_batch(self, batch):
...
-        should_step : boolean
-            Whether optimizer.step() was called or not.
-        """
-        pass
+    def check_gradients(self):
+        """ Checks if the gradients are finite. If not, it will emit a warning and set them to zero."""
+        for param in self.modules.parameters():
+            if param.requires_grad and param.grad is not None:
+                if not torch.isfinite(param.grad).all():
+                    param.grad = torch.zeros_like(param.grad)
```
pplantinga, line 1098: I think part of the original reason for checking the loss instead of the gradient was out of concern for the speed of the operation (under the theory that checking millions/billions of parameters could be slow). Do you know if this operation adds a noticeable amount of time to each step or is it fast?

Adel-Moumen: One note: I used `param.grad = None` instead of `param.grad = torch.zeros_like(param.grad)` (should be faster). And also, I don't know if you saw it, but NeMO is doing something a little bit different see: https://github.com/NVIDIA/NeMo/blob/1c10b3677040cde4785ea73f7d9ec887578885c7/nemo/collections/asr/models/asr_model.py#L120. What do you think about how they are handling NaN/infs gradients?

lucadellalib:
A potential minor issue with skipping the `backward()` if loss is NaN/Inf is that DDP could get out of sync. The loss might be finite in one process, but NaN/Inf in another one due to some bad input. This could lead to misalignment between the processes and end up in a deadlock because one process might have already consumed all the data, while another one is still waiting for synchronization.

speechbrain/core.py
Adel-Moumen: Btw, ce could turn `check_gradients` optional. We could add a new flag e.g., `skip_grad_nans` in SB to activate the check+replacement of grads.

pplantinga:
> A potential minor issue with skipping the `backward()` if loss is NaN/Inf is that DDP could get out of sync. The loss might be finite in one process, but NaN/Inf in another one due to some bad input. This could lead to misalignment between the processes and end up in a deadlock because one process might have already consumed all the data, while another one is still waiting for synchronization.

This is definitely something we should test. However I don't immediately see how this would get out of sync if we're just setting bad gradients to 0.

speechbrain/core.py
pplantinga: While 0.04s per iteration doesn't sound like a lot, perhaps it would still make sense to turn `check_gradients` to be optional just in case there's a weird scenario that it matters a lot.

Adel-Moumen: Yep did it.

Adel-Moumen:
ESPNet way to handle NaN grads: https://github.com/espnet/espnet/blob/5d0758e2a7063b82d1f10a8ac2de98eb6cf8a352/espnet/asr/pytorch_backend/asr.py#L301
Basically, they compute the grad norm (can be efficiently obtained with gradient clipping), and if it's nan then they skip the current batch (so they clear the gradients). This could be a better way to handle nan grads.

Adel-Moumen:
One note: they are not checking the loss value. Maybe it means that sometimes it is expected that the loss is NaN/inf?

Adel-Moumen:
Found something really strange. 

Basically, doing:
```python
grad_norm = torch.nn.utils.clip_grad_norm_(
            self.modules.parameters(), self.max_grad_norm
        )
```
will leads `grad_norm` to NaN/inf on a lot of recipes (e.g. CommonVoice CTC w2v), and will explode the training while  doing: 
```python
for opt in valid_optimizers.values():
            grad_norm = torch.nn.utils.clip_grad_norm_(
                opt.param_groups[0]["params"], self.max_grad_norm
            )
```
Will work. No NaNs/Infs.

pplantinga:
I wonder if the difference is whether a global norm is computed over all the parameters vs local norms computed for each layer.

Adel-Moumen:
> I wonder if the difference is whether a global norm is computed over all the parameters vs local norms computed for each layer.

Good question. I have to think about it ,but it might the case. 

Btw, I have removed the `check_loss_isfinite` (and thus, the patience mechanism) I think it doesn't make sense to have it. The primary reason is that in DDP, you will potentially have a deadlock as suggested by Luca, furthermore, you might end-up having less gradients accumulated than what you are expecting to have. And finally, if there's issues during the bwd, the gradscaler is supposed to deal with it. In a normal case when you are using fp32 and you are reaching nan loss, I believe the issue should be solved by having a look at the model structure/data, not by throwing away the issue through a check/patience mechanism (and I have been looking on many projects, no one is checking the loss, the only thing that is being done is to skip the batch if the loss is very huge).

speechbrain/core.py
```diff
@@ -1050,100 +1068,42 @@ def fit_batch(self, batch):
...
+            ):
+                outputs = self.compute_forward(batch, sb.Stage.TRAIN)
 
-    def on_fit_batch_end(self, batch, outputs, loss, should_step):
-        """Called after ``fit_batch()``, meant for calculating and logging metrics.
+        with self.no_sync(not should_step):
+            # The objectives are removed from Autocast to avoid
+            # potential numerical instabilities with CTC loss, etc.
+            loss = self.compute_objectives(outputs, batch, sb.Stage.TRAIN)
+            loss = loss / self.grad_accumulation_factor
```
lucadellalib, line 1087: Should'nt this be `scaled_loss = self.scaler.scale(loss / self.grad_accumulation_factor)`? Otherwise we return the loss value divided by `self.grad_accumulation_factor` instead of the original value

Adel-Moumen: True. Indeed. Thanks for the catch!

Adel-Moumen: One question: should we return the scaled_loss or the current loss with the gradscaler scaling ?

Adel-Moumen:
Ok good news for w2v CTC CommonVoice FR: 

With the new refactoring, in fp16 we finish one epoch in 1h33 (with a good train loss/wer) while in fp32 it takes up to 4h. With the previous version, it was impossible to use fp16 as it was exploding at the very beginning of the training. 

I am still working on transducer. I don't know yet if the changes are sufficient or if we need to have a better look at the transducer itself rather than the training component. 

Some training logs:
speechbrain.utils.train_logger - epoch: 1, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 3.67e-01 - valid loss: 1.51e-01, valid CER: 3.84, valid WER: 12.42
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-28+22-43-18+00
speechbrain.utils.epoch_loop - Going into epoch 2
speechbrain.utils.train_logger - epoch: 2, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.73e-01 - valid loss: 1.33e-01, valid CER: 3.54, valid WER: 11.46
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+00-22-34+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-28+22-43-18+00
speechbrain.utils.epoch_loop - Going into epoch 3
speechbrain.utils.train_logger - epoch: 3, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.51e-01 - valid loss: 1.30e-01, valid CER: 3.39, valid WER: 10.98
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+02-02-17+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+00-22-34+00
speechbrain.utils.epoch_loop - Going into epoch 4
speechbrain.utils.train_logger - epoch: 4, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.36e-01 - valid loss: 1.26e-01, valid CER: 3.28, valid WER: 10.61
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+03-41-31+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+02-02-17+00
speechbrain.utils.epoch_loop - Going into epoch 5
speechbrain.utils.train_logger - epoch: 5, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.27e-01 - valid loss: 1.26e-01, valid CER: 3.25, valid WER: 10.45
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+05-20-48+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+03-41-31+00
speechbrain.utils.epoch_loop - Going into epoch 6
speechbrain.utils.train_logger - epoch: 6, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.21e-01 - valid loss: 1.25e-01, valid CER: 3.17, valid WER: 10.20
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+06-59-58+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+05-20-48+00
speechbrain.utils.epoch_loop - Going into epoch 7
speechbrain.nnet.schedulers - Changing lr from 1 to 0.8
speechbrain.nnet.schedulers - Changing lr from 0.0001 to 9e-05
speechbrain.utils.train_logger - epoch: 7, lr_model: 1.00e+00, lr_wav2vec: 1.00e-04 - train loss: 1.15e-01 - valid loss: 1.25e-01, valid CER: 3.20, valid WER: 10.24
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+08-39-09+00
speechbrain.utils.epoch_loop - Going into epoch 8
speechbrain.utils.train_logger - epoch: 8, lr_model: 8.00e-01, lr_wav2vec: 9.00e-05 - train loss: 1.06e-01 - valid loss: 1.24e-01, valid CER: 3.16, valid WER: 10.20
speechbrain.utils.checkpoints - Saved an end-of-epoch checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+10-18-25+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+08-39-09+00
speechbrain.utils.checkpoints - Deleted checkpoint in results/wav2vec2_ctc_fr_fp16/1234/save/CKPT+2023-09-29+06-59-58+00
speechbrain.utils.epoch_loop - Going into epoch 9


100%|██████████| 17829/17829 [1:33:58<00:00,  3.16it/s, train_loss=0.367]

mravanelli:
Hi @Adel-Moumen ,
thank you for this PR. I've done an initial code inspection, but I still have to perform extensive testing. However, I'd like to share my preliminary feedback:
1. I would request that @pplantinga , @lucadellalib , and @Gastron ro conduct a careful review of this PR and provide their valuable insights. This is important, as any potential bugs introduced here could have a significant impact on our toolkit's functionality.
2. To mitigate the risks, I would recommend a substantial expansion of unit tests for the brain class. You can reference our existing unit tests in [this location](https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_core.py). Currently, we only perform basic tests, but we need to comprehensively test and validate the various options available.
3. It is essential to ensure that our recipes are compatible with both GPUs and CPUs. Running the recipes on CPUs is particularly important for our recipe tests, as we may need to run them on machines without GPUs in the future. Additionally, users may wish to train models using multiple CPUs if they have access to such resources. I'm fixing the recipes to make CPU-compliant, as evidenced in [this pull request](https://github.com/speechbrain/speechbrain/pull/2177).  Within this PR, I recommend revising the use of autocast and calling torch.cuda.amp.GradScaler only if the device is CUDA-enabled. I've implemented a similar approach in my own PR, which you can find [here](https://github.com/speechbrain/speechbrain/blob/3c69722c2d8d2a20d62c1fd3e20323237d699c87/speechbrain/core.py#L696). 
4. Don't forget to add comprehensive docstrings for the optimizers_step and AutocastConfig functions.
5. Lastly,  address any failing tests

speechbrain/core.py
```diff
@@ -1050,60 +1068,89 @@ def fit_batch(self, batch):
...
-                if should_step:
-                    self.optimizer.step()
-                    self.zero_grad()
-                    self.optimizer_step += 1
+        amp = AutocastConfig.from_name(self.autocast)
+        should_step = (self.step % self.grad_accumulation_factor) == 0
+
+        # We wrap the forward pass in no_sync to avoid
+        # unnecessary synchronization of gradients.
+        # Empirically, we found that this improves performance by a lot (2x speedup).
```
lucadellalib, line 1076: I'm not sure about this, according to https://discuss.pytorch.org/t/whats-no-sync-exactly-do-in-ddp/170259, forward pass and backward pass should be in the same `with self.no_sync` block to correctly skip gradient synchronization

Adel-Moumen: Yes, I do know. But as pointed out in the comments, some DDP models (e.g. HF w2v2) are slower when the `self.no_sync(not should_step)` is wrapper around the fwd. There's a real difference between `self.no_sync(not should_step)` and `self.no_sync()` around the fwd.

mravanelli:
Hi Adel,
if not already done, you can write a comment in the code to justify this
choice.

speechbrain/core.py
lucadellalib: I double checked using https://github.com/speechbrain/speechbrain/tree/develop/recipes/LibriSpeech/ASR/transducer, according to my tests this code leads to wrong results as it never synchronizes the gradients (gradient values should be identical in each rank every `grad_accumulation_factor` steps):

```python
with self.no_sync():
    with torch.autocast(
        dtype=amp.dtype, device_type=torch.device(self.device).type,
    ):
        outputs = self.compute_forward(batch, sb.Stage.TRAIN)

with self.no_sync(not should_step):
    # The objectives are removed from Autocast to avoid
    # potential numerical instabilities with CTC loss, etc.
    loss = self.compute_objectives(outputs, batch, sb.Stage.TRAIN)

    scaled_loss = self.scaler.scale(
        loss / self.grad_accumulation_factor
    )
    scaled_loss.backward()

for p in self.modules.parameters():
    print("RANK: " + os.environ["RANK"], "STEP: " + str(self.step), "GRAD: " + str(p.grad.flatten()[0:5]))
    break

if should_step:
    self.optimizers_step()
```

```bash
python -m torch.distributed.launch --nproc_per_node 2 \
train.py hparams/conformer_transducer.yaml --data_folder LibriSpeech \
--grad_accumulation_factor 4 --d_model 128 --use_torchaudio True \
--distributed_launch --find_unused_parameters
```

```
...
RANK: 0 STEP: 1 GRAD: tensor([-0.2559, -0.2251, -0.1947, -0.1615, -0.1247], device='cuda:0')
RANK: 1 STEP: 1 GRAD: tensor([-0.1381, -0.1090, -0.0870, -0.1038, -0.0626], device='cuda:1')
RANK: 0 STEP: 2 GRAD: tensor([-0.5477, -0.4795, -0.3994, -0.3519, -0.27181, device='cuda:0')
RANK: 1 STEP: 2 GRAD: tensor([-0.5832, -0.5016, -0.4143, -0.3648, -0.2632], device='cuda:1')
RANK: 0 STEP: 3 GRAD: tensor([-0.7123, -0.6151, -0.5048, -0.4761, -0.3532], device='cuda:0')
RANK: 1 STEP: 3 GRAD: tensor([-0.8210, -0.6935, -0.5572, -0.5350, -0.3820], device='cuda:1')
--> RANK: 0 STEP: 4 GRAD: tensor([-0.7952, -0.6520, -0.5222, -0.5171, -0.3215], device='cuda:0")
--> RANK: 1 STEP: 4 GRAD: tensor([-0.8469, -0.6987, -0.5467, -0.5705, -0.3754], device='cuda:1')
```

```python
with self.no_sync(not should_step):
    with torch.autocast(
        dtype=amp.dtype, device_type=torch.device(self.device).type,
    ):
        outputs = self.compute_forward(batch, sb.Stage.TRAIN)

    # The objectives are removed from Autocast to avoid
    # potential numerical instabilities with CTC loss, etc.
    loss = self.compute_objectives(outputs, batch, sb.Stage.TRAIN)

    scaled_loss = self.scaler.scale(
        loss / self.grad_accumulation_factor
    )
    scaled_loss.backward()

for p in self.modules.parameters():
    print("RANK: " + os.environ["RANK"], "STEP: " + str(self.step), "GRAD: " + str(p.grad.flatten()[0:5]))
    break

if should_step:
    self.optimizers_step()
```

```bash
python -m torch.distributed.launch --nproc_per_node 2 \
train.py hparams/conformer_transducer.yaml --data_folder LibriSpeech \
--grad_accumulation_factor 4 --d_model 128 --use_torchaudio True \
--distributed_launch --find_unused_parameters
```

```
...
RANK: 0 STEP: 1 GRAD: tensor([-0.2559, -0.2251, -0.1947, -0.1615, -0.1247], device='cuda:0')
RANK: 1 STEP: 1 GRAD: tensor([-0.1381, -0.1090, -0.0870, -0.1038, -0.0626], device='cuda:1')
RANK: 0 STEP: 2 GRAD: tensor([-0.5477, -0.4795, -0.3994, -0.3519, -0.27181, device='cuda:0')
RANK: 1 STEP: 2 GRAD: tensor([-0.5832, -0.5016, -0.4143, -0.3648, -0.2632], device='cuda:1')
RANK: 0 STEP: 3 GRAD: tensor([-0.7123, -0.6151, -0.5048, -0.4761, -0.3532], device='cuda:0')
RANK: 1 STEP: 3 GRAD: tensor([-0.8210, -0.6935, -0.5572, -0.5350, -0.3820], device='cuda:1')
--> RANK: 0 STEP: 4 GRAD: tensor([-0.8211, -0.6754, -0.5344, -0.5438, -0.3485], device='cuda:0')
--> RANK: 1 STEP: 4 GRAD: tensor([-0.8211, -0.6754, -0.5344, -0.5438, -0.3485], device='cuda:1')
```

lucadellalib: We should further investigate why we have a slowdown with w2v2, do you have a script/recipe to reproduce it?

Adel-Moumen: Thanks Luca for the details. I was not aware of this behaviour w.r.t gradients. 

Yes, you just have to run the w2v CommonVoice CTC recipe with and without the no_sync() around the fwd. You need to use DDP ofc (i.e., more than 2gpus).

speechbrain/core.py
```diff
@@ -1050,60 +1068,89 @@ def fit_batch(self, batch):
...
+
+        # 3. clip gradients
+        for opt in valid_optimizers.values():
+            torch.nn.utils.clip_grad_norm_(
+                opt.param_groups[0]["params"], self.max_grad_norm
+            )
+
+        # no need to activate this flag if you are in fp16
+        # since GradScaler is automatically handling the nonfinite gradients
+        if self.skip_nan_grad:
```
lucadellalib, line 1136: `self.skip_nan_grad` is undefined

Adel-Moumen: Yes. But as I said, I did not implemented everything. See: https://github.com/speechbrain/speechbrain/pull/2010#issuecomment-1740850106 The idea was just to provide a higher view of what is expecting to change. I wanted to first hear from the core team.

speechbrain/core.py
```diff
@@ -1050,60 +1068,89 @@ def fit_batch(self, batch):
...
+        # We wrap the forward pass in no_sync to avoid
+        # unnecessary synchronization of gradients.
+        # Empirically, we found that this improves performance by a lot (2x speedup).
+        with self.no_sync():
+            with torch.autocast(
+                dtype=amp.dtype, device_type=torch.device(self.device).type,
+            ):
+                outputs = self.compute_forward(batch, sb.Stage.TRAIN)
+
+        with self.no_sync(not should_step):
```
lucadellalib, line 1083: Referring to [this](https://github.com/Adel-Moumen/speechbrain/blob/c1a23a1a672ac7c8f7bb270b99b7282daa9cfa72/speechbrain/core.py#L1600), the current version of `no_sync` leads to a silent bug when using DDP + gradient accumulation. We `break` from the loop as soon as we find a module without `require_backward_grad_sync`, i.e. not wrapped in DDP:

```python
if use:
    old_values_list = []
    for module in self.modules.values():
        if not hasattr(module, "require_backward_grad_sync"):
            # if not using DDP
            break
        old_values_list.append(module.require_backward_grad_sync)
        module.require_backward_grad_sync = False
    yield
    for module, old_value in zip(
        self.modules.values(), old_values_list
    ):
        if not hasattr(module, "require_backward_grad_sync"):
            break
        module.require_backward_grad_sync = old_value
else:
    yield
```

The assumption here is that either all modules are wrapped in DDP or none of them. But this is not the case in the majority of the recipes. We typically include in the YAML `modules` the feature extractor, the global normalization, the augmentation, frozen wav2vec2 etc., which are `torch.nn.Module` but without parameters or with all parameters frozen. These modules are not wrapped in DDP (see [here](https://github.com/Adel-Moumen/speechbrain/blob/c1a23a1a672ac7c8f7bb270b99b7282daa9cfa72/speechbrain/core.py#L1484)). Therefore `no_sync` unsets `require_backward_grad_sync` for each DDP module until a non-DDP module is found while iterating `self.modules` => when using DDP + gradient accumulation, some gradients are correctly synchronized every `grad_accumulation_factor` steps, other gradients are unnecessarily synchronized at each step. This can cause slowdowns and maybe even convergence issues. Something like this should fix the issue:

```python
if use:
    old_values_list = []
    for module in self.modules.values():
        if not hasattr(module, "require_backward_grad_sync"):
            # if not using DDP
            continue
        old_values_list.append(module.require_backward_grad_sync)
        module.require_backward_grad_sync = False
    yield
    i = 0
    for module in self.modules.values():
        if not hasattr(module, "require_backward_grad_sync"):
            continue
        module.require_backward_grad_sync = old_values_list[i]
        i += 1
else:
    yield
```

speechbrain/core.py
lucadellalib: I think it's OK to return the current loss in every case

speechbrain/core.py
```diff
@@ -296,16 +315,11 @@ def parse_arguments(arg_list=None):
...
-        default=None,
-        action="store_true",
-        help="This flag enables training with automatic mixed-precision.",
-    )
-    parser.add_argument(
-        "--bfloat16_mix_prec",
-        default=None,
-        action="store_true",
-        help="This flag enables training with bfloat16 mixed-precision.",
+        "--autocast",
```
lucadellalib, line 318: As for naming, I would prefer the old `auto_mix_prec` flag and `AutoMixPrecConfig` or `AMPConfig` for a few reasons:
- the meaning of `autocast` is less intuitive than `auto_mix_prec` for non-expert users
- if we keep the old `auto_mix_prec` flag name, it might be easier to make the new changes backward compatible at some extent
- automatic mixed precision and autocast are not exactly the same thing. You need both autocast and a gradient scaler to implement AMP training. Furthermore, you could train in standard precision and use autocast at inference time only to reduce memory/latency (https://pytorch.org/tutorials/recipes/recipes/amp_recipe.html#inference-evaluation). In our case the goal of this flag is to enable AMP training, but not to enforce a specific precision for validation or test.

speechbrain/core.py
Adel-Moumen: I think you are right about convergence issues. See the following tests that I did with CommonVoice FR + w2v2 + 2x GPUs (rtx 2080 ti):
DDP with no_sync(not should_step): 
  6%|██████▊                                                                                                                  | 999/17829 [11:10<3:08:17,  1.49it/s, train_loss=2.17]
DDP with no_sync(not should_step) + fix 
  6%|██████▊                                                                                                                  | 999/17829 [11:13<3:09:14,  1.48it/s, train_loss=2.15]
DDP with no_sync()
  6%|██████▊                                                                                                                  | 999/17829 [10:05<2:50:01,  1.65it/s, train_loss=3.62]

I stopped all training at 999 steps. The first is when we are wrapping the fwd inside the no_sync(not should_step) without your fix, the second test is with wrapping + your fix, and the last one is what I initially proposed (i.e. fwd in self.no_sync()). 

We can see that in the first two cases, the train loss is much lower (2.17) than the last one (3.62). However, I'm not seeing any improvements in terms of speedup yet…

Adel-Moumen: With 1x rtx 2080 ti:
  6%|██████▋                                                                                                                 | 1999/35658 [20:14<5:40:57,  1.65it/s, train_loss=1.53]
  
  I stopped at 1999 to simulate the same number of batches.

Adel-Moumen: So it seems that having the `no_sync` is a bit faster but not that much. However, it yields to a very bad train loss.

speechbrain/core.py
```diff
@@ -542,9 +585,9 @@ def __init__(  # noqa: C901
...
             "compile_using_fullgraph": False,
             "compile_using_dynamic_shape_tracing": False,
-            "auto_mix_prec": False,
-            "bfloat16_mix_prec": False,
+            "precision": "fp32",
             "max_grad_norm": 5.0,
+            "skip_nonfinite_grads": False,
             "nonfinite_patience": 3,
             "noprogressbar": False,
             "ckpt_interval_minutes": 0,
```
pplantinga, line 593: I'd like to see the run opts become a dataclass at some point, not sure if this PR is the right time for it.

speechbrain/core.py
```diff
@@ -1024,60 +1093,92 @@ def fit_batch(self, batch):
...
+        """Performs a step of gradient descent on the optimizers. This method
+        is called every ``grad_accumulation_factor`` steps."""
+        # 1. get the valid optimizers, i.e., the ones that are not frozen during this step
+        valid_optimizers = self.freeze_optimizers(self.optimizers_dict)
+
+        # 2. unscale the gradients of the valid optimizers
+        for opt in valid_optimizers.values():
+            self.scaler.unscale_(opt)
+
+        # 3. clip gradients
```
pplantinga, line 1156: We should add a note here about why we're clipping this way.

speechbrain/core.py
```diff
@@ -1024,60 +1093,92 @@ def fit_batch(self, batch):
...
+            torch.nn.utils.clip_grad_norm_(
+                opt.param_groups[0]["params"], self.max_grad_norm
+            )
+
+        # no need to activate this flag if you are in fp16
+        # since GradScaler is automatically handling the nonfinite gradients
+        if not self.scaler.is_enabled() and self.skip_nonfinite_grads:
+            self.check_gradients()
+
+        # 4. step the valid optimizers
```
pplantinga, line 1167: Can we add a note here that if the scaler is disabled, it simply calls the underlying optimizer's step function? I was confused at first about what was going on in the normal (non-scaled) case.

speechbrain/core.py
pplantinga: Maybe even put in a separate file.

speechbrain/core.py
```diff
@@ -1024,60 +1093,92 @@ def fit_batch(self, batch):
...
+        Override this method if you want to freeze some optimizers
+        during training. To do so, return a of active optimizers.
+        """
+        return optimizers
+
+    def on_optimizers_step_end(self):
+        """Gets called at the end of ``optimizers_step()``.
+        Override this method to do something at the end of ``optimizers_step()``.
+        """
+        pass
```
pplantinga, line 1144: Are there cases that `on_optimizers_step_end()` solves that `on_fit_batch_end()` does not? For learning rate updates etc, the latter should work just fine, no?

pplantinga (commented):
I'm always happy to see improvements to `core.py` and I think this definitely qualifies. I'd be happy to see more changes -- moving `Brain` class to its own file, separating `run_opts` into a dataclass, etc. But I'm not sure if this is the right place for those changes.

speechbrain/core.py
Adel-Moumen: Good point. What surprises me the most is that no one actually is using `on_fit_batch_end()`. Thanks for raising this point. I will use the function and delete `on_optimizers_step_end()`. 

One question though: do you think we should move https://github.com/speechbrain/speechbrain/pull/2010/files#diff-d2ad56ee927e543eceb0f9528eea488134956a963c60d845d72a92d69ae0187aR1118 inside `on_fit_batch_end()`? I guess it make sense from a semantic point of view. We are using opt steps at the end of the fit_batch.

pplantinga: The `on_fit_batch_end` was added some time after the first release, so possibly at the time that some overrides were created, the function was not available yet.

pplantinga: If we put the call to `optimizer_step` inside `on_fit_batch_end` then I suppose people could override to add code right before it's called as well as right after. I'm not really sure how often that would be needed.

pplantinga: The nice part about having small (or empty) functions to be overridden is that it reduces code copying

Adel-Moumen: Hmm Yes. Actually what I'm thinking about is that, in some cases people are using LR annealing. Which means, they have to override `on_fit_batch_end`. If we move `optimizer_step` inside `on_fit_batch_end` then it means they will have to add `optimizer_step` as well. Maybe it's better to leave it in `fit_batch` and use `on_fit_batch_end` only for specific cases like LR annealing. What do you think?

Adel-Moumen: BTW, side note, I changed the name of the flag to be `--precision` instead of `--autocast`. I think this is more intuitive to everyone, furthermore, autocast for many people, including me, means fp16/bf16 not fp32. PyTorch Lightning is also using `--precision`.

speechbrain/core.py
```diff
@@ -1024,122 +1088,117 @@ def fit_batch(self, batch):
...
+                    logger.warning(
+                        f"Gradients {param.name} contain NaN or Inf. Setting to None."
+                    )
+
+    def freeze_optimizers(self, optimizers):
+        """By default, this method returns the passed optimizers.
+        Override this method if you want to freeze some optimizers
+        during training. To do so, return a of active optimizers.
+        """
+        return optimizers
```
Gastron, line 1133: It looks like this step is not really necessary in the recipes that this PR changes. The `init_optimizers` step already drops the wav2vec 2.0 optimizer.

This sort of dynamic checking at every batch could support some wacky training schemes I guess, or is there a reason for this that I don't see?

recipes/AISHELL-1/ASR/transformer/train.py
```diff
@@ -122,37 +122,17 @@ def compute_objectives(self, predictions, batch, stage):
...
-            self.check_gradients(loss)
-
-            self.optimizer.step()
-            self.optimizer.zero_grad()
-
-            # anneal lr every update
+    def on_fit_batch_end(self, batch, outputs, loss, should_step):
+        """At the end of the optimizer step, apply noam annealing."""
+        if should_step:
             self.hparams.noam_annealing(self.optimizer)
```
Gastron, line 134: When the annealing happens at the end of `fit_batch`, doesn't the very first batch use a non-annealed LR?

I think I ran into some problems with this in my own recipes, where the first step uses a large LR and takes way too big of a step (gradients can be very large at first step too).

This connected to the issue of SpeechBrain not using the PyTorch scheduler interface.

Gastron: Not that this applies to the other recipes as well, I just used this as an example, so if I am right, this could be an important thing to fix.

Gastron (commented):
Nice work, this solves a common pain point, where we end up overriding fit_batch in so many places.

This could be the time to move to the [PyTorch standard LR scheduling interface](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate). This would be a big change though, requiring that the scheduler objects are created after the optimizers (i.e. probably schedulers would also be created during `init_optimizers`). It's just unfortunate that some schedulers need to be `step()`ed every batch and some every epoch (I wish there were e.g. separate batch_step() and epoch_step() methods on them that just handled this in a unified way).

There's a related issue, that I've pointed out below, where the Noam scheduler is called after the first optimizer step (so that the first step is done with a big LR, I am not mistaken). The PyTorch interface indeed calls schedulers after optimizer steps, but if we want to keep our interface, I think we should just call the lr scheduler during `on_fit_batch_start`.

pplantinga:
> the first step is done with a big LR, I am not mistaken

Show of confidence from the man himself

speechbrain/core.py
Adel-Moumen: The main reason is for recipes that are using warmup for the opt. Many w2v recipes are waiting N steps before doing w2v opt steps. So the purpose of this fn is to deal with this exact problem.

recipes/AISHELL-1/ASR/transformer/train.py
Adel-Moumen: Hmmmm how would you deal with it? Doing it at the beginning of the `fit_batch` ? Also, the LR annealing is only applied every `should_step`.

Adel-Moumen: OK. Sorry I didn't read at first your big comment :p. It could be a way. Using LR annealing at the beginning of the fit_batch fn. I propose with this PR to solely target the mentioned issue (i.e., DDP, precision, overriding). We could think about adding a better interface with PyTorch scheduling in another PR. But I tend to fully agree with you.

pplantinga:
> So many recipes 0.0

😂

recipes/AISHELL-1/ASR/transformer/train.py
Gastron: Yeah, I think the simple fix would be to update the LR at the beginning of fit_batch ( conditional on should_step of course, but before the first optimizer step).

Adel-Moumen:
I fixed all the recipes + the pre-commit are passing.

Now, I'm going to run the recipe tests and will try to solve every little issues (I spotted some). 

Feel free to do another round of review (@mravanelli @pplantinga ) if possible please so that we can gain some time ;) Thanks!! :)

speechbrain/nnet/activations.py
```diff
@@ -33,15 +35,19 @@ class Softmax(torch.nn.Module):
...
 
-    def __init__(self, apply_log=False, dim=-1, reshape=True):
+    def __init__(
+        self, apply_log=False, dim=-1, reshape=True, dtype=torch.float32
+    ):
         super().__init__()
 
         if apply_log:
-            self.act = torch.nn.LogSoftmax(dim=dim)
+            self.act = F.log_softmax
```
mravanelli, line 44: This should be changes as it does not allow users to specify the dimension on which softmax should be applied.

mravanelli (changes requested):
I did another quick code inspections and I only noticed the issue on log softmax mentioned in the comment. To further test this PR, I plan to do the following:
1- run recipe tests
2- run full training of TIMIT recipe and compare results with current Dev
3- run the first epochs of some LibriSpeech recipes and compare the results wit Dev

speechbrain/nnet/activations.py
mravanelli: (there is something similar below)

Adel-Moumen: > This should be changes as it does not allow users to specify the dimension on which softmax should be applied.

Are you sure? Below, in the fwd,  I used `x_act = self.act(x, dim=self.dim, dtype=self.dtype)` so I do specify the dim which can be changed in the `self.__init__` method.

speechbrain/nnet/activations.py
```diff
@@ -61,7 +67,7 @@ def forward(self, x):
             if len(dims) == 4:
                 x = x.reshape(dims[0] * dims[1], dims[2], dims[3])
 
-        x_act = self.act(x)
+        x_act = self.act(x, dim=self.dim, dtype=self.dtype)
```
Adel-Moumen, line 70: CC: @mravanelli

mravanelli:
Regarding this pull request, I recommend the following steps:
1. We've recently merged a few new recipes, including TedLium, CVSS, and the new fastspeech hifigan with quantization. To proceed, we should merge the development branch into both the unstable and unstable branches and update the new recipes accordingly.
2. I have conducted recipe tests, and everything appears to be working correctly, except  the known issues caused by the outdated HF repository. There are a few other tests that have failed:
- DVoice_row_03 DVoice_row_06 CommonVoice_row: `IndexError: Out of range - the piece ID is out of range`.

- CommonVoice18: UnicodeEncodeError -` 'ascii' codec can't encode character '\u062a' at position 1: ordinal not in range(128)`.
However, these errors are unrelated to this pull request, and I suggest addressing them in a separate pull request.

I am currently running additional tests for TIMIT and LibriSpeech training to verify that everything performs as expected.

mravanelli:
Hi @Adel-Moumen, I ran the recipe test and pushed some fixes. However, there are still some recipe tests failing (mainly because the new recipes should be converted to the new fit_batch). In particular, I detected the following issues:
- `recipes/CVSS/S2ST/hparams/train_fr-en.yaml` => convert to new fit_batch
- `recipes/IWSLT22_lowresource/AST/transformer/hparams/*` => convert to new fit_batch
- `recipes/LJSpeech/TTS/vocoder/hifi_gan_unit/hparams/train.yaml`  => convert to new fit_batch
- `recipes/Tedlium2/ASR/transformer/hparams/branchformer_large.yaml` => convert to new fit_batch

- ` recipes/LibriTTS/TTS/mstacotron2/hparams/train.yaml`. Here the issue is the following:

```
  File "/workspace/speechbrain/recipes/LibriTTS/TTS/mstacotron2/train.py", line 644, in <module>
    tacotron2_brain.fit(
  File "/workspace/speechbrain/speechbrain/core.py", line 1421, in fit
  File "/workspace/speechbrain/speechbrain/core.py", line 1335, in _fit_valid
    self.on_stage_end(Stage.VALID, avg_valid_loss, epoch)
  File "/workspace/speechbrain/recipes/LibriTTS/TTS/mstacotron2/train.py", line 363, in on_stage_end
    train_stats=self.last_loss_stats[sb.Stage.TRAIN],
                ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: <Stage.TRAIN: 1>
```

Adel-Moumen:
> Hi @Adel-Moumen, I ran the recipe test and pushed some fixes. However, there are still some recipe tests failing (mainly because the new recipes should be converted to the new fit_batch). In particular, I detected the following issues:
> 
> * `recipes/CVSS/S2ST/hparams/train_fr-en.yaml` => convert to new fit_batch
> * `recipes/IWSLT22_lowresource/AST/transformer/hparams/*` => convert to new fit_batch
> * `recipes/LJSpeech/TTS/vocoder/hifi_gan_unit/hparams/train.yaml`  => convert to new fit_batch
> * `recipes/Tedlium2/ASR/transformer/hparams/branchformer_large.yaml` => convert to new fit_batch
> * ` recipes/LibriTTS/TTS/mstacotron2/hparams/train.yaml`. Here the issue is the following:
> 
> ```
> Traceback (most recent call last):
>   File "/workspace/speechbrain/recipes/LibriTTS/TTS/mstacotron2/train.py", line 644, in <module>
>     tacotron2_brain.fit(
>   File "/workspace/speechbrain/speechbrain/core.py", line 1421, in fit
>     self._fit_valid(valid_set=valid_set, epoch=epoch, enable=enable)
>   File "/workspace/speechbrain/speechbrain/core.py", line 1335, in _fit_valid
>     self.on_stage_end(Stage.VALID, avg_valid_loss, epoch)
>   File "/workspace/speechbrain/recipes/LibriTTS/TTS/mstacotron2/train.py", line 363, in on_stage_end
>     train_stats=self.last_loss_stats[sb.Stage.TRAIN],
>                 ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
> KeyError: <Stage.TRAIN: 1>
> ```

Hello @mravanelli, thanks for running the tests! 

For CVSS : the issue is not related to this PR but from the fact that the HF hub is not yet sync with the new beam search : 
> ImportError: There is no such class as speechbrain.decoders.seq2seq.S2STransformerBeamSearch

For IWSLT22_lowresource, one issue was related to this PR, and the others were related to the keyword `--skip_prep` that was set twice in `tests/recipes` : 
> (1/6) Running test for IWSLT22_lowresource_row_02...
        ... 27.61s
(2/6) Running test for IWSLT22_lowresource_row_03...
        ... 91.50s
(3/6) Running test for IWSLT22_lowresource_row_04...
        ... 245.26s
(4/6) Running test for IWSLT22_lowresource_row_05...
        ... 96.67s
(5/6) Running test for IWSLT22_lowresource_row_06...
        ... 146.97s
(6/6) Running test for IWSLT22_lowresource_row_07...
        ... 386.77s
TEST PASSED

For LJSPEECH : done. It was related to a missing dict. 

For Tedlium : done. We were using the wrong Beam Search class / passing an outdated parameter in the yaml file / using device in the load_collected() method / using device in the avg checkpoint method / and we were not handling the return output of the new beam search correctly.  
> (1/2) Running test for Tedlium2_row_02...
        ... 4.56s
(2/2) Running test for Tedlium2_row_03...
        ... 28.34s
TEST PASSED



For mstacotron2 : it does not seems to be linked to my PR :/

mravanelli:
I did the last tests and everything looks now fine to me. Thank you @Adel-Moumen for working on this PR. This is another remarkable step toward speechbrain 1.0.

PR #2015, opened 2023-06-06
Fixing multiple device bug : Update interfaces.py

Siddharth1698:
Fixing Error :
expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument tensors in method wrapper_CUDA_cat)

# Contribution in a nutshell
Fixing bug in codebase. 

# Notes for reviewing 
 When this part of code is activated, it tries to use both cuda and cpu if we have used cuda in hyperparams. This is because, this part of code has hardcoded with cpu and only runs with cpu. So this fix will take care of this issue.

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -1457,7 +1457,14 @@ class VAD(Pretrained):
         # Fix edge cases (when a speech starts in the last frames)
         if (prob_th == 1).nonzero().shape[0] % 2 == 1:
             prob_th = torch.cat(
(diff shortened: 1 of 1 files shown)


Adel-Moumen:
Hello @Siddharth1698, 

thanks for this PR! Could you please share a working example of this issue/fix please?

Many thanks.

Adel-Moumen (approved):
LGTM! 

I was able to test and reproduce the issue. This PR is indeed fixing the issue.

Siddharth1698:
Thanks for the merge. 😄✌

Issue #1936, opened 2023-04-12
[Feature Request]: Non-finite gradient can be accumulated during the gradient accumulation


jaeyeonkim99:
### 🚀The feature

![image](https://user-images.githubusercontent.com/84221374/231334266-18acba61-d621-46a5-abd7-ca52b7a59766.png)

In the current `fit_batch()` function of the `Brain` class, `check_graidents()` is called only on the `should_step`(the step for optimization after the gradient is accumulated for the gradient accumulation steps).
However, if the loss is not finite on non-should steps, 'check_gradients()' isn't called so the non-finite gradients are accumulated. Then if the loss is finite on the following 'should_step', the 'optimzation.step()' will run with accumulated non-finite gradients. 

### Solution outline

Now, the 'check_graidents()' performs 1. check if the loss is finite 2. gradient clipping.
I think these two functionalities should be separated. Checking if the loss is finite should be done for every step(including both should and non-should steps) before `backward()` is called. Then, gradient clipping should be called for the `should_step` only, 

It should be something like this. 
![image](https://user-images.githubusercontent.com/84221374/231342602-cea933f1-2474-4096-af35-623478502237.png)

Adel-Moumen:
Hi @jaeyeonkim99,

Thanks for spotting this. 

Could you please open a PR on that? 

Many thanks

jaeyeonkim99:
No problem.
I will open a PR on this soon.

jaeyeonkim99:
I was busy, and just open a PR on this.

jaeyeonkim99:
I've just open the PR on this issue.

Adel-Moumen:
Solved in #2018. Thanks gain for the PR! :-)

PR #2018, opened 2023-06-07
Change gradient accumulation logic

jaeyeonkim99:
This is the pull request corresponding to issue https://github.com/speechbrain/speechbrain/issues/1936.

I changed the logic to check the loss for every forward step to prevent non-finite gradient being accumulated.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -930,7 +930,7 @@ class Brain:
         detached loss
         """
         valid_loss = False
-    
+
         # Managing automatic mixed precision
         if self.auto_mix_prec:
             with torch.autocast(device_type=torch.device(self.device).type):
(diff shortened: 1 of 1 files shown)


Adel-Moumen:
Many thanks! It does also solve the issue #1803

speechbrain/core.py
```diff
@@ -1026,13 +1039,19 @@ def check_gradients(self, loss):
...
+        Automatically clips large gradients.
+        """
         # Clip gradient norm
         if self.max_grad_norm > 0.0:
             torch.nn.utils.clip_grad_norm_(
                 (p for p in self.modules.parameters()), self.max_grad_norm
             )
 
-        return True
+        return
```
Adel-Moumen, line 1054: why do you return something? You can remove this line.

Adel-Moumen (changes requested):
Small change but everything else looks good. 

As explained earlier, this fix also another bug with fp16. I'm currently adding the new changes on all the others recipes such as CommonVoice.

Adel-Moumen:
Please have a look at #2010 I added your change and I changed every recipes that was relying on the check_loss_isfinite fn by adding the gradient clipping.

PR #2037, opened 2023-06-19
Reach 2.0 on conformer and retrain all (code from Samsung AI Cambridge)

TParcollet:
# Contribution in a nutshell

Reach 2.0 on test-clean with the conformer. The key was label smoothing, but I also changed a few hyperparameters.

# Scope
* [x] Reach 2.0 on test-clean with conformer.
* [x] Re-train all models from LibriSpeech and update the numbers and models.


recipes/LibriSpeech/ASR/transformer/hparams/branchformer_large.yaml
```diff
@@ -255,10 +256,10 @@ augmentation: !new:speechbrain.lobes.augment.SpecAugment
     freq_mask: True
     n_freq_mask: 2
     time_mask: True
-    n_time_mask: 3
+    n_time_mask: 10
```
popcornell, line 259: this is interesting, that is a lot of masks. It is fine I guess because LibriSpeech mean utternace length is like around 5-6 seconds IRCC. 
A better approach would be variable length SpecAugment. 
So the main issue was overfitting. Label smoothing and more specaugment fixed it.

popcornell: I have an implementation for variable length specaugm on local. I am too lazy for a PR. Do you want to do it ?

recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml
```diff
@@ -191,7 +192,7 @@ model: !new:torch.nn.ModuleList
     - [!ref <CNN>, !ref <Transformer>, !ref <seq_lin>, !ref <ctc_lin>]
 
 # define two optimizers here for two-stage training
-Adam: !name:torch.optim.Adam
+Adam: !name:torch.optim.AdamW
```
popcornell, line 195: do you also use weight decay ?

popcornell (approved):
LGTM, made a small change

recipes/LibriSpeech/ASR/transformer/hparams/transformer.yaml
TParcollet: AdamW uses it by default yeah.

recipes/LibriSpeech/ASR/transformer/hparams/branchformer_large.yaml
TParcollet: This one I don't know, maybe it's too much... I'd love a variable spec augment. Could you PR and we update?

Adel-Moumen (approved):
LGTM! Model on the Hub does work.

PR #2041, opened 2023-06-21
Basic SECURITY.md

Gastron:
It is good to have at least some rudimentary security policy set up for SpeechBrain. The GitHub template suggests two sections: Supported versions and Reporting a vulnerability.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,10 @@
+# Security Policy
+
+## Supported Versions
+
+SpeechBrain aims to support the latest major version (at x.x level) with security updates.
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,10 @@
+# Security Policy
+
+## Supported Versions
+
+SpeechBrain aims to support the latest major version (at x.x level) with security updates.
```
Adel-Moumen, line 5: "(at x.x level)" ?

Gastron: Ah maybe x.y is better. I mean like the latest 0.5 will be supported until we release 0.6, and then that will be the one we focus on.

Gastron:
I have made the security updates paragraph a bit clearer.

mravanelli (approved):

PR #2045, opened 2023-06-22
Fix deprecation warning for STFT

mravanelli:
This PR fixes the following deprecation warning: 
`/home/torch/functional.py:641: UserWarning: stft with return_complex=False is deprecated. In a future pytorch release, stft will return complex tensors for all inputs, and return_complex=False will raise an error.`

I got ii with torch 2.0, but it might show up even with earlier version. I tested and everything seems to work properly.


Adel-Moumen (approved):
LGTM. Ty.

Issue #1914, opened 2023-03-30
[Feature Request]: Early stopping supported by fit


Jeronymous:
### 🚀The feature

The validation dataset can be used for model selection, but also for early stopping.
In particular, early stopping prevents from guessing a fixed number of epochs in advance : training stops when the validation loss/error rate decreased for a certain amount of steps.
I think it's a must-have feature (for things like AutoML) that should be straightforward to use.

### Solution outline

There is already `EpochCounterWithStopper`, but passing it to `speechbrain.Brain.fit()` does not do anything special.
It just iterates over the epoch count until the end
```
        # Iterate epochs
        for epoch in epoch_counter:
            self._fit_train(train_set=train_set, epoch=epoch, enable=enable)
            self._fit_valid(valid_set=valid_set, epoch=epoch, enable=enable)
            ...
```
A solution would be to add the method `should_stop(self, current, current_metric)` to the base class `EpochCounter` (always returning False for this class), and use it where appropriate in the `fit()` function.

### Additional context

There was this issue opened a while ago: https://github.com/speechbrain/speechbrain/issues/962

But the link given by @mravanelli is broken
> As far as I remember we already have some recipes using early stopping. You can take a look at this one for instance:
> https://github.com/speechbrain/speechbrain/blob/develop/recipes/ERPCore/P3_decoding/train.py

I looked at the current recipes and did not find "real" early stopping implemented (only early stopping when the loss are not finite).

Adel-Moumen:
Hello @Jeronymous, thank you once again for submitting new feature requests.

Based on your request, it seems like @mravanelli and @TParcollet would be the most appropriate individuals to address your needs.

TParcollet:
@pplantinga and @Gastron might want to have a look as well.

Issue #1939, opened 2023-04-12
[Bug]: DDP with epochCounterWithStopper leads to CUDA timeouts


matthewkperez:
### Describe the bug

I'm currently using DDP and EpochCounterWithStopper in order to speed up training. However, on my experiments I am running into timeout errors,

`Watchdog caught collective operation timeout: WorkNCCL(SeqNum=588758, OpType=ALLREDUCE, Timeout(ms)=3600000) ran for 3602438 milliseconds before timing out.`

From tinkering around with it, I think this is due to the fact that I am using DDP and where I have placed the epoch_counter.should_stop() as it seems to run this for each GPU and the timeout occurs when one GPU hits a early stop while the other does not.

### Expected behaviour

EpochCounterWithStopper to work on multiple GPUs using DDP.

### To Reproduce

Here is my fit function where I have the early stopping

I am currently running on 2 GPUS

`def fit(
        self,
        epoch_counter,
        train_set,
        valid_set=None,
        progressbar=None,
        train_loader_kwargs={},
        valid_loader_kwargs={},
    ):
        """Iterate epochs and datasets to improve objective.

        Relies on the existence of multiple functions that can (or should) be
        overridden. The following methods are used and expected to have a
        certain behavior:

        * ``fit_batch()``
        * ``evaluate_batch()``
        * ``update_average()``

        If the initialization was done with distributed_count > 0 and the
        distributed_backend is ddp, this will generally handle multiprocess
        logic, like splitting the training data into subsets for each device and
        only saving a checkpoint on the main process.

        Arguments
        ---------
        epoch_counter : iterable
            Each call should return an integer indicating the epoch count.
        train_set : Dataset, DataLoader
            A set of data to use for training. If a Dataset is given, a
            DataLoader is automatically created. If a DataLoader is given, it is
            used directly.
        valid_set : Dataset, DataLoader
            A set of data to use for validation. If a Dataset is given, a
            DataLoader is automatically created. If a DataLoader is given, it is
            used directly.
        train_loader_kwargs : dict
            Kwargs passed to `make_dataloader()` for making the train_loader
            (if train_set is a Dataset, not DataLoader).
            E.G. batch_size, num_workers.
            DataLoader kwargs are all valid.
        valid_loader_kwargs : dict
            Kwargs passed to `make_dataloader()` for making the valid_loader
            (if valid_set is a Dataset, not DataLoader).
            E.g., batch_size, num_workers.
            DataLoader kwargs are all valid.
[...]

PR #2051, opened 2023-06-24
Updating early stopping for more intuitive use

pplantinga:
# Contribution in a nutshell
This addresses issue #1914 . Early stopping was already available with `EpochCounterWithStopper` but the use was quite unintuitive. I've updated to be more intuitive and updated the single recipe that used it.

### DIFF
--- a/recipes/LibriSpeech/G2P/train.py
+++ b/recipes/LibriSpeech/G2P/train.py
@@ -533,10 +533,7 @@ class G2PBrain(sb.Brain):
                 if self.hparams.enable_metrics:
                     self._write_reports(epoch, final=False)
 
-            if self.epoch_counter.should_stop(
-                current=epoch, current_metric=per,
-            ):
-                self.epoch_counter.current = self.epoch_counter.limit
+            self.epoch_counter.update_metric(per)
 
         if stage == sb.Stage.TEST:
             test_stats = {"loss": stage_loss}
--- a/speechbrain/utils/epoch_loop.py
+++ b/speechbrain/utils/epoch_loop.py
@@ -91,11 +91,19 @@ class EpochCounterWithStopper(EpochCounter):
     >>> epoch_counter = EpochCounterWithStopper(limit, limit_to_stop, limit_warmup, direction)
     >>> for epoch in epoch_counter:
     ...     # Run training...
-    ...     # Track a validation metric,
+    ...     # Track a validation metric, (insert calculation here)
     ...     current_valid_metric = 0
-    ...     # get the current valid metric (get current_valid_metric)
-    ...     if epoch_counter.should_stop(current_metric=current_valid_metric):
-    ...         epoch_counter.current = epoch_counter.limit  # skipping unpromising epochs
+    ...     # Update epoch counter so that we stop at the appropriate time
+    ...     epoch_counter.update_metric(current_valid_metric)
+    ...     print(epoch)
+    1
+    2
+    3
+    4
+    5
+    6
+    7
+    8
     """
 
     def __init__(self, limit, limit_to_stop, limit_warmup, direction):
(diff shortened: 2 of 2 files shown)


pplantinga:
Wondering if this PR should attempt to address #1939 as well. The issue I guess is that `on_stage_end(stage==valid)` is only called on the main process, so the loop stops on the main process but not the other processes. This could be handled by synchronizing the relevant metric across all processes, but I'm trying to think if there's a simpler way.

Adel-Moumen:
I think it makes sense to add the fix in this PR too.

pplantinga:
Depends on #2053 or something like it

pplantinga:
#2059 just landed, and I think it fixes the DDP bug without having to change anything, because now the `update_metrics()` gets called on all processes. This PR should be ready for review now by @Adel-Moumen @Gastron @mravanelli

Adel-Moumen (commented):
LGTM.

I'm going to try with multiple GPUs.

mravanelli:
@Adel-Moumen, any news?

Adel-Moumen:
I don't know if it's because of `EpochCounterWithStopper` but I have an issue with the G2P recipe with DDP (didn't test with single GPU yet). Basically, the `EpochCounterWithStopper` is working properly. I set the limit to be equal to 2 and I got the following message:
> speechbrain.utils.epoch_loop - 2 epochs without improvement.
Patience of 2 is exhausted, stopping.
__main__ - Attempting to restore checkpoint for step lexicon
speechbrain.utils.checkpoints - Loading a checkpoint from results/transformer/1234/save/CKPT+2023-08-09+12-06-58+00

However, now, my training is stopped. I can't do anything. Nothing show up. Additionally, the recipe doesn't seem to save the stopping information, when I kill the program an re run it, I am going back to training while it should skip it as we reached the patience limit... 

What do you think @pplantinga ?

Adel-Moumen:
> __main__ - seq2seq, and PER stats written to file: results/transformer/1234/reports/lexicon/35/wer_lexicon.txt
speechbrain.utils.epoch_loop - 2 epochs without improvement.
Patience of 2 is exhausted, stopping.
__main__ - Attempting to restore checkpoint for step lexicon
speechbrain.utils.checkpoints - Loading a checkpoint from results/transformer/1234/save/CKPT+2023-08-09+12-42-08+00
100%|██████████████████████████████████████████████████████████████████████████████████| 17/17 [00:28<00:00,  1.69s/it]
speechbrain.utils.train_logger - Epoch loaded: 33 - test loss: 1.87e-01, test PER: 9.13
__main__ - seq2seq, and PER stats written to file: results/transformer/1234/save/wer_lexicon.txt
[E ProcessGroupNCCL.cpp:828] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=35615, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1807776 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:455] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:460] To avoid data inconsistency, we are taking the entire process down.
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 72113 closing signal SIGTERM
ERROR:torch.distributed.elastic.multiprocessing.api:failed (exitcode: -6) local_rank: 1 (pid: 72114) of binary: /users/amoumen/.conda/envs/2051/bin/python

pplantinga:
> However, now, my training is stopped. I can't do anything. Nothing show up. Additionally, the recipe doesn't seem to save the stopping information, when I kill the program an re run it, I am going back to training while it should skip it as we reached the patience limit...

Ah yes, the `EpochCounterWithStopper` currently doesn't override the saver/loader from `EpochCounter` but would need to in order to work correctly on restart (there's more info to be stored: the best score so far and the epoch that occurred).

I can add this. As for the other issue, I don't immediately know what's wrong.

Adel-Moumen:
It is working! Thanks. 

I will open an issue related to the G2P recipe. There's a lot of problems when using it with DDP....

Adel-Moumen (approved):
lgtm

matthewkperez:
Yes, confirmed it worked for me. Sorry for the late reply

PR #2053, opened 2023-06-26
Add implementation of "on_stage_end_on_main"

pplantinga:
# Contribution in a nutshell
There's a number of cases where one might want to run code in `on_stage_end(stage.VALID)` on the non-main process, where currently it only runs on main. A few examples:

* #2051 -- need this to synchronize early stopping across processes
* FSDP -- need this to synchronize the shards present on different processes in order to save a model
* For distributed validation computation for large validation sets

# Notes for reviewing
As currently written, this would require a small change in most recipes to ensure continued compatibility with DDP (they would continue to work as-is when single-threaded). The required change is to convert `on_stage_end` to `on_stage_end_on_main`, and if there's code for the train stage that would have to be moved to a new `on_stage_end` function. Reviewers should comment on if these changes are worth the effort.

### DIFF
--- a/recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
+++ b/recipes/LibriSpeech/ASR/transformer/hparams/train_hf_whisper.yaml
@@ -23,8 +23,9 @@ normalized_transcripts: True
 test_only: False # Set it to True if you only want to  do the evaluation
 
 # Data files
-data_folder: !PLACEHOLDER # e,g./path/to/LibriSpeech
-train_splits: ["train-clean-100", "train-clean-360", "train-other-500"]
+data_folder: "/Users/N752624/Documents/Repositories/LibriSpeech" # !PLACEHOLDER # e,g./path/to/LibriSpeech
+#train_splits: ["train-clean-100", "train-clean-360", "train-other-500"]
+train_splits: ["train-clean-100"]
 dev_splits: ["dev-clean"]
 test_splits: ["test-clean", "test-other"]
 skip_prep: False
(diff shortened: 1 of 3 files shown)


pplantinga:
Let me clarify the issue if I can. Simplified, the `_fit_valid()` looks like this:

```python
def _fit_valid(self, valid_set, epoch, enable):
        if valid_set is not None:
            self.on_stage_start(Stage.VALID, epoch)
            with torch.no_grad():
                for batch in tqdm(valid_set ...):
                    loss = self.evaluate_batch(batch, stage=Stage.VALID)

                # Only run validation "on_stage_end" on main process
                run_on_main(self.on_stage_end)
```

This means that if you want to run code on all processes at the end of validation (e.g. you need to save the shards present on all processes for FSDP by collecting them to the main process and saving from there) there's no way to do it without overriding `_fit_valid()` or waiting until the next train step has started and communicating from there (which would fail on the last epoch).

Two potential solutions:

* Change `_fit_valid` so that `self.on_stage_end` is run on all processes. This would require adding `run_on_main` to either all recipes or, as @TParcollet suggested, to all functions typically called by `on_stage_end` that ought to run only on main, (e.g. default checkpoint saver), if they're not already set up this way. This would require a big disclaimer on release that any code that depends on this function only being run on the main process needs to wrap it with `run_on_main`.
* Add a separate method called by `_fit_valid` that runs on all processes. The easiest way is to just add something like `on_stage_end_all_processes` which requires very little changes to the recipes. The thing I don't like about this is that it leaves the behavior of `on_stage_end` very unintuitive (sometimes is run on all processes but not always?!). The most clear approach in my mind is something like what is proposed here -- two separate methods that are both called on every stage end. One last concern with this approach: what order do you do the two methods in? There could be times where you want to run code on a single process first, then communicate to all, or communicate to all first and then save on one.

Which of these options (or a third option) would you prefer? @Gastron @TParcollet @mravanelli @Adel-Moumen @AsuMagic @ycemsubakan (and any others I may have missed).

pplantinga:
Closed in favor of #2059

PR #2054, opened 2023-06-26
Fix features unittest using pytorch 2.0

pplantinga:
There is one failing unittest when using pytorch 2.0. This fixes it and should work for all versions.


mravanelli:
thank you @pplantinga!

mravanelli (approved):

PR #2059, opened 2023-06-30
Run "on_stage_end" on all processes and save on only a single process

pplantinga:
Alternative proposal to #2053 .

In this, we just run `on_stage_end` on all processes all the time, and take care of running things on a single process at the most granular level.

There's likely still areas that need to be fixed here, like logging or something like this.

Does this seem better or worse to you @Gastron @Adel-Moumen or any others?

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -438,7 +438,7 @@ class Brain:
         ckpt_interval_minutes (float)
             Amount of time between saving intra-epoch checkpoints,
             in minutes, default: ``15.0``. If non-positive, these are not saved.
-        ckpt_inerval_steps (int)
+        ckpt_interval_steps (int)
             Number of steps between saving intra-epoch checkpoints.
             If non-positive, these are not saved. Default: ``0``.
 
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1,7 +1,7 @@
 """Core SpeechBrain code for running experiments.
 
 Authors
- * Peter Plantinga 2020
+ * Peter Plantinga 2020, 2023
  * Abdel Heba 2020
  * Mirco Ravanelli 2020
  * Aku Rouhe 2021
@@ -11,7 +11,6 @@ Authors
 import os
 import sys
 import yaml
-import time
 import torch
 import shutil
 import logging
@@ -271,10 +270,10 @@ def parse_arguments(arg_list=None):
         help="This flag disables the data loop progressbars.",
     )
     parser.add_argument(
-        "--ckpt_interval_minutes",
-        type=float,
-        help="Amount of time between saving intra-epoch checkpoints "
-        "in minutes. If non-positive, intra-epoch checkpoints are not saved.",
+        "--save_after_steps",
+        type=int,
+        help="Save an intra-epoch checkpoint after this many steps."
+        "If non-positive, intra-epoch checkpoints are not saved.",
     )
     parser.add_argument(
         "--grad_accumulation_factor",
@@ -428,9 +427,9 @@ class Brain:
             Default: ``3``.
         noprogressbar (bool)
             Whether to turn off progressbar when training. Default: ``False``.
-        ckpt_interval_minutes (float)
-            Amount of time between saving intra-epoch checkpoints,
-            in minutes, default: ``15.0``. If non-positive, these are not saved.
+        save_after_steps (int)
+            Number of steps between saving intra-epoch checkpoints.
+            If non-positive, these are not saved. Default: ``0``.
 
         Typically in a script this comes from ``speechbrain.parse_args``, which
         has different defaults than Brain. If an option is not defined here
@@ -486,7 +485,7 @@ class Brain:
             "max_grad_norm": 5.0,
             "nonfinite_patience": 3,
             "noprogressbar": False,
-            "ckpt_interval_minutes": 0,
+            "save_after_steps": 0,
             "grad_accumulation_factor": 1,
             "optimizer_step_limit": None,
             "tqdm_colored_bar": False,
@@ -1098,8 +1097,7 @@ class Brain:
         ):
             self.train_sampler.set_epoch(epoch)
 
-        # Time since last intra-epoch checkpoint
-        last_ckpt_time = time.time()
+        steps_since_save = 0
         with tqdm(
             train_set,
             initial=self.step,
@@ -1112,6 +1110,7 @@ class Brain:
                     logger.info("Train iteration limit exceeded")
                     break
                 self.step += 1
+                steps_since_save += 1
                 loss = self.fit_batch(batch)
                 self.avg_train_loss = self.update_average(
                     loss, self.avg_train_loss
@@ -1129,19 +1128,11 @@ class Brain:
 
                 if (
                     self.checkpointer is not None
-                    and self.ckpt_interval_minutes > 0
-                    and time.time() - last_ckpt_time
-                    >= self.ckpt_interval_minutes * 60.0
+                    and steps_since_save >= self.save_after_steps
                 ):
-                    # This should not use run_on_main, because that
-                    # includes a DDP barrier. That eventually leads to a
-                    # crash when the processes'
-                    # time.time() - last_ckpt_time differ and some
-                    # processes enter this block while others don't,
-                    # missing the barrier.
-                    if sb.utils.distributed.if_main_process():
-                        self._save_intra_epoch_ckpt()
-                    last_ckpt_time = time.time()
+                    # Checkpointer handles running on main process only
+                    self._save_intra_epoch_ckpt()
+                    steps_since_save = 0
 
         # Run train "on_stage_end" on all processes
         self.zero_grad(set_to_none=True)  # flush gradients
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -33,7 +33,6 @@ from torch.utils.data import IterableDataset
 from torch.utils.data import DistributedSampler
 from torch.nn.parallel import DistributedDataParallel as DDP
 from hyperpyyaml import resolve_references
-from speechbrain.utils.distributed import run_on_main
 from speechbrain.utils.optimizers import rm_vector_weight_decay
 from speechbrain.dataio.dataloader import LoopedLoader
 from speechbrain.dataio.dataloader import SaveableDataLoader
@@ -1177,12 +1176,8 @@ class Brain:
                     if self.debug and self.step == self.debug_batches:
                         break
 
-                # Only run validation "on_stage_end" on main process
                 self.step = 0
-                run_on_main(
-                    self.on_stage_end,
-                    args=[Stage.VALID, avg_valid_loss, epoch],
-                )
+                self.on_stage_end(Stage.VALID, avg_valid_loss, epoch)
 
     def fit(
         self,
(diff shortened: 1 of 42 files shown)


Gastron:
This seems better to me. But in the current implementation any registered save hooks have expected to be called on the main process. I don't know if Checkpointer should wrap saving in `run_on_main` or the individual objects. Maybe Checkpointer should do it? 

Can saving still only run on the main process under FSDP, or do the other processes need to explicitly do something?

Also, the intra-epoch checkpoint saving as currently implemented (based on time) is not compatible with this: https://github.com/speechbrain/speechbrain/blob/5d479fc6713f3aa0e218802019b7f7bceb46146c/speechbrain/core.py#L1136
This makes me realise that if we put run_on_main somewhere, then all processes must hit that code, so if the run_on_main goes into classes, we must ensure that all processes run that (and in the same order, i.e. if one process runs into save barrier and another into gradient sync barrier then it ends up in deadlock). Maybe we don't need the barrier inside run_on_main in many cases? There is already the `if_main_process()` (which really should be called `is_main_process()`), that might be enough in most cases. The other processes will wait at the next barrier, while main process saves or does whatever errands we need.

Note: Checkpointer should also wrap the checkpoint deletion in `run_on_main`.

pplantinga:
> This seems better to me. But in the current implementation any registered save hooks have expected to be called on the main process. I don't know if Checkpointer should wrap saving in `run_on_main` or the individual objects. Maybe Checkpointer should do it?
> 
> Can saving still only run on the main process under FSDP, or do the other processes need to explicitly do something?
 
You've correctly deduced the reason why I put this in individual classes -- some (including FSDP) may want to run code on all processes. Perhaps a more backwards-compatible way of doing this would be to add some sort of argument (to `mark_as_saver`?) that would by default run only on main but would allow running on all if passed.... Thinking aloud on this one.

> Also, the intra-epoch checkpoint saving as currently implemented (based on time) is not compatible with this:
> 
> https://github.com/speechbrain/speechbrain/blob/5d479fc6713f3aa0e218802019b7f7bceb46146c/speechbrain/core.py#L1130-L1144
> 
> 
> This makes me realise that if we put run_on_main somewhere, then all processes must hit that code, so if the run_on_main goes into classes, we must ensure that all processes run that (and in the same order, i.e. if one process runs into save barrier and another into gradient sync barrier then it ends up in deadlock). Maybe we don't need the barrier inside run_on_main in many cases? There is already the `if_main_process()` (which really should be called `is_main_process()`), that might be enough in most cases. The other processes will wait at the next barrier, while main process saves or does whatever errands we need.
> Note: Checkpointer should also wrap the checkpoint deletion in `run_on_main`.

Perhaps by default we should call `if_main_process()` unless the barrier is really needed. Wondering if there would be a reason for the barrier that I haven't thought of -- a timeout in the gradient sync or something like this.

Gastron:
Continuing on the checkpoints with time intervals: basically, time cannot be relied on for multi-process synchronisation (almost by definition). If FSDP needs Checkpointer code to run on all nodes then I think that this checkpoint with time interval is not possible with FSDP.

With the coming changes in SpeechBrain, I think we can afford to break some backwards compatibility and move to checkpoints every x steps. Some people also requested that anyway.

Gastron:
I think the mark_as_saver decorator could indeed have an optional parameter (something like `@mark_as_saver(main_process_only=False)`), and checkpointer could then keep track of this.

pplantinga:
> With the coming changes in SpeechBrain, I think we can afford to break some backwards compatibility and move to checkpoints every x steps. Some people also requested that anyway.

Should that be done as part of this PR or could it be a separate one?

> I think the mark_as_saver decorator could indeed have an optional parameter (something like `@mark_as_saver(main_process_only=False))`, and checkpointer could then keep track of this.

Implementation attempted here -- does this look reasonable?

Gastron:
I think the two separated sets of default save hooks (main proc only vs. all procs) will lead to many errors. When searching for a default hook, the code looks through the class hierarchy, where it will often find e.g. torch.Module, and use that default hook. While we could change the order so that all_procs hooks are prioritized (as they are less common), I fear that this will lead to similar weird and hard to find bugs.

I think we'd want to keep one set of save hooks. Perhaps the checkpoint registering could wrap the saver hook callable in a wrapper that only runs on the main process:
```python
from functools import wraps
def main_process_only(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        if sb.utils.distributed.if_main_process():
            func(*args, **kwargs)
    return wrapper
```
Or we could keep more information about those hooks, e.g. the hooks dict could map types to some Hook dataclass with entries like `callable`, `run_on_all_procs`, etc. but then that might require changing all existing default hooks.

> > With the coming changes in SpeechBrain, I think we can afford to break some backwards compatibility and move to checkpoints every x steps. Some people also requested that anyway.
> Should that be done as part of this PR or could it be a separate one?
It's something that breaks from this, so it could be done as part of this change or if we want to be tidy it could be a separate change implemented before this.

pplantinga:
> I think we'd want to keep one set of save hooks. Perhaps the checkpoint registering could wrap the saver hook callable in a wrapper that only runs on the main process

This is a neat solution, I wish I'd thought of it first!

> With the coming changes in SpeechBrain, I think we can afford to break some backwards compatibility and move to checkpoints every x steps. Some people also requested that anyway.

Okay, I've made the change in `core.py`. Pretty much every recipe uses it, so it'll take some time to change over, but fortunately it's only a one-line change. I looked and the number of minutes varies from recipe to recipe. I'm not sure how to select a sane value for the checkpoint steps... any ideas?

I was thinking of assuming there's 5 iterations per second (a tremendously inaccurate assumption) and converting the minutes to steps (* 60 * 5), but there's probably a better way to do this.

Gastron:
> Okay, I've made the change in `core.py`. Pretty much every recipe uses it, so it'll take some time to change over, but fortunately it's only a one-line change. I looked and the number of minutes varies from recipe to recipe. I'm not sure how to select a sane value for the checkpoint steps... any ideas?

Oof. Well, one idea is to keep the time based checkpointing, it is kind of useful and works in most cases. We could also try to detect FSDP and raise an error, though I guess at the moment we cannot easily detect FSDP. Maybe a hack for now, something like checking if the name "FSDP" is in globals, I don't know.

Gastron:
Otherwise, I think this could be a good implementation. You need to remember to wrap torch_recovery with main_process_only, and same for checkpoint deletion. Also, I wonder if many recipes do things like write some WER file in an on_stage_end, expecting this to happen only on main?

pplantinga:
> You need to remember to wrap torch_recovery with main_process_only, and same for checkpoint deletion.

Ah yes, I've added this for checkpoint deletion. However, for `torch_recovery` I'm pretty confident we *don't* want it to happen only on main, but rather on all processes, since we use this to load models on all devices (and there's no contested resource).

> Also, I wonder if many recipes do things like write some WER file in an on_stage_end, expecting this to happen only on main?

These should be fixed:

```bash
$ grep -r 'with open(self.hparams.wer_file' recipes/ | wc -l
      33
```

But there are some cases of saving that don't necessarily need to be fixed:

```bash
grep -r 'torchaudio.save(' recipes/ | wc -l
      63
```

Looking at these, they mostly happen in places other than `on_stage_end` so it seems like the recipes are not expected to work on multiple processes.

Gastron:
> > You need to remember to wrap torch_recovery with main_process_only, and same for checkpoint deletion.
> 
> Ah yes, I've added this for checkpoint deletion. However, for `torch_recovery` I'm pretty confident we _don't_ want it to happen only on main, but rather on all processes, since we use this to load models on all devices (and there's no contested resource).

Oops yea sorry, I meant torch_save

speechbrain/core.py
```diff
@@ -1151,6 +1163,21 @@ def _fit_train(self, train_set, epoch, enable):
...
+        """Determines if an intra-epoch checkpoint should be saved.
+
+        Returns True if there's a checkpointer and time or steps has exceeded limit.
+        """
+        if self.checkpointer is None:
+            return False
+
+        # Check if we've run for the requested amount of time
+        if 0 < self.ckpt_interval_minutes * 60.0 < time.time() - last_ckpt_time:
+            return True
```
Gastron, line 1176: I still think here we would want to check if we are on the main process. In the current approach we are letting all processes call the checkpointer at different times and just expecting the checkpointer to work right. I think this invites bugs where users think they can rely on the different processes calling the checkpointer at the same time.

pplantinga: Okay, I've added this, and updated the warning accordingly. Do you think the warning is still needed or should we remove it altogether?

speechbrain/core.py
```diff
@@ -432,6 +438,10 @@ class and override any methods for which the default behavior does not
         ckpt_interval_minutes (float)
             Amount of time between saving intra-epoch checkpoints,
             in minutes, default: ``15.0``. If non-positive, these are not saved.
+        ckpt_inerval_steps (int)
```
Gastron, line 441:
```suggestion
        ckpt_interval_steps (int)
```

speechbrain/core.py
```diff
@@ -1151,6 +1165,24 @@ def _fit_train(self, train_set, epoch, enable):
...
+        # Check if we've run for the requested amount of time
+        if 0 < self.ckpt_interval_minutes * 60.0 < time.time() - last_ckpt_time:
+            # Only save on the main process to avoid race conditions
+            return if_main_process()
+
+        # Save after requested # of steps. This option is the only one that
+        # allows saving on multiple processes. The logic for whether saving
+        # is run only on the main process is handled by the checkpointer.
+        return 0 < self.ckpt_interval_steps <= steps_since_ckpt
+
```
Gastron, line 1185: With the separate early returns, I think a tricky bug could happen if someone uses both `ckpt_interval_minutes` and `ckpt_interval_steps`. There could be cases where the time interval and step interval happen simultaneously.

I suggest:
```suggestion
        should_save = False
        # Check if we've run for the requested amount of time
        if 0 < self.ckpt_interval_minutes * 60.0 < time.time() - last_ckpt_time:
            # Only save on the main process to avoid race conditions
            should_save = if_main_process()

        # Save after requested # of steps. This option is the only one that
        # allows saving on multiple processes. The logic for whether saving
        # is run only on the main process is handled by the checkpointer.
        if 0 < self.ckpt_interval_steps <= steps_since_ckpt:
            should_save = True
        
        return should_save

```

Gastron (commented):
Other than two very minor things I think this is ready to merge!

speechbrain/core.py
Gastron: Actually now that I think about it, resetting both steps since last checkpoint and the time since last checkpoint will make the main process quickly out of sync when both methods are used. Should we make steps and time exclusive? Or maybe just run two different checks in the loop?

Gastron (approved):
Alright thanks for the persistent work! Everything looks good to me now.

Gastron:
Side note: I was going to nitpick about using sys.exit to crash Brain in case of wrong arguments, but I noticed that this is done for some other options as well and you were using the same convention. I propose a separate PR to change all those to raising an 
 error like a proper respectful Python class should.

pplantinga:
There is an issue with this PR, as reported by @Adel-Moumen:

"I trained a model with 3 GPUs and basically, I got [three] save folders"

and

"they all have:
```
# yamllint disable
WER: 5.784713797286864
end-of-epoch: true
unixtime: 1691803456.4516888
```
except one with the right `.ckpt`.  however, the problem is that when loading the new ckpt we are getting an error saying that we are failing to load the .ckpt as there is no ckpt, and indeed, the issue is that the last ckpt (the most recent one) is lacking from this .ckpt"

Issue #2066, opened 2023-07-06
Comments on the ESC50 recipes


mravanelli:
### 🚀 The feature

Hi @fpaissan and @ycemsubakan,

I have tried the ESC50 recipes, and everything is running smoothly. As we discussed privately, I would like to share some comments to help improve it:

1. **Comments on `/ESC50/classification/`**:

- Please mention the performance and computing time for both CNN14 and conv2d-based models. Additionally, include references to our papers, including Zhepei's paper.
- We need to make the HF repository (https://huggingface.co/speechbrain/soundrec-cnn14-esc50) public. Also, please add a working example, similar to other HF interfaces. It would be helpful to upload an audio signal as an example in the repository as well.

2. **Comments on `/recipes/ESC50/interpret`**:

- When introducing PIQ, please add a reference to our paper. Additionally, at the end, you can include a section on citing PIQ with the necessary BibTeX information.
- Create a new interface and a corresponding HF repository for the PIQ Model. In this case, the input should be a signal, and the output should be the audio for interpretation, correct?
- For this interface, please upload one or two example files (maybe one positive and one negative example) and briefly explain the outcomes in the text.
- In `/ESC50/interpret/train_piq.py`, please add missing docstrings. All functions should have at least a short description.

- The YAML files should use the pretrained model form the speechbrain HF repository (remember to make it public). Currently, it is using `fpaissan/conv2d_us8k` or `cemsubakan/cnn14-esc50/`.

- It would be helpful to write a short tutorial on interpretability with a special focus on PIQ.

### Solution outline

-

### Additional context

-

PR #2070, opened 2023-07-10
PR to address the comments on the ESC50 recipe

ycemsubakan:
This is a PR to address the comments of Mirco in https://github.com/speechbrain/speechbrain/issues/2066

Here are the original comments and the progress:

1. **Comments on `/ESC50/classification/`**:

- Please mention the performance and computing time for both CNN14 and conv2d-based models. [done] Additionally, include references to our papers, including Zhepei's paper. [done]
- We need to make the HF repository (https://huggingface.co/speechbrain/soundrec-cnn14-esc50) public. Also, please add a working example, similar to other HF interfaces. It would be helpful to upload an audio signal as an example in the repository as well. [done] 
   ---> note: this repository is called speechbrain/cnn14-esc50 now.  

2. **Comments on `/recipes/ESC50/interpret`**:

- When introducing PIQ, please add a reference to our paper. Additionally, at the end, you can include a section on citing PIQ with the necessary BibTeX information. [done]
- Create a new interface and a corresponding HF repository for the PIQ Model. In this case, the input should be a signal, and the output should be the audio for interpretation, correct? [done]
- For this interface, please upload one or two example files (maybe one positive and one negative example) and briefly explain the outcomes in the text. [done]
- In `/ESC50/interpret/train_piq.py`, please add missing docstrings. All functions should have at least a short description. [done]

- The YAML files should use the pretrained model form the speechbrain HF repository (remember to make it public). Currently, it is using `fpaissan/conv2d_us8k` or `cemsubakan/cnn14-esc50/` [done].

- It would be helpful to write a short tutorial on interpretability with a special focus on PIQ [I think this might stay out of scope of this PR].


mravanelli:
Excellent work! I tested both the training recipes and HF interfaces and everything works. I only have the following comments:
1. It looks like there are a couple of wrong links in the interpret/README.md. The log folder of PIQ is empty. The CNN14 log links seem to point to the conv2D one.
2. It could be great to conclude this work with a nice tutorial on interpretability. We can merge this PR meanwhile, but we will need to add a link to the tutorial in the README files once available.

mravanelli (approved):

PR #2074, opened 2023-07-13
corrected the right import

abdouaziz:
This PR is corrected the right import to `LinearNorm` instead of `Tacotron2`


mravanelli:
Thank you for the fix @abdouaziz

mravanelli (approved):

PR #2081, opened 2023-07-19
Fix Recipe Tests

mravanelli:
With this PR, we run the recipe tests and  fix recipes whose tests are failing.
To check:

- [x] AISHELL-1
- [x] Aishell1Mix
- [x] AMI
- [x] BinauralWSJ0Mix
- [x] CommonLanguage
- [x] CommonVoice
- [x] DVoice
- [x] ESC50
- [x] Fisher-Callhome-Spanish
- [x] fluent-speech-commands
- [x] Google-speech-commands
- [x] IEMOCAP
- [x] IWSLT22_lowresource
- [x] KsponSpeech
- [x] LibriMix
- [x] LibriParty
- [x] LibriSpeech
- [x] LibriTTS
- [x] LJSpeech
- [x] MEDIA
- [x] REAL-M
- [x] RescueSpeech
- [x] SLURP
- [x] Switchboard
- [x] timers-and-such
- [x] TIMIT
- [x] UrbanSound8k
- [x] Voicebank
- [x] VoxCeleb
- [x] VoxLingua107
- [x] WHAMandWHAMR
- [x] WSJ0Mix
- [x] ZaionEmotionDataset

**Note**: We need to determine the reason for skipping some recipe tests. For example, in Librispeech, certain tests are not done due to the absence of "test_field=test_debug_flags". However, in LibriSpeech.csv, all entries have the test_debug_flag specified.


mravanelli:
It looks like there are several recipe tests that are failing. We need to fix them before releasing the upcoming minor version.

1. @ycemsubakan, It seems that the recipe tests for REALM, WHAMandWHAMR, and BinauralWSJ0Mix are failing. Could you please investigate and address this issue?

You can run the recipe test (from the root SpeechBrain folder) with the following command:

```
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Dataset"], filters=[["REALM", "WHAMandWHAMR", "BinauralWSJ0Mix"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```

Regarding REALM, it appears that there might be an issue with the enforced version of `pyroomacoustics`, which is not compatible with a recent version of `numpy`. Could you explore the possibility of converting all the recipes to work with the latest version of `pyroomacoustics`?

2. **@BenoitWang**, the recipe tests for `fastspeech2` and `emotion diarization` are also failing. To run the recipe test, use the following command from the root SpeechBrain folder:

```
print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Dataset"], filters=[["LJSpeech", "ZaionEmotionDataset"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")
```

It appears that the flags are not set correctly, resulting in the use of incorrect data and labels.

3. **@Adel-Moumen**, please take a look at the other issues as well. Many of them seem to be related to the pretrainer not saving elements in the right folder (`collect_in`) when specifying a file instead of a URL. Feel free to investigate further. Tomorrow, I'll also work on resolving these issues.

Adel-Moumen:
Hey @Gastron,

Could you please have a look at this commit please : https://github.com/speechbrain/speechbrain/pull/2081/commits/629b1b1a6fdb7699674e7bf78e03affcf28b165e ?

Basically, we had some issues with the recipe Fisher-Callhome-Spanish when using the recipe test tool:
> python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Dataset"], filters=[["Fisher-Callhome-Spanish"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'

Indeed, the recipe in the test tool is overriding the 'tokenizer_file' with the relative path '--tokenizer_file=tests/tmp/Fisher-Callhome-Spanish_row_2/30_bpe.model', however, this leads to an issue because the Pretrainer is unable to find the location of the tokenizer because we are using the relative path. When we give the absolute path to the tokeniser, there's no issues. Everything is working, but not when the relative path is given. The commit tries to solve this issue by first checking if the loadable path exists (so the relative path), if not, it uses the absolute path. I prefer to ask you if it's correct because maybe there's something else that we could do. 

(see: https://github.com/speechbrain/speechbrain/blob/83c6ac3db41fd1abbd6d9c6ec37b2ba7a99faee2/tests/recipes/Fisher-Callhome-Spanish.csv#L3) 

Thanks!

mravanelli:
Hi @pplantinga ,
It looks like the recipe tests for Voicebank (including the mimic loss one) fail. You can run the tests by installing the version of speech brain in this PR and running the following command from the speechbrain root folder:

` python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Dataset"], filters=[["Voicebank"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'`

Do you think you can help us fix these issues?

ycemsubakan:
The tests for WHAMR, WHAMandWHAMR and REAL-M should pass now. Upgrading to the newer version of pyroomacoustics solves the issue.

mravanelli:
All the recipe tests are now passing. Thank you @Adel-Moumen, @pplantinga, and @ycemsubakan for your collaboration!

PR #1778, opened 2023-01-02
[WIP] Streamable Voice Activity Detection

fpaissan:
Integrate streamable Voice Activity Detection with script to run on laptop via `ffmpeg`.

Missing:

- [ ] choose model, train and deploy on HF Hub;
- [ ] test VAD_stream and perform last consistency checks;
- [ ] update README.md

PR #2082, opened 2023-07-21
[WIP] Streamable VAD

fpaissan:
# Contribution in a nutshell
An updated version of #1778. Closing that one to avoid excessive conflicts for merge.

# Notes for reviewing
Please also review the (private) [HF card](https://huggingface.co/speechbrain/stream-vad-crdnn-libriparty).

# Pre-review
* [x] (if applicable) add an `extra_requirements.txt` file
* [x] (if applicable) add database preparation scripts & use symlinks for nested folders (to the level of task READMEs)
* [x] (if applicable) add a recipe test entry in the depending CSV file under: tests/recipes
* [x] (if applicable) run a recipe test for each yaml/your recipe dataset
* [x] (optional) run `tests/.run-doctests.sh` & `tests/.run-unittests.sh`

### DIFF
--- a/recipes/LibriParty/streamable_VAD/hparams/streamable_cnn.yaml
+++ b/recipes/LibriParty/streamable_VAD/hparams/streamable_cnn.yaml
@@ -1,0 +1,167 @@
+# ##################################################################################################
+# Model: VAD on LibriParty with CRDNN.
+# This code heavily relis on on-the-fly data augmentation using external data.
+# Before running the code, please download the needed datasets:
+#
+# - LibriParty: https://drive.google.com/file/d/1--cAS5ePojMwNY5fewioXAv9YlYAWzIJ/view?usp=sharing
+# - Musan: https://www.openslr.org/resources/17/musan.tar.gz
+# - CommonLanguage: https://zenodo.org/record/5036977/files/CommonLanguage.tar.gz?download=1
+# ##################################################################################################
+
+# Seed and output folders
+seed: 1986
+__set_seed: !apply:torch.manual_seed [!ref <seed>]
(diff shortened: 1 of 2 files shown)


mravanelli:
I just fixed a backward incompatibility issue that was pretty problematic and caused the tests to fail. The bias of the DNN_block (used in all CRDNN) was set to false. Even though this can make sense because the bias term is already included in the batch norm part, that had an impact on the many recipes using the CRDNN. I modified the code to make the proposed change backward compatible.

mravanelli:
Thank you for @fpaissan for this PR! This contribution is crucial as we plan to release many more streamable models for various tasks like keyword spotting, speech recognition, and possibly speech enhancement in the future.

After conducting an initial inspection, I have made the modifications above, and I have some minor comments to address:
1. We have to switch from Gdrive to Dropbox. Please, grant me access to the Gdrive folder so that I can download the output folder and upload it to the Speechbrain Dropbox.
2. The header of `inference.py` could be improved. I suggest adding instructions on how to run the code with additional details similar to what you provided in the HF repository.
3. I propose creating a short video (around 2-3 minutes) that demonstrates the process of performing streamable inference. This would be valuable for users trying to understand the implementation better.

Feel free to take care of the previous comments. My next step will be to run the training and inference scripts along with the recipe test.

mravanelli:
I did some fixes and now training is running on my side. My next step is to check the streamable inference part.

mravanelli:
https://huggingface.co/speechbrain/stream-vad-crdnn-libriparty/blob/main/README.md => It looks like the inference.py script only works with Apple devices, due to a limitation in torchaudio (in fact, it doesn't work on my Linux machine). If so, we should mention that here as well.

PR #2087, opened 2023-07-24
Bug in env corruption:  $rir_root variable added without '/' 

mravanelli:
This is a little bug that prevents all the recipes using env corruption to run.


TParcollet (approved):
LGTM

PR #2089, opened 2023-07-25
add documentation for HF API testing

TParcollet:
# Contribution in a nutshell
Adding information on how to test the HF API and add missing interfaces.

### DIFF
--- a/tests/PRE-RELEASE-TESTS.md
+++ b/tests/PRE-RELEASE-TESTS.md
@@ -42,12 +42,34 @@ tests/.run-recipe-tests.sh
 ```
 tests/.run-HF-checks.sh
 ```
+10. Make sure all HuggingFace API Interfaces are up to date and working (see [here](#huggingface-api-testing)])
 11. Check URLs
 ```
 tests/.run-url-checks.sh
 ```
 
 Make sure all the tests are passing. Also, make sure to check that the tutorials are working (we might set up an automatic test for that as well in the future).
+
+# HuggingFace API testing
+
+API testing cannot be automated within SpeechBrain, however, it is already done within HuggingFace repository!
+Steps to test them are:
+1. Go to the [api-inference-community](https://github.com/huggingface/api-inference-community) and clone it.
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you @TParcollet. It is very good to have tests for the HF browser APIs. I followed the instructions, but it looks like there are a few things to fix on the HF side:
1. When I installed the requirement on a new conda environment I got a conflict error: 
```
ERROR: Cannot install -r docker_images/speechbrain/requirements.txt (line 3), -r docker_images/speechbrain/requirements.txt (line 5) and huggingface_hub==0.5.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested huggingface_hub==0.5.1
    api-inference-community 0.0.23 depends on huggingface-hub>=0.0.12
    transformers 4.30.0 depends on huggingface-hub<1.0 and >=0.14.1

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```
2. When running the tests, they all fail (even when the browser API is working well). Many errors are like this:  
`FAILED docker_images/speechbrain/tests/test_api_audio_classification.py::AudioClassificationTestCase_1_speechbrain_google_speech_command_xvector::test_simple - AssertionError: 400 != 200`

I think we need to follow up with the HF team for that.

TParcollet:
Try with the new requirements that we have in the PR I did in the HF repo. Also, make sure that SpeechBrain is installed

BenoitWang:
Hi @TParcollet I've made an audio_diarization pipeline for diarization tasks (speaker/emotion/other speech events, for now it's only for emotion), could you give me access that I can push it to your HF pr branch?

mravanelli:
I think this PR is fine. We can follow up in the PR that @TParcollet open on HF.

mravanelli (approved):

PR #2091, opened 2023-07-26
[WIP] Support for full-inference recipe tests

mravanelli:
This PR expands the existing standard recipe tests to include support for full-inference recipe tests as well. 

In a full-inference test, we follow these steps: 
1. First, we download specific data (e.g., 'test-clean' from LibriSpeech) and relevant output folders from Dropbox. 
2. Using the downloaded data and checkpoint, we run inference.
3. Finally, we verify if the performance aligns with the expected results.

Having this functionality will be particularly important in the coming weeks, as we plan to carry out significant refactoring in the search part of SpeechBrain.


mravanelli:
@TParcollet,  I think this PR is now ready for a quick review (even though I'm running the last tests to make sure everything is fine).
You can run the tests with:
`
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Task"], filters=[["full_inference"]], do_checks=True, run_opts="--device=cuda")) else print("TEST PASSED")'`

We would need this to be merged soon as we will use a full-inference test to conclude that PR (e.g., https://github.com/speechbrain/speechbrain/pull/751) where the decoding part has been modified.

mravanelli:
As discussed privately with @TParcollet, we need to merge now this PR to proceed safely with the review of https://github.com/speechbrain/speechbrain/pull/751.

PR #2099, opened 2023-08-05
Add the ability to optionally pass encoded representation lengths during transformer decoding (beam search / greedy search)

flexthink:
# Contribution in a nutshell
Hey, this could help our community 🌱

# Scope
* [x] Add the ability to optionally pass encoded representation lengths during transformer decoding (beam search / greedy search)

# Pre-review
* [ ] (if applicable) add an `extra_requirements.txt` file
* [ ] (if applicable) add database preparation scripts & use symlinks for nested folders (to the level of task READMEs)
* [ ] (if applicable) add a recipe test entry in the depending CSV file under: tests/recipes
* [ ] (if applicable) run a recipe test for each yaml/your recipe dataset
* [ ] (optional) run `tests/.run-doctests.sh` & `tests/.run-unittests.sh`

---


Adel-Moumen:
Hey, thanks for the PR! 

I don't like the idea of having a parameter to set whether or not we pass the enc lens in the decoder. I believe, if it's an error from our side to have omitted this arg, then we should fix it. And it doesn't seems to be backward incompatible. So why not always giving to the decoder the enc_lens ? 

Could you please try to run a pretrained model which is using the Transformer decode method and see if we are still getting the same WER? (e.g. LibriSpeech pretrained model see dropbox) 

Have you run the recipe test tool? Is everything working? 

Thanks again for the PR and spotting the issue!
Adel

flexthink:
> Hey, thanks for the PR!
> 
> I don't like the idea of having a parameter to set whether or not we pass the enc lens in the decoder. I believe, if it's an error from our side to have omitted this arg, then we should fix it. And it doesn't seems to be backward incompatible. So why not always giving to the decoder the enc_lens ?
> 
> Could you please try to run a pretrained model which is using the Transformer decode method and see if we are still getting the same WER? (e.g. LibriSpeech pretrained model see dropbox)
> 
> Have you run the recipe test tool? Is everything working?
> 
> Thanks again for the PR and spotting the issue! Adel

I don't know if there are any recipes that:
- Rely on lengths **not** being passed
- Use models that don't support the length parameter

I also prefer not having a parameter... but I'm not sure if it's possible to make the change without breaking any recipes or significantly changing their performance stats...

Adel-Moumen:
> Use models that don't support the length parameter

That is why we have a recipe test tool.  Please run it (see: https://github.com/speechbrain/speechbrain/blob/develop/tests/.run-recipe-tests.sh).

> Rely on lengths not being passed

Please download a pretrained LibriSpeech ASR model from our released ones (see dropbox links) and check if the WER is the same or better when we add the enc_lens.

flexthink:
> > Use models that don't support the length parameter
> 
> That is why we have a recipe test tool. Please run it (see: https://github.com/speechbrain/speechbrain/blob/develop/tests/.run-recipe-tests.sh).
> 
> > Rely on lengths not being passed
> 
> Please download a pretrained LibriSpeech ASR model from our released ones (see dropbox links) and check if the WER is the same or better when we add the enc_lens.

OK I will run a test and report the results shortly

mravanelli:
I conducted further experimentation with additional flags, including passing the 'enc_len' parameter during transformer beamsearch. I also tested with a batch size greater than 1,and the results remained consistent. Everything appears to be in good shape and is ready for merging. Thank you @flexthink

mravanelli (approved):

PR #2112, opened 2023-08-09
Update sampler.py (fix typo error)

mohamedAnterGad:
fix simple type error in sampler.py
I suggest this to be fixed in the docs as well, in order not to cause frustration.


mravanelli:
Thank you @mohamedAnterGad for this fix!

mravanelli (approved):

PR #2116, opened 2023-08-10
HF interface

mhn226:
This PR simplifies the #1596 PR following @TParcollet 's ideas. 
Briefly, we create an interface class called HuggingFaceTransformer which only implements several esensial methods, for example, modeling loading using HuggingFace's Auto Classes (https://huggingface.co/docs/transformers/model_doc/auto).
The users would need to customize their own HF's class which inherits  HuggingFaceTransformer. The adapted huggingface_wav2vec.py and huggingface_whisper.py would serve as examples of how to make such classes.


speechbrain/lobes/models/huggingface.py
```diff
@@ -0,0 +1,361 @@
...
+except ImportError:
+    MSG = "Please install transformers from HuggingFace to use wav2vec2 / Hubert\n"
+    MSG += "E.G. run: pip install transformers"
+    raise ImportError(MSG)
+
+logger = logging.getLogger(__name__)
+
+
+class HuggingFaceTransformer(nn.Module):
+    """This lobe provides AutoClass architecture loading into SpeechBrain modules.
```
TParcollet, line 55: Maybe slightly more precise about what it means: being able to load any model from HuggingFace transformer.

speechbrain/lobes/models/huggingface.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+
+    def __init__(
+        self,
+        source,
+        save_path,
+        tokenizer=False,
+        for_pretraining=False,
+        freeze=False,
+        sampling_rate=16000,
```
TParcollet, line 91: Do we need the sampling_rate here?

speechbrain/lobes/models/huggingface.py
```diff
@@ -0,0 +1,361 @@
...
+            for File in files:
+                if File.rfilename.endswith(".bin"):
+                    checkpoint_filename = File.rfilename
+                    is_sb = False
+                    return is_sb, checkpoint_filename, is_local
+
+        err_msg = f"{path} does not contain a .bin or .ckpt checkpoint !"
+        raise FileNotFoundError(err_msg)
+
+    def _modify_state_dict_partial_fn(self):
```
TParcollet, line 249: What is this function for?

speechbrain/lobes/models/huggingface.py
TParcollet: It would be good to give information about how to use this class as well. Like it gives an ensemble of functions that the user must override by inheriting from this class to support any HuggingFace model.

speechbrain/lobes/models/huggingface.py
```diff
@@ -0,0 +1,361 @@
...
+        model.eval()
+        for param in model.parameters():
+            param.requires_grad = False
+
+    def override_config(self, config):
+        """Users should modify this function according to their own tasks."""
+        return config
+
+
+def make_padding_masks(src, wav_len=None, pad_idx=0):
```
TParcollet, line 326: Don't we have a function in SpeechBrain to do this already? I believe we do, maybe in util, or attention or somewhere.

speechbrain/lobes/models/huggingface.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+    src_key_padding_mask = None
+    if wav_len is not None:
+        abs_len = torch.round(wav_len * src.shape[1])
+        src_key_padding_mask = length_to_mask(abs_len).bool()
+
+    return src_key_padding_mask
+
+
+def make_masks(src, wav_len=None, pad_idx=0):
```
TParcollet, line 345: Same here.

TParcollet (changes requested):
Thanks for the work! I will request ideas from @Adel-Moumen and @pplantinga here. I think I like this easy to use abstraction. The main remaining question here being: do we put everything in a single file called HuggingfaceTransformers.py or do we create a directory with multiple files, with each file containing our implementation of a model? Also we have to think about one thing: dropping the wav2vec2 in the naming of our implementation, because it also works with Hubert and wavlm. So maybe something like DefaultHuggingfaceTransformer? wdyt @Adel-Moumen  @pplantinga and @mravanelli

speechbrain/lobes/models/huggingface.py
mhn226: I put it there because whisper puts it: https://github.com/speechbrain/speechbrain/blob/024478de11a75d8b8fa0dfbdd0cddcf5423e1358/speechbrain/lobes/models/huggingface_whisper.py#L100

speechbrain/lobes/models/huggingface.py
mhn226: it for modifying the state_dict when needed. For example with wav2vec, loading sb model would require some modifications to be compatible (https://github.com/speechbrain/speechbrain/blob/024478de11a75d8b8fa0dfbdd0cddcf5423e1358/speechbrain/lobes/models/huggingface_wav2vec.py#L203)
This should be overwritten by users depending on the model in use.

mhn226:
> Thanks for the work! I will request ideas from @Adel-Moumen and @pplantinga here. I think I like this easy to use abstraction. The main remaining question here being: do we put everything in a single file called HuggingfaceTransformers.py or do we create a directory with multiple files, with each file containing our implementation of a model? Also we have to think about one thing: dropping the wav2vec2 in the naming of our implementation, because it also works with Hubert and wavlm. So maybe something like DefaultHuggingfaceTransformer? wdyt @Adel-Moumen @pplantinga and @mravanelli

what is default huggingface transformer exactly? I agree that "wav2vec" is not representative for hubert and wavlm, but  DefaultHuggingfaceTransformer is a bit too much? :D

speechbrain/lobes/models/huggingface.py
TParcollet: This could be solved by passing a **kwarg instead. This is to avoid having 1 billion arguments once we have many model implemented

speechbrain/lobes/models/huggingface.py
TParcollet: Can we remove the partial_fn and add a description to explain what it is for?

TParcollet:
Well, I have no idea actually, trying to build a name out of the fact that a lot of models share the same default encoder structure, and this is why they can use the same class.

Adel-Moumen:
> Thanks for the work! I will request ideas from @Adel-Moumen and @pplantinga here. I think I like this easy to use abstraction. The main remaining question here being: do we put everything in a single file called HuggingfaceTransformers.py or do we create a directory with multiple files, with each file containing our implementation of a model? Also we have to think about one thing: dropping the wav2vec2 in the naming of our implementation, because it also works with Hubert and wavlm. So maybe something like DefaultHuggingfaceTransformer? wdyt @Adel-Moumen @pplantinga and @mravanelli

I like the idea of having a new folder (e.g. called `huggingface_transformers`)  with multiple files. We need to separate each model. So, having `wav2vec.py`, `wavlm.py`, etc... 

Why should we drop the `wav2vec2` naming ? We could just leave it as it is and then for `wavlm.py` create a new class which inherits from the class in `wav2vec.py` (e.g. `WavLM(Wav2vec2)`). Even though we are not going to change anything (e.g. fwd etc), but from an end user perspective, it could be cleaner to do that.

speechbrain/lobes/models/huggingface.py
mhn226: Took these two from huggingface_wav2vec.py

mhn226:
I added a huggingface_transformers directory inside speechbrain/lobes/models which hosts the interface huggingface.py, wav2vec class huggingface_wav2vec.py, and whisper class huggingface_whisper.py. 
This would slightly affects some recipes which were modified accordingly.
There should be a huggingface_wavlm.py and huggingface_hubert.py added if we follow @Adel-Moumen  and @TParcollet 's suggestion. They should technically resemble huggingface_wav2vec.py. I will soon add them.

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,372 @@
...
+        self.feature_extractor = AutoFeatureExtractor.from_pretrained(
+            source, cache_dir=cache_dir, **kwarg
+        )
+
+    def load_tokenizer(self, source, **kwarg):
+        """Load model's tokenizer from the hub"""
+        self.tokenizer = AutoTokenizer.from_pretrained(source, **kwarg)
+
+
+def make_padding_masks(src, wav_len=None, pad_idx=0):
```
TParcollet, line 337: Did you double-check that we don't have this somewhere already? I really think we do.

TParcollet (commented):
I went again throughout the code. I like the current state of it. I think it is close to be ready. I will just ask @Adel-Moumen to run recipe testing to make sure that nothing breaks. Also @mhn226 could you please modify **all** the tutorials? You will need to manually check unfortunately ...

Adel-Moumen:
I think we should change the name of `huggingface_whisper` to only `whisper`. The class is already located in `speechbrain/lobes/models/huggingface_transformers` so it is already obvious that whisper is coming from HuggingFace... This comment applies to all models including wav2vec. 

Furthermore, @TParcollet, you said you wanted to separate `WavLM` from `Hubert` etc... this is not the current state of the code. We are still creating a `WavLM` obj through the `huggingface_wav2vec` class! What about creating new files for them  ? (e.g. `wavlm.py` which is inheriting from `wav2vec` etc...)

Adel-Moumen:
Note: when merging, please select merge with git stash!

mhn226:
> I think we should change the name of `huggingface_whisper` to only `whisper`. The class is already located in `speechbrain/lobes/models/huggingface_transformers` so it is already obvious that whisper is coming from HuggingFace... This comment applies to all models including wav2vec.
> 
> Furthermore, @TParcollet, you said you wanted to separate `WavLM` from `Hubert` etc... this is not the current state of the code. We are still creating a `WavLM` obj through the `huggingface_wav2vec` class! What about creating new files for them ? (e.g. `wavlm.py` which is inheriting from `wav2vec` etc...)

I somehow don't have problems with the file names because huggingface.py would cease to nothing.py following your logic, unless you have another suggestion for it?
I have huggingface_wavlm.py and huggingface_hubert.py already. I will push them soon!

Adel-Moumen:
> > I think we should change the name of `huggingface_whisper` to only `whisper`. The class is already located in `speechbrain/lobes/models/huggingface_transformers` so it is already obvious that whisper is coming from HuggingFace... This comment applies to all models including wav2vec.
> > Furthermore, @TParcollet, you said you wanted to separate `WavLM` from `Hubert` etc... this is not the current state of the code. We are still creating a `WavLM` obj through the `huggingface_wav2vec` class! What about creating new files for them ? (e.g. `wavlm.py` which is inheriting from `wav2vec` etc...)
> 
> I somehow don't have problems with the file names because huggingface.py would cease to nothing.py following your logic, unless you have another suggestion for it? I have huggingface_wavlm.py and huggingface_hubert.py already. I will push them soon!

No, it could be a special case or could be renamed to something like `base_transformers.py` or `base_huggingface.py` idk. I think this is really redundant to name them `huggingface_*` while there are already in the `huggingface` folder. Everything gets longer for 0 amount of informations. In a yaml file its 12 chars in more. What do you think @TParcollet @mravanelli ?

mhn226:
> > > I think we should change the name of `huggingface_whisper` to only `whisper`. The class is already located in `speechbrain/lobes/models/huggingface_transformers` so it is already obvious that whisper is coming from HuggingFace... This comment applies to all models including wav2vec.
> > > Furthermore, @TParcollet, you said you wanted to separate `WavLM` from `Hubert` etc... this is not the current state of the code. We are still creating a `WavLM` obj through the `huggingface_wav2vec` class! What about creating new files for them ? (e.g. `wavlm.py` which is inheriting from `wav2vec` etc...)
> > 
> > 
> > I somehow don't have problems with the file names because huggingface.py would cease to nothing.py following your logic, unless you have another suggestion for it? I have huggingface_wavlm.py and huggingface_hubert.py already. I will push them soon!
> 
> No, it could be a special case or could be renamed to something like `base_transformers.py` or `base_huggingface.py` idk. I think this is really redundant to name them `huggingface_*` while there are already in the `huggingface` folder. Everything gets longer for 0 amount of informations. In a yaml file its 12 chars in more. What do you think @TParcollet @mravanelli ?

Ok I see your point. I'll modify the model files as your suggestion. I'll keep huggingface.py as it is because "base" would create undesirable confusion with the model size in my opinion.

Adel-Moumen:
> > > > I think we should change the name of `huggingface_whisper` to only `whisper`. The class is already located in `speechbrain/lobes/models/huggingface_transformers` so it is already obvious that whisper is coming from HuggingFace... This comment applies to all models including wav2vec.
> > > > Furthermore, @TParcollet, you said you wanted to separate `WavLM` from `Hubert` etc... this is not the current state of the code. We are still creating a `WavLM` obj through the `huggingface_wav2vec` class! What about creating new files for them ? (e.g. `wavlm.py` which is inheriting from `wav2vec` etc...)
> > > 
> > > 
> > > I somehow don't have problems with the file names because huggingface.py would cease to nothing.py following your logic, unless you have another suggestion for it? I have huggingface_wavlm.py and huggingface_hubert.py already. I will push them soon!
> > 
> > 
> > No, it could be a special case or could be renamed to something like `base_transformers.py` or `base_huggingface.py` idk. I think this is really redundant to name them `huggingface_*` while there are already in the `huggingface` folder. Everything gets longer for 0 amount of informations. In a yaml file its 12 chars in more. What do you think @TParcollet @mravanelli ?
> 
> Ok I see your point. I'll modify the model files as your suggestion. I'll keep huggingface.py as it is because "base" would create undesirable confusion with the model size in my opinion.

Yes. Good.

Adel-Moumen:
Coud you please sync your PR with the latests commits in the dev branch @mhn226 ?

speechbrain/lobes/models/huggingface_transformers/hubert.py
```diff
@@ -0,0 +1,88 @@
...
+import logging
+
+from speechbrain.lobes.models.huggingface_transformers.wav2vec import (
+    HuggingFaceWav2Vec2,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class HuggingFaceHuBERT(HuggingFaceWav2Vec2):
```
Adel-Moumen, line 24: What about changing the name as well ? Makes more sense to call it 'HuBert' than 'HuggingFaceHuBERT' since it is already part of the folder 'huggingface_transformers'. (the comment applies to all the models e.g. wav2vec whisper etc.)

mhn226:
@Adel-Moumen : I changed some classes' names following your suggestion:

- HuggingFaceTransformer -> HFTransformersInterface
- HuggingFaceWav2Vec2 -> Wav2Vec2
- HuggingFaceWavLM -> WavLM
- HuggingFaceHuBERT -> HuBERT
- HuggingFaceWhisper -> Whisper
I adapted accordingly the yaml files of affected recipes.

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+    >>> model = HFTransformersInterface(model_hub, save_path=save_path)
+    """
+
+    def __init__(
+        self,
+        source,
+        save_path,
+        for_pretraining=False,
+        freeze=False,
+        cache_dir: Union[str, pathlib.Path, None] = "pretrained_models",
```
Adel-Moumen, line 94: you are using typing, please use it everywhere if possible... :)

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+            self.model.train()
+
+    def _from_pretrained(
+        self, source, config, model, save_path, cache_dir,
+    ):
+        """This function manages the source checking and loading of the params.
+        # 1. Is the model from HF or a local path
+        # 2. Is the model pretrained with HF or SpeechBrain
+        # 3. Download (if appropriate) and load with respect to 1. and 2.
+        Arguments
```
Adel-Moumen, line 141: add a return line to space the docstring with the arguments

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+            )
+            self.freeze_model(self.model)
+        else:
+            self.model.gradient_checkpointing_disable()  # Required by DDP
+            self.model.train()
+
+    def _from_pretrained(
+        self, source, config, model, save_path, cache_dir,
+    ):
+        """This function manages the source checking and loading of the params.
```
Adel-Moumen, line 137: return line to give some space

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+                # To be modified in the future to support more complicated scenerios
+                # For example fine-tuning in the SSL manner
+                self.model = model
+            else:
+                self.model = model.from_pretrained(
+                    source, config=config, cache_dir=save_path
+                )
+
+    def _check_model_source(self, path, save_path):
+        """Checks if the pretrained model has been trained with SpeechBrain and
```
Adel-Moumen, line 182: I would add a line explaining why we are checking the model source.

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        is hosted locally or on a HuggingFace hub.
+        Called as static function in HFTransformersInterface._from_pretrained.
+        Arguments
```
Adel-Moumen, line 185: return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        ---------
+        path : str
+            Used as "source"; local path or HuggingFace hub name: e.g "facebook/wav2vec2-large-lv60"
+        save_path : str
+            norm_output (dir) of the downloaded model.
+        Returns
```
Adel-Moumen, line 191: return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        -------
+        is_sb : bool
+            Whether/not the model is deserializable w/ SpeechBrain or not (then, model conversion is needed).
+        checkpoint_filename : str
```
Adel-Moumen, line 195: We are returning `return is_sb, checkpoint_filename, is_local`, please update the return docstring accordingly . We are missing the `is_local` variable

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+                if File.rfilename.endswith(".bin"):
+                    checkpoint_filename = File.rfilename
+                    is_sb = False
+                    return is_sb, checkpoint_filename, is_local
+
+        err_msg = f"{path} does not contain a .bin or .ckpt checkpoint !"
+        raise FileNotFoundError(err_msg)
+
+    def _modify_state_dict(self, path, **args):
+        """A custom loading ensures SpeechBrain compatibility for pretrain and model
```
Adel-Moumen, line 259: a return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        For example, wav2vec2 model pretrained with SB (Wav2Vec2Pretrain) has slightly different keys from Wav2Vec2.
+        This method handle the compatibility between the two.
+        Users should modify this function according to their own tasks.
+        Arguments
```
Adel-Moumen, line 263: line for space

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        path : str
+            Used as "source"; local path or HuggingFace hub name: e.g "facebook/wav2vec2-large-lv60"
+        save_path : str
+            norm_output (dir) of the downloaded model.
+        Returns
+        -------
+        is_sb : bool
+            Whether/not the model is deserializable w/ SpeechBrain or not (then, model conversion is needed).
+        checkpoint_filename : str
+            as of HuggingFace documentation: file name relative to the repo root (guaranteed to be here).
```
Adel-Moumen, line 196: add a Raise docstring please (see: https://github.com/speechbrain/speechbrain/blob/91cac2feee51febcbc6a34f7675234baac6ad1a6/speechbrain/pretrained/fetching.py#L112)

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _load_sb_pretrained_parameters(self, path, modify_state_dict_fn):
+        """Loads the parameter of a HuggingFace model pretrained with SpeechBrain
+        and the HuggingFace Pretrain Object. It is necessary to perform a custom
+        loading because HuggingFace adds a level to the checkpoint when storing
+        the model breaking the compatibility Pretrain and model de/serialization.
+
+        For example, a typical Wav2Vec2 checkpoint for a given parameter
+        would be: model.conv.weight.data while for Wav2Vec2Pretrain it
+        is: model.wav2vec2.weight.data (wav2vec2 must be removed before loading).
```
Adel-Moumen, line 278: return line please

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        model.eval()
+        for param in model.parameters():
+            param.requires_grad = False
+
+    def override_config(self, config):
+        """Users should modify this function according to their own tasks."""
+        return config
+
+    def load_feature_extractor(self, source, cache_dir, **kwarg):
+        """Load model's feature_extractor from the hub
```
Adel-Moumen, line 347: return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+            HuggingFace hub name: e.g "facebook/wav2vec2-large-lv60"
+        cache_dir : str
+            Path (dir) in which a downloaded pretrained model configuration should be cached.
+        """
+        self.feature_extractor = AutoFeatureExtractor.from_pretrained(
+            source, cache_dir=cache_dir, **kwarg
+        )
+
+    def load_tokenizer(self, source, **kwarg):
+        """Load model's tokenizer from the hub
```
Adel-Moumen, line 360: return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        source : str
+            HuggingFace hub name: e.g "facebook/wav2vec2-large-lv60"
+        cache_dir : str
+            Path (dir) in which a downloaded pretrained model configuration should be cached.
+        """
+        self.feature_extractor = AutoFeatureExtractor.from_pretrained(
+            source, cache_dir=cache_dir, **kwarg
+        )
+
+    def load_tokenizer(self, source, **kwarg):
```
Adel-Moumen, line 359: if you are using `**kwarg` then you have to put it in the docstring

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        model.eval()
+        for param in model.parameters():
+            param.requires_grad = False
+
+    def override_config(self, config):
+        """Users should modify this function according to their own tasks."""
+        return config
+
+    def load_feature_extractor(self, source, cache_dir, **kwarg):
```
Adel-Moumen, line 346: if you are using `**kwarg` then you have to put it in the docstring

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+            for File in files:
+                if File.rfilename.endswith(".bin"):
+                    checkpoint_filename = File.rfilename
+                    is_sb = False
+                    return is_sb, checkpoint_filename, is_local
+
+        err_msg = f"{path} does not contain a .bin or .ckpt checkpoint !"
+        raise FileNotFoundError(err_msg)
+
+    def _modify_state_dict(self, path, **args):
```
Adel-Moumen, line 258: `**args` -> `**kwargs` + put it in the docstring by adding a small line explaining on which case it might be useful to give more parameters

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        This method handle the compatibility between the two.
+        Users should modify this function according to their own tasks.
+        Arguments
+        ---------
+        path : str
+            Checkpoint path, file name relative to the repo root.
+        """
+        return
+
+    def _load_sb_pretrained_parameters(self, path, modify_state_dict_fn):
```
Adel-Moumen, line 270: I don't understand the usefulness of the argument `modify_state_dict_fn`. From my understanding, this is always a `self` method. Why not just calling `self._modify_state_dict` inside?

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        ---------
+        path : pathlib.Path
+            The full path to the checkpoint.
+        modify_state_dict_fn : method
+            A customized method which modifies the checkpoint's state_dict.
+        """
+        if modify_state_dict_fn is not None:
+            modified_state_dict = modify_state_dict_fn(path)
+        else:
+            modified_state_dict = torch.load(path, map_location="cpu")
```
Adel-Moumen, line 289: It seems that this line is never used. `modify_state_dict_fn` is always different to `None`

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        ---------
+        source : str
+            HuggingFace hub name: e.g "facebook/wav2vec2-large-lv60"
+        """
+        self.tokenizer = AutoTokenizer.from_pretrained(source, **kwarg)
+
+
+def make_padding_masks(src, wav_len=None, pad_idx=0):
+    """This method generates the padding masks.
+    Arguments
```
Adel-Moumen, line 371: a return line

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        The relative length of the wav given in SpeechBrain format.
+    pad_idx : int
+        The index for <pad> token (default=0).
+    """
+    src_key_padding_mask = None
+    if wav_len is not None:
+        abs_len = torch.round(wav_len * src.shape[1])
+        src_key_padding_mask = length_to_mask(abs_len).bool()
+
+    return src_key_padding_mask
```
Adel-Moumen, line 385: please add a Return docstring

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+def make_masks(src, wav_len=None, pad_idx=0):
+    """This method generates the padding masks.
```
Adel-Moumen, line 389: 1. return line 
2. add Returns docstring

speechbrain/lobes/models/huggingface_transformers/wav2vec.py
```diff
@@ -96,7 +82,7 @@ class HuggingFaceWav2Vec2(nn.Module):
     >>> inputs = torch.rand([10, 600])
     >>> model_hub = "facebook/wav2vec2-base-960h"
     >>> save_path = "savedir"
-    >>> model = HuggingFaceWav2Vec2(model_hub, save_path)
+    >>> model = Wav2Vec2(model_hub, save_path)
```
Adel-Moumen, line 85: add import in the example

speechbrain/lobes/models/huggingface_transformers/wav2vec.py
```diff
@@ -398,7 +226,7 @@ class HuggingFaceWav2Vec2Pretrain(nn.Module):
     >>> inputs = torch.rand([10, 32000])
     >>> model_hub = "facebook/wav2vec2-base-960h"
     >>> save_path = "savedir"
-    >>> model = HuggingFaceWav2Vec2Pretrain(model_hub, save_path)
+    >>> model = Wav2Vec2Pretrain(model_hub, save_path)
```
Adel-Moumen, line 229: add import

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        Arguments
+        ---------
+        model : from AutoModel.from_config
+            Valid HuggingFace transformers model object.
+        """
+        model.eval()
+        for param in model.parameters():
+            param.requires_grad = False
+
+    def override_config(self, config):
```
Adel-Moumen, line 342: the docstring please

speechbrain/lobes/models/huggingface_transformers/huggingface.py
```diff
@@ -0,0 +1,404 @@
...
+        """A custom loading ensures SpeechBrain compatibility for pretrain and model
+        For example, wav2vec2 model pretrained with SB (Wav2Vec2Pretrain) has slightly different keys from Wav2Vec2.
+        This method handle the compatibility between the two.
+        Users should modify this function according to their own tasks.
+        Arguments
+        ---------
+        path : str
+            Checkpoint path, file name relative to the repo root.
+        """
+        return
```
Adel-Moumen, line 268: `return` -> `return None`. Better to explicit it.

speechbrain/lobes/models/huggingface_transformers/wav2vec.py
```diff
@@ -533,8 +348,8 @@ class WeightedSSLModel(torch.nn.Module):
     """
 
     def __init__(self, hub, num_layers, layernorm=False):
-        super().__init__()
-        self.encoder = AutoModel.from_pretrained(hub, output_hidden_states=True)
+        super().__init__(source=hub, save_path=".")
```
Adel-Moumen, line 351: why is `save_path` force to be `"."` ?

speechbrain/lobes/models/huggingface_transformers/wavlm.py
```diff
@@ -0,0 +1,86 @@
...
+"""
+
+import logging
+
+from speechbrain.lobes.models.huggingface_transformers.wav2vec import Wav2Vec2
+
+logger = logging.getLogger(__name__)
+
+
+class WavLM(Wav2Vec2):
```
Adel-Moumen, line 22: add a line explaining why we are using `Wav2vec2`

Adel-Moumen (changes requested):
HI Ha, 

Thanks for the PR this is a really good work. I left some comments that I think should be addressed. 

I just had one question: have you checked that everything is working? Recipe tests? Pretrained inference test? 

If no, I suggest to run the recipe tests and to wget a dropbox pretrained model and carefully check that we still are getting the same WER. 

One additional note: please change the base of your PR to unstable-v6 and sync everything. 

Thanks for your work.

mhn226:
Hi, so this PR passed the recipe tests (thanks to @Adel-Moumen for doing the tests). The remaining problem is that it doesn't seem to give the same scores when loading some pre-trained (old) models we published on dropbox. For example, this model for [Italian](https://www.dropbox.com/sh/46z6xbydqjaxbb2/AACyQ_NaErAfev6JVjeaIWyea?dl=0) from [this recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/seq2seq/hparams/train_it_with_wav2vec.yaml) gave super weird WER. And it's the same if I do the test with the develop branch. Retraining the model from scratch gave the same outcome for both this PR and the develop branch.

With this recipe, [Italian model](https://www.dropbox.com/sh/0v2o2hmrv1j33p6/AAA3xUiqKbSKsX88fWfptPmFa?dl=0) from [this recipe](https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/ASR/CTC/hparams/train_it_with_wav2vec.yaml), we have relatively the same scores as what we published (tested with both this PR and the develop branch). So it's good here.

But wav2vec+CTC doesn't work with the [french model](https://www.dropbox.com/sh/wytlbeddrt8oe4n/AAAY59qMsDlWy5F017bmBeVua?dl=0). same problem!

So I don't know how we can conclude? I personally think that if we see the same scores between this PR and the develop branch, we can say it is OK? The underlying problem doesn't come from the PR? Can you give your thoughts? @Adel-Moumen  @TParcollet  @mravanelli

speechbrain/lobes/models/huggingface_transformers/whisper.py
```diff
@@ -83,60 +80,66 @@ def __init__(
...
         self.output_all_hiddens = output_all_hiddens
 
-        self.tokenizer = None
-        # Download the tokenizer only if we are going to use the Decoder.
-        if not encoder_only:
-            self.tokenizer = WhisperTokenizer.from_pretrained(source)
+        if encoder_only:
+            self.tokenizer = None
+        else:
+            self.load_tokenizer(source)
```
Adel-Moumen, line 98: why ?

mhn226 (commented):
Adapt changes to comments from @TParcollet and @Adel-Moumen

mravanelli:
Hi @mhn226, I suggest merging the latest version of develop. We recently merge  a new recipes that used HF to get a pretrained version of GPT2 (see [here](https://github.com/speechbrain/speechbrain/blob/develop/recipes/MultiWOZ/response_generation/hparams/train_gpt.yaml#L91)). Do you think we should integrate it here?

mhn226:
> Hi @mhn226, I suggest merging the latest version of develop. We recently merge a new recipes that used HF to get a pretrained version of GPT2 (see [here](https://github.com/speechbrain/speechbrain/blob/develop/recipes/MultiWOZ/response_generation/hparams/train_gpt.yaml#L91)). Do you think we should integrate it here?

Hi Mirco, I also plan to implement text-based pretrained models like mbart, labse, nllb, etc. too.
What are your opinions about merging to develop? @TParcollet  @Adel-Moumen

mravanelli:
Note that @poonehmousavi is working on an HF interface to integrate Llama2. Maybe you guys can talk to each other to better coordinate your efforts.

Note: we need to merge in unstable. We just need to update the unstable branch and the code in your PR with the latest develop such that you have have the new GPT2 recipe

poonehmousavi:
> > Note that @poonehmousavi is working on an HF interface to integrate Llama2. Maybe you guys can talk to each other to better coordinate your efforts.
> > Note: we need to merge in unstable. We just need to update the unstable branch and the code in your PR with the latest develop such that you have have the new GPT2 recipe
> 
> OK I see. Thanks for the clarification!

@mhn226 , The gpt recipe is already merged, it is not that much different from other HF models that we previously have. But it might be different slightly for LLAMA2. I will add you to the LLAMA2 PR when I feel more confident about it. Let me know if I could be of any help.

mravanelli:
I merged the latest version of dev, which includes the GPT2 fine-tuning recipe implemented by @poonehmousavi. @mhn226, could you please revise it and integrate into the HF refactor?

mhn226:
> I merged the latest version of dev, which includes the GPT2 fine-tuning recipe implemented by @poonehmousavi. @mhn226, could you please revise it and integrate into the HF refactor?

Hi Marco, yes I'm working on that. Thank you for the merging!

recipes/MultiWOZ/response_generation/hparams/train_gpt.yaml
```diff
@@ -88,7 +88,7 @@ test_dataloader_options:
 padding_mask: !name:speechbrain.lobes.models.transformer.Transformer.get_key_padding_mask
 
 # gpt model
-gpt_model: !new:speechbrain.lobes.models.huggingface_gpt.HuggingFaceGPT
+gpt_model: !new:speechbrain.lobes.models.huggingface_transformers.gpt.GPT
```
mhn226, line 91: @poonehmousavi : this got changed because we created a dir for all HF models (huggingface_transformers), in which gpt.py is located. Long story but we omitted "huggingface" everywhere in the py file names and also the name of the model classes because we believe that the huggingface_transformers dir implicitly indicates the HF nature of these models.

recipes/MultiWOZ/response_generation/train_with_gpt.py
```diff
@@ -426,9 +425,7 @@ def input_and_token_type_pipeline(
     )
 
     # Load tokenizer and add special tokens
-    tokenizer = GPT2Tokenizer.from_pretrained(
```
mhn226, line 429: @poonehmousavi : I removed these lines because the tokenizer will be automatically loaded by the GPT class.

recipes/MultiWOZ/response_generation/train_with_gpt.py
```diff
@@ -426,9 +425,7 @@ def input_and_token_type_pipeline(
     )
 
     # Load tokenizer and add special tokens
-    tokenizer = GPT2Tokenizer.from_pretrained(
-        hparams["gpt_hub"], pad_token=None
-    )
+    tokenizer = hparams["gpt_model"].tokenizer
```
mhn226, line 428: @poonehmousavi : here I call the tokenizer of the GPT model

speechbrain/lobes/models/huggingface_transformers/gpt.py
```diff
@@ -0,0 +1,151 @@
...
+
+from speechbrain.lobes.models.huggingface_transformers.huggingface import (
+    HFTransformersInterface,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class GPT(HFTransformersInterface):
```
mhn226, line 23: @poonehmousavi : This class defines the GPT model. Could you please complete the docstring? I think it lacks quite a lot of the parameters used by the class. I could do it but I think you're more familiar with the class.

speechbrain/lobes/models/huggingface_transformers/gpt.py
```diff
@@ -0,0 +1,151 @@
...
+    ) -> None:
+        super().__init__(
+            source=source, save_path=save_path, freeze=freeze, with_lm_head=True
+        )
+        self.max_new_tokens = max_new_tokens
+        self.min_length = min_length
+        self.top_k = top_k
+        self.top_p = top_p
+        self.num_beams = num_beams
+        self.early_stopping = early_stopping
```
mhn226, line 74: @poonehmousavi : I made eos_token_id a param because different gpt models might have different eos_token_id. This especially true when a multilingual model is used.

speechbrain/lobes/models/huggingface_transformers/gpt.py
```diff
@@ -0,0 +1,151 @@
...
+
+        self.load_tokenizer(source=source, pad_token=None, use_fast=False)
```
mhn226, line 76: @poonehmousavi : here the tokenizer is loaded. "use_fast" is set False because I see that you specifically use GPT2Tokenizer instead of GPT2TokenizerFast which is loaded by default by AutoTokenizer.

mhn226 (commented):
Hi @poonehmousavi : I adapt your recipe to this HF interface PR. Could you please take a look at the changes, especially at my comments? Please leave some comments if needed.

P/s: I see that you're using directly the generate() method from HF. I don't have any problem with that but I wonder if we want to use our all decoding method in the future as @Adel-Moumen has been making wonderful effort in the scoring and searching part?

Adel-Moumen:
> P/s: I see that you're using directly the generate() method from HF. I don't have any problem with that but I wonder if we want to use our decoding method in the future as @Adel-Moumen has been making wonderful effort in the scoring and searching part?

No. This is a different beam search  :)

mravanelli:
Thank you @mhn226 and all the other contributors for this PR. I did some tests and everything LGTM. It significantly improves the way we fetch models from HuggingFace.

PR #2119, opened 2023-08-11
add FastSpeech2 with internal alignment

BenoitWang:
# Contribution in a nutshell
This recipe allows training FastSpeech2 without forced aligner referring to [One TTS Alignment To Rule Them All](https://arxiv.org/pdf/2108.10447.pdf). The alignment can be learned by an internal alignment network that is added to FastSpeech2. 

This PR aims to:
1. Provide a better alignment for punctuation
2. Facilitates FastSpeech2 training when using custom data (everything can be done inside speechbrain).

As for the phonemes, they are automatically obtained with the speechbrain g2p model during preparation.

# Scope
* [x] Benchmark the training period with binary alignment loss
* [x] Benchmark if a higher dropout is needed
* [x] Docstrings and examples
* [x] Dropbox + HF


mravanelli:
Thank you, @BenoitWang , for your PR contribution! t's great to see this valuable feature being added to the FastSpeech model. 
I'm running the recipe, but I still have some comments to share with you:

1. **Recipe Tests:**
The recipe tests are currently failing. Please run the following command to check:
```bash
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Dataset"], filters=[["LJSpeech"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```
I encountered this error:
```
RuntimeError: These keys are still unaccounted for in the data pipeline: phonemes
```

2. **FastSpeech2.py:**
In the `speechbrain/lobes/models/FastSpeech2.py` file, please consider the following improvements:
- Add a docstring example for the `LossWithAlignment` class.
- Describe the arguments and add a docstring example for the `ForwardSumLoss` class.
- Describe the arguments and add a docstring example for the `BinaryAlignmentLoss` class.
- Add a docstring example for the `maximum_path_numpy` function.
- Consider whether it makes sense to write a docstring example for the `TextMelCollateWithAlignment` class.

3. **interfaces.py:**
In the `speechbrain/pretrained/interfaces.py` file, I noticed the addition of `FastSpeech2InternalAlignment`. Is there are reason for that? Ideally, we should strive to reuse the existing `FastSpeech2` interface if possible and convenient.

4. **text_to_sequence.py:**
For the `g2p_keep_punctuations` function in `speechbrain/utils/text_to_sequence.py`, please describe the arguments and add a docstring for clarity.

BenoitWang:
Hi @mravanelli , thanks for the comments! I've fixed the recipe test and added the necessary docstrings and examples.
However, for `LossWithAlignment`, since it's a summary of the other losses, an example takes too many inputs and seems repeated because an example can be found in each component loss.
As for the new interface, I created a new one mainly because the outputs of the models are different, the internal alignment model outputs the alignments in addition. The inference processes are very different as well, for example the original one needs the spn prediction while the new one focuses on keeping the punctuations during g2p.

mravanelli (approved):

PR #2123, opened 2023-08-13
Specify min_key in evaluate for ASR recipes

mravanelli:
In some ASR recipes, we did not specify the min_key when calling the final evaluation. This means that we evaluate the last checkpoint rather than the one with the best performance on the validation set. 

This PR fixes this issue for ASR recipes. I ran recipe tests, and all seem to work.


Adel-Moumen:
All checks are passing. Thanks @mravanelli for your PR! :)

Adel-Moumen (approved):
LGTM!

PR #2125, opened 2023-08-13
replacements applied only on wav column

mrouvier:
The "load_data_csv" function aims to replace all occurrences of variables (begin with the symbol $) with their corresponding values across all columns of the CSV file.

The problem is that the function can mistakenly identify a variable when it was just a word (a word that starting with $ and followed by a few characters). And these misinterpretations can frequently occur when the text undergoes transliteration.

Therefore, I propose to limit the application of this substitution exclusively to the row "wav".

### DIFF
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -143,16 +143,17 @@ def load_data_csv(csv_path, replacements={}):
             if data_id in result:
                 raise ValueError(f"Duplicate id: {data_id}")
             # Replacements:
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you @mrouvier for pointing out this issue and sketching a solution. The problem with the proposed approach is that the field "wav" is not mandatory. There might be recipes with other names for the input file field where replacement is needed. In some cases  multiple fields must be replaces. I'm wondering if there is another solution for the issue you pointed.

PR #2128, opened 2023-08-14
fix ddp issue with loading a key

Adel-Moumen:
revert commit https://github.com/speechbrain/speechbrain/commit/629b1b1a6fdb7699674e7bf78e03affcf28b165e which is making our DDP recipes to fail when loading a key in an HuggingFace repo (e.g. speech-to-speech translation or librispeech). 

Note: it will make our recipe tests to fail (only the recipes which are using relative paths for loading tokenizer etc). 

Note: to reproduce please see: https://github.com/speechbrain/speechbrain/pull/2081#issuecomment-1644007780

### DIFF
--- a/speechbrain/utils/parameter_transfer.py
+++ b/speechbrain/utils/parameter_transfer.py
@@ -61,7 +61,6 @@ class Pretrainer:
         conditions=None,
     ):
         self.loadables = {}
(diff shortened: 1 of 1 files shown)


Gastron:
LGTM!

Gastron (approved):

PR #2130, opened 2023-08-14
[WIP] DDP recipe tests and fixes

mravanelli:
Ideally, we want our recipes to work with different devices. Ideally, they should work with:

- [ ] Single GPU (cuda:0)
- [ ] Single GPU (cuda:1) => This is to test if there are some recipes with cuda:0 hard-coded
- [ ] Single CPU
- [ ] Multiple GPUs on the same node
- [ ] Multiple CPUs on the same node
- [ ] Multiple GPUs on the different nodes
- [ ] Multiple CPUs on the different nodes

The goal of this PR is allow recipe tests to run on multiple GPUs and fix the issues that cause tests to fail.

### DIFF
--- a/recipes/SLURP/direct/hparams/train.yaml
+++ b/recipes/SLURP/direct/hparams/train.yaml
@@ -60,9 +60,7 @@ epoch_counter: !new:speechbrain.utils.epoch_loop.EpochCounter
     limit: !ref <number_of_epochs>
 
 # Models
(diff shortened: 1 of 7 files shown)


mravanelli:
I limited this PR to fix the issues emerging when using a not default cuda device (e.g, cuda:1). I will follow up in the future with other PRs to extend tests on multi-GPU training.

PR #2132, opened 2023-08-15
Save the checkpoint folder and meta only on the main process

pplantinga:
PR #2059 has an issue where checkpoint folders are created on all processes, this PR fixes the issue by saving them only on the main process.


speechbrain/utils/checkpoints.py
```diff
@@ -561,14 +561,20 @@ def save_checkpoint(
...
+                ckpt_dir = self._new_checkpoint_dirpath()
+            else:
+                ckpt_dir = self._custom_checkpoint_dirpath(name)
+            os.makedirs(ckpt_dir)  # May raise FileExistsError, let it.
+            saved_meta = self._save_checkpoint_metafile(
+                ckpt_dir / METAFNAME, meta, end_of_epoch
+            )
+
+        # Communicate ckpt_dir to all procs
+        torch.distributed.broadcast_object_list([ckpt_dir], src=0)
```
pplantinga, line 576: I'm a little unsure if this works when using multiple machines, perhaps we would need to use the concept of process groups and only communicate within the group somehow.

speechbrain/utils/checkpoints.py
```diff
@@ -590,9 +596,13 @@ def save_checkpoint(
...
-        ckpt_type = "end-of-epoch" if end_of_epoch else "intra-epoch"
-        logger.log(verbosity, f"Saved an {ckpt_type} checkpoint in {ckpt_dir}")
-        return Checkpoint(ckpt_dir, saved_meta, saved_paramfiles)
+
+        if if_main_process():
+            ckpt_type = "end-of-epoch" if end_of_epoch else "intra-epoch"
+            logger.log(
+                verbosity, f"Saved an {ckpt_type} checkpoint in {ckpt_dir}"
+            )
+            return Checkpoint(ckpt_dir, saved_meta, saved_paramfiles)
```
pplantinga, line 605: Returns `None` on all but the main process.

Adel-Moumen:
One note: with DDP we also have some little issues with our Loggers e.g. wandb. Each process is running the init of the logger while it should only be on the main process. CC @Chaanks

pplantinga:
> One note: with DDP we also have some little issues with our Loggers e.g. wandb. Each process is running the init of the logger while it should only be on the main process.

For any methods/functions that should only run on main, we have a decorator: `@main_process_only` that can be added. We could open a new PR to fix the loggers if that's the issue.

speechbrain/utils/checkpoints.py
Gastron: This is fine, but should be documented.

Gastron (changes requested):
Could you document the way this works in the docstring for save_checkpoint (or for the whole class). Other than that, looks good to me.

As discussed privately, we should look into running multi-process tests. Perhaps there could multi-process unittests for classess like the Checkpointer.

pplantinga:
I attempted to write a multi-process test but found it difficult due to the way my computer is set up. I'm wondering if it may be hard to write a general enough multi-process test that works on most machines. Anyone else have ideas about how this could be done?

Gastron (approved):

Gastron:
Looks good to me, I recommend we merging soon as this is an important fix. 

Thanks for looking into multi-process tests! I guess we should take a look at what others are doing, perhaps someone has writetn something on multi-process unittests that we should read. Let's leave that as a separate quest.

This is marked as a draft though, is that

Adel-Moumen (commented):
Sounds good to me. Is it ready to be tested or merged?

mravanelli:
I think we need to test it before merging it. Do you think we can create
some unittest?) as well?

pplantinga:
The trouble I ran into with trying to run DDP on my machine (which may or may not also be the case in the GitHub test environment) was that to initialize DDP you have to specify a web address (e.g. "localhost") and a port (e.g. 12355), but these were blocked on my computer. I could add the test I mocked up and just see if it works without testing locally.

mravanelli:
that makes sense.

tests/unittests/test_checkpoints.py
```diff
@@ -385,3 +381,44 @@ def test_torch_defaults(tmpdir, device):
...
     checkpointer.load_checkpoint(ckpt)
     assert torch.allclose(module(inp), prev_output)
+
+
+def parallel_checkpoint(rank, world_size, tmpdir):
+    import os
+    from speechbrain.utils.checkpoints import Checkpointer
+
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = "12355"
```
pplantinga, line 391:
```suggestion
    os.environ["MASTER_PORT"] = "12355"
    os.environ["RANK"] = str(rank)
    os.environ["LOCAL_RANK"] = str(rank)
```

speechbrain/utils/checkpoints.py
```diff
@@ -559,16 +566,28 @@ def save_checkpoint(
...
+                ckpt_dir / METAFNAME, meta, end_of_epoch
+            )
+
+        # Communicate ckpt_dir to all procs
+        communication_list = [ckpt_dir]
+        if torch.distributed.is_initialized():
+            torch.distributed.broadcast_object_list(communication_list, src=0)
+        ckpt_dir = communication_list[0]
+
+        print(ckpt_dir)
```
pplantinga, line 589:
```suggestion
```

pplantinga:
Is it normal for this test to hang sometimes? `test_categorical_encoder.py`

mravanelli:
I re-ran the tests, and unfortunately, the same issue persists. I'm uncertain about the cause (I can attempt to reproduce it on my local environment as well). It might be worthwhile to investigate whether this issue exists in the latest development branch as well. There's a possibility that the problem comes from updates in certain dependencies.

On another note, I'd like to discuss the possibility of upgrading the Python versions we test against. Currently, we're testing on relatively outdated Python versions. Could we consider expanding our testing to include Python 3.9, 3.10, and 3.11? This would help ensure compatibility with more modern Python environments.

mravanelli:
I tried with a local environment with python 3.8 and it looks like the tests are passing. I'm not sure why that unittest hanging here. My suggestion is to change the CI such that we test with python 3.9-3.11 (I don't remember what we need to change). Then, we can merge this PR.

mravanelli:
ok, that's annoying. It looks like now everything gets stuck with python 3.10. This should be something related to this PR as the tests on dev branch are passing. Any idea?

pplantinga:
Wondering if it is related to this: https://github.com/pytorch/pytorch/issues/75097

If so, I need to investigate if its okay for us to just leave off this part.

mravanelli:
Interesting. That looks related. Curiously, on my local computer I'm able to run all the tests successfully.

tests/unittests/test_checkpoints.py
```diff
@@ -385,3 +381,46 @@ def test_torch_defaults(tmpdir, device):
...
+    os.environ["RANK"] = str(rank)
+    os.environ["LOCAL_RANK"] = str(rank)
+
+    # initialize the process group
+    torch.distributed.init_process_group(
+        "gloo", rank=rank, world_size=world_size
+    )
```
pplantinga, line 398:
```suggestion
    os.environ["RANK"] = str(rank)
    os.environ["LOCAL_RANK"] = str(rank)

    # initialize the process group
    sync_file = f"file://{tmpdir}/sync"
    torch.distributed.init_process_group(
        "gloo", rank=rank, world_size=world_size, init_method=sync_file
    )
```

tests/unittests/test_checkpoints.py
```diff
@@ -385,3 +381,45 @@ def test_torch_defaults(tmpdir, device):
...
+
+    model = torch.nn.Linear(10, 10, device="cpu")
+    checkpointer = Checkpointer(tmpdir, recoverables={"model": model})
+    ckpt = checkpointer.save_checkpoint()
+
+    if rank != 0:
+        assert ckpt is None
+    if rank == 0:
+        # Check that only a single checkpoint is saved, even in ddp
+        assert len(os.listdir(tmpdir)) == 1
```
pplantinga, line 407:
```suggestion
        # Check that only a single checkpoint is saved, even in ddp
        # Second file is the DDP synchronization file
        assert len(os.listdir(tmpdir)) == 2
```

pplantinga:
> Interesting. That looks related. Curiously, on my local computer I'm able to run all the tests successfully.

I think this is due to all 3 versions sharing the same port. I've edited the DDP to use file-based synchronization instead, which should avoid resource contention.

tests/unittests/test_checkpoints.py
```diff
@@ -385,3 +381,46 @@ def test_torch_defaults(tmpdir, device):
...
+        # Second file is the DDP synchronization file
+        assert len(os.listdir(tmpdir)) == 2
+
+        # Check that the model is saved
+        inp = torch.randn((3, 10), device="cpu")
+        prev_output = model(inp)
+        checkpointer.load_checkpoint(ckpt)
+        assert torch.allclose(model(inp), prev_output)
+
+    torch.distributed.destroy_process_group()
```
pplantinga, line 416:
```suggestion
```

pplantinga:
Tests are passing, so I guess this is good to go. The error did indeed seem to be related to `destroy_process_group()` which might be necessary for some cases but doesn't seem necessary here when the machine is shutting down after tests are done.

mravanelli (approved):

mravanelli:
Thank you @pplantinga. It looks like now all the tests pass. I think we can merge it!

PR #2134, opened 2023-08-15
Initialize external loggers only on main process

pplantinga:
In PR #2059 the valid stage end is run on all processes, which required changes in loggers and checkpointers. There are some remaining areas where loggers ought to run only on one process, and this is an attempt to clean this up.


Adel-Moumen (approved):
LGTM! Thanks Peter. :)

PR #2138, opened 2023-08-17
Update python versions tested in CI

pplantinga:
Simple change to CI tests: update Python versions from 3.7, 3.8, 3.9 to 3.9, 3.10, 3.11

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -14,7 +14,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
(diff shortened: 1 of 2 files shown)


pplantinga:
I'm wondering if at some point we want to update other CI dependencies -- black, yamllint, etc. though this may require a much bigger PR for the necessary changes.

mravanelli:
it looks like all the tests are passing (which means that our toolkit is already compatible with more recent version of python). I think this is safe to merge now...

mravanelli (approved):

Issue #1970, opened 2023-05-04
[Feature Request]: Streaming ASR (WIP)


AsuMagic:
### 🚀 The feature

Streaming ASR being a goal for SpeechBrain has already been discussed internally among the core team members, but it seems worthwhile to share the info/personal notes/WIP I have and to keep this publicly updated. I'll try to semi-regularly update the issue according to how things evolve, and it will be closed once a working implementation is merged.

### Solution outline

The current plan is to base off work on #1782, first meaning changes to certain modules have to be done to enable streaming and having a working recipe. Implementing the actual interfaces for streaming ASR can come after once there is a proven to be working model. Certain features (like beam search decoding or LM rescoring) are non-essential for a proof-of-concept.

No implementation was started yet but I'm hoping to start work on that soon.

tl;dr of my own notes with all the technical details I could figure out yet, which is still a WIP:

- investigating the WIP conformer transducer LS PR as a reference: #1782
- most (all?) Conformer-based models use causal convolutions for streamability in toolkits
    - this already seems to be the case in SB, but need to double check
- in SB, the first applied convolution (ConvolutionFrontEnd) does not seem to be causal, but should be made so to be for streaming
- NeMo, ESPnet seem to use limited right context for streaming attention (to double check)
- k2/icefall appears to use https://arxiv.org/pdf/2012.05481.pdf aka a form of "chunked attention"
    - during training, at a given batch, a random chunk size is emulated by aligning visible future context for an output token to the chunk size (the paper explains it better)
    - because the model is trained with multiple chunk sizes, it can be adjusted for a latency-performance tradeoff at runtime
- pretrained RNNLM should be streamable
- at first, because it is not essential for testing/training, let's do greedy search instead of beam search
    - could include LM rescoring early enough, still. RNNLM should be streamable with little to no change
    - beam search would probably need some bigger changes (should eventually check what )
- random design thoughts:
    - streaming is footgun galore
        - we could add some tooling to detect unwanted dependencies and test masking to check if a model is "streaming-proof", which should not need training a whole model
        - ideally, it should be difficult to incorrectly train a streaming model (e.g. forgetting about enabling causal convolutions)
- remaining questions:
    - need to check if any toolkit can just stream infinitely (can they?), i.e.
        - deal with positional encodings
            - might be what the relative positional embeddings do in the first place?
        - cutting off the transformer history
            - otherwise the perf/memory cost grows ~linearly
            - unless the solution is merely to use endpointing

[...]

PR #2140, opened 2023-08-21
Streamable Conformer-Transducer ASR model for LibriSpeech

asumagic:
# Contribution in a nutshell
This PR implements a streaming ASR model for SpeechBrain.

Resolves: #1970 

This PR will (probably) not be implementing the high-level interfaces (think `interfaces.py`) to perform inference ~~as my work ends at the end of August, but I'm hoping to get everything in a functional state with a streaming inference proof-of-concept done.~~ The high level interfaces will be implemented separately.

# Scope
* [x] Implement the Dynamic Chunk Training method to mask transformer keys according to chunks of time, and vary chunk size across batches.
* [x] Implement AWS' [Dynamic Chunk Convolution](https://www.amazon.science/publications/dynamic-chunk-convolution-for-unified-streaming-and-non-streaming-conformer-asr) technique for the Conformer.
* [x] Allow training a model with fully streaming compliant constraints and emulate streaming at inference through masking.
* [x] Provide a tool to empirically analyze dependencies between output tokens and input tokens, in order to validate whether the intended time constraints are respected.
* [x] Write a LibriSpeech Conformer-Transducer recipe.
* [x] Provide a fully trained streaming model.
* [x] Implementing a live microphone demo (using PyAudio)

Additionally, still TODO for me:
* [x] Upload the model somewhere: https://drive.google.com/drive/folders/1vL0Wr15eQZWS8SHp6RJQ_acj__7iri6J?usp=sharing to be transferred to Dropbox
* [x] Update README
* [x] Solve essential TODOs
* [x] Solve pre-commit issues
* [x] Change the experiment seed
* [x] Document new things heavily
* [x] Ensure the code isn't breaking other things
* [x] Probably figuring out what to do with `custom_tgt_module`, again -- probably needs undoing and patching the model same as the other conformer-transducer
* [x] Implement a fully functional proof-of-concept of chunked streaming (WIP)
* [x] Implementing transducer greedy search

**Out-of-scope** for this PR (will be done separately):
* Implementing high level interfaces (for HF demos etc.)

~~See thread, there is an accuracy issue for streaming decoding ATM.~~ Solved

## Explanations and technical details

This should go quite in a bit of detail for the big picture. It could probably eventually be turned into a general documentation on streaming once SB's support matures out.

https://gist.github.com/asumagic/aaf875929d1c03a96fd08a897599aaa8

## Modifying the existing recipe?

Streaming has required very little hparams changes so far. I figure that considering the only current difference between a non-streaming model and a streaming model is changing one hyperparameter, it made more sense to keep it at one file for now.

Additionally, as per the above technical doc, the model is also trained with a context extremely close to non-streaming for ~40% of the batches. The result is that the streaming model can operate in a non-streaming fashion, and for LibriSpeech, the non-streaming performance is very close between a `streaming: False` and a `streaming: True` model, at least in my testing.

## Self-attention logic changes

The `attention.py` change which moves the scaling before in order to reduce the likelihood of float16 overflow should be reviewed thoroughly as it is extensively used. Namely, even though it looks mathematically correct, the `rel_shift` trick scares me a bit and someone with a better understanding than me on the subject should double-check it.

Performing the softmax in fp32 space did not seem to affect performance for me and it seems more correct.

Ultimately I trained the model using bf16. I am not sure if training in fp16 is fully stable yet, but I know that these changes reduced the amount of training failures significantly in fp16 mode.

# Pre-review

* [x] N/A ~~(if applicable) add an `extra_requirements.txt` file~~
* [x] N/A ~~(if applicable) add database preparation scripts & use symlinks for nested folders (to the level of task READMEs)~~
* [x] (if applicable) add a recipe test entry in the depending CSV file under: tests/recipes
* [x] (if applicable) run a recipe test for each yaml/your recipe dataset
* [x] (optional) run `tests/.run-doctests.sh` & `tests/.run-unittests.sh`
* [x] exhausted patience before clicking « Ready for review » in the merge box 🍄


asumagic:
Note: The consistency checks are failing because of code from `develop`, not due to changes introduced by this PR.

asumagic:
I probably won't be able to spend much more time on this since my work ends today, and I'll share my notebook with the WIP streaming code today. The PR and code is generally cleaned up.

There is however an issue... The WER% when using the streaming code is bad (when it is fairly decent when emulating streaming), so something is broken somewhere, and I am not sure what. I edited the technical document with a section on what I have attempted so far to debug the issue, and things to try.

So unless my priorities change, someone will have to pick up work from this. Hopefully a new set of eyes would spot the issues that I couldn't find in time.  
With this issue solved (and I don't know how much effort that would be), there technically isn't a lot remaining to have streaming working, though: some cleanups and proper generic interfaces that could be reused for other models, and stuff to actually push it to HF.

I can still answer questions, clarify things and provide some help/guidance if needed of course.

So will all that said, I will switch the PR to "ready to review" once I upload the notebook somewhere, but it is up to you how to figure out what's the next best course of action.

asumagic:
[Archive containing the streaming test jupyter notebook](https://github.com/speechbrain/speechbrain/files/12486184/streaming-tests.zip). Should be relatively straightforward to run (checkout this branch in a new environment, edit the paths, get the model directory from gdrive and place it in the proper results subdirectory for the conformer-transducer recipe).

TParcollet:
Hey @asumagic please ping me when it's ready to review again :)

asumagic:
Trying to fix fp16 training NaN loss issues so that I can retrain a model after fixing a mismatch in the new convolution code, and rebasing to be up to date with the core changes.

asumagic:
> Trying to fix fp16 training NaN loss issues so that I can retrain a model after fixing a mismatch in the new convolution code, and rebasing to be up to date with the core changes.

Status update:

- I gave up on trying to fix fp16 for now, I'm training using bf16 again on 4x A100 (on narval).
- I had to merge #2271 and #2267 (in a currently not pushed version) to fix issues related to numba bf16 handling and checkpointing.

Not sure when the training will be done especially as narval will go under maintenance tomorrow, but at least I finally got the ball rolling there.

TParcollet:
Did you try removing the loss computation from the mixed training + the last softmax after the joiner network? 

One thing to check as well, is the RNNT loss a sum or an average? If it's a sum, dynamic batching might be slightly less stable.

asumagic:
> Did you try removing the loss computation from the mixed training + the last softmax after the joiner network?
> 
> One thing to check as well, is the RNNT loss a sum or an average? If it's a sum, dynamic batching might be slightly less stable.

I think I did try forcing the loss computation to fp32 to no avail. I didn't try touching the joiner network/softmax though. For now I'm tempted to keep training with bf16 as I started one (.. once compute canada goes out of maintenance) and to try to investigate the fp16 problem after that so that the PR is ready ASAP.

The RNN-T loss uses a sum here.

TParcollet:
The reason could be the sum here then. Could you please try to initialise the grad scaler by a much smaller value? Like 10 instead of 65536? Or even 2.0?

asumagic:
cc @TParcollet @Adel-Moumen: nevermind, a mean is actually used in the transducer loss itself, I was looking at the wrong thing (the joiner network).

I've tried reducing the initial gradscaler scale very significantly, but it did not help. Printing its state_dict to monitor what's going on shows that the scale is vanishing, so it seems to be endlessly trying to reduce scale.

I'll be retrying to force different parts to float32 to diagnose the issue, since it's possible I had done it wrong the last time I tried. I *think* disabling autocast with `enabled=False` might not actually suffice as the inputs may need to be converted back to float explicitly, see: https://pytorch.org/docs/stable/amp.html#autocasting

Looking into this now.

TParcollet:
@asumagic right ... what about the rest of the recipe, is it ready?

asumagic:
> @asumagic right ... what about the rest of the recipe, is it ready? If so, I'll review one last time and try it.

The recipe itself should™ be ready, other than that fp16 is still not fixed. The new training (with bf16) should end tomorrow or on Monday. Once done, I will test accuracy and everything once again.

Notes and reminders:
- This PR depends on #2268 for training.
- In `TransformerASR`, I still have to launch training with the `custom_tgt_module` line commented out. Because that field is not used when only encoding, DDP raises an error during backpropagation. To work around it, when the model is done training, I manually add it back to the model to avoid missing key errors. This issue was mentioned back with the other Conformer PR but we postponed a fix.
- The interfaces to conveniently do chunked streaming are not currently part of the PR. This PR only includes the recipe changes and the model changes. I'll update and reupload my test notebook when it's done.

The high-level interfaces shouldn't take too long to implement (comparatively) but I kind of wanted to make sure everything works before trying to design something that may also end up being reused for other streaming models (or for adapting non-streaming models to streaming contexts like whispercpp etc. do IIRC).

TParcollet:
wouldn't --find_unused_parameters solve the issue without commenting out the code? Can't we just have a check on the number of layers for the tgt module?

asumagic:
> wouldn't --find_unused_parameters solve the issue without commenting out the code?

No, all this does is to print which parameters were unused, but you still get an error. If you don't specify it, it does not tell you which parameter was the problem.

(I imagine the default just checks a counter, whereas this option also tracks and sends the list of parameter names over DDP, which is probably why it's not default).

asumagic:
One final rebase against `unstable-v0.6` to pull in the DDP fixes..

asumagic:
Which recipes are you referring to?

Currently, in this conformer-transducer recipe, the whole `TransformerASR` is used along with an `EncoderWrapper` specified in the hparams. This recipe is the only user of that specific `EncoderWrapper` class, but maybe other recipes might be using `TransformerASR` in an encoding-only scenario too.

The thing is that `TransformerASR` does abstract some useful stuff but that it currently has this problem when only being used for encoding.

To be fair we can probably change this if we swap out the whole class in the hparams without breaking existing models, as those pretrained models would still refer to the "old" class in the inference hparams... It would need some manual patching of the model state dict I'm training but that's feasible.

asumagic:
Side note, I was a bit worried about whether the `* scale` change in the RelPosMHAXL was correct. Ideally, someone should still sanity check it because of the shifting trick I don't fully understand, but from the WER alone on the bf16-trained model, it seems to be within margin of error (or maybe even slightly better at fp16).

```
Full;Full;3.201080340839927   -- autocast fp16
Full;Full;3.210590383444918   -- autocast fp16, without PR scale change
Full;Full;3.202982349360925   -- autocast bf16
Full;Full;3.2048843578819235  -- fp32
Full;Full;3.2048843578819235  -- fp32, without PR scale change
```

asumagic:
[Updated archive containing a notebook with inference streaming logic, WER evaluation code, demos](https://github.com/speechbrain/speechbrain/files/13646995/streaming-tests.zip).

Will reupload the model now (the bf16 one) and update the README with results.

The code successfully transcribes (non-spontaneous) 15 minute speech.

recipes/LibriSpeech/ASR/transducer/train.py
```diff
@@ -69,10 +68,62 @@ def compute_forward(self, batch, stage):
...
+            # TODO: while this is fairly small logic, it may make sense to
+            # extract it to its own class orchestrating dynamic chunk training,
+            # partly because explantions are beneficial
+            if stage == sb.Stage.TRAIN:
+                # When training for streaming, for each batch, we have a
+                # `dynamic_chunk_prob` probability of sampling a chunk size
+                # between `dynamic_chunk_min` and `_max`, otherwise output
+                # frames can see anywhere in the future.
+                # NOTE: We use torch random to be bound to the experiment seed.
+                if torch.rand((1,)).item() < self.hparams.dynamic_chunk_prob:
```
TParcollet, line 88: Couldn't we put this on the audio_pipeline function to generate these masks on CPU and benefit from parallel workers?

asumagic: In theory, yes, in practice, it's currently only <5% of the runtime in my testing without even computing the loss or the backward, so idk if it is worth the micro-optimization (as it won't affect streaming where we don't need to mask).

TParcollet:
Ok just, ping me when the new Transformer Class is integrated and I'll review / test in depth!

Thanks.

asumagic:
cc @Adel-Moumen could you reupload this model to Dropbox? https://drive.google.com/drive/folders/1QtQz1Bkd_QPYnf3CyxhJ57ovbSZC2EhN?usp=sharing

Thanks!

asumagic:
Ok, I think it's finally ready cc @TParcollet. After the refactor I get the same WER% as before in both masking and streaming so nothing should be broken. Training also seems to still run fine.

FWIW, there's a small WER% difference between masking and the streaming code, but in favor of the streaming code (e.g. 3.07% vs 3.18%).  
I had checked for any discrepancy given the same features (post the downsampling CNN) and the tensors were basically equal, so I think it has to do with the how the streaming code path inserts padding and has slight differences due to features getting chunked, so I believe it doesn't indicate a problem.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -414,9 +417,9 @@ def forward(
             The mask for the src keys per batch.
         pos_embs: torch.Tensor, torch.nn.Module, optional
             Module or tensor containing the input sequence positional embeddings
-        chunk_size: int
-            Whether to preform convolution chunking to hide future context,
-            useful for chunked conformers in a dynamic chunk training setting
+        dct_config: Optional[DCTConfig]
```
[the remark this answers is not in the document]
asumagic: What do you mean exactly?

When *actually* streaming, `dct_config` is `None` here (as the streaming function is used instead). This function is specifically the masking-capable code path used during training.

At this specific point in the code, the input shape is identical whether you're in a DCT batch or not (because in the transformer, DCT merely causes masking to prevent attention towards the future).  
When DCT is active for the current batch, this will split up the input in order to apply the DCConv logic. At the end, everything gets concatenated back.

So when training, during a DCT batch, the input and the output of a conformer layer have a strictly identical shape to a non-DCT batch, simply with additional constraints on which frames could attend to which frames.

recipes/LibriSpeech/ASR/transducer/README.md
```diff
@@ -25,14 +25,46 @@ If your GPU effectively supports fp16 (half-precision) computations, it is recom
...
+|:-------------:|:---------------------------:|:-:| :------:| :-----------:| :------------------:| :------------------:| :------------------:| :--------:| :-----------:|
+| 2023-12-12 | conformer_transducer.yaml `streaming: True` | bf16 | 2.56% | 2.72% | 6.47% | \* | \* | https://drive.google.com/drive/folders/1QtQz1Bkd_QPYnf3CyxhJ57ovbSZC2EhN?usp=sharing | [4x A100SXM4 40GB](https://docs.alliancecan.ca/wiki/Narval/en) |
+
+<sub>\*: not evaluated due to performance issues, see [issue #2301](https://github.com/speechbrain/speechbrain/issues/2301)</sub>
+
+## Streaming model
+
+### WER vs chunk size & left context
+
+**Note:** High-level streaming inference code is not currently available.
```
TParcollet, line 43: Remove.

TParcollet, line 43: Remove.

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -63,14 +63,33 @@ precision: fp32 # bf16, fp16 or fp32
...
-avg_checkpoints: 1 # Number of checkpoints to average for evaluation
+avg_checkpoints: 10 # Number of checkpoints to average for evaluation
 
 # Feature parameters
 sample_rate: 16000
 n_fft: 512
 n_mels: 80
 win_length: 32
 
+# Streaming & dynamic chunk training options
```
TParcollet, line 74: Any "DCT" acronym should be replaced by DynChunkTrain or something similar. The reason is that DCT means "Discrete Cosine Transform".

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -63,14 +63,33 @@ precision: fp32 # bf16, fp16 or fp32
...
+# At least for the current architecture on LibriSpeech, we found out that
+# non-streaming accuracy is very similar between `streaming: True` and
+# `streaming: False`.
+streaming: True  # controls all DCT & chunk size & left context mechanisms
+
+dct_config_sampler: !new:speechbrain.lobes.models.transformer.DCT.DCTConfigRandomSampler # yamllint disable-line rule:line-length
+   dct_prob: 0.6
+
```
TParcollet, line 82: remove the line space and a few in-line comments on what these params are.

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -63,14 +63,33 @@ precision: fp32 # bf16, fp16 or fp32
...
+   chunk_size_min: 8
+   chunk_size_max: 32
+
+   limited_left_context_prob: 0.75
+   left_context_chunks_min: 2
+   left_context_chunks_max: 32
+
+   # valid_config
```
TParcollet, line 90: remove?

recipes/LibriSpeech/ASR/transducer/train.py
```diff
@@ -69,10 +68,20 @@ def compute_forward(self, batch, stage):
             )
 
         current_epoch = self.hparams.epoch_counter.current
+
+        # Old models may not have the streaming hparam, we don't break them in
+        # any other way so just check for its presence
+        if hasattr(self.hparams, "streaming") and self.hparams.streaming:
+            dct_config = self.hparams.dct_config_sampler(stage)
```
TParcollet, line 75: DCT -> rename.

speechbrain/decoders/transducer.py
```diff
@@ -149,18 +151,43 @@ def transducer_greedy_decode(self, tn_output):
...
+        Tuple of 4 or 5 elements (if `return_hidden`).
+
+        First element: List[List[int]]
+            List of decoded tokens
+
+        Second element: torch.Tensor
             Outputs a logits tensor [B,T,1,Output_Dim]; padding
             has not been removed.
+
+        Third element: None
```
TParcollet, line 175: Then why do we return it?

speechbrain/decoders/transducer.py
```diff
@@ -149,18 +151,43 @@ def transducer_greedy_decode(self, tn_output):
...
+            nbest; irrelevant for greedy decode
+
+        Fourth element: None
```
TParcollet, line 178: Same.

speechbrain/lobes/models/transformer/Branchformer.py
```diff
@@ -335,6 +336,7 @@ def forward(
             If custom pos_embs are given it needs to have the shape (1, 2*S-1, E)
             where S is the sequence length, and E is the embedding dimension.
         """
+        assert dct_config is None, "DCT unsupported for this encoder"
```
TParcollet, line 339: Sad :-(

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -17,11 +20,48 @@
     MultiheadAttention,
     PositionalwiseFeedForward,
 )
+from speechbrain.lobes.models.transformer.DCT import DCTConfig
```
TParcollet, line 23: Naming --- I won't repeat, last mention, but must be broadcasted.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -17,11 +20,48 @@
...
 from speechbrain.lobes.models.transformer.hypermixing import HyperMixing
 from speechbrain.nnet.normalization import LayerNorm
 from speechbrain.nnet.activations import Swish
 
 
+@dataclass
+class ConformerEncoderLayerStreamingContext:
+    """Streaming metadata and state for a `ConformerEncoderLayer`.
+
+    The multi-head attention and Dynamic Chunk Convolution require to save some
```
TParcollet, line 33: Should we add the Dynamic Chunk Convoluation paper as a reference here .

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -99,20 +142,149 @@ def __init__(
             nn.Dropout(dropout),
         )
 
-    def forward(self, x, mask=None):
+    def _do_conv(self, x, inhibit_padding: bool):
```
TParcollet, line 145: I don't understand what this is about.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -17,11 +20,48 @@
     MultiheadAttention,
     PositionalwiseFeedForward,
 )
+from speechbrain.lobes.models.transformer.DCT import DCTConfig
 from speechbrain.lobes.models.transformer.hypermixing import HyperMixing
 from speechbrain.nnet.normalization import LayerNorm
 from speechbrain.nnet.activations import Swish
 
 
+@dataclass
```
TParcollet, line 29: Why do we need this context stuff? It reads like something that will complexity the understanding of the architecture?

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -99,20 +142,149 @@ def __init__(
...
+            out = [self.bottleneck(chk) for chk in out]
+            out = [chk.transpose(1, 2) for chk in out]
+
+            # TODO: experiment around reflect padding, which is difficult
+            # because small chunks have too little time steps to reflect from
+            out = [
+                F.pad(
+                    out[i],
+                    (
+                        # channel dims, we do not to pad these
```
TParcollet, line 226: unclear

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -99,20 +142,149 @@ def __init__(
...
+            out = self._do_conv(out, inhibit_padding=True)
+
+            # -> [batch_size * num_chunks, chunk_size + rpad, out_channels]
+            out = out.transpose(1, 2)
+
+            out = self.after_conv(out)
+
+            # -> [batch_size, num_chunks, chunk_size, out_channels]
+            out = torch.unflatten(out, dim=0, sizes=(batch_size, -1))
+
```
TParcollet, line 265: can't we do both in a single call?

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -263,11 +440,87 @@ def forward(
         )
         x = x + skip
         # convolution module
-        x = x + self.convolution_module(x, conv_mask)
+        x = x + self.convolution_module(x, conv_mask, dct_config=dct_config)
+        # ffn module
+        x = self.norm2(x + 0.5 * self.ffn_module2(x))
+        return x, self_attn
+
+    def forward_streaming(
```
TParcollet, line 448: Maybe a docstring here? What is the difference with the standard forward.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -263,11 +440,87 @@ def forward(
...
+        x = x + skip
+
+        # truncate outputs corresponding to the MHA left context (we only care
+        # about our chunk's outputs); see above to-do
+        x = x[..., -orig_len:, :]
+
+        # TODO: this is slightly suboptimal as this will add left padding inside
+        # the convolution code that we do not need. it would be better to
+        # manually add the right-padding ourselves and disable padding inside
+        # the convolution module for this usecase, but it would need some
```
TParcollet, line 490: Isn't this what you do with inhibit_padding?

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -385,12 +641,55 @@ def forward(
...
                 src_key_padding_mask=src_key_padding_mask,
                 pos_embs=pos_embs,
+                dct_config=dct_config,
             )
             attention_lst.append(attention)
         output = self.norm(output)
 
         return output, attention_lst
 
+    def forward_streaming(
```
TParcollet, line 651: Docstring and difference as well ;-) Is it very necessary for this class to have a custom forward? I see many changes for the EncoderLayer, so it makes sense, but here?

speechbrain/lobes/models/transformer/DCT.py
```diff
@@ -0,0 +1,130 @@
+"""Configuration and utility classes for classes for Dynamic Chunk Training, as
+often used for the training of streaming-capable models in speech recognition.
```
TParcollet, line 2: Reference to paper.

speechbrain/lobes/models/transformer/DCT.py
```diff
@@ -0,0 +1,130 @@
...
+        prob : float
+            Probability (0..1) to return True (False otherwise)."""
+        return torch.rand((1,)).item() < prob
+
+    def __call__(self, stage: "sb.core.Stage") -> DCTConfig:
+        """Samples a random (or not) DCT configuration depending on the current
+        stage.
+
+        Arguments
+        ---------
```
TParcollet, line 97: Maybe an example to this function?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -541,6 +542,8 @@ def forward(
         src_key_padding_mask : tensor
             The mask for the src keys per batch (optional).
         """
+        assert dct_config is None, "DCT unsupported for this encoder"
```
TParcollet, line 545: But wait, then we have it defined in TransformerASR?

speechbrain/lobes/models/transformer/DCT.py
```diff
@@ -0,0 +1,130 @@
+"""Configuration and utility classes for classes for Dynamic Chunk Training, as
```
TParcollet, line 1: Why in transformer? Why not in speechbrain/utils?

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -17,6 +18,143 @@
...
+    """DCT configuration holding chunk size and context size information."""
+
+    encoder_context: Any
+    """Opaque encoder context information. It is constructed by the encoder's
+    `make_streaming_context` method and is passed to the encoder when using
+    `encode_streaming`.
+    """
+
+
+def make_asr_src_mask(
```
TParcollet, line 38: I don't like the ASR here.

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -17,6 +18,143 @@
...
+
+                # end range is exclusive, so there is no off-by-one here
+                src_mask[t, :frame_remaining_context] = True
+
+        return src_mask
+
+    return None
+
+
+def make_asr_masks(
```
TParcollet, line 114: The naming of these two functions is not ideal IMHO. Can we have a name that explicitly says what it masks?

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -152,9 +290,11 @@ def __init__(
             ),
             torch.nn.Dropout(dropout),
         )
-        self.custom_tgt_module = ModuleList(
-            NormalizedEmbedding(d_model, tgt_vocab)
-        )
+
+        if num_decoder_layers > 0:
```
TParcollet, line 294: El Famoso.

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -333,11 +454,95 @@ def encode(self, src, wav_len=None, pad_idx=0):
...
+        src : torch.Tensor
+            The sequence (chunk) to the encoder.
+
+        context : TransformerASRStreamingContext
+            Mutable reference to the streaming context. This holds the state
+            needed to persist across chunk inferences and can be built using
+            `make_streaming_context`. This will get mutated by this function.
+
+        Returns
+        -------
```
TParcollet, line 480: Can we have an example? Or too complicated maybe.

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -83,14 +83,18 @@ test_left_context_size: 64
...
+   dct_prob: 0.6
+
+   chunk_size_min: 8
+   chunk_size_max: 32
+
+   limited_left_context_prob: 0.75
+   left_context_chunks_min: 2
+   left_context_chunks_max: 2
+
+   # valid_config
```
TParcollet, line 90: cleaning.

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -83,14 +83,18 @@ test_left_context_size: 64
...
-dynamic_chunk_min: 8
-dynamic_chunk_max: 32
-
-# Left context gets expressed as a number of chunks
-dynamic_left_context_prob: 0.75
-dynamic_left_context_min: 2
-dynamic_left_context_max: 32
+dct_config_sampler: !new:speechbrain.utils.DCT.DCTConfigRandomSampler # yamllint disable-line rule:line-length
+   dct_prob: 0.6
+
```
TParcollet, line 82: Line space + comments

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -333,11 +454,95 @@ def encode(self, src, wav_len=None, pad_idx=0):
...
+        # to let the encoder get the pos embedding itself and have a way to
+        # cache it.
+        #
+        # additionally, positional encoding functions take in a whole source
+        # tensor just to get its attributes (size, device, type) but this is
+        # sort of silly for the embeddings that don't need one.
+        # so we craft a dummy empty (uninitialized) tensor to help...
+        known_left_context = context.encoder_context.layers[0].mha_left_context
+        if known_left_context is None:
+            # print(f"no lc known: using {src.shape}")
```
TParcollet, line 504: remove print

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -333,11 +454,95 @@ def encode(self, src, wav_len=None, pad_idx=0):
...
+            pos_encoding_dummy = src
+        else:
+            target_shape = list(src.shape)
+            # print(f"computing posemb shape: from {target_shape} with lc {known_left_context.shape}")
```
TParcollet, line 508: remove print

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -99,20 +142,149 @@ def __init__(
...
+                bias=self.conv.bias,
+                stride=self.conv.stride,
+                padding=0,
+                dilation=self.conv.dilation,
+                groups=self.conv.groups,
+            )
+
+        return out
+
+    def forward(self, x, mask=None, dct_config: Optional[DCTConfig] = None):
```
TParcollet, line 167: This function and the whole streaming thing is fairly complex. I propose that you write an integration test following the current one we have for Transducer but only to test streaming.,

TParcollet (changes requested):
Once the changes are done, I'll run it. Great work! Thank you so much!

speechbrain/decoders/transducer.py
asumagic: To match the beam search API, AFAIK. This is not new, though, it was just undocumented before.

speechbrain/lobes/models/transformer/Branchformer.py
asumagic: For now :D

speechbrain/lobes/models/transformer/DCT.py
asumagic: This is what I did before, but there's a nasty circular dependency because `core.py` imports something from `speechbrain.utils`, and `utils/__init__.py` pulls every `.py` file there (aaaa), and `DCT.py` needs `sb.core.Stage`. *Somehow,* no util prior to that broke and depended on the core...

The lazy workaround would be to bring the `Stage` import inside of the function, which is ugly. I guess we really need a proper fix but I'm not sure what.

I thought it'd be fine since this is largely applied to transformers, but on second thought, nothing about the concept is really specific to it...

speechbrain/lobes/models/transformer/Transformer.py
asumagic: The confusing part is that `TransformerEncoder` is specifically "the" classic transformer encoder class, not an interface like `TransformerInterface` is.

`ConformerEncoder` (and the others) have no relationship with `TransformerEncoder`.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: The context objects hold the state that has to be remembered for streaming inference (i.e. mostly left context). I wasn't sure what would be the best approach here, so I ended up settling on this pure data design that should be easy to inspect and extend.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: In regular, non-DynChunkTrain mode, the entire input enters this function, and the convolution is applied with zero-padding on the sides (automatically, through the parameters of the convolution).

In DynChunkTrain mode, a tensor with all stacked chunks enters this function. Conceptually, for each chunk, we want the left "padding" to be the injected left context (of `(chunk_size - 1) // 2` frames) for this chunk, and the right padding to be 0s.

We would have two ways of doing this:
- simply inject the left context, do the convolution, then truncate it out, which means the convolution gets computed for all the frames in the left context even if we're throwing them away immediately after
- or disable the built-in padding from the convolution (and inject zeros for the right padding manually).

The DynChunkTrain code does the latter.

I will add explanations on this as comments.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: I'm not sure but IMO it is clearer this way (since we're "undoing" individual shape transforms) and should have no performance implication since this is only manipulating views and not doing concrete transforms in memory.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: Yes it is, not sure how I forgot this comment.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: I guess the question in general is whether the streaming forward functions should be named `forward_streaming` (and that the streaming behavior would be triggered by having the context object non-`None`).

Part of the problem is that it needs to call into layers' `forward_streaming` and not merely the forward as well, and that, down the road, other encoders may need to do something more involved that would justify a separate forward.

TBH, I think it's better this way for now.

TParcollet:
@asumagic also, it looks like when streaming = True, the training speed slowdown seriously - on 8 GPUs it goes from 10min per epoch to 15-20. Is this expected? I guess it's due to the processing per chunk with all the for loops? Should we try to @jit this?

asumagic:
I'll try to profile it. It's possible that some of the list comprehensions inside of the convolution module are what harm performance, especially with longer utterances and smaller chunk sizes.  
I might have underestimated their effect on performance, but I think most of that could be mitigated if it's that.

Should this be solved as part of this PR or another? I imagine the change would be not too extensive, but might take a bit to properly implement and test.

asumagic:
On a side note, I did try poking with `torch.compile` a few days ago and impressively enough, it succeeds compiling the entire `compute_forward`. However, the dynamic shape support is still too lacking as compilation crashes hard with it enabled, and the BS fails to compile with it (but not the GS). So right now it's a dead end, but it's promising.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: Made it refer to `ConvolutionModule` which now cites the paper.

speechbrain/lobes/models/transformer/TransformerASR.py
asumagic: Added a docstring example

asumagic:
Yet again updated the notebook due to the renaming refactor: 
[streaming-tests.zip](https://github.com/speechbrain/speechbrain/files/13705355/streaming-tests.zip)

TParcollet (approved):
LGTM. I even obtained 2.58 on test-clean with the pre-last-change model. I tried to train again. it works

PR #2150, opened 2023-09-01
Vits [Work in progress]

bloodraven66:
Hi,

Adding Implementation for VITS TTS. 
The plan is to build it for ljspeech, then extend it for multispeaker, multilingual, vc etc

TODO:
- [ ] complete mas aligner
- [ ] add masks
- [ ] add all args + config
- [ ] add losses

### DIFF
--- a/recipes/LJSpeech/TTS/VITS/hparams/train.yaml
+++ b/recipes/LJSpeech/TTS/VITS/hparams/train.yaml
@@ -1,0 +1,182 @@
+############################################################################
+# Model: FastSpeech2
+# Tokens: Raw characters (English text)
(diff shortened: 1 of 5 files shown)


csukuangfj:
I find that onnx export support is not listed in the TODOs.
Do you plan to add it?

bloodraven66:
> I find that onnx export support is not listed in the TODOs. Do you plan to add it?

I will be adding torchscript, do not plan on onnx as of now

csukuangfj:
> * [x]  complete mas aligner
> * [x]  add masks
> * [x]  add all args + config
> *

I see. Thanks for the quick reply.

Issue #1937, opened 2023-04-12
[Docs]: `torch.distributed.launch` is deprecated, `torchrun` should be recommended


asumagic:
### Describe the bug

Many places in our documentation recommends to use `torch.distributed.launch`.

However, that option is being deprecated by upstream PyTorch: <https://pytorch.org/docs/stable/distributed.html#launch-utility>

The recommendation is to use `torchrun`, which is supposed to be functionally equivalent: <https://pytorch.org/docs/stable/elastic/run.html#launcher-api>

### Expected behaviour

SpeechBrain should recommend using `torchrun` in all the documentation - but we should check which version introduced it, and whether we still support versions that might not have it.  
Plus, we should check whether we might encounter any issues with that migration.

Quite a lot of the documentation in SB still recommends using the deprecated option, including the multigpu docs and some recipe READMEs.

### To Reproduce

Run any SB training job with DDP enabled (as documented in multigpu docs)

### Versions

At the time of writing, the `develop` tip `e353e95` is affected.

### Relevant log output

```shell
FutureWarning: The module torch.distributed.launch is deprecated
and will be removed in future. Use torchrun.
Note that --use_env is set by default in torchrun.
If your script expects `--local_rank` argument to be set, please
change it to read from `os.environ['LOCAL_RANK']` instead. See 
https://pytorch.org/docs/stable/distributed.html#launch-utility for 
further instructions
```

Adel-Moumen:
Solved in #2158. Thanks @asumagic :)

PR #2158, opened 2023-09-10
Use torchrun instead of torch.distributed.launch

pplantinga:
`torch.distributed.launch` is deprecated, `torchrun` is recommended. This simple change updates to make speechbrain compatible with `torchrun`.

In addition, this PR removes the `--distributed_launch` flag by assuming if environment variables `RANK` and `LOCAL_RANK` arguments are set, the user wants to run the experiment in a distributed way.


Adel-Moumen:
Linked issue #1937

Adel-Moumen:
Hey @lucadellalib, could you please have a look? Thanks :)

lucadellalib (changes requested):
Looks good, we should also update the documentation in `docs/multigpu.md`. On a minor note, do you think it's possible to remove the command line flag `distributed_launch` to avoid redundancy?

pplantinga:
I suppose we could assume that if `RANK` and `LOCAL_RANK` env variables are set, that the program was intended to be run in distributed fashion.

pplantinga:
Okay @lucadellalib this is ready for review again.

lucadellalib (approved):
Looks good, I could kick off a distributed training on both a single node and multiple nodes without issues.

PR #2159, opened 2023-09-12
[Feat] add save_filename into Pretrained interface from_hparams class…

choiHkk:
Based on version 0.5.15, when we try to use a pre-downloaded checkpoint without downloading it from the HUB, 
the download codes is executed that accesses the HUB, which can be limiting when used locally.
* codes: 
```python
        # line 149 ~ 154
        try:
            urllib.request.urlretrieve(sourcefile, destination)
        except urllib.error.URLError:
            raise ValueError(
                f"Interpreted {source} as web address, but could not download."
            )

        # line 160 ~ 173
        try:
            fetched_file = huggingface_hub.hf_hub_download(
                repo_id=source,
                filename=filename,
                use_auth_token=use_auth_token,
                revision=revision,
                cache_dir=cache_dir,
            )
            logger.info(f"HF fetch: {fetched_file}")
        except HTTPError as e:
            if "404 Client Error" in str(e):
                raise ValueError("File not found on HF hub")
            else:
                raise
```

I didn't change anything except adding 'save_filename=None' and changing 'save_filename=None' to 'save_filename=save_filename' at 'from_hparams' classmethod.

```python
    @classmethod
    def from_hparams(
        cls,
        source,
        hparams_file="hyperparams.yaml",
        pymodule_file="custom.py",
        overrides={},
        savedir=None,
        use_auth_token=False,
        revision=None,
        download_only=False,
        save_filename=None, 
        **kwargs,
    ):

...

        hparams_local_path = fetch(
            filename=hparams_file,
            source=source,
            savedir=savedir,
            overwrite=False,
            save_filename=save_filename,
            use_auth_token=use_auth_token,
            revision=revision,
        )
        try:
            pymodule_local_path = fetch(
                filename=pymodule_file,
                source=source,
                savedir=savedir,
                overwrite=False,
                save_filename=save_filename,
                use_auth_token=use_auth_token,
                revision=revision,
            )
            sys.path.append(str(pymodule_local_path.parent))
...
```

### DIFF
--- a/speechbrain/pretrained/interfaces.py
+++ b/speechbrain/pretrained/interfaces.py
@@ -406,6 +406,7 @@ class Pretrained(torch.nn.Module):
         use_auth_token=False,
         revision=None,
         download_only=False,
+        save_filename=None, 
         **kwargs,
     ):
         """Fetch and load based from outside source based on HyperPyYAML file
@@ -464,7 +465,7 @@ class Pretrained(torch.nn.Module):
             source=source,
             savedir=savedir,
             overwrite=False,
-            save_filename=None,
+            save_filename=save_filename,
             use_auth_token=use_auth_token,
             revision=revision,
         )
@@ -474,7 +475,7 @@ class Pretrained(torch.nn.Module):
                 source=source,
                 savedir=savedir,
                 overwrite=False,
-                save_filename=None,
+                save_filename=save_filename,
                 use_auth_token=use_auth_token,
                 revision=revision,
             )


mravanelli:
Thank you @choiHkk! @Gastron, what is your opinion on this?

Gastron (approved):
This makes sense, thanks! Please just run the pre-commit hooks or otherwise fix the linter errors.

mravanelli:
Thank you @Gastron, are you sure that is a convenient solution? If I am not missing anything, it looks like we now use the same save_filename variable for two files that are different (hparams_file  and pymodule_file).

PR #2170, opened 2023-09-26
Fix dynamic batching

Adel-Moumen:
This PR aims at fixing dynamic batching in SpeechBrain. Indeed, as explained in #1881, some parameters are not actually passed in the `DynamicBatchSampler` object.  

This PR is also a clean up of #1891.

### DIFF
--- a/speechbrain/dataio/sampler.py
+++ b/speechbrain/dataio/sampler.py
@@ -485,7 +485,13 @@ class DynamicBatchSampler(Sampler):
         self._max_batch_ex = max_batch_ex
         # Calculate bucket lengths - how often does one bucket boundary fit into max_batch_length?
         self._bucket_lens = [
-            max(1, int(max_batch_length / self._bucket_boundaries[i]))
+            min(
+                self._max_batch_ex,  # tops max_duration_per_batch
+                max(
+                    1,  # and at least 1
+                    int(self._max_batch_length / self._bucket_boundaries[i]),
+                ),
+            )
             for i in range(len(self._bucket_boundaries))
         ] + [1]
         self._epoch = epoch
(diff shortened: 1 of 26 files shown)


Adel-Moumen:
What should we do about:
``` 
Furthermore, "length" as a variable name seems inconclusive as there were two args (length as duration & length as number of examples). As a result, the max_batch_len can be confused with max_batch_ex. The former term has been renamed here to make clear, it is about the duration. This change breaks the interface of the dynamic batch sampler, which is relevant only during training, for only few recipes (five of them have dynamic batching on). If that's an issue -> we can put this PR also on the unstable-v0.6 branch.
```  

?

Would you like to change the name? CC: @mravanelli @TParcollet

speechbrain/dataio/sampler.py
```diff
@@ -485,7 +486,13 @@ def __init__(
         self._max_batch_ex = max_batch_ex
         # Calculate bucket lengths - how often does one bucket boundary fit into max_batch_length?
         self._bucket_lens = [
-            max(1, int(max_batch_length / self._bucket_boundaries[i]))
+            min(
```
TParcollet, line 489: This change is good to me.

speechbrain/dataio/sampler.py
```diff
@@ -578,6 +585,15 @@ def _generate_batches(self):
             item_len = self._ex_lengths[str(idx)]
             # bucket to fill up most padding
             bucket_id = np.searchsorted(self._bucket_boundaries, item_len)
+
+            if (
+                len(bucket_batches[bucket_id])
```
TParcollet, line 590: This change is a bit more mysterious to me

Adel-Moumen: This is this code: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/dataio/sampler.py#L595  that has been moved here following Andreas PR.

PR #2176, opened 2023-09-27
Add Cspell

calvinballing:
Successor PR to: https://github.com/speechbrain/speechbrain/pull/1353

The first four commits are the same (except for addressing merge conflicts from the past 1.5 years), and the last 5 address new typos introduced since then.

I separated out this change into its own commit so that someone more familiar with it can confirm it is correct:
`lr_whisperc -> lr_whisper`
________

Possible future work, if cspell is added: 
1) Adding a GitHub action for Cspell
2) working more on the custom dictionary.  Currently it has 6 sections, the last of which could be sorted into the rest of the dictionary.:
 - British (British English spellings of words)


Adel-Moumen:
Thanks @calvinballing ! 

We will have a closer look to his PR after having merged everything. This is great that you are still active. Sorry that it took so long but now we are now converging to SpeechBrain 1.0 and in the following days/weeks we should be able to review your PR and add to it to SpeechBrain :)

Best,
Adel

calvinballing:
May want to look into test flake.  I had two test failures, then re-ran and got passes.

mravanelli:
Hi @calvinballing, thank you for this contribution. I think this is the last things to do before the release of speechbrain 1.0. Right now, we are changing so many files, so it will be for sure necessary re-running these fixes right before the big release.

Adel-Moumen:
> May want to look into test flake. I had two test failures, then re-ran and got passes.

Hello, yes, I addressed the issues in a different PR that has been merged.

pplantinga:
Can we re-open this with a target of the develop branch (I guess deleting the unstable branch closed this PR).

pplantinga:
asdf

PR #2177, opened 2023-09-28
[WIP] Fix recipe tests on CPU

mravanelli:
In this PR. I've addressed an issue where certain recipe tests were failing when executed on CPU. The cause of this problem was that automatic mixed precision is designed to work exclusively on CUDA devices. To resolve this, I made the following modifications:

1. core.py: I added an additional check within the core.py file to ensure that automatic mixed precision only runs on CUDA devices, thus preventing the failures on CPU.

2. Recipe Modifications: I also had to make adjustments to all the recipes that were previously overriding the fit_batch function. Ideally, we aim to reduce the need for such overrides, and I believe that after merging pull request #2010, these modifications may no longer be necessary.

@Adel-Moumen, I recommend considering the changes made in this pull request, as they are related to the improvements proposed in pull request #2010.

### DIFF
--- a/recipes/Aishell1Mix/separation/train.py
+++ b/recipes/Aishell1Mix/separation/train.py
@@ -120,7 +120,7 @@ class Separation(sb.Brain):
         if self.hparams.num_spks == 3:
             targets.append(batch.s3_sig)
 
-        if self.auto_mix_prec:
+        if self.auto_mix_prec and "cuda" in self.device:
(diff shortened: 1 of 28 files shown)


mravanelli:
As discussed with @Adel-Moumen, I think we can close this PR and implement the changes to make our recipes CPU-compliant in #2010. This will avoid annoying conflicts.

PR #2183, opened 2023-10-01
Fix multi-head attention when return_attn_weights=False

lucadellalib:


mravanelli (approved):

mravanelli:
Thank you @lucadellalib. I reviewed the code and ran the recipe tests. Everything looks fine. However, the presence of issues like the one that you spotted makes me think that we need to work more on unit tests. In the future, I would like to coordinate a collective effort for improving our unit tests.

PR #2185, opened 2023-10-01
[WIP] Testing Data Preparation in Recipe Tests

mravanelli:
We are currently not testing the data preparation part during recipe tests. To address this limitation, this PR suggests the following approach:

1. Create a tiny dataset for each dataset, containing only a few samples, and upload it to our Dropbox repository. This small dataset should maintain the same structure as the original one.

2. Modify the data preparation scripts to include an additional argument called "--small_dataset." When this argument is activated, the script will download the small dataset and use it for data preparation.

Note:
- I have already implemented an example with the TIMIT dataset. If we agree with this solution, we will need to extend it to all the datasets. While this may require a substantial amount of work, I believe it will be worthwhile in the end.


Adel-Moumen:
I dont fully understand why we are not doing something like the full_inference tests (see: https://github.com/speechbrain/speechbrain/blob/26e60724f337129fc6a9e6074cce4bd7d25c3e06/tests/recipes/full_inference.csv). The idea would be to download a subsample of the dataset, and then run the data preparation scripts to test that everything is correct. I would highly prefer that as it does not introduce new flags, new arguments and is exclusively located in the tests folder and not inside of the recipes. 

What do you think @mravanelli ?

mravanelli:
That's another option. One issue with that approach is that it does not
test the training and dev data manifest file.
Ideally, to make our tests more comprehensive, we should have something
like this:
1- Training test on the sample data. Ideally, we should do an overfitting
test for each recipe or at least train the recipe for a few epochs and
check if the training performance decreases over time (which is a good sign)
2- With the pretrained model available on Dropbox, we should do an
inference test on the sample data.

This requires a lot of work (I regret we did not do that from the
beginning), but I think in the long term we should aim for something like
that. What do you think?

PR #2192, opened 2023-10-05
Broadcast the decision to checkpoint to all processes

pplantinga:
This PR fixes some DDP hangs during checkpointing where some processes try to save and others don't by broadcasting the decision to all processes.


mravanelli:
@lucadellalib, could you please take a look?

Adel-Moumen (approved):
lgtm. 

Thanks @pplantinga! It is indeed solving our issue with checkpointing (`ckpt_interval_minutes`) and multiple GPUs. 

I tried it with the same setup as before (transducer ASR librispeech), and with `ckpt_interval_minutes`/`ckpt_interval_steps` and it now work as intended!

Adel-Moumen:
Hey @pplantinga,

Could you please sync your PR with the latest commits in develop? Thanks. :)

Adel-Moumen:
Thanks @pplantinga for the PR!

lucadellalib:
Looks good! A minor note, if this functionality is disabled we should directly return False without the unnecessary call to `torch.distributed.broadcast_object_list`, which requires synchronizing the processes. For the same reason, we might also check the condition every `grad_accumulation_factor`. I don't think this has an impact on performance, but you never know and I think avoiding unnecessary operations is always good.

PR #2193, opened 2023-10-05
Attempting to fix failing test (with pytorch 2.1)

mravanelli:
In this PR, I'm investigating solutions to fix the test that is failing (it fails on the CI server, on my machine everything runs). Feel free to jump in if you have some insights.


Adel-Moumen (approved):
Sounds good to me. Thanks Mirco for solving the issue....

PR #2199, opened 2023-10-08
Broadcast checkpoint decision

pplantinga:
Slight update to #2192 addressing last remaining comment. Basically we can avoid the sync if mid-epoch checkpoints are disabled, for a potential speedup in cases where communication is slow.


pplantinga:
Let me start from fresh copy of develop

PR #2200, opened 2023-10-08
Avoid sync if mid-epoch checkpoints are disabled

pplantinga:
Follow-up to #2192 in order to address the last comment. Basically, this PR avoids sync-ing when mid-epoch checkpoints are disabled which could theoretically be slow if communicating over a slow network.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1228,6 +1228,10 @@ class Brain:
         if self.checkpointer is None:
             return False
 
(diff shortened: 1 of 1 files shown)


mravanelli:
@lucadellalib are you fine with this?

lucadellalib:
Looks good!

lucadellalib (approved):

PR #2201, opened 2023-10-09
New Pitch Computation in FastSpeech 2

mravanelli:
In torchaudio 2.1, the function we used for pitch computation (`torchaudio.functional.compute_kaldi_pitch`) was deleted. We need to find another one. This PR is an attempt to fix this backward incompatibility issues. Tests are need to make sure the new pitch computation gives comparable performance.


recipes/LJSpeech/TTS/ljspeech_prepare.py
```diff
@@ -48,7 +49,7 @@ def prepare_ljspeech(
     pitch_n_fft=1024,
     pitch_hop_length=256,
     pitch_min_f0=65,
-    pitch_max_f0=2093,
+    pitch_max_f0=400,
```
BenoitWang, line 52: can 400 be a bit aggressive? I'd suggest a little bit higher like 1000

recipes/LJSpeech/TTS/ljspeech_prepare.py
```diff
@@ -461,14 +462,35 @@ def prepare_json(
...
+                        frame_shift=(pitch_hop_length / fs * 1000),
+                        min_f0=pitch_min_f0,
+                        max_f0=pitch_max_f0,
+                    )[0, :, 0]
+                else:
+                    pitch = torchaudio.functional.detect_pitch_frequency(
+                        waveform=audio,
+                        sample_rate=fs,
+                        frame_time=(pitch_hop_length / fs),
+                        win_length=3,
```
BenoitWang, line 480: how did you choose 3 here please?

recipes/LJSpeech/TTS/ljspeech_prepare.py
```diff
@@ -461,14 +462,35 @@ def prepare_json(
...
+                        freq_low=pitch_min_f0,
+                        freq_high=pitch_max_f0,
+                    ).squeeze(0)
+
+                    # Concatenate last element to match duration.
+                    pitch = torch.cat([pitch, pitch[-1].unsqueeze(0)])
+
+                    # Mean and Variance Normalization
```
BenoitWang, line 488: and how did you fix these values please?

recipes/LJSpeech/TTS/ljspeech_prepare.py
```diff
@@ -491,15 +513,37 @@ def prepare_json(
...
+                    mean = 256.1732939688805
+                    std = 328.319759158607
```
BenoitWang, line 541: and how did you choose these values please?

recipes/LJSpeech/TTS/ljspeech_prepare.py
mravanelli: The highest pitch for humans doing normal conversations should be around 400 Hz. Higher values might be just noise. I'm running a full training experiment though and I will thus check if this setting provides good performance.

recipes/LJSpeech/TTS/ljspeech_prepare.py
mravanelli: I just run the data prep script once and computed the mean and std of the pitch computed with this new function (using the training set). We can do that automatically in the data prep script, but this requires adding another loop over the files...

mravanelli:
I think this is the best solution fro now. I did some tests and the two pitch algorithms gives similar results in both standard fastspeech and fastspeech with internal alignments.

PR #2206, opened 2023-10-12
Refactor Augmentation

mravanelli:
I've created this PR to refactor the data augmentation component of our project.

 To better understand the changes, please review the example located at `templates/speech_recognition/ASR/` for an overview of how the updated code will look. 

Here are the key improvements proposed:

1. **Enhanced Configurability**: In the current solution, all data augmentation functions (e.g., addnoise, addrev, dropfreq, speed change) and their respective hyperparameters are defined directly in the YAML configuration. In the previous version, these were nested within a separate layer, which may have made it less intuitive for users. However, this change enhances transparency for users and simplifies the process of integrating their custom augmenters into the augmentation pipeline.

2. **Augmentation Combinator Class**: I've introduced a new class that allows us to mix these augmenters in various ways. The philosophy behind this approach is similar to our separation of the scorer from the beam searcher, where we defined scorers and enabled users to pipeline them into the beam searcher.

3. **Improved Dataset**: In the previous version, the `open_rir dataset `was downloaded and generated behind the scenes. With this solution, the download location is now explicit, giving users greater control.

4. **Enhanced open-rirs Dataset**: I've created a new version of the open-rirs dataset and uploaded it to our Dropbox repository. In this new version, noise sequences and impulse responses have been separated. This change eliminates the need for users to download impulse responses when they only require noise sequences.

5. **Transparent Data Preparation**: In the earlier version, the data preparation for the noisy dataset occurred behind the scenes. With this update, it is performed in `train.py`, aligning it with the data preparation process for other training datasets.

**Note**:
- As a result of these changes,tests are currently failing as I will need to convert all the recipes. Before proceeding with that, I would need your feedback.
- Additional tests and ablations are required, such as comparing FastDropChunk vs. DropChunk, experimenting with larger bandwidth for frequency drop, varying the number of augmentations, and exploring shuffled augmentations.
- The support for speed perturbation in the augmenter, with variable-length augmentation, is still intact.


speechbrain/augment/augmenter.py
```diff
@@ -176,6 +184,9 @@ def forward(self, x, lengths):
             input to augment.
         lengths : torch.Tensor
             The length of each sequence in the batch.
+        total_augmentation : int
+            The total number of augmentations perform for each signal
+            (with the orignal signal included in the count).
```
pplantinga, line 189: It seems like the motivation for this return value is to record how many copies of the data there are when `parallel_augment` is enabled. But this wouldn't change from run to run, would it? I'm wondering if this is better done as a property or method (e.g. `augmenter.get_num_augmentations()` or `augmenter.get_output_replication_count()` or something like this). If you want to keep it here this should be in a `Returns` section rather than `Arguments`.

speechbrain/augment/speech_augment.py
pplantinga, line 1: What's the difference between `signal_augment` and `speech_augment`? Should they all be in one file? Or each in their own file?

templates/speech_recognition/ASR/train.yaml
pplantinga, line 1: This is a definite improvement, looks good Mirco!

pplantinga (commented):
Overall I really like the change, only a few minor design questions to answer here.

speechbrain/augment/preparation.py
```diff
@@ -0,0 +1,208 @@
...
+    URL: str
+        The URL of the dataset to download.
+    dest_folder : str
+        The local folder where the noisy dataset will be downloaded.
+    ext: str
+        File extensions to search for within the downloaded dataset.
+    csv_file : str
+        The path to store the prepared noise CSV file.
+    max_length : float
+        The maximum recirdubg length in seconds.
```
Adel-Moumen, line 35: Recirdubg ?

speechbrain/augment/augmenter.py
```diff
@@ -75,7 +77,7 @@ def __init__(
         max_augmentations=None,
         shuffle_augmentations=False,
         repeat_augment=1,
-        **augmentations,
+        augmentations=list(),
```
Adel-Moumen, line 80: we should add an assert on the size of `augmentations` (e.g. check that the provided list is not empty)

speechbrain/augment/signal_augment.py
```diff
@@ -0,0 +1,526 @@
...
+    >>> from speechbrain.dataio.dataio import read_audio
+    >>> clipper = DoClip(clip_low=0.01, clip_high=0.01)
+    >>> signal = read_audio('tests/samples/single-mic/example1.wav')
+    >>> clipped_signal = clipper(signal.unsqueeze(0))
+    >>> "%.2f" % clipped_signal.max()
+    '0.01'
+    """
+
+    def __init__(self, clip_low=0.5, clip_high=1, clip_prob=1):
+        super().__init__()
```
Adel-Moumen, line 36: missing clip_prob from doctstring

speechbrain/augment/signal_augment.py
```diff
@@ -0,0 +1,526 @@
...
+        The number of segments to combine.
+
+    Example
+    -------
+    >>> signal = torch.ones((4, 256, 22)) * torch.arange(4).reshape((4, 1, 1,))
+    >>> cutcat =  CutCat()
+    >>> output_signal = cutcat(signal)
+    """
+
+    def __init__(self, min_num_segments=2, max_num_segments=10):
```
Adel-Moumen, line 279: missing `max_num_segments` from docstring

speechbrain/augment/signal_augment.py
```diff
@@ -0,0 +1,526 @@
...
+
+    Arguments
+    ---------
+    waveforms : torch.Tensor
+        The original waveform. It is just used to infer the shape.
+    pink_alpha_low : float
+        The minimum value for the alpha spectral smoothing factor.
+    pink_alpha_high : float
+        The maximum value for the alpha spectral smoothing factor.
+    bb_offset_low:
```
Adel-Moumen, line 426: missing docstring for this and `bb_offset_high`

speechbrain/augment/signal_augment.py
```diff
@@ -0,0 +1,526 @@
...
+            torch.rand(waveforms.shape[0], device=waveforms.device) * vrange
+            + low
+        )
+
+        return var.unsqueeze(1)
+
+    alpha = sample_params(pink_alpha_low, pink_alpha_high)
+    bb_offset = sample_params(bb_offset_low, bb_offset_high)
+
+    def emg_spectral_mask(f, device="cpu"):
```
Adel-Moumen, line 451: docstring of args

templates/speech_recognition/ASR/train.py
```diff
@@ -126,7 +129,7 @@ def is_ctc_active(self, stage):
         current_epoch = self.hparams.epoch_counter.current
         return current_epoch <= self.hparams.number_of_ctc_epochs
 
-    def prepare_features(self, stage, wavs):
+    def prepare_features(self, stage, wavs, N_augments=1):
```
Adel-Moumen, line 132: is `N_augments` == `total_augmentations` ?

Adel-Moumen: make sure to add the docstring associated with `N_augments`

Adel-Moumen:
Hey @mravanelli, I really like the new changes. This improvement and the refactoring of the `fit_batch` fn will help a lot!

speechbrain/augment/augmenter.py
mravanelli: Returning this variable is needed to figure out in train.py how many augmentations we apply to the input batch and augment the labels accordingly. For instance, if we apply 3 augmentations in parallel, the label tensor should be augmented by a factor of 3. It can definitely be computed with a specific method, or simply access it with augmenter.num_augmentations.

speechbrain/augment/speech_augment.py
mravanelli: I preferred to split the two as putting them in a single file was too much in terms on number of lines. in signal_augment I put more basic data augmentation strategies that we can used for other sequences (such as EEG).

speechbrain/augment/preparation.py
mravanelli: I will fix

speechbrain/augment/augmenter.py
mravanelli: If the list is empty, no augmentation should be performed. Maybe I can raise a warning when that happen.

speechbrain/augment/signal_augment.py
mravanelli: I will fix

speechbrain/augment/signal_augment.py
mravanelli: I will fix

speechbrain/augment/signal_augment.py
mravanelli: ok

speechbrain/augment/augmenter.py
pplantinga: I wonder if we could simply pass the labels here too. Or have another method like `replicate_labels` that takes care of it.

mravanelli: Yes, we can do that actually!

recipes/CommonVoice/self-supervised-learning/wav2vec2/hparams/wav2vec2_base.yaml
```diff
@@ -58,7 +58,7 @@ batch_size: 12
 test_batch_size: 8
 gradient_accumulation: 8
 num_workers: 4
-
+sorting: ascending
```
Adel-Moumen, line 61: ? where is it being used ?

recipes/LibriParty/VAD/hparams/train.yaml
```diff
@@ -71,54 +80,52 @@ output_neurons: 1
...
+    clean_sample_rate: 16000
+    num_workers: !ref <num_workers>
+
+add_speech_musan: !new:speechbrain.augment.time_domain.AddNoise
+    csv_file: !ref <speech_csv>
+    snr_low: -5
+    snr_high: 15
+    noise_sample_rate: 16000
+    clean_sample_rate: 16000
+    num_workers: !ref <num_workers>
```
Adel-Moumen, line 120: they do not match the previous hparms. -15/-20 instead of -5/15

recipes/TIMIT/ASR/transducer/hparams/train.yaml
```diff
@@ -51,7 +55,7 @@ dnn_neurons: 512
 dec_neurons: 128
 
 # Outputs
-output_neurons: 40
+output_neurons: 42
```
Adel-Moumen, line 58: why 42 ?

recipes/TIMIT/ASR/transducer/hparams/train_wav2vec.yaml
```diff
@@ -46,12 +46,12 @@ sample_rate: 16000
...
 dnn_blocks: 1
-dnn_neurons: 40
+dnn_neurons: 43
 dec_neurons: 128
 
 # Outputs
-output_neurons: 40
-joint_dim: 40
+output_neurons: 43
+joint_dim: 43
```
Adel-Moumen, line 54: why did they changed ?

recipes/UrbanSound8k/SoundClassification/hparams/train_ecapa_tdnn.yaml
```diff
@@ -94,67 +103,87 @@ epoch_counter: !new:speechbrain.utils.epoch_loop.EpochCounter
     limit: !ref <number_of_epochs>
 
 
-augment_wavedrop: !new:speechbrain.lobes.augment.TimeDomainSpecAugment
-    sample_rate: !ref <sample_rate>
-    speeds: [100]
-
-augment_speed: !new:speechbrain.lobes.augment.TimeDomainSpecAugment
-    sample_rate: !ref <sample_rate>
-    speeds: [95, 100, 105]
```
Adel-Moumen, line 103: hmmm have you forgot to put this below ?

recipes/VoxCeleb/SpeakerRec/hparams/train_ecapa_tdnn.yaml
```diff
@@ -80,67 +86,88 @@ epoch_counter: !new:speechbrain.utils.epoch_loop.EpochCounter
     limit: !ref <number_of_epochs>
 
 
-augment_wavedrop: !new:speechbrain.lobes.augment.TimeDomainSpecAugment
-    sample_rate: !ref <sample_rate>
-    speeds: [100]
-
-augment_speed: !new:speechbrain.lobes.augment.TimeDomainSpecAugment
-    sample_rate: !ref <sample_rate>
-    speeds: [95, 100, 105]
```
Adel-Moumen, line 89: same ?

Adel-Moumen:
Please make sure to update the PR with the latest commits in unstable-v0.6 (it includes a new recipe Tedlium2)

speechbrain/alignment/ctc_segmentation.py
```diff
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
-# 2021, Technische Universität München, Ludwig Kürzinger
 """Perform CTC segmentation to align utterances within audio files.
 
 This uses the ctc-segmentation Python package.
 Install it with pip or see the installing instructions in
 https://github.com/lumaku/ctc-segmentation
+
+Author
+-  Ludwig Kürzinger 2021 (Technische Universität München)
```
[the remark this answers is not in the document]
mravanelli: Yes, I think having a linter that checks that  would be good for consistency. More in general, I think we need to add linter checks for all the docstrings.

recipes/AISHELL-1/ASR/CTC/hparams/train_with_wav2vec.yaml
```diff
@@ -104,21 +104,64 @@ split_tokens: !apply:operator.not_ [!ref <remove_spaces>]
...
-   time_warp_mode: bicubic
-   freq_mask: True
-   n_freq_mask: 2
-   time_mask: True
-   n_time_mask: 2
-   replace_with_zero: False
-   freq_mask_width: 30
-   time_mask_width: 40
+# Speed perturbation
+speed_changes: [95, 100, 105]  # List of speed changes for time-stretching
```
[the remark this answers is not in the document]
mravanelli: I will change

recipes/TIMIT/ASR/transducer/hparams/train.yaml
mravanelli: In my recipe tests, I had errors because the units (according the label_dictionary) are actually 42. I will check better later

recipes/TIMIT/ASR/transducer/hparams/train_wav2vec.yaml
mravanelli: Same as above. I will double check

recipes/LibriParty/VAD/hparams/train.yaml
mravanelli: Yes, I changed that on purpose because it was wrong: consistently adding -15-20dB is a crazy amount of noise that might be the cause of some issues we are observing with the VAD

recipes/VoxCeleb/SpeakerRec/hparams/train_ecapa_tdnn.yaml
mravanelli: This was done on purpose. I think applying speed change to speaker recognition is not a  good idea as it will change the pitch and the speaker identity as well.

mravanelli:
I did a bunch of tests and everything seems to work. Recipe tests are fine. TIMIT recipes achieve the expected performance. I tried LibriSpeech Transformer as well and the performance is the expected one too. I think we can merge it!

PR #1905, opened 2023-03-27
HyperConformer

florianmai:
I added an implementation for the HyperConformer architecture.
In short, HyperConformer replaces the self-attention module in Conformer with multi-head HyperMixer, an extension of [the original HyperMixer](https://arxiv.org/abs/2203.03691) proposed for NLP. Since HyperMixer has linear memory and time complexity in the sequence length, it is an efficient alternative to Conformer while attaining similar or sometimes better results.

Regarding the implementation, I made three major changes. First, I added the "hypermixing.py" module to nnet. Second, I made minimal changes to lobes/models/ASR/transformer/Transformer /Conformer and /TransformerASR such that HyperMixing is supported as a type of attention. Third, I added two new recipes for training HyperConformer on LibriSpeech.

PR #2217, opened 2023-10-23
reverting an error added by HyperConformer (code from Samsung AI Cambridge)

TParcollet:
This PR reverts an error that was introduced by this [PR](https://github.com/speechbrain/speechbrain/pull/1905). This error destroys the performance of all our conformer/transformer model for ASR, whatever the dataset. We are currently making sure that it works as expected now. CC @shucongzhang

- [x] Verify that the issue is fixed for the Tedlium PR


Adel-Moumen (approved):
:(

PR #2228, opened 2023-11-02
Fix Resample instantiation to include device argument

ywk991112:
## What does this PR do?

This pull request fixes a runtime error that occurs when the `Resample` object from `torchaudio.transforms` is instantiated without being assigned to the specified device. The lack of device assignment can cause a mismatch when the downstream operations are performed on tensors residing on non-CPU devices (e.g., CUDA devices).

The change includes the addition of the `.to(device)` method call during the `Resample` object instantiation, ensuring that the object is created on the correct device as specified by the `device` argument in the `download_vctk` function.

No dependencies are added with this change, and the fix is isolated to the instantiation of the `Resample` object.

## Before submitting

- [ ] Did you write any **new necessary tests**? (not for typos and docs) (You would mark this as complete if you added tests, otherwise explain why they are not necessary)
- [x] Did you list all the **breaking changes** introduced by this pull request? (Mark this as complete if there are no breaking changes)

## PR review

### DIFF
--- a/recipes/Voicebank/voicebank_prepare.py
+++ b/recipes/Voicebank/voicebank_prepare.py
@@ -432,7 +432,7 @@ def download_vctk(destination, tmp_dir=None, device="cpu"):
         "clean_trainset_28spk_wav",
     ]
 
-    downsampler = Resample(orig_freq=48000, new_freq=16000)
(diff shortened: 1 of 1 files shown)


mravanelli:
Thank you for the fix @ywk991112!

mravanelli (approved):

PR #2231, opened 2023-11-04
Add wrappers for Encodec and Vocos vocoders

flexthink:
Introduce wrappers for the following pretrained models:
* Encodec
  * **Repository:**: https://huggingface.co/docs/transformers/v4.31.0/en/model_doc/encodec
  * **Paper:**:  https://arxiv.org/abs/2210.13438
* Vocos
  *  **Repository:** https://huggingface.co/charactr/vocos-encodec-24khz. 
  * **Paper:** https://arxiv.org/pdf/2306.00814.pdf

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -26,11 +26,15 @@ except ModuleNotFoundError:
 try:
     from transformers import Wav2Vec2Model  # noqa: F401
 except ModuleNotFoundError:
-    collect_ignore.append("speechbrain/lobes/models/huggingface_transformers/wav2vec2.py")
+    collect_ignore.append(
+        "speechbrain/lobes/models/huggingface_transformers/wav2vec2.py"
+    )
 try:
     from transformers import WhisperModel  # noqa: F401
 except ModuleNotFoundError:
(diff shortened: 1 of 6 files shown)


mravanelli:
Thank you  @flexthink for submitting this PR! Here are my comments:

1. It appears that the encodec and vocos interfaces do not inherit from HFTransformersInterface. As you can see, in the unstable branch (https://github.com/speechbrain/speechbrain/blob/develop/tests/.run-doctests.sh#L8), all modules now inherit from it. I recommend revising your code to ensure that it inherits from that class as well to maintain consistency.

2. There seems to be a test failure due to additional requirements. To address this, I suggest adding "vocos.py" to the list of files to avoid, similar to the following example: https://github.com/speechbrain/speechbrain/blob/develop/tests/.run-doctests.sh#L8

3. I recommand including docstring examples for each model

mravanelli (approved):

mravanelli:
Thank you @flexthink for this PR. Encodec is a very popular model and having it in SpeechBrain is a plus!

PR #2233, opened 2023-11-07
Discrete SSL

poonehmousavi:
## What does this PR do?

 Recipe for Training K-Means Clustering on LibriSpeech Data
-Using Self-Supervised Model-Based Representations: Hubert, WavLM,Wav2Wec
- Adding Discrete version of HUbert, Wav2vec and WavLM

List any dependencies that are required for this change.
 - scikit-learn


mravanelli:
Thank you, Pooneh, for this PR! I've reviewed it and have some comments:

1. **GPU Compatibility**: The recipe is currently not GPU-compliant. When the device is set to 'cuda', it should compute the features on the GPU, switch them to the CPU (this step is currently missing and causes a device mismatch issue), and then perform clustering on the CPU using scikit-learn. 

2. **Batch Size**: Even after resolving the device mismatch issue, the recipe remains too demanding for any GPU due to the high batch size, leading to Out Of Memory (OOM) issues. It appears that the batch size constraint, "batch_size: 1000 # should be >= num_clusters," is causing the problem. To mitigate this, consider defining two batch sizes: one for data loading and one for clustering. You can accumulate feature vectors on the CPU until you reach the batch size needed for clustering.

3. **Memory Usage**: Please note that the recipe also imposes high memory demands on CPU machines, leading to job terminations. This issue has been observed with wav2vec and wavlm. I think you can address the problem with the approach suggested in the pont above.

4. **Docstring Example**: The docstring example for HuberDiscrete references SSL_Quantization, which doesn't exist in SpeechBrain. I suggest adding this functionality (with a different name, like "Hubert_Quantization") and marking the repository as a work in progress. This way, it can be updated once the optimal number of tokens is determined.

5. **README.md File**: The LibriSpeech Quantization repository is missing a README.md file. Similar to existing repositories, it should explain what these recipes do, the concept of quantization, when it can be beneficial, how to run an experiment, any additional requirements, links to the Hugging Face repository, Dropbox links (if necessary), and more. A comprehensive README will enhance user understanding and adoption.

6. **Recipe Tests**: Please ensure that recipe tests are included to validate the functionality of the added recipes.

7. **Models**: Once the aforementioned issues are addressed, make sure to include discrete_wav2vec and wavlm models

8. **Dropbox for Models**: While it can be postponed, we will need to provide access to the logs on Dropbox in the future.

mravanelli:
@Chaanks as discussed privately, It could be great if you and @poonehmousavi converge on a reasonable setting for the downstream task. Of course, we can change it according on the experimental evidence, but we need to start with something reasonable.

mravanelli (approved):

mravanelli:
Thank you @poonehmousavi, I ran the last tests and everything runs fine. I think we can merge it now. @poonehmousavi and @Chaanks, please follow up on the best hparams to start with. I think we need to start training the vocoders for these models soon.

PR #2237, opened 2023-11-09
Fix length_normalization in Unstable Branch

mravanelli:
After many experiments, we realized that setting `length_normalization:False` makes the beamsearch performance worse in the unstable branch. This PR sets length_normalization:True in all recipes (which is normally a good choice well aligned with what is done in the test search). This issue was introduced in the rescoring refactor PR [here](https://github.com/speechbrain/speechbrain/pull/751/files). @Adel-Moumen will investigate the cause of that, but for now, the easiest fix is to set length_normalization:True everywhere.

__________________
before the fix:
`unstable: epoch: 10, lr: 6.45e-04, steps: 60170, optimizer: Adam - train loss: 17.08 - valid loss: 10.89, valid ACC: 9.58e-01, valid WER: 7.89`
`dev: epoch: 10, lr: 6.45e-04, steps: 60170, optimizer: Adam - train loss: 18.03 - valid loss: 11.98, valid ACC: 9.56e-01, valid WER: 5.72`

after the fix:
`ustable: epoch: 10, lr: 7.40e-04, steps: 45610, optimizer: Adam - train loss: 17.69 - valid loss: 10.24, valid ACC: 9.56e-01, valid WER: 5.59`
`dev: epoch: 10, lr: 6.45e-04, steps: 60170, optimizer: Adam - train loss: 18.03 - valid loss: 11.98, valid ACC: 9.56e-01, valid WER: 5.72`


Adel-Moumen (approved):
LGTM! Thanks @mravanelli.

PR #2240, opened 2023-11-11
Fixes for Encodec

flexthink:
* Decouple the extraction of tokens from embeddings
* Register buffers for tensors in order to avoid CPU/GPU errors


mravanelli:
Thank you @flexthink for this PR. Everything works on my side.

PR #2246, opened 2023-11-13
Sg/dac

shubham-gupta-30:
## What does this PR do?

This PR incorporates Descript-Audio-Codec in speechbrain. 

<details>

</details>

## PR review

<details>

</details>


poonehmousavi:
Thank you @shubham-gupta-30  for this valuable PR.
Following are some of my comments :

1.  the pre-commit check is failed. You need to follow this doc (https://speechbrain.readthedocs.io/en/latest/contributing.html) to make sure that all the checks are passed. You need to run pre-commit run [your-file] to make sure everything works.
3. You need to provide a docstring for all your methods(That is why there are some errors in the CI workflow). Also, please check other recipes and provide the docstring in a consistent format.
4. The docstring of your main model (class DAC(nn.Module) should contain some example. Please refer to this file and follow the same format. (https://github.com/speechbrain/speechbrain/blob/unstable-v0.6/speechbrain/lobes/models/huggingface_transformers/discrete_hubert.py)
5. The init function of your main module should load the weights. So, when we define the DAC in the hyper-param.yaml file of the recipe, we have a pre-trained model that is ready to be used without the need to call any other function. For example, when we want to define the DAC module in our YAML file, we should be able to pass the   model_type and  model_bitrate, and we get the pre-trained models that the weights are already loaded.
6. It would be better to have a param called freeze in the DAC module that defines if we want to freeze or fine-tune the model. In our case, I guess we should set freeze to be True, and if it is true all the computation of forward function should be done inside no_grad to be more efficient. You could wrap your forward function with this:  with a torch.set_grad_enabled(not self.freeze):
7. Right now in both existing discrete models( discrete_ssl and encoded), in the forward/encode function we return the tokens and the embedding. I noticed that the DAC model return more variables. Unless there is a reason for that,  it would be better for all models to have the same returning variables for easier integration across different recipes.
8. After all these changes, please  follow https://speechbrain.readthedocs.io/en/latest/contributing.html to make sure that all the checks are passed
Thanks for this PR and let me know if you need any help.

shubham-gupta-30:
Thanks for a detailed review @poonehmousavi 

I address all your concerns, please take another look at the PR! 

For preloading  DAC through the constructor, I introduced a constructor parameter  `load_pretrained` which when set to `True`, a pretrained model is downloaded from the right link. So the easiest way to use the model right now is simply:

`model = DAC(load_pretrained=True)` and it will load a model based on the default values of otehr params set in the constructor.

speechbrain/lobes/models/discrete_tokens/dac.py
```diff
@@ -0,0 +1,1108 @@
...
+        self.decoder = Decoder(
+            self.latent_dim, self.decoder_dim, self.decoder_rates,
+        )
+        self.apply(init_weights)
+
+        if load_pretrained:
+            self.load_state_dict(model_dict["state_dict"], strict=strict)
+            self.metadata = metadata
+
+    def preprocess(self, audio_data: torch.Tensor, sample_rate: int):
```
poonehmousavi, line 1000: Most of the time in speech brain, the resampling and the padding have been done in the dataio_prepare of each recipe..So maybe this function is not necessary but the assertion for the sampling rate is good to be done here and also it would be good that you mention the required SR  in the docstring.  also, it might be better to do this check inside the forward fiction.

Also, you need to set the default value for sample_rate to None.. otherwise dac.preprocess(audio_data) yields " missing 1 required positional argument: 'sample_rate"

speechbrain/lobes/models/discrete_tokens/dac.py
```diff
@@ -0,0 +1,1108 @@
...
+    strict : bool
+        Whether to strictly enforce the state dictionary match.
+    load_pretrained : bool
+        Whether to load a pretrained model.
+
+    Examples
+    --------
+    Creating a new DAC instance:
+
+    ```python
```
poonehmousavi, line 892: You need to follow the same format for example using >>> .. Please check the other model.  It would be good to also mention the expected shape of the output(both token and embedding : E.g, [bathc_size, n_codebooks, codebook_size] for embedding.

It would be better that all required preprocessing and checking have been done inside the forward function instead of calling two functions.  It would be easier to use the DAC model later in the recipes.

poonehmousavi:
@shubham-gupta-30 I reviewed the changes. I left some minor comments, mostly formatting issues. Please apply the changes and re-run all the tests and pre-commit to be sure everything works.

shubham-gupta-30:
@poonehmousavi  Thanks for the review, I removed the `preprocess` function entirely. Please take another look!

poonehmousavi:
@shubham-gupta-30 Thanks for the changes. This PR seems fine to me. @mravanelli  Could you please make a final pass?

shubham-gupta-30:
Thanks @mravanelli  for the review. I addressed all of your comments for the PR. I added runnable examples for all the three major units of DAC - `Encoder`, `ResidualVectorQuantize` and `Decoder`. 

I also renamed `discrete_tokens` to `discrete` The import statement will now be:

```from speechbrain.lobes.models.discrete.dac import DAC```

To get rid of the `.dac`, I will need to create a `__init__.py` file in the folder, which does not pass the pre commit tests right now as the model is not being used anywhere. I can add this when i add code using this model, and so everywhere we use the model will actually have the import statement you mentioned that is : 

```from speechbrain.lobes.models.discrete import DAC```

Please take another look!

mravanelli (approved):

mravanelli:
Thank you @shubham-gupta-30 for this PR and @poonehmousavi for helping with the review. I only had to make docstring exampels much smaller to fit them into the CI server.

PR #2247, opened 2023-11-14
Kemans other dataset

poonehmousavi:
## What does this PR do?
 Adding quantization recipes for a subset of the dataset:
 - IEMOCAP
 - CommonVoice
 - LJSPEECh
 Create means class and define all k-means related function there for reusability and modify existing LibriSpeech recipe to be compatible.


recipes/CommonVoice/quantization/hparams/train_with_hubert.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+sorting: ascending
+
+# Dataloader options
+train_dataloader_opts:
+   batch_size: !ref <batch_size>
+
+ssl_model: !new:speechbrain.lobes.models.huggingface_transformers.hubert.HuBERT
+   source: !ref <ssl_hub>
+   output_norm: True
```
Chaanks, line 43: Need to be set to False otherwise the Kmeans model will not converge

recipes/CommonVoice/quantization/hparams/train_with_wav2vec.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+sorting: ascending
+
+# Dataloader options
+train_dataloader_opts:
+   batch_size: !ref <batch_size>
+
+ssl_model: !new:speechbrain.lobes.models.huggingface_transformers.wav2vec2.Wav2Vec2
+   source: !ref <ssl_hub>
+   output_norm: True
```
Chaanks, line 43: `ouput_norm`  have to be set to False otherwise the Kmeans model will not converge

poonehmousavi:
done in another PR:  https://github.com/speechbrain/speechbrain/pull/2255

PR #2252, opened 2023-11-20
Refactor Inference (files and folders)

mravanelli:
This PR refactors the files and folders of the inference interfaces.

NOTE:
- For the new version of SpeechBrain, I would also like that each model to have a working inference interface and that in each README.md file we clearly explain how users can use the trained models for inference. This will be done in a future PR.


speechbrain/inference/text.py
```diff
@@ -0,0 +1,248 @@
+""" Specifies the inference interfaces for tex-processing modules.
```
Adel-Moumen, line 1: `tex-processing` -> `text-processing`

speechbrain/inference/__init__.py
```diff
@@ -0,0 +1,11 @@
+"""Pretrained models"""
```
Adel-Moumen, line 1: we should elaborate a bit more on that

speechbrain/inference/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+import os
+
+__all__ = []
+for filename in os.listdir(os.path.dirname(__file__)):
+    filename = os.path.basename(filename)
+    if filename.endswith(".py") and not filename.startswith("__"):
+        __all__.append(filename[:-3])
```
Adel-Moumen, line 9: Do you mind explaining a bit more why we are doing that ?

speechbrain/utils/text_to_sequence.py
```diff
@@ -324,12 +324,13 @@ def _g2p_keep_punctuations(g2p_model, text):
     """do grapheme to phoneme and keep the punctuations between the words
     Arguments
     ---------
-    g2p_model: speechbrain pretrained g2p model
+    g2p_model: speechbraininference.text g2p model
```
Adel-Moumen, line 327: `speechbrain.inference`

Adel-Moumen (commented):
Hello Mirco, 

Overall everything seems great. I have only 1 or 2 remarks that should be straightforward to change.

PR #2253, opened 2023-11-21
Fix beam search

Adel-Moumen:
## What does this PR do?

This PR fixes several minor issues with the current Beam Search in unstable. 

I found one issue on `self.prefix_length` that was starting at `1` instead of `0` in the `CTCPrefixScorer`, and also, we were not blocking the `log_probs` of `blank_index` before computing the scores. 

Regarding the WER difference w.r.t `length_normalization`, the issue was due to: `final_scores = scores[index]`. Indeed, in unstable, we save the scores of an hyp that reach eos in a variable called `final_scores`, HOWEVER, doing so will not create a new variable but rather a reference to the variable `scores[index]`. What is the issue then ? The issue is that we are then setting the `scores[index]` to `-inf` because the hyp reached eos, and so `final_scores` will be equal to `-inf`... Why is it not happening in develop ? In develop, we are multiplying the `final_scores` with `length_rewarding` such as : `final_scores = scores[index] + self.length_rewarding * (timesteps + 1)`. Even if `self.length_rewarding` is equal to `0.0`, it will lead to a NEW variable even though we have the same final value, and thus, it will not be override by `-inf`....

I did the tests and it works. 

Example in python :
```python
import torch 

scores = torch.tensor([10.0, 20.0])
final_score = scores[0]
scores[0] = float('-inf')
final_score == float('-inf') # True...
```

An effective solution is to simply do `clone()`:

```python
import torch 

scores = torch.tensor([10.0, 20.0])
final_score = scores[0].clone()
scores[0] = float('-inf')
final_score == float('-inf') # False !
```

<details>

- [ x Did you verify new and **existing [tests](https://github.com/speechbrain/speechbrain/tree/develop/tests) pass** locally with your changes?

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/decoders/ctc.py
+++ b/speechbrain/decoders/ctc.py
@@ -55,7 +55,7 @@ class CTCPrefixScore:
         self.minus_inf = -1e20
         self.last_frame_index = enc_lens - 1
         self.ctc_window_size = ctc_window_size
-        self.prefix_length = 0
+        self.prefix_length = -1
 
         # mask frames > enc_lens
         mask = 1 - length_to_mask(enc_lens)
--- a/speechbrain/decoders/scorer.py
+++ b/speechbrain/decoders/scorer.py
@@ -1087,7 +1087,7 @@ class ScorerBuilder:
         weights=dict(),
         full_scorers=list(),
         partial_scorers=list(),
-        scorer_beam_scale=1.5,
+        scorer_beam_scale=2,
     ):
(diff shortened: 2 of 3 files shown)


mravanelli (approved):

mravanelli:
I did a full inference experiment with and without length normalization and everything works now. Thank you @Adel-Moumen for the fix.

PR #2120, opened 2023-08-13
[WIP] Zero-Shot Multi-Speaker Tacotron2

pradnya-git-dev:
# Contribution in a nutshell
Hey, this could help our community work with zero-shot multi-speaker text-to-speech

# Scope
* [ ] Extending single-speaker Tacotron2 with speaker encoding capabilities to generate speech based on input text and speaker identity (embedding)

PR #2261, opened 2023-11-22
multi-speaker tacotron2 enhancements

BenoitWang:
## What does this PR do?
Hi @mravanelli @pradnya-git-dev, as discussed, this PR contains some enhancements of the multi-speaker tacotron2 (#2120) based on my experiments.

The modifications are:
- [x] Use characters instead of phonemes (which shows better alignments with our tacotron)
- [x] Do 16k generation instead of 22k with the 16k hifi-gan pretrained
- [x] Limit the maximum audio length to 10 seconds
- [ ] Train with more speakers (add train-clean-360 into training set)
- [ ] Make MS-Tacotron2 a child class of Tacotron2
- [ ] Update the interface and the model
- [ ] recipe tests

### DIFF
--- a/recipes/LibriTTS/README.md
+++ b/recipes/LibriTTS/README.md
@@ -20,7 +20,9 @@ Training time required on NVIDIA A100 GPU using LibriTTS train-clean-100 and tra
 
 The training logs are available [here](https://www.dropbox.com/sh/ti2vk7sce8f9fgd/AABcDGWCrBvLX_ZQs76mlJRYa?dl=0).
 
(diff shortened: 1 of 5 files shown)


BenoitWang:
Hi @mravanelli @pradnya-git-dev , for now I'm still trying to make the alignments work with more speakers by adding `train-clean-360`, the experiments may take long, I think we can merge these first modifications while waiting.

mravanelli:
Thank you @BenoitWang for the improvements. I tested and everything runs. As mentioned, we can merge this PR and follow up in the future with another one for further changes.

Issue #2182, opened 2023-10-01
Causal transformer encoder


lucadellalib:
### Describe the bug

In `TransformerASR`, when `causal=True`, I don't see any code to handle causality in a few locations:
1) Source attention mask: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/transformer/TransformerASR.py#L229
2) 1D CNN: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/transformer/Transformer.py#L348

### Expected behaviour

I think we should:
1) Use `get_lookahead_mask` to initialize `src_mask`  when `causal=True` in https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/transformer/TransformerASR.py#L229
2) Make the 1D CNN causal when `causal=True` in https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/transformer/Transformer.py#L348

mravanelli:
This might be relevant for #2140

mhn226:
Hi @lucadellalib , could you please elaborate in which situation the causal encoder should be useful? Is it for streaming? If it is, I think it might need more treatment than simply making the encoder causal like the decoder part (Mirco referred to this streaming PR #2140).

asumagic:
#2140 uses `causal=False` because it implements chunked attention rather than pure left attention, so I am not very familiar with that code path. I don't think the problems raised in the issue affect the chunked implementation. #2140 also implements custom chunking for the convolution (DCConv's approach).

For ASR, most modern techniques avoid using left attention and causal convolutions, because chunked approaches generally seem to benefit performance significantly. I guess the question is whether we even want to keep the causal setting if it is potentially broken and unused.

lucadellalib:
@mhn226 @asumagic true, you need more than a causal encoder to make the model streamable. However sometimes it's useful to evaluate the model in causal settings to figure out what's the expected performance loss if you had to use it in a streaming fashion. The current code is supposed to make the encoder causal when `causal=True` but it doesn't. I opened a pull request to fix the issue (https://github.com/speechbrain/speechbrain/pull/2262).

asumagic:
I'm assuming all the issues were addressed by #2262, so I'm closing this. Feel free to reopen if I'm wrong.

PR #2262, opened 2023-11-22
Fix encoder causality in TransformerASR (see issue #2182)

lucadellalib:


asumagic:
For reference, using that repro code with the PR:  
with-pr

Without the PR:  
without-pr

A red cell means that changing a specific input had an effect on a specific output. So at least with this test code, no output can see beyond its corresponding input as expected.

mravanelli:
Thank you @lucadellalib for the fix and @asumagic for the review!

Issue #2250, opened 2023-11-17
Checkpoint not being fully removed


whettenr:
### Describe the bug

Randomly get this error
`OSError: [Errno 39] Directory not empty: 'location/to/CKPT'
`
and again when I look at the CKPT directory the optimizer.ckpt is still in there

### Expected behaviour

I expect the CKPT to be entirely removed.

### Environment Details

- Using DDP
- Happen on the SpeechBrain development branch an 1 other branch

### Additional Context

I think this is just an issue with DDP that rarely happens. I think sometimes when running on DDP the program tries to remove the CKPT directory before it is has removed all the files inside of it. 

I'm not too familiar with speech brain and DDP but maybe there could be a simple fix to wait or check if the CKPT folder is empty.

Adel-Moumen:
Hello @pplantinga, could you please have a look? Thanks...

pplantinga:
The directory is removed all at once with `shutil.rmtree` so at first I couldn't think of what the issue could be. But I guess if any files get written by another process into this folder while it's being deleted, that would cause it to fail. For example, if process 2 saves `optimizer.ckpt` into a checkpoint while process 1 is deleting the same checkpoint, then it would fail.

whettenr:
@Adel-Moumen  @pplantinga, 

I think that is the issue with one process deleting and another process creating, but I don't understand why anything should be being created. 

The checkpoint was from 10 epochs ago and so I believe the optimizer should have already been created. So I don't know why anything new should be being created in this folder.

Had it happen to me again btw this morning.

whettenr:
@Adel-Moumen Yes I will test it out and get back to you! Thanks everybody for taking the time to work on this!

asumagic:
Note, which Adel told me to report here: based on a recent-ish `unstable-0.6`, without #2267, I've encountered an issue with `save_checkpoint` failing at the end of a training epoch.

```
  File "/home/sdelang/sb/speechbrain/core.py", line 1420, in fit
    asr_brain.fit(
  File "/home/sdelang/sb/speechbrain/core.py", line 1420, in fit
    self._fit_train(train_set=train_set, epoch=epoch, enable=enable)
  File "/home/sdelang/sb/speechbrain/core.py", line 1267, in _fit_train
    self._save_intra_epoch_ckpt()
  File "/home/sdelang/sb/speechbrain/core.py", line 1440, in _save_intra_epoch_ckpt
    self.checkpointer.save_and_keep_only(
[...]

PR #2268, opened 2023-11-26
Sync before and after deleting

pplantinga:
To prevent the error in #2250 this PR adds a barrier before and after deletion so that no processes can write at the same time as the deletion.

### DIFF
--- a/speechbrain/utils/checkpoints.py
+++ b/speechbrain/utils/checkpoints.py
@@ -990,10 +990,22 @@ class Checkpointer:
                 )
             )
 
+        # Sync before deleting to avoid another process saving at the same time.
+        # This has led to errors as documented here:
+        # https://github.com/speechbrain/speechbrain/issues/2250
+        if torch.distributed.is_initialized():
+            torch.distributed.barrier()
+
         # Delete unprotected checkpoints
         for ckpt in potential_deletions:
             if ckpt not in protected_checkpoints:
                 Checkpointer._delete_checkpoint(ckpt, verbosity=verbosity)
+
+        # Sync after deleting to avoid another process saving at the same time.
+        # This has led to errors as documented here:
+        # https://github.com/speechbrain/speechbrain/issues/2250
+        if torch.distributed.is_initialized():
+            torch.distributed.barrier()
 
     @staticmethod
     @main_process_only
--- a/speechbrain/utils/checkpoints.py
+++ b/speechbrain/utils/checkpoints.py
@@ -60,7 +60,11 @@ import logging
 import warnings
 from packaging import version
 import speechbrain.utils._workarounds as __wa
-from speechbrain.utils.distributed import main_process_only, if_main_process
+from speechbrain.utils.distributed import (
+    main_process_only,
+    if_main_process,
+    ddp_barrier,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -993,8 +997,7 @@ class Checkpointer:
         # Sync before deleting to avoid another process saving at the same time.
         # This has led to errors as documented here:
         # https://github.com/speechbrain/speechbrain/issues/2250
-        if torch.distributed.is_initialized():
-            torch.distributed.barrier()
+        ddp_barrier()
 
         # Delete unprotected checkpoints
         for ckpt in potential_deletions:
@@ -1004,8 +1007,7 @@ class Checkpointer:
         # Sync after deleting to avoid another process saving at the same time.
         # This has led to errors as documented here:
         # https://github.com/speechbrain/speechbrain/issues/2250
-        if torch.distributed.is_initialized():
-            torch.distributed.barrier()
+        ddp_barrier()
 
     @staticmethod
     @main_process_only
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -97,6 +97,8 @@ def main_process_only(function):
         os.environ[MAIN_PROC_ENV] = "1"
         if if_main_process():
             result = function(*args, **kwargs)
+        else:
+            result = None
         os.environ[MAIN_PROC_ENV] = "0"
         return result
 
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -65,7 +65,9 @@ def run_on_main(
             # Just run on every process without any barrier.
             post_func(*post_args, **post_kwargs)
         else:
-            main_process_only(post_func)(*post_args, **post_kwargs)
+            # Do the opposite of `run_on_main`
+            if not if_main_process():
+                post_func(*post_args, **post_kwargs)
             ddp_barrier()
 
 
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -10,7 +10,7 @@ import os
 import torch
 from functools import wraps
 
-MAIN_PROC_ENV = "MAIN_PROC_ONLY"
+MAIN_PROC_ONLY = 0
 
 
 def run_on_main(
@@ -94,12 +94,13 @@ def main_process_only(function):
     @wraps(function)
     def main_proc_wrapped_func(*args, **kwargs):
         """This decorated function runs only if this is the main process."""
-        os.environ[MAIN_PROC_ENV] = "1"
+        global MAIN_PROC_ONLY
+        MAIN_PROC_ONLY += 1
         if if_main_process():
             result = function(*args, **kwargs)
         else:
             result = None
-        os.environ[MAIN_PROC_ENV] = "0"
+        MAIN_PROC_ONLY -= 1
         return result
 
     return main_proc_wrapped_func
@@ -111,7 +112,7 @@ def ddp_barrier():
     group enters this function.
     """
     # Check if we're in a single-threaded section, skip barrier
-    if os.environ.get(MAIN_PROC_ENV, "0") == "1":
+    if MAIN_PROC_ONLY >= 1:
         return
     elif torch.distributed.is_initialized():
         torch.distributed.barrier()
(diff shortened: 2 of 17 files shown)


pplantinga:
Looks like the issue here is an `if_main_process()` in the recipe which doesn't play nice with `torch.barrier()`. Although this technically could be said to be an issue with the recipe, not the core code, we should improve the situation so that these two functions work together better. I briefly looked into converting `if_main_process()` into a context manager with setting an environment variable or something to check when running `torch.barrier()` but unfortunately this doesn't allow skipping the execution of the body of the `with` statement without some very hacky workarounds. If anybody has a bright idea about a solution to this, I'm all figurative ears (even if I'm literally deaf).

speechbrain/utils/distributed.py
```diff
@@ -103,8 +94,10 @@ def main_process_only(function):
     @wraps(function)
     def main_proc_wrapped_func(*args, **kwargs):
         """This decorated function runs only if this is the main process."""
+        os.environ[MAIN_PROC_ENV] = "1"
         if if_main_process():
             return function(*args, **kwargs)
+        os.environ[MAIN_PROC_ENV] = "0"
```
TParcollet, line 100: This line will be hit by a non main process and the variable will be changed, hence the ddp barrier afterward wouldn't work? no?

TParcollet: Ok nevermind

speechbrain/utils/checkpoints.py
```diff
@@ -990,11 +994,21 @@ def delete_checkpoints(
                 )
             )
 
+        # Sync before deleting to avoid another process saving at the same time.
+        # This has led to errors as documented here:
+        # https://github.com/speechbrain/speechbrain/issues/2250
+        ddp_barrier()
```
TParcollet, line 1000: @pplantinga what will happen if torch_recovery is called outside of a run on main? These barrier would be hit and MAIN_PROC_ENV wouldn't be 1?

pplantinga: Outside of `run_on_main` the program should be operating with multiple processes, so all should hit the barrier together. The only scenario where it would still freeze is if you are inside `if if_main_process():` block, which we should discourage use of.

TParcollet:
It fixes my bug now. We still need Ryan to verify if it solves issue 2250 as well.

speechbrain/utils/distributed.py
```diff
@@ -103,8 +94,11 @@ def main_process_only(function):
     @wraps(function)
     def main_proc_wrapped_func(*args, **kwargs):
         """This decorated function runs only if this is the main process."""
+        os.environ[MAIN_PROC_ENV] = "1"
         if if_main_process():
-            return function(*args, **kwargs)
+            result = function(*args, **kwargs)
```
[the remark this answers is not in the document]
pplantinga: Ah, thanks for catching this, that's indeed a bug. I've pushed the patch.

speechbrain/utils/checkpoints.py
Gastron: I think solution could be developed to catch those bugs where you branch based on the main_process, but inside that branch you call some code which should hit a DDP barrier. So this will not automatically solve problems, but should help catch bugs. This would replace the if_main_process() (almost drop-in, just adds indentation).

```python
BARRIER_PROTECTOR = "SPEECHBRAIN_DDP_BARRIER_PROTECTOR"
os.environ[BARRIER_PROTECTOR] = 0

class DDPProtector(object):
    """Protects from running into DDP Barrier in a code block that has already branched"""
    def __enter__(self):
        # Increment so that we can support nested protectors
        os.environ[BARRIER_PROTECTOR] = str(int(os.environ[BARRIER_PROTECTOR])+1)

    def on_main_process(self):
        # ...There would be a check here...
        return  ## True if on main process, else False

    def __exit__(self, exception_type, exception_value, traceback):
        <something to possibly handle exceptions>
        os.environ[BARRIER_PROTECTOR] = str(int(os.environ[BARRIER_PROTECTOR])-1)
        return

def ddp_barrier():
    """In DDP mode, this function will synchronize all processes.
    torch.distributed.barrier() will block processes until the whole
    group enters this function.
    """
    if int(os.environ[BARRIER_PROTECTOR]) > 0:
        raise RuntimeError("DDP Barrier inside a main process only branch, this will create a deadlock or a subtle bug.")
    # Check if we're in a single-threaded section, skip barrier
    elif os.environ.get(MAIN_PROC_ENV, "0") == "1":
        return
    elif torch.distributed.is_initialized():
        torch.distributed.barrier()
```

This would be simply used to mark that you intend not to run into DDP Barriers in this part of the code:
```python
with DDPProtector() as protector:
    if protector.on_main_process():
        ...
```
So when `if_main_process()` is replaced by this, we should catch some bugs more easily.

Gastron (changes requested):
I think at least the post_func logic should be checked. I also left a suggestion, which could help catch some bugs.

TParcollet:
@pplantinga what do you think of @Gastron comments?

TParcollet:
My take is that the point of this PR is to fix bugs. We should merge it with the fixes so that unstable can be merged in develop and then we PR @Gastron idea into dev?

speechbrain/utils/distributed.py
```diff
@@ -103,8 +94,13 @@ def main_process_only(function):
     @wraps(function)
     def main_proc_wrapped_func(*args, **kwargs):
         """This decorated function runs only if this is the main process."""
+        os.environ[MAIN_PROC_ENV] = "1"
```
Gastron, line 97: Additionally I wonder if the environment variables (like `MAIN_PROC_ENV` here) are the right way to do this sort of process-wide communication. I think something like a variable in a module should be enough here. So instead of this, I think we could just have:

```python
MAIN_PROC_FLAG=0

def main_proc_wrapped_func(*args, **kwargs):
    global MAIN_PROC_FLAG
    MAIN_PROC_FLAG = 1
    ...
    MAIN_PROC_FLAG = 0
    
    
def ddp_barrier():
    global MAIN_PROC_FLAG  # Actually I think this is unnecessary even.
    if MAIN_PROC_FLAG == 0:
        ...
```

speechbrain/utils/distributed.py
```diff
@@ -54,27 +57,15 @@ def run_on_main(
...
             post_func(*post_args, **post_kwargs)
-        elif not if_main_process():
-            # Others go here
-            try:
-                post_func(*post_args, **post_kwargs)
-            finally:
-                ddp_barrier()
         else:
-            # But main comes here
+            main_process_only(post_func)(*post_args, **post_kwargs)
```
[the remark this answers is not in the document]
pplantinga: Aha, you are totally right about this... I'll go ahead and fix this.

speechbrain/utils/distributed.py
pplantinga: Ah yes, a module-level flag is better here.

speechbrain/utils/distributed.py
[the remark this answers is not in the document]
pplantinga: Should be fixed in latest commit

speechbrain/utils/distributed.py
```diff
@@ -3,12 +3,15 @@
...
  * Abdel Heba 2020
  * Aku Rouhe 2020
+ * Peter Plantinga 2023
 """
 import datetime
 import os
 import torch
 from functools import wraps
 
+MAIN_PROC_ENV = "MAIN_PROC_ONLY"
```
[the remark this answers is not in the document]
pplantinga: Changed to module-level variable, which makes this unnecessary

Gastron:
> You have some nice suggestions here @Gastron. You're right that the current setup would fail if there's a `run_on_main` inside another `run_on_main` which perhaps an increment/decrement could handle.
> 
> I was trying to avoid having double-indentation for this scenario, and the `run_on_main` already exists as a construct so I was hoping we could just extend its use a little. I'm not too crazy about the extra context manager, but could be convinced if we aren't able to accomplish the same thing using `run_on_main`.

Perhaps we could encourage all user code to use run_on_main, then any branching flags or counters or such things can be implemented behind the scenes in library code.

In the short term I think forcing run_on_main everywhere (getting rid of if_main_process) would mean a bigger refactor, since the local code (in the if if_main_process: block) would need to be moved into a new function or other callable.

TParcollet:
@Gastron and @pplantinga just to clarify one thing, if we keep using if_main_process (and we should, I agree with Aku), it will still work with this PR right? I don't see any reason why we must replace all the run_on_main? So basically, Mirco wants to release unstable in dev this week, so we need to settle this PR. We either revert with Adel's PR or we move forward with this one. @Gastron what is your opinion on merging this code and opening a new PR to develop your idea? I like the context manager if we can have something simple -- as if_main_process. @pplantinga if we merge, could you confirm that if_main_process will still work or it will break them and we must change all of them?

pplantinga:
`if_main_process` is currently incompatible with `ddp_barrier`, meaning you can't call `delete_checkpoint` without freezing. We should either convert to `run_on_main` or to Aku's suggestion. I'm fine to merge now to fix bugs but we need to address this before any releases.

TParcollet:
@pplantinga could we briefly assess what is impacted by this deadly combination? If it's minor, we could easily fix on dev.

TParcollet:
@Gastron do you agree? If so, we merge and plan a meeting next week (if you guys are available) to solve this design issue properly.

Gastron:
I think improving this incrementally makes sense, I guess we can indeed merge this and make further improvements soonish

mravanelli:
Thank you all for working on this. So, based on what is discussed here, I'm going to merge it. After that, you can discuss better solutions and implement them in another PR.

mravanelli (approved):

PR #2278, opened 2023-11-28
Fix Bug: CommonVoice Transformer Bug loading correct optimizer

poonehmousavi:
## What does this PR do?

The mechanism for loading the optimizer from checkpoints in the Common Voice transformer recipe (CommonVoice/ASR/transformer/train.py) is flawed. This recipe utilizes two distinct optimizers (Adam and SGD), and after a specified number of epochs (specified by the stage_one_epochs parameter in the YAML file), the optimizer transitions from Adam to SGD. While this transition functions correctly during training, issues arise when attempting to load the model from a checkpoint for purposes such as resuming training or inference.

To solve this issue, a two-step process is proposed. Initially, the epoch counter needs to be loaded and examined to determine whether it exceeds the threshold set by stage_one_epochs. If the condition is met, the SGD optimizer should be loaded; otherwise, the Adam optimizer should be employed. The issue in the current implementation arises from the "on_fit_start" function, where an attempt is made to obtain the epoch counter directly by calling self.checkpointer.recover_if_possible(). However, this method runs into errors because of invoking self.checkpointer.recover_if_possible() tries to load all checkpoints, including the optimizer, resulting in inconsistencies. To address this, a modification is suggested. First, the epoch counter should be loaded independently using the pretrainer. Subsequently, the condition can be evaluated, the appropriate optimizer (SGD or Adam) set accordingly, and finally, self.checkpointer.recover_if_possible() can be called to load the checkpoints, ensuring the correct optimizer is utilized without encountering errors.

The only problem is to load the the epoch-counter , an exact path to the saved checkpoint should be passed which doesn't seem so clean. 

Fixes #<issue_number>

<details>

</details>

## PR review

<details>

</details>


mravanelli:
@Gastron, I think we need your feedback here. Essentially, in this particular case where we have two optimizers, we want to first load the epoch counter from the checkpoint, then decide which optimizer we have to use, and only after load the optimizer. Any idea for an elegant solution?

poonehmousavi:
@Adel-Moumen and @mravanelli .. I found a better solution to fix the bug... instead of loading the epoch_counter using pretrainer which is not very clean. Inside "on_fit_start" method, we first need to call " self.checkpointer.recover_if_possible()" to load epoch_counter before self.init_optimizers()( since this function add optimzier to the recoverables and triggered the bug). Then after, we check the condition for optimizer we could call  self.checkpointer.recover_if_possible() to load all recoverables including the optimizer. Also, I have updated the lr of adam optimizer from 0 to lr_adam. The lr is decided by noam_scheduler anyway but it might be misleading for the users.

.vscode/launch.json
Adel-Moumen, line 1: could you please remove this file ?

recipes/CommonVoice/ASR/transformer/train.py
```diff
@@ -268,14 +269,12 @@ def on_fit_start(self):
...
         if self.checkpointer is not None:
             self.checkpointer.recover_if_possible()
 
-        # if the model is resumed from stage two, reinitialize the optimizer
+        # Initialize optimizers after parameters are configured
+        self.init_optimizers()
+
         current_epoch = self.hparams.epoch_counter.current
         if current_epoch > self.hparams.stage_one_epochs:
             self.optimizer = self.hparams.SGD(self.modules.parameters())
```
Adel-Moumen, line 280: I think it would be great to add a logging message on line 280 saying that you are loading the SGD opt (so that the user can be assured that everything works as intended)

.vscode/launch.json
poonehmousavi: done

recipes/CommonVoice/ASR/transformer/train.py
```diff
@@ -268,16 +269,17 @@ def on_fit_start(self):
...
             self.checkpointer.recover_if_possible()
 
-        # if the model is resumed from stage two, reinitialize the optimizer
+        # Initialize optimizers after parameters are configured
+        self.init_optimizers()
+
         current_epoch = self.hparams.epoch_counter.current
         if current_epoch > self.hparams.stage_one_epochs:
+            logger.info(
+                f"The checkpoint's epoch(= {current_epoch}) is greater than stage_one_epochs(= {self.hparams.stage_one_epochs}). Using SGD as the optimizer instead of Adam"
```
Adel-Moumen, line 281: I would rewrite it to something like this so that it is more general:
```
f"The checkpoint's epoch(= {current_epoch}) is greater than stage_one_epochs(= {self.hparams.stage_one_epochs}). Using the fine-tuning optimizer instead of the training one."
```

mravanelli:
Thank you @poonehmousavi for this fix and @Adel-Moumen for the help with the review!

PR #2281, opened 2023-11-30
Gradscaler flags

Adel-Moumen:
## What does this PR do?

This PR add GradScaler flags to ease the customisation of fp16 flags. Furthermore, it adds back the `check_loss_isfinite` so that the training can be automatically interrupted when the model is locked in NaN/infs losses. 

<details>

</details>

## PR review

<details>

</details>


Adel-Moumen:
Hey @asumagic, could you please have a look at this PR? Do you feel it helps to customise more the fp16 training ?

speechbrain/core.py
```diff
@@ -555,6 +582,14 @@ class and override any methods for which the default behavior does not
             The location for performing computations.
         precision (str)
             One of ``fp32``, ``fp16``, ``bf16``.
+        gradscaler_init_scale (float)
```
TParcollet, line 585: I am not in favor in adding these parameters as basic Brain parameters. They will never be used by 99.99% of our users.

speechbrain/core.py
```diff
@@ -1133,6 +1174,7 @@ def fit_batch(self, batch):
             scaled_loss = self.scaler.scale(
                 loss / self.grad_accumulation_factor
             )
+            self.check_loss_isfinite(scaled_loss)
```
TParcollet, line 1177: Checking for all parameters to be finite is redundant with the GradScaler. It already does this. Also it can be crazy expensive for very large large models. Checking if the loss is not finite makes sense, not the parameters.

TParcollet (changes requested):

mravanelli:
Personally, I'm not against adding this hparams. @Adel-Moumen what is the typical use case for that?

asumagic:
I agree it might be a bit weird to have those flags exposed directly as a CLI parameter in the core especially since it is not obvious that those relate to the gradient scaler from their name alone.

Maybe it could be specified as an optional brain parameter of some sort? Not sure how feasible that is and how ergonomic it would be to use in the CLI.

speechbrain/core.py
asumagic: The grad scaler does this but it does not care for how long the parameters have gone infinite. An idea I suggested yesterday was to occasionally check the gradscaler scale for insane values with a patience mechanism, as I've sometimes seen the scale vanish or explode when issues occurred.  
That does induce a CPU-GPU sync though, but I'm not sure how often we have one in the first place.

Adel-Moumen: I agree with @TParcollet. I will remove the part where I'm checking the NaNs/inf in the weights and will only cover the loss part. BTW, the function was also intended for other non-GradScaler use cases like fp32.

asumagic:
> They are all starting with gradscaler as the root of their name suggest that they are related to gradscaler :p

Whoops, I didn't double check the code this morning. It wasn't the case when I checked yesterday :P

mravanelli:
I think @Adel-Moumen is raising a good point here. Even if we think that most of the users won't change these options, what users are supposed to do if they want to change them? Changing core.py directly?

asumagic:
> I think @Adel-Moumen is raising a good point here. Even if we think that most of the users won't change these options, what users are supposed to do if they want to change them? Changing core.py directly?

You can hijack the `Brain` constructor in your training script to override the grad scaler defined by the core, but it's not very pretty...

mravanelli:
As discussed privately, we remove all the gradscaler options but we keep the checks on the loss. Thank you @Adel-Moumen

PR #2282, opened 2023-12-01
Fix recipe tests for TransformerASR

lucadellalib:
This PR fixes the recipe tests for TransformerASR, which broke after merging https://github.com/speechbrain/speechbrain/pull/2262. The reason is that, when using the full encoder-decoder model we set `memory_mask=src_mask`, which seems incorrect. The expected shape for `memory_mask` is `[target_sequence_length, source_sequence_length]`, while the expected shape for `src_mask` is `[source_sequence_length, source_sequence_length]`. The tests were not failing before merging https://github.com/speechbrain/speechbrain/pull/2262 because `src_mask` was hardcoded as `None` (which is the default value for `causal=False`).


mravanelli:
I ran the recipe tests and everything seems to work. Thank you @lucadellalib

Issue #2276, opened 2023-11-28
Severe BUG: `pos` is unexpectedly all zero.


chenjiasheng:
### Describe the bug

The `pos` here as well as the one bellow is unexpectedly zero-ed out by `srcmask` who has `True` values on pad positions instead of non-pad positions.
It is supposed to use `srcmask_inverted` when calling `sinusoidal_positional_embed_encoder`.
To prove this, just deliberately assign `pos` to zero with `pos = 0` after `sinusoidal_positional_embed_encoder` , and you will get exactly the same output audio.

I suggest using `length_to_mask()` to get the mask instead of `get_key_padding_mask`,  because a "\<pad\>" token in the middle of a sequence does need attention (this may happen for mel feature):
```python
def length_to_mask(lens, max_len=None):
    if max_len is None:
        max_len = lens.max()
    return torch.arange(max_len)[None, :] < lens[:, None]
```

One more thing, I think a mask should have `True` values for non-pad positions as convention, otherwise we will name it "pad_mask".

### Expected behaviour

`pos` should not be all 0.

chenjiasheng:
@turian @jnordberg @mschonwe @shawntan

Adel-Moumen:
@TParcollet could you please have a look?

chenjiasheng:
I was trying to reproduce fastspeech2.
I succeeded to run the "huggingface/speechbrain/fastspeech2_ljspeech" model, then when digging into the code I found this issue.
You can quickly confirm the issue by adding one line `pos = 0` after calling `sinusoidal_positional_embed_encoder`, you will synthesize the same waveform unexpectedly.

BenoitWang:
Hi @chenjiasheng, you are right in this case the positional embedding is not used. Even though I don't think it will improve a lot the model performance, I believe that at least it will speed up the convergence while learning the alignments. Thanks for spotting the bug.

As for the mask I agree with you that `get_key_padding_mask ` is a quite old version of the code, in the recent code we do use sequence lengths to generate masks, for example [here](https://github.com/speechbrain/speechbrain/blob/3fcbbbabc020b8917f607566121d5c5c415769b4/speechbrain/lobes/models/Tacotron2.py#L1526in). I will start a pr to fix this and add the functions into the core.

As for your last question, I understand your point and I think both are ok, what really matters is how we use it. If the masks of the padded elements are true, we can easily apply the `tensor.masked_fill_()` function.

asumagic:
[...]

PR #2283, opened 2023-12-01
fix position embedding

BenoitWang:
This pr fixes the position encoding issue in FastSpeech2 (#2276)
- [x] use speechbrain internal position encoding, use lengths to generate masks for mels 
- [x] add `get_mask_from_lengths` into the core
- [x] re-train the models and verify the performance


BenoitWang:
@mravanelli @asumagic I've made the necessary fixes and I'm now waiting to see the models' performance. This may take some time, will tell you when it's ok to merge.

BenoitWang:
Hi @mravanelli @asumagic I just verified the audio performance, generally I can't find a recognisable improvement on audio quality, but a good point is that for the internal alignment model I found some phonemes missed before that can be captured by this fix. For now this pr is ready, will update huggingface and dropbox once this branch is merged.

mravanelli:
Thank you @BenoitWang for the fix!

PR #2290, opened 2023-12-05
CTC-only training recipes for LibriSpeech (code from Samsung AI Cambridge) 

shucongzhang:
## What does this PR do?
This PR provides recipes for training CTC models from scratch (no W2V2, no Whisper). It yields to strong Branchformer-CTC and Conformer-CTC models.


speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -152,9 +152,12 @@ def __init__(
             ),
             torch.nn.Dropout(dropout),
         )
-        self.custom_tgt_module = ModuleList(
```
TParcollet, line 155: @asumagic I guess that this is the place of the problem?

TParcollet:
Thanks @shucongzhang ! Could you start by fixing the tests? It can be done locally by installing pre-commit :-)

speechbrain/lobes/models/transformer/TransformerASR.py
asumagic: Yes, this change does what I ultimately suggested to do in the discussion in my PR, with the same consequences:

> Though, now, I'm thinking that we could just skip the headache and fix the problem by gating it behind `num_decoder_layers > 0` and break models: As far as I can tell, actually, the only affected model in the SB repo seems to be the conformer-transducer model.  
> Then, just in case, we could provide a script that would remove the module from a checkpoint file.

shucongzhang:
> Thanks @shucongzhang ! Could you start by fixing the tests? It can be done locally by installing pre-commit :-)

Definitely. I have fixed the tests :)

mravanelli:
Thank you @shucongzhang! This is a great contribution. My main comment is that we need to update this PR to the new version of speechbrain that is currently available in the unstable branch (but soon will be merged in the development branch). 
You can take a look at the recipes available in LibriSpeech/ASR/CTC for a reference. @Adel-Moumen can help with that conversion. With the new version of speechbrain, you can also improve the performance using CTC beamsearch with n-gram LM rescoring.

Adel-Moumen:
Hello @shucongzhang, 

Thanks for your great work! 

We recently released in `develop` the new SpeechBrain 1.0 version (see: https://colab.research.google.com/drive/1IEPfKRuvJRSjoxu22GZhb3czfVHsAy0s?usp=sharing), and unfortunately, this PR needs to be sync with the latest commits in the `develop` branch so that we can move forward with this PR. 

Best, 
Adel

recipes/LibriSpeech/ASR/CTC/train_from_scratch.py
```diff
@@ -0,0 +1,547 @@
...
+    def evaluate(
+        self,
+        test_set,
+        max_key=None,
+        min_key=None,
+        progressbar=None,
+        test_loader_kwargs={},
+        nb_avg_ckpt=1,
+        recoverable_name="model",
+    ):
```
Adel-Moumen, line 131: Hello @shucongzhang, I have some hard time understanding the point of redefining these functions. Could you please elaborate a bit more? I'd like to move forward and update this PR with SB 1.0 but Idk if this is necessary. Thanks :) 
CC: @TParcollet maybe you do know why?

speechbrain/nnet/schedulers.py
```diff
@@ -667,6 +668,108 @@ def load(self, path, end_of_epoch=False, device=None):
         self.n_steps = data["n_steps"]
 
 
+@checkpoints.register_checkpoint_hooks
+class LinearNoamScheduler:
```
Adel-Moumen, line 672: Note: we need a unittest for that :p

recipes/LibriSpeech/ASR/CTC/train_from_scratch.py
shucongzhang: Hello @Adel-Moumen , thank you so much for the review and really sorry for the late reply. I was not feeling well and then occupied by something else. I suppose the main reason is at line 538 
 `for nb_ckpt in [1, 5, 10]:`
we can have a quick and easy control about different number of models we want to do the model averaging. Thanks :)

Adel-Moumen: Hello @shucongzhang, hope you are getting better! It's fine, health comes first :)

Yes, at the end, I did understand why you modified it. I decided to remove it as I don't think it should be like that in SpeechBrain, but I do understand why you did it.

Adel-Moumen:
Hello @shucongzhang small ping on my previous message about training time. Could you please confirm me that 500 epochs is reasonable ? Ty.

Adel-Moumen:
I'm wondering about something. Why are we using a sentencepiece 128 BPE vocab for our CTC branch/conformer ? If you are training with CTC then why not using label_encoder with chars ? Ping @TParcollet / @shucongzhang

shucongzhang:
> I'm wondering about something. Why are we using a sentencepiece 128 BPE vocab for our CTC branch/conformer ? If you are training with CTC then why not using label_encoder with chars ? Ping @TParcollet / @shucongzhang

@Adel-Moumen Hi Adel, I was in China for the Lunar New Year and just back to work. Happy Lunar New Year to you :) I used the 128 BPE to follow some previous works. I didn't test what will be the results if using using label_encoder with chars.

Adel-Moumen:
> > I'm wondering about something. Why are we using a sentencepiece 128 BPE vocab for our CTC branch/conformer ? If you are training with CTC then why not using label_encoder with chars ? Ping @TParcollet / @shucongzhang
> 
> @Adel-Moumen Hi Adel, I was in China for the Lunar New Year and just back to work. Happy Lunar New Year to you :) I used the 128 BPE to follow some previous works. I didn't test what will be the results if using using label_encoder with chars.

@shucongzhang Welcome back, and Happy Lunar New Year to you too 🎉! Thanks for the clarification. It makes sense from your point of view. I will try with label_encoder since I think it could be useful here (and from a CTC point of view, I don't really understand the need of using BPE instead of phonemes/chars). 

I let you know when I have some results to share. I might take a bit of time due to the next release of SB 1.0, but I'll let you know what happens :)

TParcollet:
@Adel-Moumen we should avoid using label_encoder imho. SentencePiece can be degenerated to char only. I don't like our recipes that use label_encoder, we have developed this class because we did not want to use SentencePiece at first, but here we are. Maybe we should deprecate it.

Adel-Moumen:
Hello,

So I had the opportunity to train a bit (~200 epochs) with each model (Branchformer and Conformer), and everything ran smoothly :) 

I will therefore merge this PR. Thanks a lot @shucongzhang for your very nice work. :)

## Tests
```python
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Hparam_file"], filters=[["recipes/LibriSpeech/ASR/CTC/hparams/conformer_large.yaml", "recipes/LibriSpeech/ASR/CTC/hparams/branchformer_large.yaml"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```
```
(1/2) Running test for LibriSpeech_row_41...
        ... 11.01s
(2/2) Running test for LibriSpeech_row_42...
        ... 15.03s
TEST PASSED
```

shucongzhang:
> LGTM. Thanks again @shucongzhang :)

Thank you so much @Adel-Moumen !

PR #2308, opened 2023-12-19
Optimize masked Dynamic Chunk Convolution

asumagic:
## What does this PR do?

#2140 introduced a streaming Conformer-Transducer recipe. @TParcollet noted a slowdown when training with `streaming: True`. This PR reworks the convolution module code to not use lists of slices and expresses the entire chunking logic as tensor operations.

- The layernorm and bottleneck are unaffected by the time axis, and are thus applied over the input tensor directly.
- The "list of slice + stack" logic was fully replaced in favor of more clever use of padding and `Tensor.unfold`.

When testing, I found the output of the function to be virtually identical without and with this PR, and the WER% remains unchanged. Training doesn't crash. No breaking changes should occur due to the PR.

I tested performance on my 7900 XT (HIP vs CUDA might have some performance characteristics differences) in a more extreme case (chunk size of `8`, longest batch in `test-clean`, no GS, forward-only):

| Scenario | Performance |
|-|-|
| No masking | `124 ms ± 212 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)` |
| Chunk size = 8, **before** | `236 ms ± 1.55 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)` |
| Chunk size = 8, **after** | `186 ms ± 645 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)` |

The difference is fairly large here as the chunk size is rather small. I don't exactly know how this would translate in training performance and whether that fully solves the performance difference observed by @TParcollet.

<details>

</details>

## PR review

<details>

- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

</details>

### DIFF
--- a/speechbrain/lobes/models/transformer/Conformer.py
+++ b/speechbrain/lobes/models/transformer/Conformer.py
@@ -197,9 +197,6 @@ class ConvolutionModule(nn.Module):
 
             chunk_size = dynchunktrain_config.chunk_size
             batch_size = x.shape[0]
-            chunk_left_context = self.padding
-
-            chunk_count = int(math.ceil(x.shape[1] / chunk_size))
 
             # determine the amount of padding we need to insert at the right of
             # the last chunk so that all chunks end up with the same size.
--- a/speechbrain/lobes/models/transformer/Conformer.py
+++ b/speechbrain/lobes/models/transformer/Conformer.py
@@ -209,56 +209,43 @@ class ConvolutionModule(nn.Module):
                 for i in range(chunk_count)
             ]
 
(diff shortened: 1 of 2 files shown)


TParcollet:
Should we unit-test this as well to make sure both methods are equivalents?

asumagic:
> Should we unit-test this as well to make sure both methods are equivalents?

Actually a good point, should've done so from the start. I've just implemented a test to test this at `ConformerLayer` level.

TParcollet:
I am re-running the experiment now and comparing it, I'll keep you posted. The code LGTM for me.

PR #2309, opened 2023-12-19
BEST-RQ implementation

whettenr:
## What does this PR do?
Adds an implementation of BEST-RQ.
Add layer dropout interface for transformer classes.
Add output hidden layers and wrappers to be able to run MP3S benchmarks.

Fixes #<issue_number>

<details>

</details>

## PR review

<details>

</details>


whettenr:
Hey @Adel-Moumen,
Yes I will work on this! Thanks!
Ryan

TParcollet:
Hello @whettenr, a few major changes before we go into the details of the code. You know that SB follows a dataset-task-oriented structure. Hence you will need to refactor the structure of your code. BEST-RQ is trained on LS, so it should be on Librispeech self-supervised-learning folder, alongside wav2vec2. For the finetuning, they should be in the different task folders, like ASR, CTC. You also have too many hparams  I think, let's keep only the most meaningful ones.

AmirHussein96:
@whettenr thank you for your contribution. I tried reproducing your results reported in https://arxiv.org/pdf/2405.04296. I followed your implementation for pretraining from here https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py using branchformer configuration. The only difference I have is that I increased the batch size from 100 to 1000 and reduced the warmup steps from 25000 to 2500 accordingly. After pretraining I used the all hidden representation following SUPURB ctc downstream task and trained single layer bilstm on 100 Libri and got  PER around  65 which is significantly worse than the WER in the paper. Any ideas what could be the issue? Could you share your pretrained checkpoint you used in the paper so I can test it in my SUPURB setup?

AmirHussein96:
> @whettenr thank you for your contribution. I tried reproducing your results from [this paper](https://arxiv.org/pdf/2405.04296). I followed your pretraining implementation from [here](https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py) using the Branchformer configuration. The only differences in my setup were increasing the batch size from 100 to 1000 and to match your setup I reduced the warmup steps from 25000 to 2500. After pretraining, I froze the model and used all hidden representation with linear combination following the SUPURB CTC downstream task and trained a single-layer BiLSTM on 100 Libri, but I got a PER of around 65, which is significantly worse than the WER reported in the paper. Do you have any ideas on what might be causing this issue? I also tried Hubert in same setup and it PER is 7.

This is my expert.py which shows how I loaded and used the pretrained BestRQ model  in s3prl: 
```
from collections import OrderedDict
from typing import Dict, List, Union
import torch.nn.functional as F
import sys
from audiotools import AudioSignal
import torch.nn as nn
from torch import Tensor
from torch.nn.utils.rnn import pad_sequence

# load speech brain model

import sys
sys.path.insert(0, '/mm2/ahussein/speechbrain/recipes/BEST-RQ') 
import speechbrain as sb
import torch
from hyperpyyaml import load_hyperpyyaml

HIDDEN_DIM = 8

import torch
from hyperpyyaml import load_hyperpyyaml

hyperparams = """
output_folder: /mm2/scratch/ahussein/speechbrain/results/best_hyperbranchconformer/2000
save_folder: !ref <output_folder>/save
pt_model_hub: /mm2/scratch/ahussein/speechbrain/results/best_hyperbranchconformer/2000/save/CKPT+2024-07-10+16-14-27+00
pt_model_output_dim: 512
sample_rate: 16000

####################### Model parameters ###########################
# Transformer
d_model: 512
nhead: 8 # table 1 https://arxiv.org/pdf/2010.10504.pdf
num_encoder_layers: 12 # section 4.1.1
num_decoder_layers: 0
d_ffn: 2048
transformer_dropout: 0.1
activation: !name:torch.nn.GELU
output_neurons: 5000


# Feature parameters
n_fft: 400
n_mels: 80
hop_length: 10
pad_to_divisible_by: 4

# quantizer parameters not using though for finetunning
p_input: 320
cb_dim: 16
cb_vocab: 8192


compute_features: !new:speechbrain.lobes.features.Fbank
    sample_rate: !ref <sample_rate>
    n_fft: !ref <n_fft>
    n_mels: !ref <n_mels>
    
normalize: !new:speechbrain.processing.features.InputNormalization
    norm_type: global
    update_until_epoch: 4

CNN: !new:speechbrain.lobes.models.convolution.ConvolutionFrontEnd
    input_shape: (8, 10, 80)
    num_blocks: 2
    num_layers_per_block: 1
    out_channels: (128, 32)
    kernel_sizes: (3, 3)
    strides: (2, 2)
    residuals: (False, False)

Transformer: !new:speechbrain.lobes.models.transformer.TransformerASR.TransformerASR # yamllint disable-line rule:line-length
    input_size: 640
    tgt_vocab: !ref <output_neurons>
    d_model: !ref <d_model>
    nhead: !ref <nhead>
    num_encoder_layers: !ref <num_encoder_layers>
    num_decoder_layers: !ref <num_decoder_layers>
    d_ffn: !ref <d_ffn>
    dropout: !ref <transformer_dropout>
    activation: !ref <activation>
    encoder_module: conformer
    attention_type: RelPosMHAXL
    normalize_before: True
    causal: False
    output_hidden_states: True
    

enc: !new:speechbrain.lobes.models.transformer.TransformerASR.EncoderWrapper
   transformer: !ref <Transformer>

# Quantizer: !new:.quantiser.RandomProjectionQuantizer
Quantizer: !new:speechbrain.nnet.quantisers.RandomProjectionQuantizer
    # projection
    input_dim: !ref <p_input>
    # codebook
    cb_dim: !ref <cb_dim>    
    cb_vocab: !ref <cb_vocab>

linear: !new:speechbrain.nnet.linear.Linear
    input_size: !ref <d_model>
    n_neurons: !ref <cb_vocab>

pt_model: !new:torch.nn.ModuleList
    - [!ref <CNN>, !ref <enc>, !ref <Quantizer>, !ref <linear>]

modules:
   normalize: !ref <normalize>
   pt_model: !ref <pt_model>
   
pretrainer: !new:speechbrain.utils.parameter_transfer.Pretrainer
   collect_in: !ref <save_folder>
   loadables:
      pt_model: !ref <pt_model>
      normalize: !ref <normalize>

   paths:
      pt_model: !ref <pt_model_hub>/model.ckpt
      normalize: !ref <pt_model_hub>/normalizer.ckpt
"""



class UpstreamExpert(nn.Module):
    def __init__(self, ckpt: str = None, model_config: str = None, **kwargs):
        """
        Args:
            ckpt:
                The checkpoint path for loading your pretrained weights.
                Can be assigned by the -k option in run_downstream.py

            model_config:
                The config path for constructing your model.
                Can be assigned by the -g option in run_downstream.py
        """
        super().__init__()
        loaded_hparams = load_hyperpyyaml(hyperparams)
        self.name = "[BestRQ]"
        
        self.model = loaded_hparams
        print(
            f"{self.name} - You can use model_config to construct your customized model: {model_config}"
        )
        print(f"{self.name} - You can use ckpt to load your pretrained weights: {ckpt}")
        print(
            f"{self.name} - If you store the pretrained weights and model config in a single file, "
            "you can just choose one argument (ckpt or model_config) to pass. It's up to you!"
        )

        # The model needs to be a nn.Module for finetuning, not required for representation extraction
        # self.model1 = nn.Linear(1, HIDDEN_DIM)
        # self.model2 = nn.Linear(HIDDEN_DIM, HIDDEN_DIM)

    def get_downsample_rates(self, key: str) -> int:
        """
        Since we do not do any downsampling in this example upstream
        All keys' corresponding representations have downsample rate of 1
        """
        return 160*4
    
    def get_output_lengths(self, input_lengths):
        """ Function to get the output length of the feature extractor this is
            necessery to compute the masks of BestRQ.
        """
        sr = self.model["sample_rate"]
        hop_length = self.model["hop_length"]

        return (input_lengths // (sr*hop_length / 1000) + 1).to(torch.long)
    
    def forward(self, wavs: List[Tensor]) -> Dict[str, Union[Tensor, List[Tensor]]]:
        """
        When the returning Dict contains the List with more than one Tensor,
        those Tensors should be in the same shape to train a weighted-sum on them.
        """
        # breakpoint()
        device = wavs[0].device
        length = torch.tensor([data.shape[0] for data in wavs], dtype=torch.int32, device=device)
        
        length = self.get_output_lengths(length)
        max_len = torch.max(length)
        length = length//max_len
        wavs = pad_sequence(wavs, batch_first=True)
        feats = self.model['compute_features'](wavs)
        feats = self.model['modules']['normalize'](feats, length, epoch=10)
        self.model['modules']['pt_model'][0] = self.model['modules']['pt_model'][0].to(device)
        src = self.model['modules']['pt_model'][0](feats)
        self.model['modules']['pt_model'][1] = self.model['modules']['pt_model'][1].to(device)
        enc_out, hidden_activations = self.model['modules']['pt_model'][1](src, length)

        # Normalize each feature vector to have L2 norm of 1 along the specified dimension
        # rep = F.normalize(rep, p=2, dim=2)
        # feature: (batch_size, max_len, hidden_dim)
        # The "hidden_states" key will be used as default in many cases
        # Others keys in this example are presented for SUPERB Challenge
        return {
            "hidden_states": hidden_activations[1:],
            # "PR": [hidden, feature],
            # "ASR": [hidden, feature],
            # "QbE": [hidden, feature],
            # "SID": [hidden, feature],
            # "ASV": [hidden, feature],
            # "SD": [hidden, feature],
            # "ER": [hidden, feature],
            # "SF": [hidden, feature],
            # "SE": [hidden, feature],
            # "SS": [hidden, feature],
            # "secret": [hidden, feature],
        }
```

whettenr:
> @whettenr thank you for your contribution. I tried reproducing your results from [this paper](https://arxiv.org/pdf/2405.04296). I followed your pretraining implementation from [here](https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py) using the Branchformer configuration. The only differences in my setup were increasing the batch size from 100 to 1000 and to match your setup I reduced the warmup steps from 25000 to 2500. After pretraining, I froze the model and used all hidden representation with linear combination following the SUPURB CTC downstream task and trained a single-layer BiLSTM on 100 Libri, but I got PER 65 on test-clean, which is significantly worse than the WER reported in the paper. Do you have any ideas on what might be causing this issue? I also tried Hubert in same setup and the resulted PER is 7.

I'm not sure what exactly could be causing it, but there few key differences that maybe could lead to bad performance.

- Changing from Conformer to Branchformer, I have not experimented thoroughly with branchformer, this might mean you need to adjust other hyperparameters to work with Branchformer.
- I'm not sure if how well branchformer and hypermixing work well together. Maybe you could try the original branchformer. I have experimented with conformer and hypermixing and got pretty good results, but I'm not sure about hypermixing with branchformer
- Changing warmup steps, I'm not sure if this could make a big difference, but it is a difference, slow warmup can help prevent getting stuck in a high local minimum
- Batch size, in my opinion this probably has the biggest effect. The **batch size, mask ratio, and learning rate** are all **very** important. Remember that the total batch size is max number of seconds * number of GPUs, * grad accumulation. It looks like you increase the batch size a lot (you might need to increase the number of buckets a bit too if you haven't). With this is might be worth trying to change the mask ratio and or learning rate.

My advice:
If you are ok with not using branchformer with hypermixing and just want to reproduce results, I would try with the conformer (I have done conformer with a bigger batch size of around 1.7 hours and got good results). If you have to use branchformer, I would try using the standard branchformer just to see if you can get good results with that combination of hyperparameters. If not I would experiment with the hyperparameters (mask size, learning rate, warmup, ...).

AmirHussein96:
Thanks, @whettenr, for the quick response. I followed your recommendation and used the Conformer setup. However, I noticed that although you have defined the Noam scheduler in the Conformer config file, it is not used in your train.py file (https://github.com/whettenr/speechbrain/blob/sync-deletion/recipes/BEST-RQ/train.py). During training, the logs showed the same value for the learning rate throughout. Was this intentional or an oversight?

I assumed it might have been an oversight, so I modified train.py slightly to incorporate the Noam scheduler. Here is the change I made:
```
    def on_fit_batch_end(self, batch, outputs, loss, should_step):
        """ Called after fit_batch(), updates learning rate and does per-step logging. """
        if should_step:
            if isinstance(
                self.hparams.noam_annealing, sb.nnet.schedulers.NoamScheduler
            ) or isinstance(
                self.hparams.noam_annealing,
                sb.nnet.schedulers.CyclicCosineScheduler,
            ):
```
I also added TensorBoard to visualize training and validation losses:
bestrq

The current setup is exactly the same as yours, except I am using 5 GPUs with a batch size of 1000 samples per GPU, buckets=100, and warmup=700. Let me know if the values I obtained for training and validation are close to the ranges of your values or if they are off. By the way, the accuracy on validation is around 20%.

Adel-Moumen:
> @whettenr thank you for your contribution. I tried reproducing your results from [this paper](https://arxiv.org/pdf/2405.04296). I followed your pretraining implementation from [here](https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py) using the Branchformer configuration. The only differences in my setup were increasing the batch size from 100 to 1000 and to match your setup I reduced the warmup steps from 25000 to 2500. After pretraining, I froze the model and used all hidden representation with linear combination following the SUPURB CTC downstream task and trained a single-layer BiLSTM on 100 Libri, but I got PER 65 on test-clean, which is significantly worse than the WER reported in the paper. Do you have any ideas on what might be causing this issue? I also tried Hubert in same setup and the resulted PER is 7.

if you want to reproduce the results, I believe @whettenr used the MP3S benchmark available here: https://github.com/speechbrain/benchmarks/tree/main/benchmarks/MP3S. As pointed out by Ryan, augmenting the batch size and reducing the warmup might affect negatively the dowstream results. 

BTW, I just saw on your GitHub profile that you were at JHU. I am currently working as part of the JSALT workshop so if you want we can have a one-to-one discussion about speechbrain and your issue :)

whettenr:
> Thanks, @whettenr, for the quick response. I followed your recommendation and used the Conformer setup. However, I noticed that although you have defined the Noam scheduler in the Conformer config file, it is not used in your train.py file (https://github.com/whettenr/speechbrain/blob/sync-deletion/recipes/BEST-RQ/train.py). During training, the logs showed the same value for the learning rate throughout. Was this intentional or an oversight?
> 
> I assumed it might have been an oversight, so I modified train.py slightly to incorporate the Noam scheduler. Here is the change I made:
> 
> ```
>     def on_fit_batch_end(self, batch, outputs, loss, should_step):
>         """ Called after fit_batch(), updates learning rate and does per-step logging. """
>         if should_step:
>             if isinstance(
>                 self.hparams.noam_annealing, sb.nnet.schedulers.NoamScheduler
>             ) or isinstance(
>                 self.hparams.noam_annealing,
>                 sb.nnet.schedulers.CyclicCosineScheduler,
>             ):
>                 self.hparams.noam_annealing(self.optimizer)
> ```
> 
> I also added TensorBoard to visualize training and validation losses: bestrq
> 
> The current setup is exactly the same as yours, except I am using 5 GPUs with a batch size of 1000 samples per GPU, buckets=100, and warmup=700, masking 10%. Let me know if the values I obtained for training and validation are close to the ranges of your values or if they are off. By the way, the accuracy on validation is around 20%.

I believe a validation accuracy around 20% is pretty good! Have you tried that model on the DS tasks? and yes I did use the MP3S benchmark.

AmirHussein96:
> > @whettenr thank you for your contribution. I tried reproducing your results from [this paper](https://arxiv.org/pdf/2405.04296). I followed your pretraining implementation from [here](https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py) using the Branchformer configuration. The only differences in my setup were increasing the batch size from 100 to 1000 and to match your setup I reduced the warmup steps from 25000 to 2500. After pretraining, I froze the model and used all hidden representation with linear combination following the SUPURB CTC downstream task and trained a single-layer BiLSTM on 100 Libri, but I got PER 65 on test-clean, which is significantly worse than the WER reported in the paper. Do you have any ideas on what might be causing this issue? I also tried Hubert in same setup and the resulted PER is 7.
> 
> if you want to reproduce the results, I believe @whettenr used the MP3S benchmark available here: https://github.com/speechbrain/benchmarks/tree/main/benchmarks/MP3S. As pointed out by Ryan, augmenting the batch size and reducing the warmup might affect negatively the dowstream results.
> 
> BTW, I just saw on your GitHub profile that you were at JHU. I am currently working there as part of the JSALT workshop so if you want we can have a one-to-one discussion about speechbrain and your issue :)

Thank you, @Adel-Moumen! I will definitely try the MP3S benchmark.

It's such a coincidence that you're at JSALT. I hope you're enjoying your time there. Unfortunately, I'm currently in Boston, but I look forward to meeting you in person someday.

AmirHussein96:
> I believe a validation accuracy around 20% is pretty good! Have you tried that model on the DS tasks? and yes I did use the MP3S benchmark.

I saw your logs here https://github.com/whettenr/bestrq/blob/main/results/best_hyperconformer/2000/log.txt and the numbers look close to mine. Do you mind sharing the `train_brq.py` and `ssl_brq_hyperconformer.yaml ` from `https://github.com/whettenr/bestrq/blob/main/run/downstream/run_base_hyperconf_v100.sh`?

AmirHussein96:
@Adel-Moumen,  how can I load models trained with SpeechBrain and saved in the directory CKPT+2024-07-12+15-00-15+00 (which contains brain.ckpt, CKPT.yaml, counter.ckpt, dataloader-TRAIN.ckpt, model.ckpt, noam_scheduler.ckpt, normalizer.ckpt, optimizer.ckpt) using the HuggingFace interface (WeightedSSLModel)?

Alternatively, is there a way to export a SpeechBrain model to be compatible with HuggingFace, including all the files it expects?

AmirHussein96:
@whettenr  @Adel-Moumen  Ok I have good news, I managed to run the pretrained Best RQ on the MR3S  and got: valid CER 6.62, valid WER: 22.21, which is much better than what I got before. Now I am wondering since I use exactly same loading of model parameters and parameters freezing, where is this difference coming from compared to S3PRL? Either this is because of the difference in downstream model (1 vs 2 layers BiLSTM) or difference in batch preparation, (can't think of anything else). @Adel-Moumen does speech brain do anything special during batch preparation (eg normalization)?

whettenr:
> S3PRL

Is the S3PRL using LibriSpeech too? If so, I don't think it would be a problem with normalization. Definitely the difference between a 1 and 2 layer BiLSTM will make a difference, but from what it sounds like results are much worse. 

If result are really bad, my suspicion would be that the weights/parameters of the SSL are not being loaded/transferred properly.

AmirHussein96:
> > S3PRL
> 
> Is the S3PRL using LibriSpeech too? If so, I don't think it would be a problem with normalization. Definitely the difference between a 1 and 2 layer BiLSTM will make a difference, but from what it sounds like results are much worse.
> 
> If result are really bad, my suspicion would be that the weights/parameters of the SSL are not being loaded/transferred properly.

I managed to reproduce the results in S3PRL that matches the MP3S on libri100. There was a bug in data normalization, I was providing the wrong length. I got test CER 5.3 and WER 18.3 on test clean.

whettenr:
> > > S3PRL
> > 
> > 
> > Is the S3PRL using LibriSpeech too? If so, I don't think it would be a problem with normalization. Definitely the difference between a 1 and 2 layer BiLSTM will make a difference, but from what it sounds like results are much worse.
> > If result are really bad, my suspicion would be that the weights/parameters of the SSL are not being loaded/transferred properly.
> 
> I managed to reproduce the results with S3PRL that matches the MP3S on libri100. There was a bug in data normalization, I was providing the wrong length. I got test CER 5.3 and WER 18.3 on test clean.

That's good! I think maybe with a some more hyper parameter tuning you could get even better results. Also I noticed that when updating this pull request to the new version of speechbrain there was indeed an issue with the learning rate schedular. As you previously asked me this:

> During training, the logs showed the same value for the learning rate throughout. Was this intentional or an oversight?

You were 100% right, I did forget at at this in when updating to the new version. It is fixed now but your solution looks great too!

karynaur:
Hope this gets merged soon!

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -224,6 +224,10 @@ class TransformerASR(TransformerInterface):
     use_linear_after_conv: bool, optional
         If True, will apply a linear transformation of size input_size//2.
         -> Branchformer
+    output_hidden_states: bool, optional
```
TParcollet, line 227: @asumagic this PR adds the possibility to return the hidden states of ANY transformer model. I find it important as SSL evaluation requires an access to these states. What do you think of the implementation in this PR?

asumagic: The idea seems ok to me, though the docs should be more explicit on return type.

TParcollet:
@asumagic I took care of cleaning, fixing, documenting the code and even retraining the models. Could you please have a look very briefly at the code to see if anything seems crazy to you? I also added the return type documentation due to this new hidden_states return. 

The models are missing from the readme, but we can do another PR for that imho. 

If @asumagic is happy, i'll merge.

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -180,6 +187,8 @@ def __init__(
                     bias=bias,
                     causal=self.causal,
                     attention_type=self.attention_type,
+                    output_hidden_states=self.output_hidden_states,
```
asumagic, line 190: Missing for `"transformer"` just above?

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -488,15 +515,20 @@ def encode(
...
             dynchunktrain_config=dynchunktrain_config,
         )
 
-        return encoder_out
+        if self.output_hidden_states:
+            encoder_out, _, hidden_states = outputs
+            return encoder_out, hidden_states
+        else:
+            encoder_out, _ = outputs
+            return encoder_out
```
asumagic, line 531: IMO there should be tests for the `output_hidden_states` which would be covering all code paths (i.e. testing with different encoders) and test for the length of the array. The control flow here is a bit messy and prone to breaking when changing the nearby code in the future.

recipes/LibriSpeech/self-supervised-learning/BEST-RQ/train.py
```diff
@@ -0,0 +1,339 @@
...
+        sr = hparams["sample_rate"]
+        hop_length = hparams["hop_length"]
+
+        return (input_lengths // (sr * hop_length / 1000) + 1).to(torch.long)
+
+    @sb.utils.data_pipeline.takes("wav")
+    @sb.utils.data_pipeline.provides("sig")
+    def audio_pipeline(wav):
+        # Speed Perturb is done here so it is multi-threaded with the
+        # workers of the dataloader (faster).
```
asumagic, line 241: It isn't?

asumagic (changes requested):
LGTM, nice! I've only made a superficial review, though. With those points fixed, if the recipe tests pass, it's ok to me.

xxchauncey:
> @whettenr thank you for your contribution. I tried reproducing your results from [this paper](https://arxiv.org/pdf/2405.04296). I followed your pretraining implementation from [here](https://github.com/whettenr/speechbrain/blob/f084239b5792baf8e993e5a57b84bc4de459a09a/recipes/BEST-RQ/train.py) using the Branchformer configuration. The only differences in my setup were increasing the batch size from 100 to 1000 and to match your setup I reduced the warmup steps from 25000 to 2500. After pretraining, I froze the model and used all hidden representation with linear combination following the SUPURB CTC downstream task and trained a single-layer BiLSTM on 100 Libri, but I got PER 65 on test-clean, which is significantly worse than the WER reported in the paper. Do you have any ideas on what might be causing this issue? I also tried Hubert in same setup and the resulted PER is 7.

hi @AmirHussein96 ,

Have you tried finetuning only with a linear and CTC loss like HuBERT, does it work?

TParcollet:
@xxchauncey it does work, you have a recipe to do that in this PR. It will get merged soon. We will train the models and get the results soon-ish as well.

speechbrain/lobes/models/transformer/TransformerASR.py
TParcollet: I agree... but I somehow don't want to do this now... what do you suggest?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -174,6 +174,7 @@ def __init__(
                     attention_type=self.attention_type,
                     kdim=self.encoder_kdim,
                     vdim=self.encoder_vdim,
+                    output_hidden_states=self.output_hidden_states,
```
asumagic, line 177: layerdrop as well

TParcollet: Stop bullying me. Fixed.

recipes/LibriSpeech/self-supervised-learning/BEST-RQ/train.py
TParcollet: Oui Sir.

speechbrain/lobes/models/transformer/TransformerASR.py
asumagic: Procrastinate, yolo

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -518,6 +533,7 @@ def __init__(
         attention_type="regularMHA",
         ffn_type="regularFFN",
         ffn_cnn_kernel_size_list=[3, 3],
+        output_hidden_states=False,
```
asumagic, line 536: layerdrop is missing here now though :)

asumagic: (and in the code beneath)

TParcollet: Why you trolling me :'( It is not! Just not at the same place ! Look a few lines above.

asumagic: oops

asumagic (approved):

TParcollet:
@asumagic I added a few docstring test for the hidden_state thing, but i guess you are right, we don't even have a single unit test for the transformerASR models outside of docstrings tests... maybe we should have a PR to fix that.

karynaur:
yayyy

PR #2315, opened 2023-12-27
New README file

mravanelli:
The current README.md is too verbose, not fully updated, and not really clean, elegant, and fancy. With this PR, I will propose a new version to use for SpeechBrain 1.0.


README.md
```diff
@@ -2,287 +2,262 @@
...
 
-# Running an experiment
-In SpeechBrain, you can run experiments in this way:
+## 🏃‍♂️ Running an Experiment
 
+In SpeechBrain, you can train a model for any task using the following steps:
+
+```bash
+cd recipes/<dataset>/<task>/
+python experiment.py params.yaml
```
Adel-Moumen, line 129: It would be great to have the Python highlight instead of bash. Maybe adding one line saying that SpeechBrain is only relying on torch/python would be great.

PR #2331, opened 2024-01-12
Fix weight_norm import in DAC

lucadellalib:
`from torch.nn.utils.parametrizations import weight_norm` does not work for older versions of PyTorch supported by SpeechBrain (e.g. `torch==1.13.1`). `from torch.nn.utils import weight_norm` should be used instead. However this throws a deprecation warning for newer versions. This PR fixes the issue.


Adel-Moumen (approved):
LGTM! Thanks :)

PR #2333, opened 2024-01-12
Fix zero_grad when opt_class is None

lucadellalib:
`zero_grad` breaks if `opt_class=None` because `self.optimizers_dict` is not defined. This PR fixes the issue.

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1067,9 +1067,11 @@ class Brain:
         Setting gradients to None should save the memory, e.g.
         during ``evaluate()`` and thus larger batch might be used.
         """
-        if self.opt_class is not None:
+        if self.optimizers_dict is not None:
             for opt in self.freeze_optimizers(self.optimizers_dict).values():
                 opt.zero_grad(set_to_none=set_to_none)
+        elif self.opt_class is not None:
+            self.optimizer.zero_grad(set_to_none=set_to_none)
 
     def on_evaluate_start(self, max_key=None, min_key=None):
         """Gets called at the beginning of ``evaluate()``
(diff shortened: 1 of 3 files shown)


mravanelli:
@Adel-Moumen, could you please take a look?

Adel-Moumen:
> `zero_grad` breaks if `opt_class=None` because `self.optimizers_dict` is not defined. This PR fixes the issue.

In which case do we have `opt_class=None` and `self.optimizers_dict=None` at the same time ?

Adel-Moumen:
I added some checks to deal with the case `opt_class is not None` and `self.optimizers_dict is None`. I also fixed a painful import message from kmeans. I'm awaiting for a reply by Luca regarding my previous message, and then, I will merge it.

lucadellalib:
The problem is that `self.optimizers_dict` is defined only if `self.opt_class` is not `None` https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/core.py#L1052:
```python
if self.opt_class is not None:
    if self.remove_vector_weight_decay:
        all_params = rm_vector_weight_decay(self.modules)

    self.optimizer = self.opt_class(all_params)

    self.optimizers_dict = {"opt_class": self.optimizer}

    if self.checkpointer is not None:
        self.checkpointer.add_recoverable("optimizer", self.optimizer)
```

Hence if `self.opt_class=None`, when you call `zero_grad` it errors out because the variable `self.optimizers_dict` doesn't exist. I think the latest commits you made address the issue.

Using a `Brain` without passing `opt_class` can be useful in some advanced use cases (e.g. in continual learning when you want to compute some gradient statistics over the whole dataset without performing any parameter update).

Adel-Moumen:
> The problem is that `self.optimizers_dict` is defined only if `self.opt_class` is not `None` https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/core.py#L1052:
> 
> ```python
> if self.opt_class is not None:
>     if self.remove_vector_weight_decay:
>         all_params = rm_vector_weight_decay(self.modules)
> 
>     self.optimizer = self.opt_class(all_params)
> 
>     self.optimizers_dict = {"opt_class": self.optimizer}
> 
>     if self.checkpointer is not None:
>         self.checkpointer.add_recoverable("optimizer", self.optimizer)
> ```
> 
> Hence if `self.opt_class=None`, when you call `zero_grad` it errors out because the variable `self.optimizers_dict` doesn't exist. I think the latest commits you made address the issue.
> 
> Using a `Brain` without passing `opt_class` can be useful in some advanced use cases (e.g. in continual learning when you want to compute some gradient statistics over the whole dataset without performing any parameter update).

OK I see... I tried to reproduce your example on the Transducer baseline, by removing init_optimizers and opt_class, it creates an issue on the side of `optimizers_step` because we do expect an optimiser. I added a check, if we do not have opt_dict and opt_class then we simply return from this function. What do you think? Could you please let me know if now everything works on your end ?

Adel-Moumen (approved):
lgtm

PR #2352, opened 2024-01-24
change _assert_len warning to debug

Adel-Moumen:
## What does this PR do?

This PR solves one issue with our TextEncoder in which we are spammed by warnings related to _assert_len. I proposed to switch to `logger.debug` instead of `warning.warns` as this message is only meaningful when you are trying to debug your model. 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/dataio/encoder.py
+++ b/speechbrain/dataio/encoder.py
@@ -717,8 +717,8 @@ class CategoricalEncoder:
                     f".expect_len({self.expected_len}) was called, "
                     f"but {real_len} categories found"
                 )
-        else:
-            warnings.warn(
+        else: 
+            logger.debug(
                 f"{self.__class__.__name__}.expect_len was never called: "
                 f"assuming category count of {len(self)} to be correct! "
                 "Sanity check your encoder using `.expect_len`. "
--- a/speechbrain/dataio/encoder.py
+++ b/speechbrain/dataio/encoder.py
@@ -9,7 +9,6 @@ import torch
 import collections
 import itertools
 import logging
-import warnings
 import speechbrain as sb
 from speechbrain.utils.checkpoints import (
     mark_as_saver,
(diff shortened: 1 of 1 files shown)


Adel-Moumen:
> > This PR solves one issue with our TextEncoder in which we are spammed by warnings related to _assert_len. I proposed to switch to logger.debug instead of warning.warns as this message is only meaningful when you are trying to debug your model.
> 
> Have there been cases where the message was incorrectly printed more than once? My understanding of our logging setup/the `warnings` module is that the warning [should be displayed at most once per source code location](https://docs.python.org/3/library/warnings.html#the-warnings-filter).
> 
> This is not the case of `logging.debug`, which will always print the message every time. Considering our logging setup is debug by default, I think it might exacerbate the issue that e.g. Mirco has encountered when using recipes that don't use the length assert mechanism.

Hello, @mravanelli reported to me that he had his terminal flooded by warnings due to this code. What I proposed is to swap from `warnings.warn` to `logger.debug` so that the message will only be displayed when entering in debug mode (which is what people should do when facing issues such as nan with SB).

asumagic:
Oh, I read [`log-config.yaml`](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/log-config.yaml) wrong. We default to `DEBUG` in the log files but the console filter is `INFO`.

I think that's fine, but I think it might be worth mentioning in the docs somewhere, if we actually have a page for training-related tips.

Adel-Moumen:
> Oh, I read [`log-config.yaml`](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/log-config.yaml?rgh-link-date=2024-01-25T11%3A01%3A53Z) wrong. We default to `DEBUG` in the log files but the console filter is `INFO`.
> 
> I think that's fine, but I think it might be worth mentioning in the docs somewhere, if we actually have a page for training-related tips.

I don't think we do have that :p

asumagic (approved):

Issue #2225, opened 2023-10-31
Pytorch 2.1, Requirements.txt and "RuntimeError: Couldn't find appropriate backend to handle uri".


jonah-ramponi:
### Describe the bug

Unless I am missing something, the requirements.txt file simply specifies

"torch>=1.9.0
torchaudio>=1.9.0"

However the majority of the tests fail using pytorch 2.1 due to "RuntimeError: Couldn't find appropriate backend to handle uri".

I am not sure why you recommend in the readme to use pytorch 1.13, suggest in the readme that it might work for pytorch 2.0, but then specify the requirements in this way. I think the requirements.txt or readme should be updated. 

After following the installiation steps, I had to run pip install torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1

### Expected behaviour

I expected the installation to work, to get it ti work I had to downgrade the pytorch version.

jonah-ramponi:
Hey,

Interesting. I'm using Linux, and the main version. Downgrading pytorch to 1.13 worked.  Before, 17 of the tests failed due to "RuntimeError: Couldn't find appropriate backend to handle uri".  Perhaps there was something up with my pytorch install! I suspect that is the case. Thanks anyways, my bad.

mravanelli:
I cannot replicate this on my side. Can you replicate it?

TParcollet:
Yes, the only solution was to downgrade torch and torchaudio to 2.0 ... Hopefully it will be fixed in a upcoming release but in the meantime I suggest that we freeze the torch version ...

mravanelli:
I suggest doing an investigation as soon as possible to understand the
cause of it. Unfortunately, I cannot help as on my side every works fine.

TParcollet:
Ok, so according to the following screen from the torchaudio 2.1.0 release, it looks like it's again a backend issue. 
![image](https://github.com/speechbrain/speechbrain/assets/11910731/bf828941-706b-4a5a-9098-148720ff48fb)

TParcollet:
@jonah-ramponi could you please try changing "soundfile" to ffmpeg after updating it and letting us know if it solved your issue? https://github.com/speechbrain/speechbrain/blob/bd27e99119edfe6b4f530be16d46cce28af260b3/speechbrain/utils/torch_audio_backend.py#L19

TParcollet:
[...]

PR #2354, opened 2024-01-25
Load audio with PyAV

pplantinga:
Torchaudio depends on one of a few backends being installed, which has caused a number of problems. See #2225 

This PR solves it by using PyAV to load the audio, because it bundles ffmpeg and makes for easy install.

https://pyav.org/docs/stable/

This is the approach taken by faster whisper (see https://github.com/SYSTRAN/faster-whisper/blob/master/faster_whisper/audio.py#L1)

### DIFF
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -10,7 +10,10 @@ Authors
  * Gaelle Laperriere 2021
  * Sahar Ghannay 2021
  * Sylvain de Langen 2022
+ * Peter Plantinga 2024
 """
+import av
+import io
 import os
 import torch
 import logging
@@ -26,6 +29,50 @@ from speechbrain.utils.torch_audio_backend import check_torchaudio_backend
 
 check_torchaudio_backend()
 logger = logging.getLogger(__name__)
+
+
+def load_audio_pyav(uri, num_frames=-1, frame_offset=0, normalize=True):
+    """Mimics torchaudio.load using PyAV which bundles ffmpeg.
+
+    Arguments
+    ---------
+    uri: str
+        Path to file
+    num_frames: int
+        Number of frames to load. Default (-1) loads all frames
+    frame_offset: int
+        Frame from which to start loading. Default (0) loads all frames
+    normalize: bool
+        Whether to normalize output to float32 if int32
+    """
+
+    raw_buffer = io.BytesIO()
+    dtype = None
+    sample_rate = None
+    channels = None
+
+    with av.open(uri, mode="r", metadata_errors="ignore") as container:
+        for frame in container.decode(audio=0):
+            sample_rate = frame.rate
+            layout = frame.layout
+            channels = len(layout.channels)
+            array = frame.to_ndarray()
+            dtype = array.dtype
+
+            raw_buffer.write(array)
+
+    audio = np.frombuffer(raw_buffer.getbuffer(), dtype=dtype)
+    audio = np.reshape(audio, [channels, -1], order="F")
+
+    if dtype.kind == "i" and normalize:
+        audio = audio.astype(np.float32) / np.iinfo(dtype).max
+
+    if num_frames >= 0:
+        audio = audio[:, frame_offset : frame_offset + num_frames]
+    elif frame_offset > 0:
+        audio = audio[:, frame_offset:]
+
+    return torch.tensor(audio), sample_rate
 
 
 def load_data_json(json_path, replacements={}):
@@ -210,7 +257,7 @@ def read_audio_info(path) -> "torchaudio.backend.common.AudioMetaData":
     # double-checking anyway. If I am wrong and you are reading this comment
     # because of it: sorry
     if info.num_frames == 0:
-        channels_data, sample_rate = torchaudio.load(path, normalize=False)
+        channels_data, sample_rate = load_audio_pyav(path, normalize=False)
 
         info.num_frames = channels_data.size(1)
         info.sample_rate = sample_rate  # because we might as well
@@ -236,9 +283,6 @@ def read_audio(waveforms_obj):
         "stop": 16000
     })
     ```
-
-    Which codecs are supported depends on your torchaudio backend.
-    Refer to `torchaudio.load` documentation for further details.
 
     Arguments
     ----------
@@ -272,7 +316,7 @@ def read_audio(waveforms_obj):
     True
     """
     if isinstance(waveforms_obj, str):
-        audio, _ = torchaudio.load(waveforms_obj)
+        audio, _ = load_audio_pyav(waveforms_obj)
     else:
         path = waveforms_obj["file"]
         start = waveforms_obj.get("start", 0)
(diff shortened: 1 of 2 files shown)


Adel-Moumen:
I tried this PR to load `.wav` / `.flac` / `.mp3` and it works as intended. There's only one warning that we should pay attention to when loading mp3: `WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate`. This will flood your console when you are looping over a lot of mp3 files (e.g. common voice): 
```
WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate
WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate
... 
WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate
WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate
```

To reproduce:
```
wget -O example_rw.mp3 "https://www.dropbox.com/scl/fi/iplkymn8c8mbc6oclxem3/example_rw.mp3?rlkey=yhmqfsn8q43pmvd1uvjo3yl0s&dl=1"
```
and then:
```python
for _ in range(10):
  load_audio_pyav("example_rw.mp3", normalize=False)[0]
```

I will try to use pyav on common voice and see how it behaves compared to Torchaudio since creating a new conda env in my cluster may lead to issues with torchaudio and mp3.

speechbrain/dataio/dataio.py
```diff
@@ -28,6 +31,50 @@
...
+
+    if dtype.kind == "i" and normalize:
+        audio = audio.astype(np.float32) / np.iinfo(dtype).max
+
+    if num_frames >= 0:
+        audio = audio[:, frame_offset : frame_offset + num_frames]
+    elif frame_offset > 0:
+        audio = audio[:, frame_offset:]
+
+    return torch.tensor(audio), sample_rate
```
Adel-Moumen, line 75: I think using `torch.from_numpy` method instead of `torch.tensor` would be better (in terms of readability/torch.from_numpy infers the dtype of your audio file while torch.tensor is an alias of torch.FloatTensor)

speechbrain/dataio/dataio.py
```diff
@@ -28,6 +31,50 @@
...
+
+            raw_buffer.write(array)
+
+    audio = np.frombuffer(raw_buffer.getbuffer(), dtype=dtype)
+    audio = np.reshape(audio, [channels, -1], order="F")
+
+    if dtype.kind == "i" and normalize:
+        audio = audio.astype(np.float32) / np.iinfo(dtype).max
+
+    if num_frames >= 0:
```
asumagic, line 70: Not sure how much of a proof-of-concept this currently is but I'll leave some comments anyway.

Would it be possible to implement `frame_offset`/`num_frames` without needing to decode the entire file? Certain datasets I have encountered have long files with utterances representing a very small segment of the file.

speechbrain/dataio/dataio.py
```diff
@@ -28,6 +31,50 @@
 logger = logging.getLogger(__name__)
 
 
+def load_audio_pyav(uri, num_frames=-1, frame_offset=0, normalize=True):
```
asumagic, line 34: Maybe we could have a more generic `load_audio` function in SB that is not necessarily tied to PyAV and could ultimately be made to support different backends (e.g. still allowing to use torchaudio if desired)?

In theory PyAV should already sort of cover our usecases but it would still be more future-proof, I think.

speechbrain/dataio/dataio.py
```diff
@@ -210,7 +257,7 @@ def read_audio_info(path) -> "torchaudio.backend.common.AudioMetaData":
     # double-checking anyway. If I am wrong and you are reading this comment
     # because of it: sorry
     if info.num_frames == 0:
-        channels_data, sample_rate = torchaudio.load(path, normalize=False)
+        channels_data, sample_rate = load_audio_pyav(path, normalize=False)
```
asumagic, line 260: If using PyAV directly most of this function can be replaced, and reading the entire audio should no longer be necessary. We probably should make it return our own metadata dataclass rather than returning the library's own metadata, though.

speechbrain/dataio/dataio.py
Adel-Moumen: I do like this. In lhoste they have a function that switch the backend if one fails:
```
@lru_cache(maxsize=1)
def get_default_audio_backend() -> "AudioBackend":
    """
    Return a backend that can be used to read all audio formats supported by Lhotse.

    It first looks for special cases that need very specific handling
    (such as: opus, sphere/shorten, in-memory buffers)
    and tries to match them against relevant audio backends.

    Then, it tries to use several audio loading libraries (torchaudio, soundfile, audioread).
    In case the first fails, it tries the next one, and so on.
    """
    return CompositeAudioBackend(
        [
            # First handle special cases: OPUS and SPHERE (SPHERE may be encoded with shorten,
            #   which can only be decoded by binaries "shorten" and "sph2pipe").
            FfmpegSubprocessOpusBackend(),
            Sph2pipeSubprocessBackend(),
            # New FFMPEG backend available only in torchaudio 2.0.x+
            TorchaudioFFMPEGBackend(),
            # Prefer libsndfile for in-memory buffers only
            LibsndfileBackend(),
            # Torchaudio should be able to deal with most audio types...
            TorchaudioDefaultBackend(),
            # ... if not, try audioread...
            AudioreadBackend(),
            # ... oops.
        ]
    )
```

see: https://github.com/lhotse-speech/lhotse/blob/master/lhotse/audio/backend.py
 
We could have something more generic like that.

TParcollet: I am against this option. More backend means more maintenance. We already had to manage one until now, it's already causing a mess.

speechbrain/dataio/dataio.py
TParcollet: What do you mean?

TParcollet:
I will try it. Something important to do would be to benchmark the reading speed of this lib. Some backend of torchaudio were awfully slow for instance leading to slower training times.

speechbrain/dataio/dataio.py
pplantinga: I think I'm with Titouan on this. Perhaps its worth noting that users are free to use lhotse with speechbrain if they want the more flexible audio loading.

speechbrain/dataio/dataio.py
pplantinga: I did think about this, and its probably important to have it, but I'm not sure how straightforward it is to do it for compressed audio files.

speechbrain/dataio/dataio.py
asumagic: Most of this function was written to work around `torchaudio.info` being broken with certain mp3 files. Since we're moving to PyAV, the workarounds become unnecessary. In any case, Peter was probably already aware this function should be rewritten with PyAV anyway.

What I mean by the second sentence is that IMO we should make our own dataclass with the same fields as `torchaudio.info` provided for compatibility and to not have the API explicitly refer to PyAV.

asumagic:
> > I tried this PR to load `.wav` / `.flac` / `.mp3` and it works as intended. There's only one warning that we should pay attention to when loading mp3: `WARNING:libav.mp3:Estimating duration from bitrate, this may be inaccurate`.
> 
> I have seen this warning before, but as far as I know it can be ignored. Perhaps we can find a way to suppress it.

The problem is that the warning shows on every call so having the user ignore it is not really an option. IMO the warning underlines a valid concern so the user should still be made aware of it in some way.  
A single warning with the rest of them suppressed would seem fair to me, but I don't know if others agree.

mravanelli:
I think we can close this PR as we decided to not support for now PyAV as it is slower than torchaudio. Free free to reopen it in the future if you find the integration with PyAV valuable.

PR #2377, opened 2024-01-31
Streaming ASR interfaces

asumagic:
## What does this PR do?

The goal of this PR is to provide high-level interfaces for streaming inference using the model introduced by #2140, and to provide some shared infrastructure for future streaming models.

I have hosted a LibriSpeech streaming ASR model at <https://huggingface.co/sdelangen/speechbrain-asr-conformer-test/tree/main>. There is currently no model card, but it will be possible to migrate this model under the speechbrain org once this is merged.

### Introduced interfaces

The `StreamingASR` inference interface was introduced.

Features:

- Streaming decode of long audio files (if ffmpeg is installed), see demos
- Streaming decode of live streams (if ffmpeg is installed), see demos
- Chunkwise decoding support (with an exposed lower-level `transcribe_chunk`/`decode_chunk` API)
- Batching support for the lower-level interfaces which makes it suitable as a basis for transcription servers
- Relatively model agnostic

<hr>

### Feature extraction
In the current conformer-transducer model, we have some feature extraction logic that **stacks filter banks and 2 layers of down-sampling convolutions**.
Ignoring normalization, we can consider feature extraction as a **"filter"** with a specific **window size** and **stride**.

Dynamic chunk training does *not* change how feature extraction is performed: it is applied over the *entire batch* at once, and is **unaware of chunking**.
This is mostly for simplicity, flexibility and training performance reasons.

Thus, when streaming, we have to be careful about how we chunk up our signal: for any given chunk, we must provide some past and future frames so that the transition across chunks remains as close to training as possible.

#### Streaming feature extraction abstraction
This logic is abstracted by `speechbrain.lobes.features.StreamingFeatureWrapper` as introduced by this PR. When provided the "filter properties" of our feature extractor (window size and stride), it is able to process a stream of audio chunk by chunk while abstracting padding and left context saving details¹.

#### Automatically determining filter properties
Those filter properties remain to be provided by the model developer. For this, we introduce `speechbrain.utils.filter_analysis`. It allows modelling simple filter properties, and provides utilities to "stack" them, i.e. model what the properties of `filter2(filter1(x))` are.

In our case, we are essentially trying to model `conv_layer_2(conv_layer_1(fbanks(waveform)))`².

This allows us to determine the filter properties of our feature extraction easily and requires little to no adjustment when changing feature extraction hyperparameters.
This PR introduces a `get_filter_propreties` method for some modules, so all that remains is providing `StreamingFeatureWrapper` our feature extractor module list.

¹ This comes with a latency cost which depends on the window size and stride. However, that overhead is generally insignificant in practice as long as the window size and stride are kept reasonable.

² ignoring normalization. Not considering normalization specifically in our chunking process does not appear to have a significant effect on WER.

<hr>

### Example programs

<details>
<summary>Commandline tool to transcribe a file or a live stream</summary>

**Decoding from a live stream using ffmpeg (BBC Radio 4):** `python3 asr.py http://as-hls-ww-live.akamaized.net/pool_904/live/ww/bbc_radio_fourfm/bbc_radio_fourfm.isml/bbc_radio_fourfm-audio%3d96000.norewind.m3u8 --model-source=sdelangen/speechbrain-asr-conformer-test --device=cpu -v`

**Decoding from a file:** `python3 asr.py some-english-speech.wav --model-source=sdelangen/speechbrain-asr-conformer-test --device=cpu -v`

```python
from argparse import ArgumentParser
import logging

parser = ArgumentParser()
parser.add_argument("audio_path")
parser.add_argument("--model-source", required=True)
parser.add_argument("--device", default="cpu")
parser.add_argument("--ip", default="127.0.0.1")
parser.add_argument("--port", default=9431)
parser.add_argument("--chunk-size", default=24, type=int)
parser.add_argument("--left-context-chunks", default=4, type=int)
parser.add_argument("--num-threads", default=None, type=int)
parser.add_argument("--verbose", "-v", default=False, action="store_true")
args = parser.parse_args()

if args.verbose:
    logging.getLogger().setLevel(logging.INFO)

logging.info("Loading libraries")

from speechbrain.inference.ASR import StreamingASR
from speechbrain.utils.dynamic_chunk_training import DynChunkTrainConfig
import torch

device = args.device

if args.num_threads is not None:
    torch.set_num_threads(args.num_threads)

logging.info(f"Loading model from \"{args.model_source}\" onto device {device}")

asr = StreamingASR.from_hparams(args.model_source, run_opts={"device": device})
config = DynChunkTrainConfig(args.chunk_size, args.left_context_chunks)

logging.info(f"Starting stream from URI \"{args.audio_path}\"")

for text_chunk in asr.transcribe_file_streaming(args.audio_path, config):
    print(text_chunk, flush=True, end="")
```
</details>

### ~~TODOs~~

<details>
<summary>Completed</summary>

- [x] Will need uploading the model and inference hparams to huggingface. (~~I probably should be made part of the SB team on HF?~~ done)
- [x] Currently, some stuff around streaming state initialization is essentially hardcoded inside of the interfaces code, which should not need to care about MHA left context etc., which should be the responsibility of the hparams in some way.
- [x] Really need to double/triple-check the `StreamingASR` interface for flexibility as we might want to have several architectures be able to use it. -- I tried to make it so, but it's probably still biased towards the architecture I made it for. For what it's worth, it can be modified down the line and extended by any external users.
- [x] The sentencepiece space hack is very annoying and should probably be moved somewhere else, and the code ~~should be made to support more than just sentencepiece~~ -- the code was moved but it still only supports sentencepiece; but adding support for other tokenizers should otherwise not be too painful.
- [x] Audio loading should be done in a streaming fashion ~~-- might wait for #2354~~ probably will use torchaudio ffmpeg streaming stuff?
- [x] Figure out why the doc generation even fails... Probably a type annotation somewhere.
- [x] Final pass of checking all docstrings (and how they generate too)

</details>

Not sure how to handle things like rescoring that could change entire sentences, but let's consider it out-of-scope for `StreamingASR`.

<hr>

<details>

</details>

## PR review

<details>

- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

</details>


asumagic:
I believe this should be ready for review @TParcollet

docs/conf.py
```diff
@@ -93,20 +93,6 @@ def run_apidoc(app):
     import better_apidoc
 
     better_apidoc.APP = app
-
```
TParcollet, line 96: Why do we remove this?

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -399,3 +399,17 @@ error_rate_computer: !name:speechbrain.utils.metric_stats.ErrorRateStats
 
 cer_computer: !name:speechbrain.utils.metric_stats.ErrorRateStats
    split_tokens: True
+
+# for the inference hparams, you will need to include and uncomment something like this:
```
TParcollet, line 403: Should we have a better documentation on how to turn a trained model to streaming? Like a colab?

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
```diff
@@ -399,3 +399,17 @@ error_rate_computer: !name:speechbrain.utils.metric_stats.ErrorRateStats
...
+
+# make_tokenizer_streaming_context: !name:speechbrain.tokenizers.SentencePiece.SentencePieceDecoderStreamingContext
+# tokenizer_decode_streaming: !name:speechbrain.tokenizers.SentencePiece.spm_decode_preserve_leading_space
+
+# fea_streaming_extractor: !new:speechbrain.lobes.features.StreamingFeatureWrapper
+#    module: !new:speechbrain.lobes.models.convolution.ConformerFeatureExtractorWrapper
+#       - !ref <compute_features>
+#       - !ref <normalize>
+#       - !ref <CNN>
+#    # don't consider normalization as part of the input filter chain
```
TParcollet, line 413: Explain why

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,390 @@ def transcribe_batch(self, wavs, wav_lens):
     def forward(self, wavs, wav_lens):
         """Runs full transcription - note: no gradients through decoding"""
         return self.transcribe_batch(wavs, wav_lens)
+
+
+@dataclass
+class ASRStreamingContext:
+    """Streaming metadata for a streaming transducer model. Initialized by
```
TParcollet, line 489: Should we change the name to TransducerASRStreamingContext? ASR context sounds like anything can work, but then the documentation says it's transducer.

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,390 @@ def transcribe_batch(self, wavs, wav_lens):
...
+
+    Holds some references to opaque streaming contexts, so the context is
+    model-agnostic to an extent."""
+
+    config: DynChunkTrainConfig
+    """Dynamic chunk training configuration used to initialize the streaming
+    context. Cannot be modified on the fly."""
+
+    fea_extractor_context: Any
+    """Opaque feature extractor streaming context."""
```
TParcollet, line 503: This comment is not sufficient to understand what we are talking about. When reading the class, it's not clear to the user what should be passed here. Especially since there is no example of the class (maybe add an example of instanciation?)

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,390 @@ def transcribe_batch(self, wavs, wav_lens):
...
+                dynchunktrain_config
+            ),
+            decoder_hidden=None,
+            tokenizer_context=None,
+        )
+
+    def get_chunk_size_frames(
+        self, dynchunktrain_config: DynChunkTrainConfig
+    ) -> int:
+        """Chunk size in actual audio samples that the user should forward to
```
TParcollet, line 711: Unclear.

speechbrain/inference/ASR.py
```diff
@@ -13,13 +13,18 @@
  * Adel Moumen 2023, 2024
  * Pradnya Kandarkar 2023
```
TParcollet, line 14: Add yourself.

speechbrain/lobes/features.py
```diff
@@ -5,7 +5,9 @@
  * Peter Plantinga 2020
  * Sarthak Yadav 2020
```
TParcollet, line 6: Add yourself, credits deserved.

speechbrain/lobes/features.py
```diff
@@ -441,3 +448,146 @@ def _check_input_shape(self, shape):
...
+
+    assert x >= 0
+
+    if (x % to) == 0:
+        return x
+
+    return x + to - (x % to)
+
+
+@dataclass
```
TParcollet, line 465: Do we want some tests for all this code maybe? I feel like the forward should be tested.

speechbrain/lobes/models/convolution.py
```diff
@@ -223,19 +233,28 @@ def __init__(
...
                 kernel_size=kernel_size,
-                stride=stride if i == num_layers - 1 else 1,
+                stride=layer_stride,
                 dilation=dilation,
                 layer_name=f"conv_{i}",
                 bias=conv_bias,
                 padding=padding,
                 conv_init=conv_init,
             )
+            self.filter_properties.append(
```
TParcollet, line 251: Are these loadables? If so, won't they break previous checkpoints?

speechbrain/lobes/models/convolution.py
```diff
@@ -264,3 +283,71 @@ def forward(self, x):
             out = out + self.reduce_conv(x)
             out = self.drop(out)
         return out
+
+    def get_filter_properties(self) -> FilterProperties:
+        return stack_filter_properties(self.filter_properties)
+
+
+class ConformerFeatureExtractorWrapper(torch.nn.Module):
```
TParcollet, line 291: I don't like having this class. This may, in the future, imply a shit ton of complexity as it will grow with the number of supported model. Why do we need this? I don't see anything special about the forward.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -62,6 +62,10 @@ class ConformerEncoderLayerStreamingContext:
 class ConformerEncoderStreamingContext:
```
TParcollet, line 62: Will we need a definition of this context for every single class we want to use? Can't we just have a generic one? Overall, having model-dependent code in a toolkit that may have hundreds of models is not a good practice imho.

speechbrain/tokenizers/SentencePiece.py
```diff
@@ -484,3 +486,53 @@ def get_spm_tokens(model_path):
...
+
+    emitted_symbol_count: int = 0
+    """The number of symbols that have been emitted for this transcription."""
+
+
+def spm_decode_preserve_leading_space(
+    tokenizer: spm.SentencePieceProcessor,
+    hyps: List[int],
+    context: SentencePieceDecoderStreamingContext,
+) -> List[str]:
```
TParcollet, line 504: the -> operator makes streaming incompatible with older version of python, is that really ok? @mravanelli @Adel-Moumen

speechbrain/utils/filter_analysis.py
```diff
@@ -0,0 +1,225 @@
...
+        ), "Dilation must be >0. NOTE: a dilation of 1 means no dilation."
+
+    @staticmethod
+    def pointwise_filter() -> "FilterProperties":
+        """Returns filter properties for a trivial filter whose output frames
+        only ever depend on their respective input frame."""
+
+        return FilterProperties(window_size=1, stride=1)
+
+    def has_overlap(self):
```
TParcollet, line 108: remove?

TParcollet (changes requested):
I've tested the code, it works well and I am really happy to see real streaming ASR with SB. However, a few classes require our attention. In particular i'd like to invite @Adel-Moumen or @mravanelli to have a very quick review, because this is a very important feature for SpeechBrain and we want to get it right from the beginning to avoid a cascade of issues in the future.

speechbrain/lobes/models/transformer/Conformer.py
asumagic: I don't think it can be made generic unless you can think of a cleaner way. I defined the streaming context objects this way largely because the data that needs to be memorized is highly module-specific (it might not even be tensors), and because it needed to be recursive in some way.

Anything that requires to be aware of streaming needs to define what data they need to hold *at some point*, so I don't think it can theoretically be improved much.

If a module knows it never actually needs to save anything for streaming, then it doesn't need to actually define a context class. (And if a module that contains it calls into `.make_streaming_context` anyway, it can just return `None`.)

I think the ugliest part is carrying along the training configuration down the hierarchy, but that's also kind of the only thing that needs to be this way.

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,393 @@ def transcribe_batch(self, wavs, wav_lens):
...
+    >>> from speechbrain.inference.ASR import StreamingASR
+    >>> from speechbrain.utils.dynamic_chunk_training import DynChunkTrainConfig
+    >>> tmpdir = getfixture("tmpdir")
+    >>> asr_model = StreamingASR.from_hparams(source="speechbrain/asr-conformer-streaming-librispeech", savedir=tmpdir,) # doctest: +SKIP
+    >>> asr_model.transcribe_file("speechbrain/asr-conformer-streaming-librispeech/test-en.wav", DynChunkTrainConfig(24, 8)) # doctest: +SKIP
+    """
+
+    HPARAMS_NEEDED = [
+        "fea_streaming_extractor",
+        "Greedysearcher",
```
Adel-Moumen, line 531: is it only working with greedy searcher ?

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,393 @@ def transcribe_batch(self, wavs, wav_lens):
...
+        Arguments
+        ---------
+        path : str
+            URI/path to the audio to transcribe. When
+            ``use_torchaudio_streaming`` is ``False``, uses SB fetching to allow
+            fetching from HF or a local file. When ``True``, resolves the URI
+            through ffmpeg, as documented in
+            :class:`torchaudio.io.StreamReader`.
+        path : str
+            Path to the audio file to trancsribe.
```
Adel-Moumen, line 612: duplicate args

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,393 @@ def transcribe_batch(self, wavs, wav_lens):
...
+
+        if use_torchaudio_streaming:
+            streamer = torchaudio.io.StreamReader(path)
+            chunks = self._get_audio_stream(streamer, chunk_size)
+        else:
+            waveform = self.load_audio(path, **kwargs)
+            batch = waveform.unsqueeze(0)  # create batch dim
+            chunks = split_fixed_chunks(batch, chunk_size)
+
+        rel_length = torch.tensor([1.0])
```
Adel-Moumen, line 641: make me think about something but does it means that we cannot transcribe a batch of audio files ?

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,393 @@ def transcribe_batch(self, wavs, wav_lens):
...
+
+    def make_streaming_context(self, dynchunktrain_config: DynChunkTrainConfig):
+        """Create a blank streaming context to be passed around for chunk
+        encoding/transcription.
+
+        Arguments
+        ---------
+        dynchunktrain_config : DynChunkTrainConfig
+            Streaming configuration. Sane values and how much time chunks
+            actually represent is model-dependent."""
```
Adel-Moumen, line 697: could you please add the return docstring ?

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,393 @@ def transcribe_batch(self, wavs, wav_lens):
...
+        list of str
+            Decoded tokens of length `batch_size`. The decoded strings can be
+            of 0-length.
+        list of list of output token hypotheses
+            List of length `batch_size`, each holding a list of tokens of any
+            length `>=0`.
+        """
+        (tokens, _scores, _, _, h,) = self.hparams.Greedysearcher(
+            x, context.decoder_hidden, return_hidden=True
+        )
```
Adel-Moumen, line 809: hmm wondering but should it work with beam search as well ?

speechbrain/lobes/features.py
```diff
@@ -441,3 +449,155 @@ def _check_input_shape(self, shape):
...
+        This is not really an issue when streaming, and might not actually
+        matter in general as the number of affected frames would usually be low.
+
+        Arguments
+        ---------
+        chunk : torch.Tensor
+            Chunk of input of shape [batch size, time]; typically a raw waveform
+        context : StreamingFeatureWrapperContext
+            Mutable streaming context object; should be reused for subsequent
+            calls in the same streaming session.
```
Adel-Moumen, line 555: adding return docstring would be valuable

speechbrain/tokenizers/SentencePiece.py
Adel-Moumen: I think it's fine. We did it on others places as well and we didn't had any issues with it.

speechbrain/tokenizers/SentencePiece.py
```diff
@@ -484,3 +486,53 @@ def get_spm_tokens(model_path):
...
+
+    if len(proto.pieces) >= 1:
+        if context.emitted_symbol_count > 0 and proto.pieces[
+            0
+        ].piece.startswith(
+            "\u2581"
+        ):  # magic spm space
+            text = " " + text
+
+        context.emitted_symbol_count += len(proto.pieces)
```
Adel-Moumen, line 536: not sure to understand what is happening here

Adel-Moumen (commented):
Thanks @asumagic! I left some comments. 

I'd like to better understand about the usability of this new streaming ASR interface. For instance, If I'd like to use a CTC based model. How can I run streaming inference with it ? What needs to be changed in order to make everything works ?

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
TParcollet: Right, but in the meantime, we might want to make it clear how to do this somewhere.

speechbrain/lobes/models/convolution.py
TParcollet: I still understand the point of it. Could you elaborate?

speechbrain/lobes/models/transformer/Conformer.py
TParcollet: I changed my mind on this, it's ok.

speechbrain/tokenizers/SentencePiece.py
TParcollet: Yeah, a bit hard to follow./

TParcollet (changes requested):
@asumagic can you run a WER validation as well to verify that this does not alter the expected WER?

speechbrain/inference/ASR.py
asumagic: Currently, no, but the equivalent non-streaming interfaces also don't support this usecase.

If an user wants to decode a bunch of files (especially of varying lengths) at once, the naive approach of using a single batch seems really impractical:

- You might want to decode more files than you can fit in a single batch due to VRAM
- You don't want to be continuously inferring on a batch where most streams are empty because we're dealing with one especially long file

Solving the latter problem also solves the issue of handling an arbitrary number of incoming streams with batching (allowing to stop and create new ones on the fly). This requires some amount of manipulation of the streaming context tensors.

This would need some more work but I probably will have to look into it sometime, but I didn't want the added complexity for this PR.

speechbrain/lobes/models/convolution.py
asumagic: The point is that I don't see how to abstract this only in the YAML. The alternative would be to make `StreamingASR` aware of how to call the Conformer feature extractor which would unnecessarily be model-specific.

These individual submodules are specific to the Conformer so forcing other models to use those would seem wrong.  
I don't think it can simply be defined as a module list in the YAML because of how `lens`, etc. need to be passed around for only some of these modules...

speechbrain/tokenizers/SentencePiece.py
asumagic: Clarified docstring and added comments.

recipes/LibriSpeech/ASR/transducer/hparams/conformer_transducer.yaml
asumagic: Added as a notice in the README.

speechbrain/lobes/models/convolution.py
TParcollet: If I am not mistaking we have a length capable sequential just for that, that can be used in the yaml. @pplantinga might know more. This wrapper really looks unnecessary :-(

asumagic: > If I am not mistaking we have a length capable sequential just for that, that can be used in the yaml. @pplantinga might know more. This wrapper really looks unnecessary :-(

Actually, yeah, this works. I had ruled it out for a reason I forgot was fixed (the default `epoch=None` param). LengthsCapableSequential seems to work, fixing.

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,391 @@ def transcribe_batch(self, wavs, wav_lens):
...
+    >>> from speechbrain.inference.ASR import StreamingASR
+    >>> from speechbrain.utils.dynamic_chunk_training import DynChunkTrainConfig
+    >>> tmpdir = getfixture("tmpdir")
+    >>> asr_model = StreamingASR.from_hparams(source="speechbrain/asr-conformer-streaming-librispeech", savedir=tmpdir,) # doctest: +SKIP
+    >>> asr_model.transcribe_file("speechbrain/asr-conformer-streaming-librispeech/test-en.wav", DynChunkTrainConfig(24, 8)) # doctest: +SKIP
+    """
+
+    HPARAMS_NEEDED = [
+        "fea_streaming_extractor",
+        "Greedysearcher",
```
Adel-Moumen, line 531: btw, in a future we will have to refactor this. For instance, if we want to use a CTC-based model, which could be the case since we will have a competitive CTC Branch/Conformer recipe on LibriSpeech, then one could use a beam search or greedy searcher in an online fashion. Maybe changing the naming to `decoding_function` or smth like that would be better.

speechbrain/inference/ASR.py
```diff
@@ -477,3 +482,391 @@ def transcribe_batch(self, wavs, wav_lens):
...
+        list of str
+            Decoded tokens of length `batch_size`. The decoded strings can be
+            of 0-length.
+        list of list of output token hypotheses
+            List of length `batch_size`, each holding a list of tokens of any
+            length `>=0`.
+        """
+        (tokens, _scores, _, _, h,) = self.hparams.Greedysearcher(
+            x, context.decoder_hidden, return_hidden=True
+        )
```
Adel-Moumen, line 807: I need to think more but maybe we will need to override this class for something like `StreamingCTCASR` so that we can write the CTC decoding logic.

asumagic:
I tested WER on test-clean with custom code and ended up with the same score as advertised (when streaming with a chunk size of 24, left context of 8).

<details>
<summary>Jank test script for reference</summary>

```python
device = "cuda"
model_path = "./librispeech-conformer"
test_csv = "/users/sdelangen/sb/recipes/LibriSpeech/ASR/transducer/results/conformer_transducer_large/3407/test-clean.csv"

import pandas as pd

test = pd.read_csv(test_csv)

import torch
torch.set_num_threads(1)

from concurrent.futures import ThreadPoolExecutor
from speechbrain.inference.ASR import StreamingASR
from speechbrain.utils.dynamic_chunk_training import DynChunkTrainConfig
from speechbrain.utils.metric_stats import ErrorRateStats
from speechbrain.utils.parallel import parallel_map

asr = StreamingASR.from_hparams(model_path, run_opts={"device": device})
config = DynChunkTrainConfig(24, 8)

wer_compute = ErrorRateStats()

def process_line(row):
    i, line = row

    path = line["wav"]

    truth = line["wrd"]
    pred = asr.transcribe_file(path, config, use_torchaudio_streaming=True)

    return pred, truth
    

for i, (pred, truth) in enumerate(parallel_map(process_line, test.iterrows(), executor=ThreadPoolExecutor(20), chunk_size=1)):
    wer_compute.append([i], [pred.split(" ")], [truth.split(" ")])
    print(wer_compute.summarize()["WER"])

# for text_chunk in asr.transcribe_file_streaming(args.audio_path, config):
#     print(text_chunk, flush=True, end="")
```

</details>

Still fixing some stuff as suggested then I will re-request a review once done.

Adel-Moumen (commented):
I would say that I have nothing much to say now. You answered all of my concerns. Could you please add the HF interface in private on the speechbrain HF hub ?

My only remark will be on the documentation part, I think we really need in the future (incoming weeks) to write good and high quality tutorials/etc to explain exactly what is needed to be changed if one wants to reuse the streamingASR interface. 

Other than that, let's wait from @TParcollet POV on that :) (bug congrats this is a big thing for SB, a lot of people (industrials/researchers) were asking for this feature!).

asumagic:
As for the docs, I've started writing some stuff but it will take some time. I already had some model details explained from the original PR that I will reuse.

I am also starting to write a simple `gradio` demo, which I believe is supposed to allow streaming ASR. Theoretically, it would be possible to use this as an HF space.

mravanelli:
Thank you @asumagic, this is a great contribution. I ran recipe tests and everything seems fine. Thank you @Adel-Moumen and @TParcollet for the review.

mravanelli (approved):

PR #2378, opened 2024-01-31
Fix EncoderASR interface that was very restrictive 

Adel-Moumen:
## What does this PR do?

This PR fix the `EncoderASR` class which was forcing the users to add a new parameter in our HF inference YAMLs `decoding_type`. Furthermore, this parameter was not letting the possibility to the end user to customise easily the type of beam search used, the hyperparams etc. 

Now, we do not need this parameter. We can automatically retrieve if it's greedy search or beam search. You can either use greedy search like that:
```yaml
decoding_function: !name:speechbrain.decoders.ctc_greedy_decode
    blank_id: !ref <blank_index>
```
Or beam search:
```yaml
decoding_function: !name:speechbrain.decoders.CTCBeamSearcher
# Decoding parameters
test_beam_search:
   beam_size: 143
   topk: 1
   space_token: ' ' # make sure this is the same as the one used in the tokenizer
   beam_prune_logp: -12.0
   token_prune_min_logp: -1.2
   prune_history: True
   alpha: 0.8
   beta: 1.2
   # can be downloaded from here https://www.openslr.org/11/ or trained with kenLM
   # It can either be a .bin or .arpa ; note: .arpa is much slower at loading
   # If you don't want to use an LM, comment it out or set it to null
   kenlm_model_path: null
```

Furthermore, for a mysterious reason, the class was expecting Sentencepiece tokenizer while most of our models expect CTCTextEncoder... So this PR is also solving this issue by checking the tokenizer before retrieving the vocab_list. 

Finally, this PR adds a small check that the input log_probs are effectively log_probs. I found that in wav2vec2 HF models, we were not passing the log softmax in the encoder which means that the logits (and not log probs) were given to the beam search decoder and ultimately led to wrong results!

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/decoders/ctc.py
+++ b/speechbrain/decoders/ctc.py
@@ -4,7 +4,7 @@ Authors
  * Mirco Ravanelli 2020
  * Aku Rouhe 2020
  * Sung-Lin Yeh 2020
- * Adel Moumen 2023
+ * Adel Moumen 2023, 2024
 """
 from itertools import groupby
 from speechbrain.dataio.dataio import length_to_mask
@@ -966,7 +966,9 @@ class CTCBaseSearcher(torch.nn.Module):
             )
 
         # check if we have log_probs
-        if not torch.allclose(torch.exp(log_probs).sum(dim=-1), torch.ones(log_probs.shape[0])):
+        if not torch.allclose(
+            torch.exp(log_probs).sum(dim=-1), torch.ones(log_probs.shape[0])
+        ):
             warnings.warn(
                 "The input `log_probs` are not log probabilities. "
                 "Going to convert them to log probabilities."
--- a/speechbrain/decoders/ctc.py
+++ b/speechbrain/decoders/ctc.py
@@ -965,6 +965,14 @@ class CTCBaseSearcher(torch.nn.Module):
                 "During decoding, going to truncate the log_probs vocab dim to match vocab_list."
             )
 
+        # check if we have log_probs
+        if not torch.allclose(torch.exp(log_probs).sum(dim=-1), torch.ones(log_probs.shape[0])):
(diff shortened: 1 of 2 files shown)


Adel-Moumen:
I tried it and got the expected output on all decoders (greedy, beam search, beam search + ngram):
CTC model greedy - THE BIRCH CANOE SLID ON THE SMOOTH PLANKS
CTC model beam search - THE BIRCH CANOE SLID ON THE SMOOTH PLANKS
CTC model beam search + ngram - THE BIRCH CANOE SLID ON THE SMOOTH PLANKS

To reproduce:
```bash
wget https://openslr.elda.org/resources/11/3-gram.pruned.3e-7.arpa.gz
gzip -d 3-gram.pruned.3e-7.arpa.gz
```

And then, modify the YAML so that you add the path to the n-gram.

speechbrain/decoders/ctc.py
```diff
@@ -965,6 +965,16 @@ def decode_beams(
                 "During decoding, going to truncate the log_probs vocab dim to match vocab_list."
             )
 
+        # check if we have log_probs
+        if not torch.allclose(
```
TParcollet, line 969: This operation is called for every call, this might be a problem for people that have speed requirements.

Adel-Moumen: Yes I dunno. I can remove it if you think it is up to the user to make sure that log probs are passed in input, otherwise, I can add an input flag that is by default True, and can be switch to False if one wants to remove this check. wdyt ?

TParcollet: Yes remove it, and add info in the doc about the requirements of the input

TParcollet (approved):
LGTM

PR #2388, opened 2024-02-04
(WIP) Focal Modulation Networks for Interpretable Sound Classification

lucadellalib:
Code for the workshop paper:
```
Focal Modulation Networks for Interpretable Sound Classification
L. Della Libera, C. Subakan, M. Ravanelli.
To appear at: 2024 IEEE International Conference on Acoustics, Speech and Signal Processing (ICASSP) XAI-SA Workshop
```


ycemsubakan:
Overall things looks good. Here are some things that could be improved, from a first scan: 
- My biggest concern is that there are too many files with many repetitions. I think it would be nicer if we can merge `train_piq_focalnet.py`, `train_piq_vit.py` under `train_piq.py`. Of course if the merge makes `train_piq.py` too complicated we might prefer having two separate files. 
- It would be even nicer if we can merge `test_vit.py` and `test_focalnet.py` under one file also.  
- `VectorQuantizedPSIFocalNet` can go inside `speechbrain/lobes/models/PIQ.py` 
- The visualization could be nicer (subplot sizes and orientations especially), but this is a comment to the original PR also. 
- At the moment the new additions are not explained in the README. I will start something now. Especially, having the default commands to run the methods in the paper would help. 
- PIQ results on focalnets do not seem very good qualitatively from what I have seen. I am investigating more.

mravanelli:
Thank you @lucadellalib for your code contribution, and @ycemsubakan  for your review. I have a few additional comments:

1. Regarding the README.MD, it would be better to include a link to your paper and a bib reference to facilitate citation. Additionally, I suggest expanding the README to cover all implemented systems.

2. The inclusion of transformers>=4.29 and pandas in the extra requirements is redundant since they are already included in the general Speechbrain requirements (see https://github.com/speechbrain/speechbrain/blob/develop/requirements.txt).

3. Once we're ready for that, we should upload the logs to the official Speechbrain folder for accessibility and organization. Please, follow up with me.

PR #2391, opened 2024-02-05
Minor fixes to UnitHIFIGAN

lucadellalib:
I found a few minor bugs in UnitHIFIGAN:
- The input tokens are modified in-place (+1 to each token)
- The minimum length of the sequence should be 4 instead of 3 (if you try with length 3 you get an error ` Padding size should be less than the corresponding input dimension ...`
- If length < 4, we should raise an exception rather than abruptly `quit()`

This PR fixes these issues.

### DIFF
--- a/speechbrain/inference/vocoders.py
+++ b/speechbrain/inference/vocoders.py
@@ -311,16 +311,16 @@ class UnitHIFIGAN(Pretrained):
             self.hparams.generator.remove_weight_norm()
             self.first_call = False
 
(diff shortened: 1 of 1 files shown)


poonehmousavi:
Thanks @lucadellalib .
@Chaanks  could you please take a look and apply the changes in your PR.

mravanelli:
Thank you @lucadellalib, I discussed with @Chaanks and he is fine with these changes.

mravanelli (approved):

PR #2392, opened 2024-02-06
TTS Eval: Add TTS evaluation (MOS estimation)

flexthink:
## What does this PR do?

Add TTS evaluation models trained on the SOMOS dataset
There should be no breaking changes

</details>

## PR review

<details>

</details>

### DIFF
--- a/recipes/SOMOS/ttseval/hparams/train_ssl_hubert_xformer.yaml
+++ b/recipes/SOMOS/ttseval/hparams/train_ssl_hubert_xformer.yaml
@@ -1,0 +1,92 @@
+# ############################################################################
+# Model: SSL with HuBERT (training from scratch)
+# Authors:  Artem Ploujnikov
+# # ############################################################################
+
+
+# Seed needs to be set at top of yaml, before objects with parameters are instantiated
+seed: 42
+__set_seed: !apply:torch.manual_seed [!ref <seed>]
+
+data_folder: !PLACEHOLDER
+output_folder: !ref results/ssl_hubert_xformer/<seed>
+save_folder: !ref <output_folder>/save
+train_log: !ref <output_folder>/train_log.txt
+train_regression_metric: False
+batch_size: 4
+num_workers: 4
+use_transcripts: False
+src_sample_rate: 24000
+tgt_sample_rate: 16000
+contrastive: False
+contrastive_loss_weight: 0.8
+contrastive_min_delta: 0.5
+
+
+lr: 0.0001
+dataloader_options:
+    batch_size: !ref <batch_size>
+    num_workers: !ref <num_workers>
+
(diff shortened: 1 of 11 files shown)


mravanelli:
Thank you @flexthink  for your contribution! Having a model for MOS estimation is valuable for SpeechBrain. Here are some comments following an initial code inspection:

1. The README.md file is missing.
2. Recipe tests are not currently implemented.
3. There seems to be an implementation of multiple systems; could we focus on integrating only the best one to help maintenance?
4. We need to put the logs in the official Speechbrain Dropbox.
5. The best model needs to be uploaded to HF (Hugging Face).
6. In `recipes/SOMOS/ttseval/contrastive_sampling.py`, I suggest writing  docstring examples for `RegressionContrastiveEnhancement` and `ContrastivePairingPipeline`.
7. For `ssl.py`, docstring examples should be added for `BaselineSSLFinetune` and `TransformerRegression`.
8. In `metric_stats.py`, docstring examples should be provided for `LinearRegressionStats` under `MetricStats`.
9. There's a duplicative presence of `somos_prepare.py`. Could you confirm if the second instance is just a soft link?

I recommend @BenoitWang  reviewing this PR as well. His insights would be valuable.

Thank you once again for your contribution.

BenoitWang:
Hi @mravanelli @flexthink,

I guess this may not be the last version of the code since we have another branch where we did some latest experiments. I agree that we keep only the best recipe. Some of my observations in my benchmark:
1. wavlm large gives the best corr for now
2. l2 loss works better than l1 loss, but sometimes it does not converge if we start with l2, so I start with l1 for 1 epoch and then shift to l2.
3. I added ljspeech into the training set and first did a ground-truth vs synthesized classification, then fine-tuned this model on the regression task, this helped gain also a little bit. To simplify the recipe we can provide the link of this classification model instead of adding the classification scripts.

As for the code, it seems good to me, the only thing I see here is that I don't find `ssl.py` very necessary, it wraps some basic speechbrain modules that we can put in the yaml file. What do you think @flexthink ?

recipes/SOMOS/ttseval/hparams/train_ssl_hubert_xformer.yaml
```diff
@@ -0,0 +1,92 @@
...
+subset: "full"
+skip_prep: False
+
+train_annotation: !ref <data_folder>/train.csv
+valid_annotation: !ref <data_folder>/valid.csv
+test_annotation: !ref <data_folder>/test.csv
+
+d_model: 512
+d_ffn: 2048
+num_layers: 4
```
BenoitWang, line 49: we may need to update the best config later for example wavlm large + 3-layer encoder, as well as the lr and the dropout, etc.

PR #2399, opened 2024-02-08
Shorten the data augmentation in YAML + cleaning (code from Samsung AI Cambridge)

TParcollet:
This PR attempts to shorten the data augmentation definition in YAML.  The idea is that these parameters are infrequently changed by people, meaning that we can pass them directly to the initialization of the lobe or directly rely on the default parameter. 

I decided to not create an other lobe as the current pipeline is easy enough once you simplify a bit the params definition. 

I am also adding a few comments in the yaml to make them more segmented and easier to read. 

Need the point of view of @mravanelli, @pplantinga, @Adel-Moumen and @asumagic before going forward as refactoring every yaml is .... long. 

I also created a 'replicate_multiple_labels' function that takes a list of labels to replicate for augmentation instead of doing N calls @mravanelli  have a look to this.

### DIFF
--- a/speechbrain/augment/augmenter.py
+++ b/speechbrain/augment/augmenter.py
@@ -464,28 +464,7 @@ class Augmenter(torch.nn.Module):
         list_of_augmented_labels = []
 
         for labels in args:
-            augmented_labels = []
-            if self.concat_original and not (self.skip_concat):
-                augmented_labels = [
-                    labels[
-                        self.concat_start_index : self.concat_end_index_batch
-                    ]
-                ]
-            selected_labels = labels[
-                self.augment_start_index : self.augment_end_index_batch
-            ]
-
-            if self.parallel_augment:
-                selected_labels = torch.cat(
-                    [selected_labels] * self.N_augment, dim=0
-                )
-
-            augmented_labels = (
-                augmented_labels + [selected_labels] * self.repeat_augment
-            )
-
-            augmented_labels = torch.cat(augmented_labels, dim=0)
-            list_of_augmented_labels.append(augmented_labels)
+            list_of_augmented_labels.append(self.replicate_labels(labels))
 
         return list_of_augmented_labels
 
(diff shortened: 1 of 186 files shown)


Adel-Moumen:
I'm ok with the changes. I think it makes sense and it reduces the size of our YAMLs.

speechbrain/augment/augmenter.py
```diff
@@ -438,20 +438,73 @@ def concatenate_outputs(self, augment_lst, augment_len_lst):
 
         return output, output_lengths
 
+    def replicate_multiple_labels(self, *args):
```
asumagic, line 441: This code is largely duplicated from `replicate_labels`, couldn't `replicate_labels` be made to call into this?

asumagic (changes requested):
Other than that, I think the yaml changes make sense and seemed correct (although I haven't tested them)

speechbrain/augment/augmenter.py
TParcollet: Oui. Fixed.

TParcollet:
@Adel-Moumen  @asumagic and @mravanelli I am running it, but someone will need to do a recipe testing on this one as going over all these files literally burnt my brain and I may have introduced errors.

mravanelli:
I'm running the recipe tests and there are a few fixes needed. I will push them soon.

mravanelli:
Thank you @TParcollet. I fixed small issues that emerge from the recipe tests.  There are actually several recipe tests that now fail for reasons not related to this PR. I will open a new PR to fix them.

TParcollet:
Do we know why the CI is failing? @Adel-Moumen

mravanelli:
yes, it looks like HF is temporary down: https://huggingface.co

TParcollet:
Wow unexpected.

mravanelli (approved):

Issue #2401, opened 2024-02-09
Possible NCCL-level deadlock during checkpointing


kokamido:
### Describe the bug

# General description
Hi! I am not sure if this is a bug. It's unclear to me how checkpointing should be used in a DDP mode on a single machine with multiple GPU. If I misunderstood something I would be grateful for explanation, but as far as I know, if I use speechbrain 0.5.16 I have only the following options:
1. Write same checkpoint multiple times because every DDP-worker will write it. I think it is not a perfect option because multiple writings are redundant and there is a possible [race condition](https://github.com/speechbrain/speechbrain/issues/2397) somewhere in the checkpointing code.
2. Get a deadlock. 

The following text is based on behavior of the repro-setup I provide in the "To Reproduce" section. I discuss a few runs of repro setup. Full logs of these runs are provided in the "Relevant Log Output" section. I implement end-of-epoch checkpointing using [this](https://github.com/speechbrain/speechbrain/blob/65c01135eff7b02ccf24fd681a5e5757bc01f1f4/recipes/ESC50/interpret/train_nmf.py#L79) speechbrain recipe as a reference.

# Multiple writings of the same checkpoint
To prove that every DDP-worker writes checkpoints I modified the source code of speechbrain==0.5.16. I added `print(f'{os.environ.get("LOCAL_RANK")}\t{name}')` [here](https://github.com/speechbrain/speechbrain/blob/65c01135eff7b02ccf24fd681a5e5757bc01f1f4/speechbrain/utils/checkpoints.py#L591). If I ran my repro as 
```shell
rm -rf experiments/ && torchrun --nnodes=1 --nproc-per-node=2 ckpt_repro.py ckpt_repro.yaml --number_of_epochs=1 --ckpt_b
ehavior=="All threads" 2>&1 | tee log_1.txt
```
I can see that both workers write the same checkpoints parts (full log provided as "log_1" in the "Relevant Log Output" section):
```shell
100%|██████████| 160/160 [00:01<00:00, 147.62it/s, train_loss=0.68] 
0       counter
0       brain
1       counter
1       brain
1       optimizer
0       optimizer
```
 
# Test setups
I ran the following setups without "print" added before for the sake of clarity.

## 1. Write intra-epoch checkpoints only
`torchrun --nnodes=1 --nproc-per-node=2 ckpt_repro.py ckpt_repro.yaml --ckpt_interval_minutes=0.001 2>&1 | tee log_2.txt`

```shell
[...]

PR #2404, opened 2024-02-10
Make a ddp broadcast clone that handles main-process-only

pplantinga:
Checkpointing with DDP uses a torch.distributed broadcast to communicate the folder for saving to all procs. However, this broadcast causes a freeze if checkpointing is called in a main-process-only section of the code. This PR addresses this by adding a `ddp_broadcast` function that should correctly handle the main-process-only limitation.

### DIFF
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -121,8 +121,20 @@ def ddp_barrier():
 
 
 def ddp_broadcast(communication_object, src=0):
-    """In DDP mode, this function will broadcast a tensor to all
+    """In DDP mode, this function will broadcast an object to all
     processes.
+
+    Arguments
+    ---------
+    communication_object: Any
+        The object to be communicated to all processes. Must be picklable.
+        See docs for ``torch.distributed.broadcast_object_list()``
+    src: int
+        The rank which holds the object to be communicated.
+
+    Returns
+    -------
+    The communication_object passed on rank src.
     """
     if MAIN_PROC_ONLY >= 1 or not torch.distributed.is_initialized():
         return communication_object
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -120,6 +120,20 @@ def ddp_barrier():
         torch.distributed.barrier()
 
 
+def ddp_broadcast(communication_object, src=0):
+    """In DDP mode, this function will broadcast a tensor to all
+    processes.
+    """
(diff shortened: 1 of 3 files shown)


pplantinga:
See issue #2401 for more info

kokamido:
Hi @pplantinga , I am the author of the issue. Could you explain the reason of the broadcasting during checkpointing? I mean, it's unclear to me why do we have to [broadcast ckpt_directory](https://github.com/speechbrain/speechbrain/blob/8c6f8ef9c8fb15ebced52bccd83bc42f1dde4550/speechbrain/utils/checkpoints.py#L606).

pplantinga:
> Hi @pplantinga , I am the author of the issue. Could you explain the reason of the broadcasting during checkpointing? I mean, it's unclear to me why do we have to [broadcast ckpt_directory](https://github.com/speechbrain/speechbrain/blob/8c6f8ef9c8fb15ebced52bccd83bc42f1dde4550/speechbrain/utils/checkpoints.py#L606).

Of course, I'm happy to explain. Basically there's some cases where checkpointing needs to be run on multiple processes, for example with FSDP there are shards of the model on each device that need to be combined before saving. In these cases, every process needs to know where the checkpointing directory will be. Given our current setup where the default checkpointing directory name is based on the time, this can't be determined ahead of time, so we must rely on inter-process communication.

This same strategy is followed by pytorch_lightning if i remember correctly

pplantinga:
Looking for the broadcast in pytorch lightning, it looks like its done in a special fsdp class, not sure if there are other places where this is done:

https://github.com/Lightning-AI/pytorch-lightning/blob/47c8f4cba089a78fa3fe31dcac6a43416bc13820/src/lightning/fabric/strategies/fsdp.py#L447

speechbrain/utils/distributed.py
```diff
@@ -120,6 +120,20 @@ def ddp_barrier():
         torch.distributed.barrier()
 
 
+def ddp_broadcast(communication_object, src=0):
+    """In DDP mode, this function will broadcast a tensor to all
+    processes.
+    """
```
Adel-Moumen, line 126: could you please add the standard SB docstring related to the arguments / returns please ? Other than that, I think this is ready to be merged.

mravanelli:
Thank you @pplantinga. I contacted @Adel-Moumen privately and he confirmed that it is fine with that. We can merge it

mravanelli (approved):

PR #2421, opened 2024-02-20
New ASR dataset VoxPopuli and Transducer recipe (Code from Samsung AI Center Cambridge)

TParcollet:
## What does this PR do?

This PR introduces a streaming ASR recipe for VoxPopuli with a transducer. I'll try to commit a transformer and CTC (SSL) recipe as well ... later. 

## To-do
- [x] add tests
- [ ] A reviewer must retrain the model on 2 - 4 GPUs (batch size between 800s and 1000s)

PR #2426, opened 2024-02-21
Optimise masking for streaming (code from Samsung AI Center Cambridge)

TParcollet:
## What does this PR do?

This PR does two things: 
1. divides by 10x 50x the time taken to generate masks for streaming training. The optimization is done by turning the loop over timesteps into a loop over chunk.s
2. Create unit tests for all the masking functions of the transformer. We had none, which is ... no gud. 

**I WOULD ADVISE A REVIEWER TO RETRAIN THE MODEL for a few epochs EVEN THOUGH I ADDED TESTS**


mravanelli:
@asumagic, could you please proceed with the review? It would be great to include it in the upcoming release

tests/unittests/test_transformer_src_tgt_masks.py
Adel-Moumen, line 1: fiou 1k5 lines for this test ? Could we simply it a bit more ? :P

recipes/VoxPopuli/ASR/transducer/hparams/conformer_transducer.yaml
Adel-Moumen, line 1: whats that? Is it a mistake you made while working on #2421 ?

tests/unittests/test_transformer_src_tgt_masks.py
TParcollet: Maybe we can divide it by two, but the process is boring because you need to copy past the exact value of the tensors ...

asumagic: Or define the expected tensor as a list of strings for easy editing, with an helper method to turn that into a bool/float tensor

mravanelli:
I ran recipe tests (including full-inference ones) and everything works fine. Thank you @TParcollet, @Adel-Moumen, and @lucadellalib .

mravanelli (approved):

PR #2433, opened 2024-02-26
New CommonVoice streaming recipes

TParcollet:
## What does this PR do?

Our CV recipes for transducer are old AF. This PR updates with the latest models + streaming. It must be run by @Adel-Moumen and results must reported. I added an example of results obtained locally, but can't share the models .......... Results are MUCH better, and not that far from SSL surprisingly. For the tests error, I don't even understand it, maybe @Adel-Moumen can help when time allows it. 

Maybe @asumagic wants to have a look as well.


Adel-Moumen (approved):
lgtm

PR #2463, opened 2024-03-18
Add cspell

calvinballing:
## What does this PR do?

Successor PR to: https://github.com/speechbrain/speechbrain/pull/2176
Which was the successor PR to: https://github.com/speechbrain/speechbrain/pull/1353

The first 3 commits are the same as the original (except for addressing merge conflicts from the past 1.5 years), the next 8 address new typos introduced between the original and the second, and the last 4 address typos in changes since then.

I separated out this change into its own commit so that someone more familiar with it can confirm it is correct:
`lr_whisperc -> lr_whisper`
________

Possible future work, if cspell is added: 
1) Adding a GitHub action for Cspell
2) working more on the custom dictionary.  Currently it has 7 sections, the last of which could be sorted into the rest of the dictionary.:
 - British (British English spellings of words)
 - Non-English (Words in languages other than English)

<details>

-- Adds cspell, but does not implement a GitHub action for it yet
-- Not needed
-- Not needed
-- no
-- Yes, none
-- Probably

</details>

## PR review

<details>

</details>


pplantinga (changes requested):
Overall, really pleased to see this change, it is needed as evidenced by the hundreds of spelling errors already found!

Just a couple things before we merge:

1. We should add some documentation to the developer guide so they know what to do when they encounter an exception that should be added to the dictionary. Perhaps we can also make an announcement in the slack to read the guide if there's new spelling errors. I'm thinking a section in `docs/contributing.md` near `pre-commit` makes the most sense.
2. The `package.json` and `yarn.lock` files are node.js-specific, unless they are necessary for `cspell` to work for `pre-commit` I would suggest we remove them.
3. There's a large number of entries in the "unsorted" section, most of which I assume occur only once in the repo. I wonder what the value is having them here vs. ignoring the specific case with a code comment. E.g. `theyre` seems like something that we should normally fix (but it might make sense to ignore sometimes).

pplantinga:
Good to see `cspell` passed in the CI

calvinballing:
Regarding point 3 above, I've ignored some additional files, sections of files, and specific words within files, which allowed for removing quite a few terms from the dictionary.  In digging into the `unsorted` section, I was also able to correct some additional typos.  I'm not sure exactly where the right tradeoff is between dumping all the remaining terms terms in the dictionary vs adding `cspell:ignore xxx` statements throughout the codebase.  Let me know if you'd like to see more work done to shrink the dictionary

I also inspected additional terms in the unsorted section, and was able

pplantinga (approved):
LGTM

PR #2465, opened 2024-03-19
Update CommonVoice transformer recipes

TParcollet:
Much like our Transducer recipes (problem addressed in #2433) our transformer recipes on CommonVoice are old AF. I updated everything here. I also removed the per language yaml as only the output vocabulary and number of epochs must be changed. 

We will need someone to retrain the models ..... at least english / french / something else?


recipes/CommonVoice/common_voice_prepare.py
```diff
@@ -304,6 +326,32 @@ def create_csv(
     logger.info(msg)
 
 
+def convert_mp3_to_wav(audio_mp3_path, sampling_rate=16_000):
+    """Convert an mp3 file to a wav file.
+
+    Parameters
+    ----------
+    audio_mp3_path : str
+        The path to the opus file to be converted.
```
Adel-Moumen, line 335: make sure to add sampling_rate in the docstring. Furthermore, maybe we should have a global variable at the top e.g. `VERBOSITY=0` which will by default remove ffmpeg output (so what you did with > /dev/null), otherwise, it will output in the stdout stream

Adel-Moumen:
Hi, 

LGTM. I uploaded the models on the dropbox. There will be shortly a follow up with the EN based model. All the recipe tests passed. 

## Test
```bash
python -c 'from tests.utils.recipe_tests import run_recipe_tests; print("TEST FAILED!") if not(run_recipe_tests(filters_fields=["Hparam_file"], filters=[["recipes/CommonVoice/ASR/transformer/hparams/conformer_large.yaml"]], do_checks=False, run_opts="--device=cuda")) else print("TEST PASSED")'
```
```
(1/1) Running test for CommonVoice_row_18...
        ... 151.56s
TEST PASSED
```

Adel-Moumen (approved):

PR #2467, opened 2024-03-20
Change CI from using `cspell` to `codespell`

pplantinga:
The `cspell` tool comes from the `node.js` ecosystem and apparently fails to install on some systems due to requiring a fairly recent version of `GLIBC`, see this issue for more details:

https://github.com/jupyterlab/jupyterlab/issues/12675

This PR suggests we switch to `codespell` which is in python. The tool is less mature but still has > 1k stars and should be well-supported. In addition, it found a bunch more spelling errors.


recipes/Fisher-Callhome-Spanish/fisher_callhome_prepare.py
```diff
@@ -710,7 +710,7 @@ def remove_labels(transcription: str):
     transcription = re.sub(r"<foreign langenglishonline", "", transcription)
     transcription = re.sub(r"<foreign langenglish", "", transcription)
     transcription = re.sub(r"</foreign", "", transcription)
-    transcription = re.sub(r"<[/]?foreing\s*\w*>", "", transcription)
+    transcription = re.sub(r"<[/]?foreign\s*\w*>", "", transcription)
```
Adel-Moumen, line 713: We should make sure that this regex is still valid

Adel-Moumen (commented):
ok, good! I left one comment. BTW, could you please fetch the latest speechbrain update? I suspect that you will need to run the linter on the new merged PR as well. I will check your PR locally once it's ready . Thanks again :)

recipes/Fisher-Callhome-Spanish/fisher_callhome_prepare.py
pplantinga: If you look at the context (before and after) all the lines are "foreign" so I'm pretty confident this one was supposed to be as well.

pplantinga: Unless there's bad tags in the data that this was removing

Adel-Moumen (approved):
LGTM! Works on my lab cluster now.

PR #2476, opened 2024-03-27
Add `fetch` `local_strategy` parameter and disable symlinks by default

### LINKED ISSUE #1070: Cannot download models locally
Hi, I tried to run the commands, which worked perfectly file.
```

asr_model = EncoderDecoderASR.from_hparams(source="speechbrain/asr-crdnn-rnnlm-librispeech")
```
But when I looked over the model files, the files downloaded are just 158 kbs, and not the full files are downloaded.

Also, when I tried to download the files directly from Files and versions, the ckpt files are downloaded as zip. which cant be used later for inference. IS there any way to resolve it or it has been purposefully made like that.

### LINKED ISSUE #1155: Pretrained.from_hparams() failing on windows when non admin
The following code is failing on windows with non-admin rights (not a big issue but sadly a "standard" case in lots of "big" companies) due to symlinks.

```python
EncoderClassifier.from_hparams(
  source=r"C:\path\to\pretrained\model",
  savedir="./tmp",
  overrides={"pretrained_path": r"C:\path\to\pretrained\model"})
```

I fixed it locally by copying models to `savedir` instead of making a symlink (`fetching.py:L102`), BUT I will not submit it as a fix.
```python
try:
  destination.symlink_to(sourcepath)
except:
  import shutil
  shutil.copyfile(sourcepath, destination)
```

I can start working on local loading feature in a PR if you're interested. When loading from local, we could even avoid copying or symlinking directories and loading directly from the local folder for example.

### LINKED ISSUE #1177: Too many levels of symlinks error when loading model in AWS Lambda function
Hi, I have been trying to containerize Speaker Recognition model in an AWS lambda function which has a read-only Linux file system with just '/tmp' folder available for write. Following issues #1001 and #1155, I have downloaded the model files and copied them into '/tmp' during build and loaded as below
`diarizer = SpeakerRecognition.from_hparams(source='/tmp', savedir='/tmp', overrides={"pretrained_path": '/tmp'})`

Getting a too many levels of symlinks error
sb-lambda-err

Please let me know how I can fix it. Thanks

### LINKED ISSUE #1278: turn off symlink for pretrained.fetching
Hi,
I have been using a pretrained model and a following line of code:
``` Python
asr_model.transcribe_file(hparams["audio_dir"] + data["file_path"][1:])
```
Generates a ton of symlink audio files in the project folder and my working space becomes completely cluttered. 
I've spotted that it happens because of this function: https://speechbrain.readthedocs.io/en/latest/API/speechbrain.pretrained.fetching.html.

Is it possible to turn it off somehow while using a pretrained model?  I would like to use real paths to files rather than symlinks, audio files are just in a different folder in my working space. The model is created with:
``` Python
asr_model = EncoderDecoderASR.from_hparams(source="speechbrain/asr-crdnn-rnnlm-librispeech", 
                                           savedir="./uncertainty_baselines/notebooks/pretrained_ASR")
```

### LINKED ISSUE #1351: wrong file loading by separate_file function
The function [separate_file](https://github.com/speechbrain/speechbrain/blob/3669d226e816bbc969246400e44ebed2824302ce/speechbrain/pretrained/interfaces.py#L1965) employs [fetch function](https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/pretrained/fetching.py#L26) to fetch files in a given path (i.e., **savedir**). The current version is okay unless **sperate_file** tries to fetch the new filename that already exists in _savedir_.  This is already addressed by  **[overwrite](https://github.com/speechbrain/speechbrain/blob/3669d226e816bbc969246400e44ebed2824302ce/speechbrain/pretrained/fetching.py#L84)** in the fetch; however, this can not be controlled within **separate_file**; then,  please provide an access to overwrite within seperate_file:

---------------------------------------------------------------------------

def separate_file(self, path, savedir=".", **overwrite**=False):
...
         path = fetch(fl, source=source, savedir=savedir, **overwrite**)

---------------------------------------------------------------------------

This problem becomes challenging in case of a huge dataset having plenty of similar filenames stored at different directories; then the user can simply set **overwrite = True** to avoid such a serious problem.

Thanks,

### LINKED ISSUE #1352: How to undo generation of softlinks of wav files when using Speaker Verification model?

### LINKED ISSUE #1586: Impossible to transcribe audio file with the same name
Due to the implementation of the [transcribe_file](https://github.com/speechbrain/speechbrain/blob/aef146861cc89482102f7ce63cda1e8d3366a229/speechbrain/pretrained/interfaces.py#L716) method and specially of the sub method [load_audio](https://github.com/mravanelli/speechbrain-1/blob/033d8cda7f237109636849ef34c0b0e83e9c4897/speechbrain/pretrained/interfaces.py#L229), it is impossible to transcribe audio file with the same name despite being located in different directories since the method copy the audio file locally at `savedir`. The new file with the same name doesn't overwrite the file, since it already exist and make the transcription impossible.

Example of file paths :

- /users/<user_name>/datasets/MyDataset/recordings/ZOWxbXGuoU/**recording_1.wav**
- /users/<user_name>/datasets/MyDataset/recordings/2h4v5o49Hj/**recording_1.wav**

It will be cool to get rid of this local file to fix this issue and improve overall performances of the transcription method.

One easy but dirty way of fixing it is to remove the file at the end of the transcription to allow further ones.

### LINKED ISSUE #1885: [Bug]: Windows saving of hugging face models fails due to using the / Override on a WindowsPath Object
### Describe the bug

Hugging Face models can't be downloaded using fetching.py

### Expected behaviour

The destination variable is initialized correctly on windows. 

### To Reproduce

    vad_model = VAD.from_hparams(source="speechbrain/vad-crdnn-libriparty", savedir="pretrained_models/vad-crdnn-libriparty")

### Versions

Name: speechbrain
Version: 0.5.13
Summary: All-in-one speech toolkit in pure Python and Pytorch
Home-page: https://speechbrain.github.io/
Author: Mirco Ravanelli & Others
Author-email: speechbrain@gmail.com
License:
Location: c:\users\dsamf\appdata\local\packages\pythonsoftwarefoundation.python.3.10_qbz5n2kfra8p0\localcache\local-packages\python310\site-packages
Requires: huggingface-hub, hyperpyyaml, joblib, numpy, packaging, scipy, sentencepiece, torch, torchaudio, tqdm
Required-by:

### Relevant log output

```shell
Exception in thread Thread-3 (process_audio_stream):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.2800.0_x64__qbz5n2kfra8p0\lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.2800.0_x64__qbz5n2kfra8p0\lib\threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "D:\Development\llm\whisper\aivsai.py", line 179, in process_audio_stream
    boundaries = vad_model.get_speech_segments(audio_buffer)
  File "C:\Users\dsamf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\speechbrain\pretrained\interfaces.py", line 1892, in get_speech_segments
    audio_file = fetch(fl, source=source)
  File "C:\Users\dsamf\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\site-packages\speechbrain\pretrained\fetching.py", line 88, in fetch
    destination = savedir / save_filename
TypeError: unsupported operand type(s) for /: 'WindowsPath' and 'int'
```

### Additional context

_No response_

### LINKED ISSUE #2197: Need admin Permissions to run the model on windows
### Describe the bug

When I use speechbrain models on windows, it crash and report that need permission. I think speechbrain use soft link for the file, please check and fix it.

### Expected behaviour

no crash.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

asumagic:
## What does this PR do?

Currently this is just mocking up APIs and writing proofs-of-concept.

The goal of this PR would ultimately be to explore ways to avoid using symbolic links in most cases in SB (which should e.g. definitely not be done when using `load_audio` against local paths in inference APIs...)

As of the current state of the PR:

- `fetch` is extended to accept a `local_strategy` parameter. This parameter controls the behavior for downloading and copying/linking files around.
- Most stuff within SB that indirectly calls `fetch` now accepts and forwards the same parameter.
- The default strategy is switched from `SYMLINK` to `NO_LINK` on all platforms (can be debated and needs extensive testing).
- Inference interfaces now explicitly use `NO_LINK` by default since polluting the working directory with audio files is largely undesired.
- Fixed up parameter transfer logic to handle `NO_LINK` behavior.
- Explicitly set the local strategy the in some recipe dataset download scripts.
- `fetch` also now has an `allow_updates` parameter, which defaults to **`True`** (TODO: what consequences does this have with no internet access?). It interrogates HF even if a local path is present in order to update any model files if required.
- `fetch` also now has an `allow_network` parameter, which defaults to **`True`**. When disabled, this fails if the URI is not available locally, or if the file isn't found locally in HF cache.

Resolves #1070, resolves #1278, resolves #2197, resolves #1155, resolves #1586, resolves #2570, resolves #1352, resolves #1177, resolves #1351

Supersedes #1303 

At the same time, probably should figure out if these can be resolved/superseded: #1253 + #1254

<details>

</details>

## PR review

<details>

</details>


Adel-Moumen:
Ping @pplantinga and @Gastron as they both have a good understanding of the current issue and SB codebase. Could you please share your thoughts on this issue so that we can move in a valuable direction please? Thanks :)

speechbrain/utils/fetching.py
```diff
@@ -58,38 +158,68 @@ def fetch(
     use_auth_token=False,
     revision=None,
     huggingface_cache_dir=None,
+    local_strategy: LocalStrategy = LocalStrategy.SYMLINK,
```
pplantinga, line 161: I'm curious if we still want the default to be `SYMLINK` here. I appreciate that we might want to keep the default the same but perhaps we could add a warning and change it in the next version?

speechbrain/utils/fetching.py
```diff
@@ -58,38 +158,68 @@ def fetch(
...
+        `overwrite == True` to ensure you're getting the latest version, as the
+        HF Hub performs caching of its own.
+
+    Effect of local file strategies when the fetch source is a **local file**:
+
+    - `LocalStrategy.SYMLINK`: Create symlink to the file in the destination.
+    - `LocalStrategy.COPY`: Create copy of the file in the destination.
+    - `LocalStrategy.COPY_SKIP_CACHE`: Create copy of the file in the
+      destination.
+    - `LocalStrategy.NO_LINK`: Returns the path to the source file.
```
pplantinga, line 184: There's many examples of backtick-ed code in this PR, I'm pretty sure sphinx documentation generation requires two backticks not just one.

But after more research it may be possible to change this behavior so that single backticks also renders as code:

https://github.com/sympy/sympy/issues/13519

pplantinga (approved):
Great change, LGTM!

speechbrain/utils/fetching.py
asumagic: > There's many examples of backtick-ed code in this PR, I'm pretty sure sphinx documentation generation requires two backticks not just one.
> 
> But after more research it may be possible to change this behavior so that single backticks also renders as code:
> 
> [sympy/sympy#13519](https://github.com/sympy/sympy/issues/13519)

You're right, but I have actually set this a while back: https://github.com/speechbrain/speechbrain/blob/develop/docs/conf.py#L91

Lots of docstrings in SB inadvertently used single backticks which were just shown as italic, but always intended code formatting.

pplantinga: Sweet, I wasn't aware this had happened but glad to see it!

speechbrain/utils/fetching.py
asumagic: I thought about this more and I think it might make sense to switch to `NO_LINK` by default. I was wary of doing such a change but I don't think most things implicitly rely on this, and it's usually a simple fix or workaround.

CI passes when changing the default after one change in parameter transfer logic.

It should be plenty documented though... I don't want users to be too surprised when they see an empty directory where the pretrained stuff should be.

asumagic:
There doesn't appear to be any regression in HF tests (although the cache might already have been warm).

asumagic:
I think the PR is in an OK state and it might make sense to merge earlyish in the v1.1 cycle so that there is plenty of time to detect and fix issues, so I opened it for review.

speechbrain/utils/fetching.py
```diff
@@ -51,56 +46,265 @@ class FetchFrom(Enum):
...
+        `allow_updates` must be `True`.
+    huggingface_cache_dir: str, optional (default: `None`)
+        Path to HuggingFace cache; if `None`, the default cache directory is
+        used: `~/.cache/huggingface` unless overridden by environment variables.
+        See `huggingface_hub documentation <https://huggingface.co/docs/huggingface_hub/guides/manage-cache#manage-huggingfacehub-cache-system>`_
+        Ignored if the local strategy is `LocalStrategy.COPY_SKIP_CACHE`.
+        (default: None)
+    local_strategy : LocalStrategy, optional
+        Which strategy to use to deal with files locally. (default:
+        `LocalStrategy.SYMLINK`)
```
pplantinga, line 307: This default was updated to `NO_LINK`, documentation should be updated to match.
```suggestion
        `LocalStrategy.NO_LINK`)
```

asumagic:
> Overall, LGTM, if a bit more complex than I'd prefer. I'm guess that's just a necessary evil though due to all the cases we are trying to handle. Only one minor doc fix to suggest, then can be merged.

I agree; I would have preferred to get rid of `fetch` under that form altogether but a large amount of things depend on that API being an automagic solution to all usecases. I do hope the complexity was spread out more evenly since the original code was kind of spaghetti though.

PR #2489, opened 2024-04-02
Fix import statement ordering using isort

pplantinga:
Adds `isort` to the linters. This is a nice way to ensure consistency and make imports easier to find.

The initial version of this PR just uses the defaults from `isort` but if anyone has opinions there are other ways to sort.

Default sorting is alphabetical within 3 sections:

1. Standard library imports
2. Third party imports
3. Internal imports

There are options, such as sort by length, different sections etc. so if anyone wants a different sort, they should speak up.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,7 @@ repos:
       - id: black
         types: [python]
         additional_dependencies: ['click==8.1.7']
+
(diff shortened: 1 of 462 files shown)


asumagic (approved):
Checked a good subset of the modified files, seems like a good improvement for me. LGTM.

PR #2492, opened 2024-04-04
Only include the `speechbrain` package in `setup.py`

### LINKED ISSUE #2491: PyPI install incorrectly ships a `tests` package
### Describe the bug

If you check the `.whl` file of SB 1.0, for instance, you will notice that there is an empty `tests/` (with only the `__init__.py`). This is not a huge deal, but it has the unintended consequence that the user now has a `tests` package at their root level.

### Expected behaviour

That package should not exist on normal or editable installs.

### To Reproduce

```python
import tests
```

```bash
$ cat env/lib/python3.xx/speechbrain-xxx.dist-info/top_level.txt 
speechbrain
tests
```

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

This is probably due to the way packages are specified in the `setup.py`.

Does not seem like it would be linked to #2484 but it's worth investigating anyway.

asumagic:
## What does this PR do?

An empty `tests` package would get shipped both with normal and editable installs (even though the `tests` package is empty anyway in the `.whl` for some reason).

This PR stops discovering packages automatically and just specifies the `"speechbrain"` package alone instead.

Fixes #2491 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
         "Programming Language :: Python :: 3",
         "License :: OSI Approved :: Apache Software License",
     ],
-    packages=setuptools.find_packages(),
+    packages=["speechbrain"],
     package_data={"speechbrain": ["version.txt", "log-config.yaml"]},
     install_requires=[
         "hyperpyyaml",


asumagic:
Nevermind, the root-package-only approach does not work with non-editable installs.

PR #2496, opened 2024-04-05
Lazy imports

asumagic:
## What does this PR do?

Solves #2500, solves #2501

Currently, we have a few (sub)packages unconditionally bring all its modules/subpackages into scope (`from . import some_module_or_subpackage`). This makes the module/subpackage in question easy to access without importing, e.g. `sb.utils.data_pipeline.takes` without ever importing `sb.utils.data_pipeline` explicitly. This approach as implemented comes with some issues:

1. `utils` importing stuff from anywhere else could itself import a different module from `utils`, and because `utils/__init__.py` tries to import everything again, this can causes circular imports and other headaches.
2. Additionally, it means that every third-party dependency (e.g. scipy) gets imported unconditionally when importing `speechbrain` if it is part of the affected modules, even for depenedencies that are not being used, which commonly doubles import times.

This PR extends the newly introduced `utils.importutils` module in order to abstract this logic under a common function, `lazy_export_all`, which scans the current package for modules and subpackages.  
These then get wrapped using the new `LazyModule` and bound to the `sys.modules`, which means that the modules remain directly accessible, but import is delayed until the module is actually accessed. Thus, this PR should not break existing code, bar some bug in the implementation.

If we want to ever deprecate this means of importing modules, this mechanism could easily be modified for it.

Possible downsides of this PR:

1. IDEs might get more confused about this and may cease to properly lint some types affected -- but to be honest, users probably should use explicit imports anyway.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/__init__.py
+++ b/speechbrain/__init__.py
@@ -3,17 +3,8 @@
 
 import os
 
-from . import alignment  # noqa
-from . import dataio  # noqa
-from . import decoders  # noqa
-from . import lm  # noqa
-from . import lobes  # noqa
-from . import nnet  # noqa
-from . import processing  # noqa
-from . import tokenizers  # noqa
-from . import utils  # noqa
 from .core import Brain, Stage, create_experiment_directory, parse_arguments
-from .utils.importutils import deprecated_redirect
+from .utils.importutils import deprecated_redirect, lazy_export_all
 
 with open(os.path.join(os.path.dirname(__file__), "version.txt")) as f:
     version = f.read().strip()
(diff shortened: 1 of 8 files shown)


asumagic:
This is currently an early WIP but I will add it to the v1.0.1 milestone as we have some circular dependencies issues this would hopefully address.

pplantinga (approved):
This is a great change, the import system has always been flaky and this is a nice way of supporting the same behavior while avoiding the annoying parts. This implementation of lazy loading is clever and should be much more performant.

PR #2507, opened 2024-04-13
Add inference to templates

pplantinga:
Our templates are lacking any mechanism for doing inference, which seems important.

This seeks to add it to the following templates:

- [x] Enhancement
- [x] Speaker ID
- [x] Speech Recognition


pplantinga:
Ready for review now

templates/speaker_id/verify_speaker.py
```diff
@@ -0,0 +1,33 @@
...
+    verifier = SpeakerRecognition.from_hparams(
+        source=".",
+        hparams_file="inference.yaml",
+        savedir=args.save_directory,
+    )
+    score, prediction = verifier.verify_files(args.sample1, args.sample2)
+    if prediction:
+        print("Model predicts SAME speaker")
+    else:
+        print("Model predicts DIFFERENT speakers")
```
Adel-Moumen, line 33: wondering if we should use `logger` instead of print.

templates/speech_recognition/ASR/transcribe_file.py
```diff
@@ -0,0 +1,58 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("speech_file")
+    parser.add_argument("save_directory")
+    args = parser.parse_args()
+
+    # Link ASR and normalizer in save folder alongside LM and tokenizer
+    source_dir = os.path.abspath(args.save_directory)
+    target_dir = os.path.dirname(source_dir)
+    link_file("model.ckpt", source_dir, target_dir)
+    link_file("normalizer.ckpt", source_dir, target_dir)
```
Adel-Moumen, line 48: maybe we should add a comment on why it is necessary to do these steps

templates/speech_recognition/ASR/transcribe_file.py
```diff
@@ -0,0 +1,58 @@
...
+
+    transcriber = EncoderDecoderASR.from_hparams(
+        source=".",
+        hparams_file="inference.yaml",
+        savedir=target_dir,
+    )
+    text = transcriber.transcribe_file(args.speech_file)
+
+    print(f"Text present in file {args.speech_file}:")
+    print(text)
```
Adel-Moumen, line 58: same Q as before. Should we use logger?

Adel-Moumen (commented):
Overall LGTM. I just left some comments. BTW, i think we should maybe start to create subfolders in templates. For instance, `speech_recognition/ASR` is very general and the provided `inference.yaml` is only targeting a `seq2seq` recipe. Maybe we should have in the future `speech_recogntion/CTC`, `speech_recogntion/seq2seq`, `speech_recogntion/transformer`, etc... CC @asumagic and @TParcollet on this remark.

templates/enhancement/train.yaml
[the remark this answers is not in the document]
pplantinga: Since the file is used to DO TRAINING clearly it needs the EXECUTABLE BIT set lol

templates/speaker_id/verify_speaker.py
pplantinga: Honestly I'm not sure about this. The point of the script is a one-off inference as an example, not something people should ever use as part of a larger system. Also, `logger.info` doesn't print anything on my system, and `logger.warning` seems wrong somehow.

Adel-Moumen: I agree with you @pplantinga :-)

Adel-Moumen (approved):
LGTM thanks @pplantinga :-)

PR #2508, opened 2024-04-15
Skip lazy imports when the caller is inspect.py

asumagic:
## What does this PR do?

This avoids having certain inspect functions import our lazy modules when we don't want them to. `getframeinfo` in particular appears to do it, and this gets called by PyTorch at some point. IPython might also be doing it but autocomplete still seems to work.

This does not appear to break anything. Added test for hyperpyyaml to ensure we're not breaking that.

Fixes the issue where the redirection warning was emitted even if you weren't using `speechbrain.pretrained`, and prevents some lazy imports from forcefully being imported.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/utils/importutils.py
+++ b/speechbrain/utils/importutils.py
@@ -6,6 +6,7 @@ Author
 """
 
 import importlib
+import inspect
 import os
 import sys
 import warnings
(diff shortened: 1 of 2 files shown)


pplantinga (commented):
I'll be honest, some of what's going on here is a bit beyond me without studying it more, but it seems to work.

Just a question though, I get wanting to prevent imports that are unnecessary but was it causing crashes or is it just to improve speed of loading etc.? Just asking because this adds some complexity to the lazy modules and I want to make sure its worth it.

asumagic:
> I'll be honest, some of what's going on here is a bit beyond me without studying it more, but it seems to work.
> 
> Just a question though, I get wanting to prevent imports that are unnecessary but was it causing crashes or is it just to improve speed of loading etc.? Just asking because this adds some complexity to the lazy modules and I want to make sure its worth it.

The problem is that the `pretrained`->`inference` lazy import raises a warning at import time because of the issue I mentioned, so the warning would often get printed even if  you were never importing `.pretrained`.

Theoretically, we could instead have a check in the warning rather than doing this, but I believe that it could result in the warning never being raised in some cases (since the lazy module will no longer do anything once the import is done, so it couldn't catch a "genuine" `import speechbrain.pretrained` if `.pretrained` already got imported "accidentally").

speechbrain/utils/importutils.py
```diff
@@ -43,9 +44,45 @@ def __init__(
...
+        code will inadvertently cause our modules to be imported, such as some
+        of PyTorch's op registering machinery.
+
+        Arguments
+        ---------
+        stacklevel : int
+            The stack trace level of the function that caused the import to
+            occur, relative to the **caller** of this function (e.g. if in
+            function `f` you call `_ensure_module(1)`, it will refer to the
+            function that called `f`).
```
Adel-Moumen, line 62: could you please add a raise docstring as well e.g.: 
```
RAISES
--------
```

PR #2513, opened 2024-04-17
Move `tests/samples` to HF?

asumagic:
## What does this PR do?

WIP. Proof-of-concept of using a submodule towards HF using LFS, which has a few practical advantages:

- HF limits are much more generous than GitHub's for LFS
- Better integration with git
- Default clone commands will not clone this submodule
- Clone commands with the recursive submodules option, by default, will only fetch the latest version of the file thanks to LFS
- We retain versioning for the samples files. Realistically, we could be fine without, but it's still a bonus if it doesn't pollute the main repo history in a significant way and keeps defaults sane

The obvious downgrade is that it makes recipe PRs more annoying.

<details>

</details>

## PR review

<details>

</details>


asumagic:
The PR seems to work in my testing. Needs more discussion to see if people are OK with this.

Then, we will need migrating that new repo under the `speechbrain/` org, documenting the procedures, updating the license details, etc.

pplantinga:
> The obvious downgrade is that it makes recipe PRs more annoying.

This is only for recipe PRs that add data to the test samples right? And the change to the process is simply to open a concurrent PR to the HF test samples repo right?

asumagic:
> > The obvious downgrade is that it makes recipe PRs more annoying.
> 
> This is only for recipe PRs that add data to the test samples right? And the change to the process is simply to open a concurrent PR to the HF test samples repo right?

Correct, but one problem that just occurred to me is that we need to think about the flow for updating the submodule from a PR.

[Apparently, submodules allow tracking branches nowadays which might make sense here.](https://stackoverflow.com/questions/1777854/how-can-i-specify-a-branch-tag-when-adding-a-git-submodule)

asumagic:
Also, on second thought, I am not sure this is so worthwhile. This makes the logistics of merging new recipes more complicated. The full history doesn't need to be cloned regularly, and options like `--filter=blob:none` would mostly mitigate the concern. Being mindful of sample size when reviewing PRs would likely solve the problem and total repo size limits are rather lenient in GitHub.

PR #2518, opened 2024-04-22
DDP at eval stage + DDP metrics

Adel-Moumen:
## What does this PR do?

blbllb life is better when everything goes fast; that's why we need DDP at eval stages. 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1418,6 +1418,9 @@ class Brain:
                 self.avg_train_loss = self.update_average(
                     loss, self.avg_train_loss
                 )
+                # TODO: need to benchmark the effect of this communication
+                # sync the avg_loss across all processes 
+                self.avg_train_loss = self.sync_average_loss(self.avg_train_loss)
                 t.set_postfix(train_loss=self.avg_train_loss)
 
                 if self.profiler is not None:
@@ -1442,7 +1445,8 @@ class Brain:
                     steps_since_ckpt = 0
 
         # Run train "on_stage_end" on all processes
-        self.zero_grad(set_to_none=True)  # flush gradients
+        # flush gradients
+        self.zero_grad(set_to_none=True)
         self.on_stage_end(Stage.TRAIN, self.avg_train_loss, epoch)
         self.avg_train_loss = 0.0
         self.step = 0
@@ -1500,6 +1504,8 @@ class Brain:
                         break
 
                 self.step = 0
+                # sync the avg_loss across all processes
+                avg_valid_loss = self.sync_average_loss(avg_valid_loss)
                 self.on_stage_end(Stage.VALID, avg_valid_loss, epoch)
 
     def fit(
@@ -1776,12 +1782,8 @@ class Brain:
                 if self.debug and self.step == self.debug_batches:
                     break
             
-            # sync avg_test_Loss
-            if torch.distributed.is_initialized():
-                avg_test_loss = torch.tensor([avg_test_loss], dtype=torch.float32, device="cuda")
-                torch.distributed.all_reduce(avg_test_loss, op=torch.distributed.ReduceOp.AVG, async_op=False)
-                avg_test_loss = avg_test_loss.item()
-                
+            # sync the avg_loss across all processes
+            avg_test_loss = self.sync_average_loss(avg_test_loss)
             self.on_stage_end(Stage.TEST, avg_test_loss, None)
         self.step = 0
         return avg_test_loss
@@ -1804,6 +1806,13 @@ class Brain:
         if torch.isfinite(loss):
             avg_loss -= avg_loss / self.step
             avg_loss += float(loss) / self.step
+        return avg_loss
+    
+    def sync_average_loss(self, avg_loss):
+        if torch.distributed.is_initialized():
+            avg_loss = torch.tensor([avg_loss], dtype=torch.float32, device=self.device)
+            torch.distributed.all_reduce(avg_loss, op=torch.distributed.ReduceOp.AVG, async_op=False)
+            avg_loss = avg_loss.item()    
         return avg_loss
 
     @contextmanager
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1002,10 +1002,10 @@ class Brain:
             loader_kwargs = self._train_loader_specifics(dataset, loader_kwargs)
         # This commented-out code block is useful when one can ensure
         # metric reporting is DDP-valid for VALID & EVAL datasets.
-        # elif self.distributed_launch:
-        #     loader_kwargs = sb.dataio.dataloader.distributed_loader_specifics(
-        #         self.distributed_launch, self.rank, dataset, loader_kwargs
-        #     )
+        elif self.distributed_launch:
+            loader_kwargs = sb.dataio.dataloader.distributed_loader_specifics(
+                self.distributed_launch, self.rank, dataset, loader_kwargs
+            )
         dataloader = sb.dataio.dataloader.make_dataloader(
             dataset, **loader_kwargs
         )
@@ -1746,6 +1746,9 @@ class Brain:
         if progressbar is None:
             progressbar = not self.noprogressbar
 
+        # Only show progressbar if requested and main_process
+        enable = progressbar and sb.utils.distributed.if_main_process()
+
         if not (
             isinstance(test_set, DataLoader)
             or isinstance(test_set, LoopedLoader)
@@ -1762,7 +1765,7 @@ class Brain:
             for batch in tqdm(
                 test_set,
                 dynamic_ncols=True,
-                disable=not progressbar,
+                disable=not enable,
                 colour=self.tqdm_barcolor["test"],
             ):
                 self.step += 1
@@ -1772,7 +1775,13 @@ class Brain:
                 # Debug mode only runs a few batches
                 if self.debug and self.step == self.debug_batches:
                     break
-
+            
+            # sync avg_test_Loss
+            if torch.distributed.is_initialized():
+                avg_test_loss = torch.tensor([avg_test_loss], dtype=torch.float32, device="cuda")
+                torch.distributed.all_reduce(avg_test_loss, op=torch.distributed.ReduceOp.AVG, async_op=False)
+                avg_test_loss = avg_test_loss.item()
+                
             self.on_stage_end(Stage.TEST, avg_test_loss, None)
         self.step = 0
         return avg_test_loss
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1418,9 +1418,6 @@ class Brain:
                 self.avg_train_loss = self.update_average(
                     loss, self.avg_train_loss
                 )
-                # TODO: need to benchmark the effect of this communication
-                # sync the avg_loss across all processes 
-                self.avg_train_loss = self.sync_average_loss(self.avg_train_loss)
                 t.set_postfix(train_loss=self.avg_train_loss)
 
                 if self.profiler is not None:
@@ -1447,6 +1444,8 @@ class Brain:
         # Run train "on_stage_end" on all processes
         # flush gradients
         self.zero_grad(set_to_none=True)
+        # sync the avg_loss across all processes 
+        self.avg_train_loss = self.sync_average_loss(self.avg_train_loss)
         self.on_stage_end(Stage.TRAIN, self.avg_train_loss, epoch)
         self.avg_train_loss = 0.0
         self.step = 0
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -79,11 +79,11 @@ def if_main_process():
     In standard mode, the process will not have `LOCAL_RANK` Unix var and will be
     authorized to run the I/O commands.
     """
-    if "LOCAL_RANK" in os.environ:
-        if os.environ["LOCAL_RANK"] == "":
+    if "RANK" in os.environ:
+        if os.environ["RANK"] == "":
             return False
         else:
-            if int(os.environ["LOCAL_RANK"]) == 0:
+            if int(os.environ["RANK"]) == 0:
                 return True
             return False
     return True
--- a/speechbrain/utils/distributed_metrics.py
+++ b/speechbrain/utils/distributed_metrics.py
@@ -74,7 +74,8 @@ def recursively_apply(func, data, *args, test_type=is_torch_tensor, error_on_oth
 def _gpu_gather_object(object: Any):
     output_objects = [None for _ in range(torch.distributed.get_world_size())]
     torch.distributed.all_gather_object(output_objects, object)
-    # all_gather_object returns a list of lists, so we need to flatten it
+    # return output_objects
+    # # all_gather_object returns a list of lists, so we need to flatten it
     return [x for y in output_objects for x in y]
 
 def _gpu_gather(tensor):
--- a/speechbrain/utils/distributed_metrics.py
+++ b/speechbrain/utils/distributed_metrics.py
@@ -1,0 +1,142 @@
+from typing import Any, Mapping
+import torch 
+
+def is_torch_tensor(tensor):
+    return isinstance(tensor, torch.Tensor)
+
+def is_namedtuple(data):
+    """
+    Checks if `data` is a `namedtuple` or not. Can have false positives, but only if a user is trying to mimic a
+    `namedtuple` perfectly.
+    """
+    return isinstance(data, tuple) and hasattr(data, "_asdict") and hasattr(data, "_fields")
+
+def honor_type(obj, generator):
+    """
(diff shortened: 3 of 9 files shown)


speechbrain/core.py
```diff
@@ -92,6 +92,57 @@
...
+    def initialized(self) -> bool:
+        "Returns whether the `DistributedState` has been initialized"
+        return self._shared_state != {}
+
+    @property
+    def use_distributed(self):
+        """
+        Whether the Trainer is configured for distributed training
+        """
+        return self.num_processes > 1
```
Adel-Moumen, line 143: This class is required since I need to retrieve the backend used by pytorch because some operations are not working on `Gloo`. For instance, using `NCCL` you'll need to use `all_gather_into_tensor` while with `Gloo` you need to use `all_gather`. 

I hope that in the future we will improve our DDP backend by improving this class.

speechbrain/utils/distributed.py
```diff
@@ -79,11 +80,11 @@ def if_main_process():
...
-        if os.environ["LOCAL_RANK"] == "":
+    if "RANK" in os.environ:
+        if os.environ["RANK"] == "":
             return False
         else:
-            if int(os.environ["LOCAL_RANK"]) == 0:
+            if int(os.environ["RANK"]) == 0:
                 return True
             return False
     return True
```
Adel-Moumen, line 90: Note for reviewer: this is temporary I needed to fix this.

speechbrain/utils/distributed.py
```diff
@@ -206,3 +207,45 @@ def ddp_init_group(run_opts):
...
+        tensor (nested list/tuple/dictionary of `torch.Tensor`):
+            The data to reduce.
+        reduction (`str`, *optional*, defaults to `"mean"`):
+            A reduction method. Can be of "mean", or "sum".
+
+    Returns
+    -------
+        The same data structure as `data` with all the tensors reduced.
+    """
+    from speechbrain.utils.distributed_metrics import recursively_apply
```
Adel-Moumen, line 227: we do the imports here due to circular import issues

speechbrain/utils/distributed_metrics.py
```diff
@@ -0,0 +1,263 @@
+"""SpeechBrain utility functions for distributed metrics.
```
TParcollet, line 1: Hum ... I struggle to see why we actually need all this. Are we using every single function in this file? If so, couldn't we just simplify it given our use case? it reads as a complicated file to do something that should be fairly simple to do? In particulare, why do we need to have something recursive here?

speechbrain/core.py
```diff
@@ -92,6 +92,57 @@
 }
 
 
+class DistributedState:
```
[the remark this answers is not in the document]
Adel-Moumen: Sure. I can move it. Maybe in `distributed.py` or `distributed_utils.py`.

speechbrain/core.py
```diff
@@ -1773,6 +1841,10 @@ def evaluate(
                 if self.debug and self.step == self.debug_batches:
                     break
 
+            # sync the avg_loss across all processes
+            avg_test_loss = sb.utils.distributed.reduce(
```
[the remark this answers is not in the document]
Adel-Moumen: Yep.

speechbrain/utils/distributed_metrics.py
Adel-Moumen: This is a valid remark. However, I don't see a way to make it lighter without harming the provided features. The purpose of using `recursively_apply` is to synchronize tensors/objects without having prior knowledge of the input data structure. If we know that we will always have a list of tensors, then it's easy; you just need to call `all_gather_into_tensor`. However, we do not have a standardized way of using metrics in SpeechBrain. Some metrics will expect a list of lists of strings, while some will expect a list of tensors. To cover such a variety, it is required to have a function that can work with arbitrary input data. The purpose of `recursively_apply` is really to first explore the data structure and to apply some function only on some objects. For instance, one purpose is to check if all the nested input data are tensors or not. If so, then you can simply call `all_gather_into_tensor`. Another is to apply the `reduce` function only on tensors. This makes things easier to deal with. Other than that, the remaining function aims to deal with objects or tensors. If you encounter a tensor, you need to use the appropriate function since there can be some mismatch with Gloo vs NCCL.

Ping @pplantinga, I'd like to get your feedback on that! thanks. 

BTW --- I will create a new `distributed_utils.py` file since `recursively_apply` and some other functions should be there because they are used in other context like for the `reduce` fn.

Adel-Moumen: Note: we see similar design patterns in other toolkits like ESPNet (https://github.com/espnet/espnet/blob/d561f1eaf1deac2aec3b86b68480de7c084f3f6e/espnet2/torch_utils/recursive_op.py#L9) but I think the way of doing it in speechbrain (hence in HF accelerate) is much better since you can reuse the code in other places without having to copy and paste some boilerplate code.

speechbrain/utils/distributed_metrics.py
```diff
@@ -0,0 +1,263 @@
...
+        The data on which to apply `func`
+    *args:
+        Positional arguments that will be passed to `func` when applied on the unpacked data.
+    main_type (`type`, *optional*, defaults to `torch.Tensor`):
+        The base type of the objects to which apply `func`.
+    error_on_other_type (`bool`, *optional*, defaults to `False`):
+        Whether to return an error or not if after unpacking `data`, we get on an object that is not of type
+        `main_type`. If `False`, the function will leave objects of types different than `main_type` unchanged.
+    **kwargs (additional keyword arguments, *optional*):
+        Keyword arguments that will be passed to `func` when applied on the unpacked data.
```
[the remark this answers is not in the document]
Adel-Moumen: done

speechbrain/core.py
```diff
@@ -1442,7 +1498,12 @@ def _fit_train(self, train_set, epoch, enable):
                     steps_since_ckpt = 0
 
         # Run train "on_stage_end" on all processes
-        self.zero_grad(set_to_none=True)  # flush gradients
+        # flush gradients
+        self.zero_grad(set_to_none=True)
+        # sync the avg_loss across all processes
+        self.avg_train_loss = sb.utils.distributed.reduce(
+            torch.tensor(self.avg_train_loss)
```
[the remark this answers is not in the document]
Adel-Moumen: Very good question. I'll need to test but I should add `self.device` in order to make sure that it works as intended.

speechbrain/core.py
```diff
@@ -1002,10 +1058,10 @@ def make_dataloader(
             loader_kwargs = self._train_loader_specifics(dataset, loader_kwargs)
         # This commented-out code block is useful when one can ensure
```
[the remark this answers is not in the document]
Adel-Moumen: yep

speechbrain/utils/distributed.py
```diff
@@ -206,3 +207,45 @@ def ddp_init_group(run_opts):
         rank=rank,
         timeout=datetime.timedelta(seconds=7200),
     )
+
+
+def reduce(tensor, reduction="mean"):
+    """Recursively reduce the tensors in a nested list/tuple/dictionary of lists of tensors
```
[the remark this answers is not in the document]
Adel-Moumen: Done. Could you please tell me if you are ok now ?

speechbrain/utils/distributed.py
```diff
@@ -12,9 +13,62 @@
...
 
+class DistributedState:
+    """
+    Singleton class that has information about the current training environment and functions to help with process
+    control.
+    """
+
+    _shared_state = dict()
+
+    def __init__(self, device: str = "cpu", distributed_backend=None):
```
pplantinga, line 29: I think the singleton pattern in python typically uses `__new__` rather than `__init__`

speechbrain/utils/distributed.py
```diff
@@ -206,3 +260,66 @@ def ddp_init_group(run_opts):
...
+        if not state.use_distributed:
+            return cloned_tensor
+        else:
+            from torch.distributed import ReduceOp
+
+            torch.distributed.all_reduce(cloned_tensor, ReduceOp.SUM)
+
+        # we cannot use 'ReduceOp.AVG` since it is unavailable for gloo backend
+        if reduction == "mean":
+            cloned_tensor /= state.num_processes
```
pplantinga, line 316: Do we need `DistributedState` here? Couldn't we use https://pytorch.org/docs/stable/distributed.html#torch.distributed.get_world_size?

speechbrain/utils/distributed_metrics.py
```diff
@@ -0,0 +1,159 @@
...
+        if tensor.ndim == 0:
+            tensor = tensor.clone()[None]
+
+        # Can only gather contiguous tensors
+        if not tensor.is_contiguous():
+            tensor = tensor.contiguous()
+
+        if (
+            state.distributed_backend is not None
+            and state.distributed_backend != "gloo"
```
pplantinga, line 47: Again, couldn't we use https://pytorch.org/docs/stable/distributed.html#torch.distributed.get_backend here rather than state?

speechbrain/utils/distributed_metrics.py
pplantinga: I think @TParcollet  was commenting on `distributed_metrics.py` rather than `recursively_apply`. And while I'm not sure about this specific file, it does seem there may be some unnecessary complexity here. See overview comment for more thoughts.

speechbrain/utils/metric_stats.py
```diff
@@ -112,7 +116,7 @@ def append(self, ids, *args, **kwargs):
                 scores = multiprocess_evaluation(
                     metric=self.metric, n_jobs=self.n_jobs, **kwargs
                 )
-
+        scores = sb.utils.distributed_metrics.gather_for_metrics(scores)
```
pplantinga, line 119: I thought we had agreed that `gather`-ing in the `summarize` method was better than in `append`

pplantinga (commented):
Supporting eval/validation with DDP is an important step, thanks for tackling it @Adel-Moumen . Tests for DDP and making metrics work across processes are crucial. However I think there may be some unnecessary complexity here. For example, I'm not convinced the `DistributedState` singleton adds anything on top of what `torch.distributed` already does.

Some options I can think of to address complexity:
* add `accelerate` as a dependency and use their version of these functions, so we don't have to maintain it.
* add `torch-metrics` as a dependency, https://github.com/Lightning-AI/torchmetrics, they have most of the metrics with distributed support already, again we don't have to maintain it.
* Force a relatively simple and consistent format for the metrics -- then we don't need much complexity to sync other than `torch.distributed`

I know we're trying to avoid adding dependencies but these ones are relatively small and well worth it.

speechbrain/utils/distributed.py
Adel-Moumen: we could but I think the point of this PoC is to show what could looks like the next SB distributed code & distributed metrics. I introduced this class mostly because I thought that we had so many places that rely on the same code and maybe it would be better for us to start refactoring a bit everything under a same class to propose convenient attributes/functions.

speechbrain/utils/distributed_metrics.py
Adel-Moumen: > I think @TParcollet was commenting on `distributed_metrics.py` rather than `recursively_apply`. And while I'm not sure about this specific file, it does seem there may be some unnecessary complexity here. See overview comment for more thoughts.

It was related to this file. I moved `recursively_apply` in a different file but at the beginning `recursively_apply` was in `distributed_metrics.py` hence titouan's comment

speechbrain/utils/metric_stats.py
Adel-Moumen: I think 'better' is the issue. Yes, it is performance-wise. However, the issue is that not all metrics have a `summarize` function, like for instance `def minDCF`. Furthermore, what would happen if the user needs to access internal states of metrics now? If they never run the `summarize` function, then it means that the internal state will only be for the current rank process. When doing unit testing, I found it particularly useful to access internal states before running `summarize`. I think doing sync in `append` is the most straightforward way to sync metrics and also the most natural way for an end-user.

I also took a look at other toolkits and found that they were also synchronizing during the `append` stage.

speechbrain/utils/distributed_metrics.py
Adel-Moumen: I think you are right on that! Maybe i can wipe up the singleton. Wdyt ? (as I said in a previosu comment, I thought the singleton to be particularly useful in the future if we want to make our codebase compatible with other features like FSDP but Im fine to revert it)

speechbrain/core.py
```diff
@@ -822,6 +822,13 @@ def __init__(  # noqa: C901
                         "all other subprocess were killed."
                     )
 
+        if self.device == "cpu" and self.distributed_backend == "nccl":
+            sb.utils.distributed.DistributedState(self.device)
+        else:
+            sb.utils.distributed.DistributedState(
+                self.device, self.distributed_backend
+            )
```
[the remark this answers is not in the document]
Adel-Moumen: I removed the class DistributedState since I agree that torch.distributed is already providing the needed features. At the beginning I was thinking about using this class for the future if we want to support other backend such as TPU, Megatron etc but I think it's not worth it to do it now. (the idea was to abstract some key functions that can differ depending on the backend used).

speechbrain/utils/distributed.py
Adel-Moumen: I removed the class DistributedState since I agree that torch.distributed is already providing the needed features. At the beginning I was thinking about using this class for the future if we want to support other backend such as TPU, Megatron etc but I think it's not worth it to do it now. (the idea was to abstract some key functions that can differ depending on the backend used).

speechbrain/utils/distributed_metrics.py
Adel-Moumen: I did the changes. Thanks for your comment. :)

speechbrain/utils/distributed.py
Adel-Moumen: I removed the class DistributedState since I agree that torch.distributed is already providing the needed features. At the beginning I was thinking about using this class for the future if we want to support other backend such as TPU, Megatron etc but I think it's not worth it to do it now. (the idea was to abstract some key functions that can differ depending on the backend used).

Adel-Moumen:
> However I think there may be some unnecessary complexity here. For example, I'm not convinced the `DistributedState` singleton adds anything on top of what `torch.distributed` already does.

Hi @pplantinga, indeed. It does not in the current stage. As I said in my replies, this PoC aimed at showing one way of handling DDP informations in the future. Indeed, depending on your backend, you may need to follow different steps to achieve the same thing. Having a convenient dataclass would have helped to have a single interface and behind the hood all the logic required by those backends. 

With respect to your comments, I decided to remove this class as it can be fully replaced by `torch.distributed` functions. 

> Some options I can think of to address complexity

Regarding the complexity, I have some hard time to understand where the complexity boils down. I am slightly biased since I am the one who spent the most time understanding how to synchronize metrics. The only complexity is due to the different backends that we can encounter in speechbrain (e.g. gloo vs nccl) which comes with different utilities functions but I don't think you can simplify unless we drop gloo support :) (which i strongly disagree). 

> add accelerate as a dependency and use their version of these functions, so we don't have to maintain it.
    add torch-metrics as a dependency, [Lightning-AI/torchmetrics](https://github.com/Lightning-AI/torchmetrics), they have most of the metrics with distributed support already, again we don't have to maintain it.
    Force a relatively simple and consistent format for the metrics -- then we don't need much complexity to sync other than torch.distributed

Again, I am still biased since I did the implementation but I really dont think we need a third package to handle the metrics. As I demonstrated, synchronizing can be summarized in < 250 lines of code (excluding docstring/tests). If you do install accelerate for metrics, you'll get thousands of lines of code that are useless mostly because they do support TPUs, Megatron, FSDP etc etc etc. I like the idea of having a SpeechBrain Trainer that does not rely on many toolkits and that can be easily hackable.

Adel-Moumen:
Please jump in the discussion (one more time please its an important topic :p) @TParcollet @asumagic :) (note: i made some changes based on your comments [thanks again guys (@TParcollet @asumagic @pplantinga) !] and feel free to get back to me if you think that now I should move on to more tests/reporting more results etc). :)

speechbrain/utils/metric_stats.py
pplantinga: Okay, you've convinced me I guess.

lucadellalib:
The small variations in CER and WER might be related to the fact that when the dataset is not divisible by the number of processes, a few samples are duplicated : https://github.com/pytorch/pytorch/issues/25162

Adel-Moumen:
> The small variations in CER and WER might be related to the fact that when the dataset is not divisible by the number of processes, a few samples are duplicated : [pytorch/pytorch#25162](https://github.com/pytorch/pytorch/issues/25162)

Indeed! Good point.

PR #2529, opened 2024-04-26
Fix keys name for Transformer

Adel-Moumen:
## What does this PR do?

This PR solves an issue due to #2489. Indeed, this PR modify the name of a key in decoder transformer `self.mutihead_attn` to `self.multihead_attn`. Doing so breaks the loading of state dict since it does not recognize the previous key. In order to solve this problem, I introduce a new function called `map_old_state_dict_weights` which is directly applied within `torch_recovery`, `average_checkpoints` and `_load_from_state_dict`. The first is when you are loading a checkpoint, the second when you are trying to avg multiple checkpoints (the issue here is that you need to make sure that before doing the avg, every ckpts has the same keys), and the latter when you are loading the state_dict directly from the object which can be the case in our codebase (i.e. bypassing checkpointer). 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/utils/checkpoints.py
+++ b/speechbrain/utils/checkpoints.py
@@ -46,6 +46,7 @@ Example
 
 Authors
  * Aku Rouhe 2020
+ * Adel Moumen 2024
 """
 
 import collections
@@ -57,6 +58,7 @@ import pathlib
 import shutil
 import time
 import warnings
+from typing import Dict
 
 import torch
 import yaml
@@ -76,6 +78,27 @@ CKPT_PREFIX = "CKPT"
 METAFNAME = f"{CKPT_PREFIX}.yaml"  # Important that this is not .ckpt
 PARAMFILE_EXT = ".ckpt"  # ...because these files will be
 
+# Some keys names change between versions, so we need to map them
+# to the correct key names.
+STATE_KEY_MAPPING: Dict[str, str] = {
+    "mutihead_attn": "multihead_attn",
+}
+
+
+def _modify_state_dict(state_dict):
+    """Modifies the state_dict to match the current key names stored in `STATE_KEY_MAPPING`"""
+    for key in list(state_dict.keys()):
+        for state_key in STATE_KEY_MAPPING:
+            if state_key in key:
+                old_state = state_dict.pop(key)
+                new_key = key.replace(state_key, STATE_KEY_MAPPING[state_key])
+                state_dict[new_key] = old_state
+                warnings.warn(
+                    f"Key {key} has been renamed to {new_key} in the state_dict."
+                    "Make sure to update your model's state_dict."
+                )
+    return state_dict
+
 
 def torch_recovery(obj, path, end_of_epoch):
     """Loads a torch.nn.Module state_dict from the given path instantly.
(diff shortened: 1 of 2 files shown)


speechbrain/utils/checkpoints.py
```diff
@@ -94,10 +131,13 @@ def torch_recovery(obj, path, end_of_epoch):
     """
     del end_of_epoch  # Unused
     device = "cpu"
+
+    state_dict = torch.load(path, map_location=device)
+    state_dict = map_old_state_dict_weights(state_dict, KEYS_MAPPING)
```
asumagic, line 136: Shouldn't we have a more generic `state_dict` fixup function that would call into `map_old_state_dict_weights`, even if the general function only calls that one for now?

speechbrain/utils/checkpoints.py
```diff
@@ -75,6 +77,41 @@
...
+    """
+    Maps the keys in the old state dictionary according to the provided mapping,
+    under the given prefix.
+
+    Parameters
+    ----------
+    state_dict : dict
+        The old state dictionary to be mapped.
+    mapping : dict
+        A dictionary specifying the mapping between old and new keys.
```
asumagic, line 98: It is unclear the extent of what this is, in particular, to my understanding:

- This can affect any part of the key, which could look like `somemodule.somesubmodule.somefield` rather than just `somefield`
- This will do partial matches on keys (e.g. `"a": "b"` will turn `_a` into `_b`)

The behavior should be documented to avoid accidentally including renames with too many side effects.

speechbrain/utils/checkpoints.py
```diff
@@ -75,6 +77,41 @@
...
+
+    Returns
+    -------
+    dict
+        The modified state dictionary with mapped keys.
+    """
+    for checkpoint_name, attribute_name in mapping.items():
+        for full_checkpoint_name in list(state_dict.keys()):
+            if checkpoint_name in full_checkpoint_name:
+                full_attribute_name = full_checkpoint_name.replace(
```
asumagic, line 108: Should we have a warning here?

speechbrain/utils/checkpoints.py
Adel-Moumen: I don't understand your question sorry haha

asumagic: Having a function where we could add other `state_dict` fixes in the future.

speechbrain/utils/checkpoints.py
Adel-Moumen: Yep. Either a warning or a log in debug mode.

speechbrain/utils/checkpoints.py
Adel-Moumen: so you mean something like:
```python
def foo(state_dict):
	state_dict(state_dict)
```

so that someone can do : 
```python
def foo(state_dict):
	bar(state_dict)
	state_dict(state_dict)
```

easily?

asumagic:
Seems like the PR current doesn't resolve issues with HF models loaded through `inference`, see https://github.com/speechbrain/speechbrain/issues/2549#issuecomment-2149528797

asumagic:
> Seems like the PR current doesn't resolve issues with HF models loaded through `inference`, see [#2549 (comment)](https://github.com/speechbrain/speechbrain/issues/2549#issuecomment-2149528797)

To solve this, I made `torch_parameter_transfer` use the same logic, which I think is OK.

Other potentially problematic calls:

- `average_checkpoints()` in `checkpoints.py` defaults to `parameter_loader=torch.load`.

Other calls that do load model `state_dict`s(?) but _shouldn't be an issue_:

- In `_cycliclrloader`

Other calls that use `torch.load` for other things that _shouldn't be an issue_:

- In `InputNormalization` (arbitrary dict)
- In `schedulers.py` (arbitrary dict)
- In `huggingface_transformers/*` (non-SB `state_dict`s)
- In `k2_integration/*` (seems like graphs and other k2 stuff)
- In `inference/TTS.py` (arbitrary dict)
- In `aligner.py` (arbitrary dict)
- In `recipes/Voicebank/ASR/CTC/train.py` (not used and local, probably not a worry)

Adel-Moumen:
sounds good to me.

asumagic (approved):

Issue #2526, opened 2024-04-26
Adapters + LLama -- re-design.


TParcollet:
### Describe the bug

It's not a bug, just a discussion. I think that people including @Adel-Moumen @poonehmousavi @mravanelli and maybe @pplantinga @asumagic may want to engage. 

Adapters, or, more generally, altering an existing pre-trained model (you can see it as an object originating from the Pretrainer or checkpointer) is something becoming more and more common. Due to this, we, imho, **must** define a proper design in SpeechBrain to do so. Recently, I implemented LoRA and Houlsby for our Transformer, on my side. But I also realised that @poonehmousavi did some work for Llama [here](https://github.com/speechbrain/speechbrain/blob/b12b6837a23590fb267b6adcd1a0640f5a1070ca/speechbrain/lobes/models/huggingface_transformers/llama2.py#L220). I don't think we are doing this correctly. The code in Llama 2, for instance, might be hard to understand, and some functions (like the one replacing a module in an existing module) should be generalised and considered a general SpeechBrain util. My strategy would be to create an Adapters.py in lobes where we could put everything relating to them, instead of having them randomly appearing in lobes files. 

What do you folks think? 

### Expected behaviour

Respect the Zen of SpeechBrain.

poonehmousavi:
I totally agree.. what we have done for LLama2 is  the QLora(Qunatization+Lora) and only applicable for LLama2. I think with this trend of public LLM released so soften, and all of them need Lora for fine-tuning, we need to have an unified class that could handle l]Lora(and Qlora) for different models. In HF, sft_trainer.py could handle all

mravanelli:
I think it is very important too. We can support at least LoRA and adapters. I'm not sure what could be the best way to go to support it elegantly. One idea could be to implement a wrapper (e.g., Adapter) to apply to our models,  which can plug the necessary new modules. However, it seems to me quite hard to create something easy to use and flexible at the same time. Any idea from the community?

pplantinga:
I have used PEFT to apply Lora to an existing model and it was pretty straightforward to use. You just pass the model and it automatically replaces all relevant layers with Lora layers. We could do something similar

pplantinga:
We could even import and use peft if the dependencies are light

Adel-Moumen:
Hi @TParcollet, I agree with your strategy. 

[...]

PR #2534, opened 2024-04-30
Adding adapters to SpeechBrain (Code from Samsung AI Center Cambridge)

TParcollet:
## What does this PR do?

Based on this #2526, this PR is a first attempt at adding Adapters to any SB model. This will only work if PreTrainer is used and not checkpointer. Indeed, the checkpointer will try to reload after the state_dict has been modified. So you need to 1. Instanciate the Brain, 2. Call the PreTrainer; 3. add the adapters; 4. call fit. An example is:

`

        asr_brain = ASR(
            modules=hparams["modules"],
            opt_class=hparams["Adam"],
            hparams=hparams,
            run_opts=run_opts,
        )

        # adding objects to trainer:
        asr_brain.tokenizer = hparams["tokenizer"]

        # Load the pretrained model.
        run_on_main(hparams["pretrainer"].collect_files)

        from speechbrain.lobes.models.Adapters import (
            add_adapters_to_linear_in_model,
        )
        from speechbrain.lobes.models.Adapters import HoulsbyAdapterLinear

        add_adapters_to_linear_in_model(
            model=asr_brain.modules.Transformer,
            adapter_class=HoulsbyAdapterLinear,
            projection_size=32,
        )

        # Training
        asr_brain.fit(
            asr_brain.hparams.epoch_counter, train_data, valid_data,
        )
`


pplantinga:
I don't think this is quite the right approach because I don't think it allows for stopping/restarting which is part of the point of checkpointing. Instead, the checkpointer should store the LoRA'd model, not the pretrained model. Ideally it would even only store the LoRA weights (and any updated weights) and not the whole model, making for very small checkpoints and faster saving. Example:

```yaml
add_adapters: !name:speechbrain.lobes.models.Adapters.add_adapters_to_linear_in_model
  adapter_class: !name:speechbrain.lobes.models.Adapters.HoulsbyAdapterLinear
  projection_size: 32

pretrainer: !new:speechbrain....Pretrainer
  loadables:
    transformer: !ref <Transformer>
```

```python
# Load the pretrained model.
run_on_main(hparams["pretrainer"].collect_files)
hparams["add_adapters"](hparams["Transformer"])

asr_brain = ASR(
    modules=hparams["modules"],
    opt_class=hparams["Adam"],
    hparams=hparams,
    run_opts=run_opts,
    checkpointer=hparams["checkpointer"], # Checkpointer loads LoRA weights only and applies them
)

asr_brain.fit(
    asr_brain.hparams.epoch_counter, train_data, valid_data,
)

TParcollet:
It does allow for stop and restart! The only problem is indeed that you store the whole model, however, I don't think it's an issue because in-fine you may simply don't know where to put the pre-trained adapters in the model if they are not applied to every linear layer for instance. How would you proceed?

pplantinga:
Tbh I think PEFT handles this perfectly, perhaps we should lift their code wholesale.

TParcollet:
You mean depend on another Huggingface library?

pplantinga:
My opinion is we should just add it as a dependency, but I understand the objections to it. So instead we could just copy the parts of the code that make sense into speechbrain.

TParcollet:
If you could give me a neat example of an integration of PEFT, I could be convinced.

poonehmousavi:
The problem with peft is when we want to load the model from the speechbrain checkpoint. It is a mess to make it work and also it could cause the problem when using different version of peft. But maybe we could find a way to do this in a cleaner way.

speechbrain/lobes/models/Adapters.py
pplantinga, line 1: Shouldn't this go in `nnet` rather than `lobes`?

TParcollet: Good question. it's unclear because Adapters can be considered as "entire models" coming from the literature. But I think I agree that they can also be seen as small components. I'd be happy if you could help with the get_model like for PEFT. From your previous PR, I liked the fact that we can rely on the larger Adapter base from PEFT -- I am wondering if there isn't a way to combine both ...

TParcollet: I personally like the fact that with my function, you can actually specify what part of the Brain.modules or whatever model you want to put Adapters on. But I'd be happy to see something else.

TParcollet:
@poonehmousavi @pplantinga @mravanelli @Adel-Moumen @asumagic what should we do here?

pplantinga:
I'm working on this, but I've been traveling this week and last. I hope to have a reasonable PR by the end of next week

pplantinga:
Superseded by #2563

PR #2543, opened 2024-05-07
Add recipe for training whisper with PEFT - LoRA

pplantinga:
Proof-of-concept PR for integrating PEFT for adapter-based training. 

Steps:

- [x] It works, you can train a model with LoRA
- [ ] Checkpointing only saves/loads trained parameters (smaller/faster checkpoints)
- [ ] Prove hackability (can add new adapter types)


TParcollet:
Hey Peter, for now I don't see any difference with what I proposed. The only ones being: 1. Your has more base model; 2. Yours also loose entirely the sovereignty of the code to a heavily changing external library (We know from experience that HF loves changing their API).

PR #2550, opened 2024-05-16
Add recipe for audio/speech LLM (ltu-as with llama3)

BenoitWang:
Hi @mravanelli, here's the ltu-as PR as discussed. I am collecting several new datasets and will start a new round of training but this may take time, so meanwhile I start this PR and carry on little by little. @poonehmousavi you are welcome to review the PR as well 😊.

## What does this PR do?

1. Add a recipe for training the [LTU-AS](https://arxiv.org/pdf/2309.14405.pdf) model (an LLM that jointly understands audio and speech).
2. Slight modifs to the LinearWarmupScheduler class.
3. Adapt the multiwoz llama2 recipe to the latest changes.

## To be done

- [ ] For now, the model is trained with only half the data than in the paper. Though access to certain datasets seems limited,  a new training round needs to be carried out with more datasets being collected.
- [ ] Prepare downloadable json files that facilitate the data preparation stage.
- [ ] Better to add a tiny validation set for stage 1 and 2.
- [ ] An evaluation needs to be implemented at the end of stage 3 and the evaluation data needs to be prepared.
- [ ] Upload training logs and prepare a huggingface interface.
- [ ] Recipe tests.
- [ ] Update results and training details in readme.

### DIFF
--- a/recipes/MultiWOZ/response_generation/llama2/hparams/train_llama2.yaml
+++ b/recipes/MultiWOZ/response_generation/llama2/hparams/train_llama2.yaml
@@ -79,6 +79,7 @@ llama2_model: !new:speechbrain.lobes.models.huggingface_transformers.llama2.LLAM
     top_k: !ref  <top_k>
     top_p: !ref <top_p>
     with_peft: True
+    use_4bit: True
 
 epoch_counter: !new:speechbrain.utils.epoch_loop.EpochCounter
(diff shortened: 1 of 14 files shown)


mravanelli:
Thank you @BenoitWang for this contribution. It looks like some tests are failing. Could you please take a look?

PR #2558, opened 2024-05-30
Add `quirks.py` for GPU/platform-specific quirks; disable CuDNN benchmarking on ROCm

### LINKED ISSUE #2552: AMD ROCm: `torch.backends.cudnn.benchmark` should be set to `False` by default on ROCm
### Describe the bug

With AMD HIP, `torch.backends.cudnn.benchmark` defaults to `True`. On CUDA, it defaults to `False`.

While this is an upstream PyTorch decision (or bug), having `benchmark` default to `True` by default is probably a bad idea in our case, because we have wildly different input shapes. Since CNN kernel benchmarking is performed on a per-shape basis, it performs those slow benchmarks very often, so training times are worsened rather than improved.

Probably should add some common per-backend quirks file for this kind of change in SB?

### Expected behaviour

On AMD ROCm, it should be set to `False`, probably with a warning since we break the default.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

asumagic:
## What does this PR do?

Fixes #2552

The code should be fairly well documented. The idea is that SpeechBrain now would take a more structured/documented approach to overriding PyTorch's global configuration.

So-called "quirks" represent an individual configuration change, and it may be unconditional or depend on hardware. The user is notified by an info log when quirks are applied.

This adds a quirk to disable CuDNN benchmarking on ROCm as it degrades performance for a majority of cases in SpeechBrain.

Quirks can be explicitly opted out from by setting an environment variable, `SB_DISABLE_QUIRKS`.

Example:

```
$ SB_DISABLE_QUIRKS=disable_cudnn_benchmarking python3
Python 3.9.18 (main, Sep 11 2023, 13:41:44) 
[GCC 11.2.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging; logging.basicConfig(level=logging.INFO); import speechbrain
INFO:speechbrain.utils.quirks:Applied quirks (see `speechbrain.utils.quirks`): [disable_jit_profiling, allow_tf32]
INFO:speechbrain.utils.quirks:Excluded quirks specified by the `SB_DISABLE_QUIRKS` environment (comma-separated list): [disable_cudnn_benchmarking]
```

<details>

</details>

## PR review

<details>

</details>


asumagic:
Some code already sets things like PyTorch globals, e.g.:

```python
torch._C._jit_set_profiling_executor(False)
torch._C._jit_set_profiling_mode(False)
```

Maybe all of this stuff should be grouped in the `quirks.py` added in this PR so that it's much easier to search for (perhaps also the `torchaudio` backend check stuff). Those things should be logged as debug too.

asumagic:
Actually this needs some more thoughts behind logging. A debug log level would be more preferable, but even in the case of training this is called _before_ the logging is properly set up.

asumagic:
BTW heads up @Adel-Moumen since this is moving the `allow_tf32` code. Nothing should be functionally different by default, though.

Adel-Moumen:
LGTM. Before merging, you need to add a proper documentation as it's seems to be something that the user should be aware of.

asumagic:
> LGTM. Before merging, you need to add a proper documentation as it's seems to be something that the user should be aware of.

Where do you suggest that documentation to be/what do you suggest that it covers?  
I'd agree (though I think the file is fairly self-descriptive on its own), but I'm not sure where to put it in a way that's easily discovered.

Adel-Moumen:
In `docs/experiment.md` I would say

PR #2560, opened 2024-06-01
Updating CommonVoice CTC English and Conformer English

TParcollet:
The CTC en recipes and Conformer EN recipes were way out of date. 
Now we have something a bit better.
Models are training, i'll update the results.

- [x] update code
- [x] train models
- [x] verify dynamic batching does not break legacy code
- [x] update readme with results


TParcollet:
@Adel-Moumen is ready to review and merge. The conformer WER is not there yet, but I'll edit the readme directly on the main repo once the decoding is done :p

recipes/CommonVoice/ASR/transformer/train.py
```diff
@@ -269,15 +269,21 @@ def dataio_prepare(hparams, tokenizer):
...
         csv_path=hparams["test_csv"],
         replacements={"data_root": data_folder},
     )
 
     # We also sort the validation data so it is faster to validate
-    test_data = test_data.filtered_sorted(sort_key="duration")
+    test_data = test_data.filtered_sorted(
+        sort_key="duration",
+        key_max_value={"duration": hparams["avoid_if_longer_than_val_test"]},
+    )
```
Adel-Moumen, line 286: I think we should definitely add a warning/log saying that we are skipping files that have a duration > to `avoid_if_longer_than_val_test`.

Adel-Moumen (commented):
hi, overall lgtm. I leaved one comment about `avoid_if_longer_than_val_test`.

recipes/CommonVoice/ASR/transformer/train.py
TParcollet: Do it :p

TParcollet:
I merge cuz I'm the boss blblblblb.

PR #2540, opened 2024-05-06
Fix recipe test errors in Latent Diffusion and pre-commit tests

flexthink:
## What does this PR do?

Fixes #<issue_number>

<details>
  <summary><b>Before submitting</b></summary>

- [x] Did you read the [contributor guideline](https://speechbrain.readthedocs.io/en/latest/contributing.html)?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing [tests](https://github.com/speechbrain/speechbrain/tree/develop/tests) pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Does your code adhere to project-specific code style and conventions?

</details>

## PR review

<details>
  <summary>Reviewer checklist</summary>

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified
- [ ] Confirm that the changes adhere to compatibility requirements (e.g., Python version, platform)
- [ ] Review the self-review checklist to ensure the code is ready for review

</details>

PR #2598, opened 2024-07-09
Add simple wrappers for unet/diffwave for diffusion

asumagic:
## What does this PR do?

Linter-induced refactors have had the consequences that some models which used `**kwargs` and discarded those would fail due to additional keyword arguments they did not expect.

The prior solution for this issue was #2540 but after some discussion with @tparcollet we agreed to try this wrapper approach.

The motivation towards this wrapper approach IMO is that the proposed solution to fix this at diffusion level would indeed, theoretically, work out of the box.  
However, to my understanding, it essentially relied on implicitly dropping extraneous parameters, even if they were passed through by the caller of the diffusion module.  
This is problematic, because this may not always result in an error in case of a typo (e.g. `length` vs `lengths`, where the wrapped module only takes it optionally).  
So we preferred a more explicit solution that is less likely to accidentally eat arguments.

Tested:

- training TTS on LJSpeech for a few steps with no errors
- training on AudioMNIST for a few epochs with no errors (with both yaml)

Supersedes #2540 (EDIT: was typo'd, oops)

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/recipes/AudioMNIST/diffusion/hparams/train.yaml
+++ b/recipes/AudioMNIST/diffusion/hparams/train.yaml
@@ -192,7 +192,7 @@ emb_speaker: !new:speechbrain.nnet.embedding.Embedding
     embedding_dim: !ref <speaker_emb_dim>
 
 diffusion: !new:speechbrain.nnet.diffusion.DenoisingDiffusion
-    model: !ref <unet>
+    model: !ref <unet.diffusion_forward>
     timesteps: !ref <train_timesteps>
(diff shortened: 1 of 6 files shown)


TParcollet:
Args are much more explicit and safe with that implementation, I like it.

mravanelli (approved):

mravanelli:
Thank you @asumagic, 
I tested it and everything now works!

PR #2609, opened 2024-07-12
Pin transient dep conllu to 4.x to workaround broken upstream

asumagic:
## What does this PR do?

CI is currently broken on develop due to a transient dependency of flair, which is installed as part of the CI process.

Reported the issue upstream, can unpin once it is solved there: https://github.com/EmilStenstrom/conllu/issues/96

If the upstream issue is fixed before this is merged feel free to close.


asumagic:
Should be fixed upstream, rerunning CI to check

PR #2617, opened 2024-07-17
Update scorer.py

raotnameh:
This PR does not do anything Just avoids an edge case. 

- This is just to avoid an edge case in the case if candidates in log_probs is less than int(beam_size * self.scorer_beam_scale)

- Error I faced.  

                    _, candidates = log_probs.topk(                                                                                                                                                                                                                                                        
                        RuntimeError: selected index k out of range

### DIFF
--- a/speechbrain/decoders/scorer.py
+++ b/speechbrain/decoders/scorer.py
@@ -1202,9 +1202,19 @@ class ScorerBuilder:
             score, new_memory[k] = impl.score(inp_tokens, memory[k], None, attn)
             log_probs += score * self.weights[k]
 
+        # This is just to avoid an edge case in the case if candidates in log_probs is less than int(beam_size * self.scorer_beam_scale)
+        '''
+        Error I faced.  
+                    _, candidates = log_probs.topk(                                                                                                                                                                                                                                                        
+                        RuntimeError: selected index k out of range  
+        '''
+        if log_probs.shape[1] < int(beam_size * self.scorer_beam_scale):
+            sbc = int(log_probs.shape[1]) # scorer beam scale.
+        else: 
+            sbc = int(beam_size * self.scorer_beam_scale) # scorer beam scale.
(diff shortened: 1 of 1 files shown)


speechbrain/decoders/scorer.py
```diff
@@ -1252,10 +1252,18 @@ def score(self, inp_tokens, memory, attn, log_probs, beam_size):
...
             log_probs += score * self.weights[k]
 
+        # This is just to avoid an edge case in the case if candidates in log_probs is less than int(beam_size * self.scorer_beam_scale)
+        """
+        Error I faced.
+                    _, candidates = log_probs.topk(
+                        RuntimeError: selected index k out of range
+        """
+        if log_probs.shape[1] < int(beam_size * self.scorer_beam_scale):
+            sbc = int(log_probs.shape[1])  # scorer beam scale.
```
pplantinga: Can we use `-1` here for the dimension, just in case this function has to handle a different shape?
```suggestion
        if log_probs.shape[-1] < int(beam_size * self.scorer_beam_scale):
            sbc = int(log_probs.shape[-1])  # scorer beam scale.
```

speechbrain/decoders/scorer.py
```diff
@@ -1252,10 +1252,18 @@ def score(self, inp_tokens, memory, attn, log_probs, beam_size):
             score, new_memory[k] = impl.score(inp_tokens, memory[k], None, attn)
             log_probs += score * self.weights[k]
 
+        # This is just to avoid an edge case in the case if candidates in log_probs is less than int(beam_size * self.scorer_beam_scale)
+        """
+        Error I faced.
+                    _, candidates = log_probs.topk(
+                        RuntimeError: selected index k out of range
+        """
```
pplantinga: I don't think this comment is needed in the code.

```suggestion
```

PR #2510, opened 2024-04-16
Feature/ksponspeech

ddwkim:
## What does this PR do?

This updates KsponSpeech recipe according to the recent changes in Speechbrain.

Not only conformer-medium model is updated, this PR includes new models, such as conformer-small, branchformer-medium, which can serve as new benchmarks for the dataset and models.

Tokenizers and transformer language model used previously trained ones.

Other updates include metadata build script using multiprocessing, and unzip script that follows recent KsponSpeech directory structure updates.

These are all CTC-Attention models, and training process of conformer-transducer model is underway for streaming cases.

We are consistently seeing the demand for KsponSpeech models from the community, as can be seen in issues and model downloads from the huggingface hub.

I believe the continuous updates of the KsponSpeech models will benefit the research community and the industry as a whole.

There are several other things that should be accomplished before this PR is completed.

- [x] Update huggingface speechbrain models. Models in my repository are already updated, and hence simply copying them to the speechbrain hub will be enough
- [x] Upload train logs to speechbrain dropbox
- [x] Update Colab and documentations to meet the current changes

<details>
  <summary><b>Before submitting</b></summary>

- [ ] Did you read the [contributor guideline](https://speechbrain.readthedocs.io/en/latest/contributing.html)?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing [tests](https://github.com/speechbrain/speechbrain/tree/develop/tests) pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Does your code adhere to project-specific code style and conventions?

</details>

## PR review

<details>
  <summary>Reviewer checklist</summary>

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified
- [ ] Confirm that the changes adhere to compatibility requirements (e.g., Python version, platform)
- [ ] Review the self-review checklist to ensure the code is ready for review

</details>

PR #2622, opened 2024-07-19
Revert "Feature/ksponspeech"

mravanelli:
Reverts speechbrain/speechbrain#2510


mravanelli:
As discussed privately, this PR has the be reverted as not deeply testes.

PR #2630, opened 2024-07-26
Update pythonapp.yml

fabiocat93:
Extending tests to python v.3.10, 3.11

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -14,7 +14,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                python-version: [3.8, 3.12]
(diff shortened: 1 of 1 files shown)


asumagic:
The issue is that GitHub actions aren't unlimited. If their docs are to be believed, the free tier for orgs on GitHub allows for 2000 minutes of CI time per month.

After some optimizations, (passing) tests take about ~22 minutes total... 90 runs a month is not a ton. Enabling tests for `3.10` and `3.11` would nearly halve those and might be an actual concern in busy months.

So far, I don't think we have encountered issues that reproduce for `3.10` or `3.11` but not `3.9` or `3.12`. This kind of breakage would be unusual and unlikely to be our fault if it happens, so I am closing this for now.

PR #2644, opened 2024-08-14
Fix CV data prepration for CV 18+ 

TParcollet:
Commonvoice data prep is broken with CV 18 because they changed the order of the csv fields. This PR fix this issue.


cvl01:
I was also looking at this, but suggest to use the header line from the csv. See #2647

asumagic:
> I was also looking at this, but suggest to use the header line from the csv. See #2647

Yeah, I discussed this with @TParcollet and brought up the same concern. I will close this PR so we can review yours instead.

Issue #2657, opened 2024-09-02
v1.0.1 docs are failing


asumagic:
### Describe the bug

```
...
  Downloading beautifulsoup4-4.12.3-py3-none-any.whl.metadata (3.8 kB)
Collecting omegaconf<2.1 (from fairseq@ git+https://github.com/facebookresearch/fairseq.git@v0.12.3->-r docs/docs-requirements.txt (line 3))
  Downloading omegaconf-2.0.6-py3-none-any.whl.metadata (3.0 kB)
WARNING: Ignoring version 2.0.6 of omegaconf since it has invalid metadata:
Requested omegaconf<2.1 from https://files.pythonhosted.org/packages/d0/eb/9d63ce09dd8aa85767c65668d5414958ea29648a0eec80a4a7d311ec2684/omegaconf-2.0.6-py3-none-any.whl (from fairseq@ git+https://github.com/facebookresearch/fairseq.git@v0.12.3->-r docs/docs-requirements.txt (line 3)) has invalid metadata: .* suffix can only be used with `==` or `!=` operators
    PyYAML (>=5.1.*)
            ~~~~~~^
Please use pip<24.1 if you need to use this version.
  Downloading omegaconf-2.0.5-py3-none-any.whl.metadata (3.0 kB)
WARNING: Ignoring version 2.0.5 of omegaconf since it has invalid metadata:
Requested omegaconf<2.1 from https://files.pythonhosted.org/packages/e5/f6/043b6d255dd6fbf2025110cea35b87f4c5100a181681d8eab496269f0d5b/omegaconf-2.0.5-py3-none-any.whl (from fairseq@ git+https://github.com/facebookresearch/fairseq.git@v0.12.3->-r docs/docs-requirements.txt (line 3)) has invalid metadata: .* suffix can only be used with `==` or `!=` operators
    PyYAML (>=5.1.*)
            ~~~~~~^
Please use pip<24.1 if you need to use this version.
INFO: pip is looking at multiple versions of hydra-core to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r docs/docs-requirements.txt (line 3) and fairseq because these package versions have conflicting dependencies.

The conflict is caused by:
    fairseq 0.12.3 depends on omegaconf<2.1
    hydra-core 1.0.7 depends on omegaconf<2.1 and >=2.0.5

To fix this you could try to:
1. loosen the range of package versions you've specified
[...]

PR #2672, opened 2024-09-06
Doc fix attempt again by bumping omegaconf version

asumagic:
## What does this PR do?

Take two, see #2657 #2671 -- if this passes it should be ok even if fairseq might not fully be compatible, as long as it succeeds to build the docs...


asumagic:
nope

PR #2682, opened 2024-09-11
Add TF32 flags + general improvements in DDP backend

Adel-Moumen:
## What does this PR do?

This PR tries to improve our DDP backend. Most of the changes tries to improve some corner case of SpeechBrain.

List of additions/modifications:
- add context manager to make sure that a block of code is always entering in `__enter__` and `__exit__` which makes sure that `MAIN_PROC_ONLY` is always increased/decreased even when a sub-function is throwing an exception (which in the past could lead to an issue). 
- fix an issue with the `collect_files()`. Indeed, in DDP you expect to see in `save/` a symlink of your pre-trained model. This indeed, led to an issue where only the main proc could load the LM while the other processes where throwing an error because they couldn't  load the collected models. One effective solution was to turn `internal_ddp_handling` flag to `True` but it is not necessary as `run_on_main` does also work on a system that is not distributed.
- Add a new `get_logger` function as a drop-in replacement to `logger.get_logger`. The key idea here is to have the possibility to choose if we want to log on all the processes or only the main proc. Generally speaking, we only want the main proc to log informations (e.g. loading of ckpts etc), in some cases like `seed_everything` you would like to know the seed for each proc but these are corner case which can be solved by just passing `main_process_only=False`. This new logger leaves us also the possibility to only emit one warning as wanted in some previous PRs. 
- By default, now, we are leveraging Tensor Floats 32 which is absolutely critical and doesn't leads to any instabilities. 
- Improvement of DDP barrier function to support the case where we use `NCCL`.
- Some new distributed functions that can be helpful.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
     description="All-in-one speech toolkit in pure Python and Pytorch",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    author="Mirco Ravanelli & Others",
+    author="Mirco Ravanelli, Titouan Parcollet & Others",
     author_email="speechbrain@gmail.com",
     classifiers=[
         "Programming Language :: Python :: 3",
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -74,19 +74,9 @@ def run_on_main(
             ddp_barrier()
 
 
-def if_main_process():
-    """Checks if the current process is the main process and authorized to run
-    I/O commands. The main process is the one with `RANK == 0`. In standard mode,
-    the process will not have `RANK` Unix var and will be authorized to run the I/O commands.
-    """
-    if "RANK" in os.environ:
-        if os.environ["RANK"] == "":
-            return False
-        else:
-            if int(os.environ["RANK"]) == 0:
-                return True
-            return False
-    return True
+def if_main_process() -> bool:
+    "Returns whether the current process is the main process."
+    return torch.distributed.get_rank() == 0
 
 
 def main_process_only(function):
@@ -111,15 +101,22 @@ def main_process_only(function):
 
 
 def ddp_barrier():
-    """In DDP mode, this function will synchronize all processes.
-    torch.distributed.barrier() will block processes until the whole
-    group enters this function.
     """
-    # Check if we're in a single-threaded section, skip barrier
-    if MAIN_PROC_ONLY >= 1:
-        return
-    elif torch.distributed.is_initialized():
-        torch.distributed.barrier()
+    Synchronize all processes in distributed data parallel (DDP) mode.
+
+    This function blocks the execution of the current process until all 
+    processes in the distributed group have reached the same point. It ensures 
+    that no process moves ahead until every other process has also reached this 
+    barrier. If DDP is not being used (i.e., only one process is running), 
+    this function has no effect and immediately returns.
+
+    Example
+    -------
+    >>> ddp_barrier()
+    >>> print("hello world")
+    hello world
+    """
+    torch.distributed.barrier()
 
 
 def ddp_broadcast(communication_object, src=0):
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -74,7 +74,10 @@ def run_on_main(
 
 def if_main_process() -> bool:
     "Returns whether the current process is the main process."
-    return torch.distributed.get_rank() == 0
+    if torch.distributed.is_initialized(): 
+        return torch.distributed.get_rank() == 0
+    else:
+        return True
 
 
 def main_process_only(function):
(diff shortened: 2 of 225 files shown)


speechbrain/utils/logger.py
```diff
@@ -61,6 +62,118 @@
...
+    If `log_level` is provided or found in the environment, both the logger and the root
+    logger will have their log levels set accordingly.
+    """
+    if log_level is None:
+        log_level = os.environ.get("SB_LOG_LEVEL", None)
+    logger = logging.getLogger(name)
+    if log_level is not None:
+        logger.setLevel(log_level.upper())
+        logger.root.setLevel(log_level.upper())
+    return MultiProcessLoggerAdapter(logger, {})
```
Adel-Moumen, line 175: Hey @asumagic @TParcollet @pplantinga,

what do you think of this new logger prototype ? The key idea is to allow to perform logging only on the main proc (you can turn down this option by setting `main_proc_only=False`) and also to only emit on one warning if wanted (which was something that we wanted to have in some previous prs). 

The drop-in replacement is quite easy as it follows the same API of `logging` with `get_logger`.

asumagic: The implementation looks reasonable enough to me. As for the environment variable it feels strange to only have it documented here, but I don't know where else to at the moment (maybe we should have a debugging tutorial/tips page in the doc sometime).

Also did you check how it interacts with how logging is configured by default in SB (i.e. whether this actually works at all)? Since for instance we have different log levels for the standard output and the recipe file output.

Adel-Moumen: > The implementation looks reasonable enough to me. As for the environment variable it feels strange to only have it documented here, but I don't know where else to at the moment (maybe we should have a debugging tutorial/tips page in the doc sometime).
I agree on this. We should have a documentation about all the tips such as using mixed precision etc. I feel that having an environment variable can be useful given the fact that modifying the log level through the python scripts might be inconvenient. 

> Also did you check how it interacts with how logging is configured by default in SB (i.e. whether this actually works at all)? Since for instance we have different log levels for the standard output and the recipe file output.
I am currently testing it :p

Adel-Moumen: I made some slight changes to remove how to setup logger level so that we only rely on `setup_logging`.

Adel-Moumen:
After some checks, I confirm that DDP works in multinodes and mononode on Compute Canada. (which was the case but this PR doesn't introduce regression on DDP.)

setup.py
```diff
@@ -22,7 +22,7 @@
     description="All-in-one speech toolkit in pure Python and Pytorch",
     long_description=long_description,
     long_description_content_type="text/markdown",
-    author="Mirco Ravanelli & Others",
+    author="Mirco Ravanelli, Titouan Parcollet & Others",
```
pplantinga, line 25: Shouldn't this be "Mirco Ravanelli, Titouan Parcollet, Adel Moumen & Others" ? 😂

speechbrain/utils/distributed.py
```diff
@@ -116,19 +116,42 @@ def run_on_main(
...
+
+def if_main_process() -> bool:
+    "Returns whether the current process is the main process."
+    if is_distributed_initialized():
+        return torch.distributed.get_rank() == 0
+    else:
+        return True
+
+
+class MainProcessGuard:
```
pplantinga, line 137: I'm a little unsure about the name `MainProcessGuard` because this context manager only manages tracking state rather than guarding anything. Would `MainProcessContext` or something make more sense?

speechbrain/utils/logger.py
```diff
@@ -62,24 +64,105 @@
...
+        main_process_only = kwargs.pop("main_process_only", True)
+        kwargs.setdefault("stacklevel", 2)
+
+        if self.isEnabledFor(level):
+            if self._should_log(main_process_only):
+                msg, kwargs = self.process(msg, kwargs)
+                self.logger.log(level, msg, *args, **kwargs)
+
+    @functools.lru_cache(None)
+    def warning_once(self, *args: tuple, **kwargs: dict):
```
pplantinga, line 131: Neat trick, are there any places in the code where we should replace the warning with this one?

pplantinga (approved):
Looks good to me!

setup.py
Adel-Moumen: Not yet 👀 One day it will be "Adel Moumen, Peter Plantinga, & Others" dw 👀

speechbrain/utils/distributed.py
Adel-Moumen: I agree. I was a bit puzzled by the name and I didn't knew which one to use. I prefer yours. Thanks.

speechbrain/utils/logger.py
Adel-Moumen: Yes --- I added one call of this function in the repo. :)

PR #2686, opened 2024-09-15
Critical: Fix DDP barriers

Adel-Moumen:
## What does this PR do?

This PR aims at fixing an issue with DDP barrier. Indeed, in the current develop version of SpeechBrain this current code snippet doesn't work:
```python
  from speechbrain.utils.distributed import get_rank, rank_prefixed_message, is_distributed_initialized, if_main_process, ddp_barrier

  logger.info(rank_prefixed_message(f"main={if_main_process()}"), main_process_only=False)
  
  if if_main_process():
      import time 
      time.sleep(10)
  logger.info(rank_prefixed_message("barrier 1 reached"), main_process_only=False)
  ddp_barrier()
  if if_main_process():
      import time 
      time.sleep(10)
  logger.info(rank_prefixed_message("barrier 2 reached"), main_process_only=False)
  ddp_barrier()
  
  logger.info(rank_prefixed_message("finished"), main_process_only=False)


  torch.distributed.destroy_process_group()
```
while it should! In the current dev version, the rank1 is reaching the end of the program and doesn't wait in the second DDP barrier. It does wait in the first for the main proc, but does not in the second. I went into debugging mode and found that the LOCAL_RANK of each proc is 0, which happen as we do not instantiate `Brain()` and thus the cuda device stream is never set to the actual LOCAL_RANK. I found that explicitly setting `torch.cuda.set_device(device)` in `ddp_init_group` was solving the issue. When you call `torch.cuda.current_device()` you get the right LOCAL_RANK value (which was not the case as it was just returning the LOCAL_RANK 0). 

Another example with `run_on_main`:
```python
from speechbrain.utils.distributed import get_rank, rank_prefixed_message, is_distributed_initialized, if_main_process, ddp_barrier

def bar():
    import time
    time.sleep(10)
    logger.info(rank_prefixed_message("SB is so good."), main_process_only=False)

def foo():
    bar()

run_on_main(
    foo
)
run_on_main(
    foo
)
logger.info(rank_prefixed_message("end."), main_process_only=False)
```
The second rank is rightly awaiting for `foo` to get executed and wait for the DDP barriers. 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -197,7 +197,6 @@ def ddp_barrier():
         return
 
     if torch.distributed.get_backend() == torch.distributed.Backend.NCCL:
-        # if NCCL, we can retrieve device_ids through this way
         torch.distributed.barrier(device_ids=[torch.cuda.current_device()])
     else:
         torch.distributed.barrier()
@@ -274,12 +273,9 @@ def ddp_init_group(run_opts):
             + " communication protocol doesn't exist."
         )
 
-    # Switch to the right context ; this is needed for DDP barriers
-    # otherwise, everything before an instance of `Brain(...)`
-    # is going to have `LOCAL_RANK` equal to 0.
-    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
-    device = torch.device(f"cuda:{local_rank}")
-    torch.cuda.set_device(device)
+    if run_opts["distributed_backend"] == "nccl":
+        device = torch.device(f"cuda:{local_rank}")
+        torch.cuda.set_device(device)
 
     # rank arg is used to set the right rank of the current process for ddp.
     # if you have 2 servers with 2 gpu:
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -276,7 +276,7 @@ def ddp_init_group(run_opts):
 
     # Switch to the right context ; this is needed for DDP barriers
     # otherwise, everything before an instance of `Brain(...)`
-    # is going to have `LOCAL_RANK` equal to 0. 
+    # is going to have `LOCAL_RANK` equal to 0.
     local_rank = int(os.environ.get("LOCAL_RANK", "0"))
     device = torch.device(f"cuda:{local_rank}")
     torch.cuda.set_device(device)
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -196,10 +196,9 @@ def ddp_barrier():
     if MAIN_PROC_ONLY >= 1 or not is_distributed_initialized():
         return
 
-    if torch.distributed.get_backend() == "nccl":
+    if torch.distributed.get_backend() == torch.distributed.Backend.NCCL:
         # if NCCL, we can retrieve device_ids through this way
-        device_ids = [int(os.environ.get("LOCAL_RANK"))]
-        torch.distributed.barrier(device_ids=device_ids)
+        torch.distributed.barrier(device_ids=[torch.cuda.current_device()])
     else:
         torch.distributed.barrier()
 
@@ -274,6 +273,13 @@ def ddp_init_group(run_opts):
             run_opts["distributed_backend"]
             + " communication protocol doesn't exist."
         )
+
+    # Switch to the right context ; this is needed for DDP barriers
+    # otherwise, everything before an instance of `Brain(...)`
+    # is going to have `LOCAL_RANK` equal to 0. 
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    device = torch.device(f"cuda:{local_rank}")
+    torch.cuda.set_device(device)
 
     # rank arg is used to set the right rank of the current process for ddp.
     # if you have 2 servers with 2 gpu:


pplantinga:
I thought `LOCAL_RANK` was supposed to be set by `torchrun` is that not the case?

speechbrain/utils/distributed.py
```diff
@@ -196,10 +196,9 @@ def ddp_barrier():
     if MAIN_PROC_ONLY >= 1 or not is_distributed_initialized():
         return
 
-    if torch.distributed.get_backend() == "nccl":
+    if torch.distributed.get_backend() == torch.distributed.Backend.NCCL:
         # if NCCL, we can retrieve device_ids through this way
-        device_ids = [int(os.environ.get("LOCAL_RANK"))]
-        torch.distributed.barrier(device_ids=device_ids)
+        torch.distributed.barrier(device_ids=[torch.cuda.current_device()])
```
pplantinga, line 201: This doesn't seem right to me, shouldn't the device_ids contain all the devices in the group? As is it seems like it wouldn't block at all

Adel-Moumen:
> I thought `LOCAL_RANK` was supposed to be set by `torchrun` is that not the case?

It is but at the cuda level, it seems that torch barrier is using torch.cuda.device to synchronize processes but since the variable is never set properly (meaning that LOCAL_RANK is not used here) then pytorch fails to synchronize barriers. I believe it might be only located to NCCL backend.

speechbrain/utils/distributed.py
```diff
@@ -275,6 +274,13 @@ def ddp_init_group(run_opts):
             + " communication protocol doesn't exist."
         )
 
+    # Switch to the right context ; this is needed for DDP barriers
+    # otherwise, everything before an instance of `Brain(...)`
+    # is going to have `LOCAL_RANK` equal to 0.
```
pplantinga, line 279: The local rank is set by torchrun, not brain. Isn't this instead about whether the device is set?

speechbrain/utils/distributed.py
```diff
@@ -275,6 +274,13 @@ def ddp_init_group(run_opts):
...
+    local_rank = int(os.environ.get("LOCAL_RANK", "0"))
+    device = torch.device(f"cuda:{local_rank}")
+    torch.cuda.set_device(device)
```
pplantinga, line 282: Looks like PyTorch docs do not recommend this func https://pytorch.org/docs/stable/generated/torch.cuda.set_device.html

pplantinga (commented):
I have some questions about the approach, I'll take a closer look tomorrow

pplantinga:
> torch barrier is using torch.cuda

That is a very odd choice by the PyTorch team....

speechbrain/utils/distributed.py
Adel-Moumen: I agree that it seems a bit "weird". I would invite you to read about this conversation by the PyTorch Lightning team: https://github.com/Lightning-AI/pytorch-lightning/pull/8165 and see code from multiple projects (https://github.com/Lightning-AI/pytorch-lightning/blob/7d2b85916dcefd91df92edcb69411e97e9ac4a65/src/lightning/pytorch/strategies/ddp.py#L291-L299 ; https://github.com/pytorch/tnt/pull/397 ).

I think what is happening when you do that, is that you are for forcing the barrier to be exactly on this process (see: https://github.com/pytorch/pytorch/issues/53658) which might not be the case.

speechbrain/utils/distributed.py
Adel-Moumen: I know this is not set by Brain but instead by torchrun (btw, one thing we could do is to remove torchrun from SB and directly dynamically set all the flags etc something like pytorch lightning does).

I was maybe unclear but I meant the device retrieved by torch. When you are doing `torch.cuda.current_device()` you'll get a different value (always 0) than the LOCAL_RANK. After creating a Brain instance (hence setting torch.cuda.device), you'll get the cuda device to be LOCAL_RANK. I believe this is the reason why we had troubles. 

Behind the hood torch might be using the cuda device and not the LOCAL_RANK for the barriers. (this is the most likely hypothesis given my experiments that showed that setting torch.device to the LOCAL_RANK was solving our issues).

speechbrain/utils/distributed.py
Adel-Moumen: yes i saw it too... but at the same time I'm seeing everyone setting it this way (and this is what we are doing in core.py). I will look into that to see if there's better ways to do it.

speechbrain/utils/distributed.py
Adel-Moumen: BTW, i really confirm that if you are not setting the cuda device, `torch.cuda.current_device()` will always return `0`. And this makes a difference when using torch.distributed.barrier. You really need to make sure that it returns the same value as `LOCAL_RANK` otherwise the barrier wont work properly.

Adel-Moumen: I might have found the exact explanation of the issue: 

Capture d’écran 2024-09-15 à 21 09 12
Source: https://pytorch.org/docs/stable/distributed.html#launch-utility

Adel-Moumen: BTW, simplifying `ddp_barrier` to:
```
def ddp_barrier():
    r"""
    Synchronize all processes in distributed data parallel (DDP) mode.

    This function blocks the execution of the current process until all
    processes in the distributed group have reached the same point. It ensures
    that no process moves ahead until every other process has also reached this
    barrier. If DDP is not being used (i.e., only one process is running),
    this function has no effect and immediately returns.

    Returns
    -------
    None


    Example
    -------
    >>> ddp_barrier()
    >>> print("hello world")
    hello world
    """
    if MAIN_PROC_ONLY >= 1 or not is_distributed_initialized():
        return

    torch.distributed.barrier()
```

Does also work! Adding `device_ids` is not what is making `torch.distributed.barrier` work. This is when we are setting the cuda device. I believe having `device_ids` is solving some issue that you can have when you are only using e.g. gpu 1 and 3 from a node with more GPUs.

pplantinga: So the PyTorch docs are inconsistent about the use of this function. Love it.

speechbrain/utils/distributed.py
pplantinga: Again this seems to be a weird choice by the PyTorch team but looks like you did proper research and this is the right approach.

pplantinga (approved):
I'm gonna mark my concerns up to weird behavior and documentation from PyTorch, it looks like this is actually the right approach

Issue #2658, opened 2024-09-02
Github release actions failing


pplantinga:
### Describe the bug

The release action is failing due to github action sunset:

```
UNSUPPORTED GITHUB ACTION VERSION \n\n You are using . The branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use . If you feel adventurous, you may opt to use use instead. A more general recommendation is to pin to exact tags or commit shas. \n\n \n\n ### Oh, and while you are here — #StandWithUkraine \n\n [![SWUbanner]][SWUdocs] \n\n [SWUbanner]: https://raw.githubusercontent.com/vshymanskyy/StandWithUkraine/main/banner-direct-single.svg \n\n [SWUdocs]: https://github.com/vshymanskyy/StandWithUkraine/blob/main/docs/README.md
```

See this run: https://github.com/speechbrain/speechbrain/actions/runs/10668639313

### Expected behaviour

This GH action should trigger when creating a release and auto-upload the build to PyPI

asumagic:
Tentatively closing this since we fixed the direct reason, we can reopen if it still requires manual intervention for the v1.0.2 release.

PR #2690, opened 2024-09-18
`gh-action-pypi-publish@master` was deprecated, switch to `release/v1`

asumagic:
## What does this PR do?

See #2658. It does **not** resolve it (see slack/the primary mail warning), but it fixes a warning brought up by it.  
The `gh-action-pypi-publish` action strongly recommends against targeting `@master` and suggests targeting `@release/v1` instead.  
This targets the latest version of the action in the `v1` series. I am assuming the intent of that branch is to have reasonable backwards compatibility...


pplantinga (approved):
LGTM

PR #2696, opened 2024-09-24
Tokotron: Tokenized TTS

flexthink:
## What does this PR do?

Introduces a simple TTS architecture based on discrete speech representations from self-supervised models

</details>

## PR review

<details>

</details>


speechbrain/inference/eval.py
pplantinga, line 1: I'm a little fuzzy on the value-add for this file on top of our current metric engines. What is this file accomplishing?

recipes/LJSpeech/TTS/tokotron/hparams/arpabet.txt
pplantinga, line 1: This file isn't really hparams. I guess its fine to put it here, but perhaps we could consider using another directory called `tokens` or `data` or something.

pplantinga (changes requested):
Overall I think this will make a valuable addition to the toolkit, tokenized TTS is a great project. This is quite a large PR and hard to review in its entirety but I have a few initial thoughts.

* I tried to run LibriTTS using "lite" configuration, and the training time was surprisingly reasonable (3 minutes or so) per epoch, but the validation time was very long, >2 hours per epoch. I wonder if it has anything to do with the warning message: `.../site-packages/torch/nn/functional.py:5193: UserWarning: Support for mismatched key_padding_mask and attn_mask is deprecated. Use same type for both instead.`
* I don't see any readme/documentation/tutorial file that explains tokotron and what this model is trying to accomplish. We may want to at the very least have a README with a link to a paper explaining the model and any results.
* I'm wondering if some of these changes (like `eval.py` or `preparation.py`) may make sense to move to a separate PR. They seem like bigger changes that are a bit unrelated to Tokotron.

Again, I believe this will be a nice addition to the toolkit, but it still requires some work and more review due to the size.

speechbrain/utils/hparams.py
```diff
@@ -20,6 +20,10 @@ def choice(value, choices, default=None):
         to the corresponding return values
     default: any
         the default value
+    apply: bool
+        if set to true, the value is expected to
```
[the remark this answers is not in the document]
flexthink: Actually, in this case it is not an index... The way choice is used is as follows.

```yaml
spk_emb_discrete_src: !apply:speechbrain.utils.hparams.choice
    value: !ref <ssl_model_type>
    choices:
        wavlm: flexthink/discrete_wavlm_spk_rec_ecapatdn_lite
        hubert: flexthink/discrete_hubert_spk_rec_ecapatdn_lite
        wav2vec2: flexthink/discrete_wav2vec2_spk_rec_ecapatdn_lite
```

So the `value` indicates the value to be mapped using choices... Perhaps it could be renamed to `key` by analogy to dictionary keys but it is not a numeric index.

The most typical use case for this is this:

* You have an hparam indicating, for instance, the type of base model to use
* Multiple other params will be automatically selected based on this choice

Without `choice` you'd have to create different hparams files for all the permutations and combinations.

PR #2702, opened 2024-09-26
Fix template inference links

### LINKED ISSUE #2701: Template inference script broken
### Describe the bug

When trying to run inference on `templates/speech_recognition/ASR` I get the following error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/pplantinga/Documents/Repositories/speechbrain/docs/tutorials/nn/speechbrain/templates/speech_recognition/ASR/lm.ckpt'
```

The inference class is looking for the model file `lm.ckpt` in the `source` folder rather than the `savedir` folder, which is different behavior from before. Perhaps this is related to the changes to `fetch` in #2476 but I haven't looked into this enough to be sure. Perhaps this is worth looking into more so we can at the very least alert users that this behavior is different, and possibly consider changes to revert to old behavior.

### Expected behaviour

Running the template inference file should not crash.

### To Reproduce

```
cd templates/speech_recognition/ASR/
python train.py train.yaml
python transcribe_file.py audio.wav results/...
```

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

pplantinga:
Fixes #2701 

The current approach just adapts the templates to match the new behavior of the inference class, but we should also consider changes to the inference class behavior to avoid needing these changes.


TParcollet:
@asumagic could you have a look here next week? It's one of the last before 1.0.2. THanks!

templates/speech_recognition/ASR/train.py
```diff
@@ -471,7 +472,7 @@ def text_pipeline(words):
     # you can train from scratch and avoid this step.
     # We download the pretrained LM from HuggingFace (or elsewhere depending on
     # the path given in the YAML file). The tokenizer is loaded at the same time.
-    hparams["pretrainer"].collect_files()
+    hparams["pretrainer"].collect_files(local_strategy=LocalStrategy.SYMLINK)
```
asumagic, line 475: Now unneeded, as this is the default again.

```suggestion
    hparams["pretrainer"].collect_files()
```

templates/enhancement/enhance_file.py
```diff
@@ -22,8 +23,7 @@
     args = parser.parse_args()
 
     enhancer = SpectralMaskEnhancement.from_hparams(
-        source=".",
-        hparams_file="inference.yaml",
-        savedir=args.save_directory,
+        source=os.path.abspath(args.save_directory),
+        hparams_file=os.path.join(os.getcwd(), "inference.yaml"),
```
asumagic, line 27: This doesn't work because the hparams path is appended to the source path. Although maybe it should detect absolute paths and stop trying to concatenate them...

asumagic:
Actually, is this PR now still relevant as implemented? It would be better to make them not use savedirs with symlinks for Windows compat reasons but at least it's not flat out broken now.

pplantinga:
Agreed, we can look in a separate issue about avoiding symlinks

PR #2707, opened 2024-10-01
Add adapter tutorial

### LINKED ISSUE #2698: Write tutorials for Adapters
### Describe the bug

Making a formal issue to attach to the milestone.

Related:

- #2563
- #2526

### Expected behaviour

N/A

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

pplantinga:
Closes #2698

This PR adds a tutorial for adapters.

While writing it, I discovered that the adapters implementation does not play very nice with `Pretrainer`, as the changes to the model structure prevent loading the pretrained weights. Thus in addition to the tutorial, this PR makes a small change to the `speechbrain.nnet.adapters.AdaptedModel` class that allows one to manually insert the adapters at a later time, which would typically look like:

```python
pretrainer.collect_files()
pretrainer.load_collected()
hparams["model"].insert_adapters()
```

Note: this behavior is optional, by default the model inserts the adapters at initialization.


speechbrain/nnet/adapters.py
```diff
@@ -39,6 +39,10 @@ class AdaptedModel(nn.Module):
         Supports Unix shell-style wildcards `(*, ?, [seq], [!seq])` with `fnmatch`.
     adapter_kwargs: dict
         Ensemble of parameters that should be given to the adapter.
+    manual_adapter_insertion: bool
```
TParcollet, line 42: This docstring is a bit unclear to me.

TParcollet (changes requested):
Thanks @pplantinga ! I only have two minor changes:
1. Could you summarize at the very beginning of the tutorial how to add an adapter to the yaml? This could be useful to our users that do not want to go into all the details. Like maybe just the piece of yaml code showing the decoration of the module.
2. In know that we must go with python modification of the yaml so it works in a notebook, but that *may* be confusing to some. Could you put, at the end of the tutorial, the full yamls? 

Once this is changed, and if @asumagic is happy, then i'll merge.

docs/tutorials/nn.rst
```diff
@@ -32,6 +33,22 @@ coming from HuggingFace. Any wav2vec 2.0 / HuBERT / WavLM or Whisper model integ
 SpeechBrain to approach a speech-related task: automatic speech recognition, speaker recognition,
 spoken language understanding ...
 
+.. rubric:: `🔗 # Neural Network Adapters for faster low-memory fine-tuning <nn/neural-network-adapters.html>`_
```
asumagic, line 36:
```suggestion
.. rubric:: `🔗 Neural Network Adapters for faster low-memory fine-tuning <nn/neural-network-adapters.html>`_
```

docs/tutorials/nn.rst
```diff
@@ -32,6 +33,22 @@ coming from HuggingFace. Any wav2vec 2.0 / HuBERT / WavLM or Whisper model integ
...
+   :header-rows: 0
+
+   * - Plantinga P.
+     - Sept. 2024
+     - Difficulty: easy
+     - Time: 20m
+     - `🔗 Google Colab <https://colab.research.google.com/github/speechbrain/speechbrain/blob/develop/docs/tutorials/nn/neural-network-adapters.ipynb>`__
+
+
+TThis tutorial covers the SpeechBrain implementation of adapters such as LoRA. This includes how to integrate either SpeechBrain implemented adapters, custom adapters, and adapters from libraries such as PEFT into a pre-trained model.
```
asumagic, line 50:
```suggestion
This tutorial covers the SpeechBrain implementation of adapters such as LoRA. This includes how to integrate either SpeechBrain implemented adapters, custom adapters, and adapters from libraries such as PEFT into a pre-trained model.
```

docs/tutorials/nn/neural-network-adapters.ipynb
asumagic, line 1: Should run `tutorial-cell-update.py`, see https://github.com/speechbrain/speechbrain/tree/develop/docs#integration-in-documentation

This should automatically add the header and footer for colab + the SB citation we have on every tutorial.

asumagic, line 1: Install step is broken, should specify `!python -m pip install -e speechbrain` not `.`

asumagic, line 1: Should indicate somewhere that CUDA should be available (and so the proper runtime on Colab should be selected, if running from that).

The `train.py` command otherwise fails, and `--device=cpu` is too slow (hours).

asumagic, line 1: I'll keep a Colab running to make sure it's working properly and will update on that front. In the mean time, pushing this review. Seemed fine for me otherwise.

asumagic (changes requested):

asumagic:
Ah, yeah, this is also plagued with the issue where creating a `speechbrain` directory means that `import`s later in the notebook are relative to the wrong directory (`./speechbrain/` instead of `./speechbrain/speechbrain`). Might need a workaround or rename.

docs/tutorials/nn/neural-network-adapters.ipynb
pplantinga: I have run this now, thanks

pplantinga: Ah, good catch. I had only run this locally, but have now run it on colab and fixed the imports

pplantinga: Good point, added

speechbrain/nnet/adapters.py
pplantinga: Good to know, hope it is clear now.

pplantinga:
> * Could you summarize at the very beginning of the tutorial how to add an adapter to the yaml? This could be useful to our users that do not want to go into all the details. Like maybe just the piece of yaml code showing the decoration of the module.

Good idea. I have added a TL;DR section at the beginning with just the yaml snippet, for those who are looking for this.

> * In know that we must go with python modification of the yaml so it works in a notebook, but that _may_ be confusing to some. Could you put, at the end of the tutorial, the full yamls?

The yamls are available in the file browser after running the tutorial. Putting it in the notebook itself may be overwhelming since they are quite long.

asumagic (approved):

PR #2712, opened 2024-10-09
Fetch take two

### LINKED ISSUE #2711: Fetch Semantics
### Describe the bug

The recent changes to `fetch` have changed the defaults that some templates (at least) depend on, see #2701. In that issue, some general changes to the default were discussed, this issue should track those changes.

### Expected behaviour

* Code that explicitly sets the fetch savedir behave the same (default to SYMLINK or perhaps an environment variable).
* Code can now explicitly opt in to any of the new strategy options.
* Code that does not set savedir are now assumed to not rely on where the files are saved (default and enforce NO_LINK).
* Examples and uses in the codebase edited to not specify savedir.
* Ensure that at least one message sums up what any fetch operation has resulted into, at INFO level.
* Detect and warn for leftover (old) pretrainer cache directories.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

pplantinga:
Fixes #2711 

This changes `fetch` behavior in a number of ways to make it a little more intuitive, as well as to prevent creating extra files by default if they're not needed.


TParcollet:
This is a lot of changes for a pretty used function. @asumagic may want to have a look as well. Maybe we should wait for 1.0.3 or 1.1 for this one. @pplantinga what do you think?

pplantinga:
We can wait on this, though note that the behavior of this function has already changed due to the recent PR by @asumagic

speechbrain/inference/interfaces.py
asumagic, line 1: What about the `Pretrained` class? It also sets the `savedir` in a similar way as `foreign_class`. Should that also be modified?

Otherwise, LGTM

pplantinga: Good call! This is changed now.

asumagic (approved):
LGTM, will look into the template fixes separately afterwards.

PR #2716, opened 2024-10-10
Ensure adapted models save their parameters

### LINKED ISSUE #2715: `AdaptedModel` does not save weights
### Describe the bug

The saved `state_dict` is always empty for `AdaptedModel` because the `state_dict` does not contain gradient information, it only has the actual parameter values.

```python
@checkpoints.mark_as_saver
    def saver(self, path):
        """Saves only the trainable parameters."""
        state_dict = {
            n: p for n, p in self.state_dict().items() if p.requires_grad
        }
        torch.save(state_dict, path)
```

`state_dict()` is not right here because it doesn't have the correct `requires_grad` value.

### Expected behaviour

Saving an `AdaptedModel` should save the adapted parameters.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

_No response_

### Additional Context

_No response_

pplantinga:
Fixes #2715 

Adapted models was not saving parameters properly due to PyTorch detaching all parameters in the state dict, setting requires_grad to False. This PR fixes the saving by preserving the parameters in the state dict for iteration and then detaching manually.

### DIFF
--- a/speechbrain/nnet/adapters.py
+++ b/speechbrain/nnet/adapters.py
@@ -129,8 +129,12 @@ class AdaptedModel(nn.Module):
     @checkpoints.mark_as_saver
     def saver(self, path):
         """Saves only the trainable parameters."""
+        # NOTE: In order to preserve the gradient info, we have to prevent `state_dict` from detaching
+        # all the parameters and buffers. The `keep_vars=True` does this, then we detach manually
         state_dict = {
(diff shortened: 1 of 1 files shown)


TParcollet (commented):
LGTM. Being tested by the person who found the bug. If all good, i'll merge.

PR #2734, opened 2024-10-28
Make calls to "open" deterministic (code from Samsung AI Center Cambridge)

rogiervd:
## What does this PR do?

This PR makes calls to `open` use UTF-8 encoding.

The current behaviour is that the encoding depends on the local environment. Under Linux, if environment variable `LC_ALL=C.UTF-8` is set, files are read as UTF-8. However, we have found that in practice, behaviour changes from user to user or machine to machine, and transcriptions with non-ASCII characters get messed up. This PR makes the behaviour deterministic.

When opening files for `csv`, this also add `newline=""`, as required by <https://docs.python.org/3/library/csv.html#csv.reader>.

*Change after review*: the PR now also adds a check for the encoding to the pre-commit checks.

## Breaking changes

For users/machines that have `LC_ALL=C.UTF-8` set, nothing changes. For users/machines that have `LC_ALL` set differently, the behaviour changes to the correct one.

## Remarks

* I believe I have changed all relevant calls to `open` but it is possible I have missed some. This highlights that it might have been better to have all parsing, including calls to `open`, in one place instead of scattered around.
* This is a bit of a big PR that touches many files. I'd be happy to split it up into multiple changes per subdirectory or something if that's more useful.
* I have not been able to run all recipes.

cc @TParcollet for visibility.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,10 @@ repos:
     hooks:
       - id: flake8
         types: [python]
-        additional_dependencies: ['pydoclint==0.4.1', 'pycodestyle==2.11.0']
+        additional_dependencies:
+          - pydoclint==0.4.1
+          - pycodestyle==2.11.0
+          - flake8-encodings==0.5.1
 
   - repo: https://github.com/adrienverge/yamllint
     rev: v1.35.1
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,6 +1,7 @@
 black==24.3.0
 click==8.1.7
 flake8==7.0.0
+flake8-encodings==0.5.1
 isort==5.13.2
 pycodestyle==2.11.0
 pydoclint==0.4.1
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,7 +1,6 @@
 black==24.3.0
 click==8.1.7
 flake8==7.0.0
-flake8-encodings==0.5.1
 isort==5.13.2
 pycodestyle==2.11.0
 pydoclint==0.4.1
--- a/recipes/AISHELL-1/ASR/CTC/train_with_wav2vec.py
+++ b/recipes/AISHELL-1/ASR/CTC/train_with_wav2vec.py
@@ -166,7 +166,7 @@ class ASR(sb.Brain):
                 stats_meta={"Epoch loaded": self.hparams.epoch_counter.current},
                 test_stats=stage_stats,
             )
-            with open(self.hparams.cer_file, "w") as w:
+            with open(self.hparams.cer_file, "w", encoding="utf-8") as w:
                 self.cer_metric.write_stats(w)
 
     def init_optimizers(self):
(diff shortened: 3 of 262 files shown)


asumagic:
Although the CRLF thing wrt. CSV file reading is interesting... Didn't know about that gotcha.

> If newline='' is not specified, newlines embedded inside quoted fields will not be interpreted correctly, and on platforms that use \r\n linendings on write an extra \r will be added. It should always be safe to specify newline='', since the csv module does its own ([universal](https://docs.python.org/3/glossary.html#term-universal-newlines)) newline handling.

rogiervd:
> I think the linter we use is able to detect the `open` with missing encoding, maybe we should look into making it into a CI error down the line.

That would be best. I'm happy to try and switch this on, but I don't see it. Where can I do this?

asumagic:
I remember encountering it while running linting tools locally, but upon closer inspection, I can only find it being implemented as part of a flake8 plugin, [`flake8-encodings`](https://pypi.org/project/flake8-encodings/). I didn't install it manually and it doesn't seem to be used in SpeechBrain, so I don't know what would have pulled it.

If I'm reading this right this might just be a matter of adding that package to the `lint-requirements.txt`?

TBH it wouldn't be a blocker for this PR but it would be better to have down the line.

asumagic:
+ potentially in `.flake8` adding `ENC` to `select =`

rogiervd:
Let me try this, first in a separate PR.  It would be best to solve this once and for all, if I can.

rogiervd:
Is limiting supposed to be run in CI? Because I don’t see where, and my commit attempting to switch on the detection of encodings hasn’t triggered a CI failure in the way I hoped it would.

asumagic:
Seems like the [pre-commit workflow (`pre-commit.yaml`)](https://github.com/speechbrain/speechbrain/blob/develop/.github/workflows/pre-commit.yml) does not actually honor the `lint-requirements` unlike [the test workflow (`pythonapp.yml`)](https://github.com/speechbrain/speechbrain/blob/develop/.github/workflows/pythonapp.yml#L20-L33).

(Noting that `requirements.txt` does include `lint-requirements.txt`.)

Seems like an oversight, will address this in a separate PR quickly.

asumagic:
Actually it might just be simpler to include the dependency in `.pre-commit-config.yaml` under `flake8`. `pre-commit` kind of does its own dependency management so it might be better to avoid pulling all SB dependencies there.

Then, it would probably make sense to remove it from `lint-requirements.txt`.

asumagic:
Seems like this works as expected now.

rogiervd:
Great. The linting ([e34eab0](https://github.com/speechbrain/speechbrain/pull/2734/commits/e34eab0f477b2451033e7afe3056f0307a110780)) is now causing the CI to fail as expected.

TParcollet:
@asumagic I also don't see how this could break anything serious from the files touched. Feel free to merge when you want.

pplantinga (approved):
Good cleanup of the repo, making it more usable on Windows! LGTM

PR #2750, opened 2024-11-07
Add Pyre type checker as a pre-commit hook (code from Samsung AI Center Cambridge)

rogiervd:
## What does this PR do?

This PR adds the Pyre type checker as a pre-commit hook. This was discussed in #2749.

This PR should not be merged yet until the test failures are removed (in separate PRs). But the log should be useful to see the errors to fix.

The reason this uses Pyre is that it is fast. On my desktop computer it takes 8 seconds. I have therefore added it to the pre-commit checks.

Currently, I see ~2000 failures, which it is my plan to address in separate PRs.

For now, please let me know any problems you see with this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,3 +52,11 @@ repos:
     rev: 5.13.2
     hooks:
       - id: isort
+
+  # There is a pre-commit hook defined in the official repo at
+  # https://github.com/facebook/pyre-check but this does not actually work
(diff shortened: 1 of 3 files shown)


rogiervd:
This gives errors like `Undefined import [21]: Could not find a module corresponding to import torch.`. So I will instead try and add a separate workflow for type checking in which all appropriate packages are installed.

PR #2762, opened 2024-11-21
allow local strategy to be passed to collect_files()

mirofedurco:
## What does this PR do?

This change enables for the custom local strategy to be propagated down to the collect_files method.


pplantinga (commented):
If we're gonna apply it here, there's another place in the file we could add it, in the `foreign_class` function

pplantinga (approved):
LGTM

PR #2765, opened 2024-11-21
Librilight data preparation for SpeechBrain SSL (code from Samsung AI Center Cambridge)

shucongzhang:
## What does this PR do?

This PR contains a script that will create a train.csv file for the LibriLight dataset. The train.csv can be directly used as the "train_csv" in any SpeechBrain SSL yaml file.

### DIFF
--- a/recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
+++ b/recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
@@ -1,15 +1,15 @@
 """ Create speebrain csv files for the Libri-Light dataset for self-supervised-learning
 
-1. download the Libri-Light dataset through the toolkit in the Libri-Light github repo 
+1. Download the Libri-Light dataset through the toolkit in the Libri-Light github repo 
     "https://github.com/facebookresearch/libri-light/"
 
-2. use the data_preparation/cut_by_vad.py script of the Libri-Light repo to do the vad. For example, 
+2. Use the data_preparation/cut_by_vad.py script of the Libri-Light repo to do the vad. For example, 
     "python cut_by_vad.py --input_dir path_to_Libri-Light_small --output_dir Libri-Light_small_vad --target_len_sec 20"
     
-3. use the output_dir in step 2 as the input_dir in this step to generate the csv file. For example,
+3. Use the vad output_dir in step 2 as the input_dir in this step to generate the csv file. For example,
     "python make_librilight_csv.py --input_dir Libri-Light_small_vad --output_dir results --max_length 20 --n_processes 128" to generate the train.csv for the SSL pretraining 
     
-4. now, you can use the train.csv as the "train_csv" in any SpeechBrain SSL pre-training yaml file
+4. Now, you can use the generated train.csv in the output_dir in step 3 as the "train_csv" in any SpeechBrain SSL pre-training yaml file
 
 Authors
  * Shucong Zhang 2024
--- a/recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
+++ b/recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
@@ -1,0 +1,139 @@
+""" Create speebrain csv files for the Libri-Light dataset for self-supervised-learning
+
+1. download the Libri-Light dataset through the toolkit in the Libri-Light github repo 
+    "https://github.com/facebookresearch/libri-light/"
+
+2. use the data_preparation/cut_by_vad.py script of the Libri-Light repo to do the vad. For example, 
+    "python cut_by_vad.py --input_dir path_to_Libri-Light_small --output_dir Libri-Light_small_vad --target_len_sec 20"
+    
+3. use the output_dir in step 2 as the input_dir in this step to generate the csv file. For example,
+    "python make_librilight_csv.py --input_dir Libri-Light_small_vad --output_dir results --max_length 20 --n_processes 128" to generate the train.csv for the SSL pretraining 
+    
+4. now, you can use the train.csv as the "train_csv" in any SpeechBrain SSL pre-training yaml file
+
+Authors
+ * Shucong Zhang 2024
+ * Titouan Parcollet 2024
+"""
+
+
+import pathlib
+import torchaudio
+import tqdm
+import multiprocessing
+import csv
+from pathlib import Path
+import sys
+import os
+import argparse
+
+def make_csv_for_each(
+    subpath_1_csv_file_folder_max_length: tuple,
+) -> None:
+    """Prepare the csv fils for each subfolder of the Libri-Light splits
+
+    Arguments
+    ---------
+    subpath_1_csv_file_folder_max_length : tuple
+        A (subfolder_path, csv_output_path, max_utt_length) tuple
+    """
+    subpath_1, csv_file_folder, max_length = subpath_1_csv_file_folder_max_length
+    for i, flac_file in enumerate(subpath_1.glob('**/*.flac')):
+        flac_file_name = flac_file.stem
+        waveform, sample_rate = torchaudio.load(str(flac_file))
+        num_frames = waveform.size(1)
+        duration_seconds = num_frames / sample_rate
+        if duration_seconds > max_length:
+            continue
+        audio_length_seconds = waveform.shape[1] / sample_rate
+        csv_file = f"{csv_file_folder}/{flac_file.parent.stem}.csv"
+        with open(csv_file, mode='a', newline='') as csvfile:
+            csv_writer = csv.writer(
+                csvfile, delimiter=",", quotechar='"', quoting=csv.QUOTE_MINIMAL
+            )
+            csv_writer.writerow([flac_file_name,audio_length_seconds,str(flac_file)])
+
+
+def processes_folder(
+    data_path: str, 
+    csv_file_folder: str, 
+    max_length: float, 
+    n_processes: int,
+) -> None:
+    """Prepare the csv fils for one of the Libri-Light splits
+
+    Arguments
+    ---------
+    data_path : str
+        Path to the Libri-Light split after vad
+    csv_file_folder: str
+        Path to the output folder of generated csv files
+    max_length:
+        The max length for each prepared audio clip
+    n_processes:
+        Number of parallel processes
+    """
+    print("Processing each subfolder of this split")
+    os.makedirs(csv_file_folder, exist_ok=True)
+    os.makedirs(f"{csv_file_folder}/tmp", exist_ok=True)
+    list_dir = pathlib.Path(data_path)
+    tasks = []
+    for x in list_dir.iterdir():
+        tasks.append((x, f"{csv_file_folder}/tmp", max_length))
+    with multiprocessing.Pool(processes=n_processes) as pool:
+        for _ in tqdm.tqdm(pool.imap_unordered(make_csv_for_each, tasks), total=len(tasks)):
+            pass
+
+def merge_csv_files(
+    csv_file_folder: str, 
+) -> None:
+    """Merges the csv files for each subfolder into one single file
+
+    Arguments
+    ---------
+    csv_file_folder: str
+        Path to the output folder of generated csv files
+    """
+    print("Merging the csvs of each subfolder into one csv")
+    file_list = [str(x) for x in Path(f"{csv_file_folder}/tmp").glob('*.csv')]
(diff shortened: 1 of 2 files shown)


recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+4. Now, you can use the generated train.csv in the output_dir in step 3 as the "train_csv" in any SpeechBrain SSL pre-training yaml file
+
+Authors
+ * Shucong Zhang 2024
+ * Titouan Parcollet 2024
+"""
+
+import argparse
+import csv
+import multiprocessing
```
Adel-Moumen, line 21: Could you please use instead `parallel_map` from our toolkit (see:  https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/common_voice_prepare.py#L368)

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+    data_path : str
+        Path to the Libri-Light split after vad
+    csv_file_folder: str
+        Path to the output folder of generated csv files
+    max_length: float
+        The max length (seconds) for each prepared audio clip
+    n_processes: int
+        Number of parallel processes
+    """
+    print("Processing each subfolder of this split")
```
Adel-Moumen, line 80: could you please use our loggers instead of using print? (e.g. https://github.com/speechbrain/speechbrain/blob/develop/recipes/CommonVoice/common_voice_prepare.py#L24)

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+    os.makedirs(f"{csv_file_folder}/tmp", exist_ok=True)
+    list_dir = pathlib.Path(data_path)
+    tasks = []
+    for x in list_dir.iterdir():
+        tasks.append((x, f"{csv_file_folder}/tmp", max_length))
+    with multiprocessing.Pool(processes=n_processes) as pool:
+        for _ in tqdm.tqdm(
+            pool.imap_unordered(make_csv_for_each, tasks), total=len(tasks)
+        ):
+            pass
```
Adel-Moumen, line 91: this code will have to change according to the `parallel_map`

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+    csv_file_folder: str,
+) -> None:
+    """Merges the csv files for each subfolder into one single file
+
+    Arguments
+    ---------
+    csv_file_folder: str
+        Path to the output folder of generated csv files
+    """
+    print("Merging the csvs of each subfolder into one csv")
```
Adel-Moumen, line 104: same logger

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+            with open(file_path, mode="r", encoding="utf-8") as infile:
+                reader = csv.reader(infile)
+
+                # filter out bad rows
+                for row in reader:
+                    if len(row) == 3 and os.path.exists(row[-1]):
+                        new_row = [row[-1], row[1], row[2]]
+                        csv_writer.writerow(new_row)
+                    else:
+                        print(f"bad row {row}")
```
Adel-Moumen, line 124: can you add in the docstring of the function what does it means to have "bad rows" ?

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+
+    import shutil
+
+    shutil.rmtree(f"{csv_file_folder}/tmp")
```
Adel-Moumen, line 128: oof?

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+    ---------
+    subpath_1_csv_file_folder_max_length : tuple
+        A (subfolder_path, csv_output_path, max_utt_length) tuple
+    """
+    subpath_1, csv_file_folder, max_length = (
+        subpath_1_csv_file_folder_max_length
+    )
+    for i, flac_file in enumerate(subpath_1.glob("**/*.flac")):
+        flac_file_name = flac_file.stem
+        waveform, sample_rate = torchaudio.load(str(flac_file))
```
Adel-Moumen, line 45: I think it would be better to instead use the `read_audio_info` (see: https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/dataio/dataio.py#L166-L221)

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
```diff
@@ -0,0 +1,168 @@
...
+        help="The max length for each prepared audio clip" "(default is 20.2)",
+    )
+    parser.add_argument(
+        "--n_processes",
+        type=int,
+        default=32,
+        help="Number of parallel processes",
+    )
+    args = parser.parse_args()
+    return args
```
Adel-Moumen, line 160: @TParcollet do you have an opinion on that? Usually the data prep is being done directly in the train.py ,BUT,  given how large the dataset is, I would say that 99% of the folks will prefer to first process the dataset, and then, train models (which means a clear way of executing only the data prep). In this regard, I would say we could stick to argparse for this recipe, but I'd like to get the goat opinion on this design choice. thanks

Adel-Moumen (commented):
Hey @shucongzhang  , how are you doing? Nice to see a data prep for Librilight! I left some early comments in case you wanna take a look. Also, do you plan to add a recipe e.g. best-rq ? would be nice to have a working prototype alongside this data prep. thanks :)

recipes/Libri-Light/self-supervised-learning/make_librilight_csv.py
TParcollet: This is an excellent question as I have been dealing with recipes with 10k+ hours recently and i've had this exact same issue. I resorted in adding a yaml argument to just do the dataprep. 

I think that the key here is to find the potential issue that it may cause. If one starts a training with a very long data prep on a single process, personally, I don't think it is a problem and it will work just fine.  Problem is ... these very long data prep imply that people will want to use multiple GPUs, hence DDP, hence time out on the dataprep. There is also the cluster occupation issue, as it is inefficient to occupy multiple GPUs for a few hours.

I think that what @shucongzhang proposes here makes sense, and we should convert Gigaspeech to do the same. However, we would also need to discuss more about that with people like @pplantinga, @mravanelli and @asu.

shucongzhang: Hi @Adel-Moumen How are you. Thank you so much for the quick and detailed comments! I really appreciate it! For the script, I've rewritten it with SpeechBrain utils functions. For having a separate  data preparation, I think the reasons are

1. Before the SpeechBrain data preparation, the user needs to use external VAD to pre-process the data, which is already a separate step. 
2. There is no dev set for LibriLight. Thus, the dev.csv should be taken from somewhere else,  which is also already a separate data preparation. 
3. As you said, the data is huge and prepare it first before training may be a good option. 

I'm also discussing with Titouan about this. I'll update this PR after there's a conclusion about how should we do the data preparation for this dataset. 

Thank you again!

Adel-Moumen: > Hi @Adel-Moumen How are you. Thank you so much for the quick and detailed comments! I really appreciate it! For the script, I've rewritten it with SpeechBrain utils functions. For having a separate data preparation, I think the reasons are
> 
>     1. Before the SpeechBrain data preparation, the user needs to use external VAD to pre-process the data, which is already a separate step.
> 
>     2. There is no dev set for LibriLight. Thus, the dev.csv should be taken from somewhere else,  which is also already a separate data preparation.
> 
>     3. As you said, the data is huge and prepare it first before training may be a good option.
> 
> 
> I'm also discussing with Titouan about this. I'll update this PR after there's a conclusion about how should we do the data preparation for this dataset.
> 
> Thank you again!

Hi! 

>     1. Before the SpeechBrain data preparation, the user needs to use external VAD to pre-process the data, which is already a separate step.

can you please elaborate on this? I didn't knew you had to run VAD on top of Libri-Light. According to the official paper, they said that they already ran a VAD no?

>     2. There is no dev set for LibriLight. Thus, the dev.csv should be taken from somewhere else,  which is also already a separate data preparation.

For the dev set, I guess we can directly in `train.py` or something like that take `N%` of training set as dev set? 

Let me know the outcome of the discussion. 

Do you plan to add a recipe e.g. best-rq or wav2vec? Woudl be nice to have that with this PR.

recipes/Libri-Light/self-supervised-learning/librilight_prepare.py
```diff
@@ -0,0 +1,270 @@
...
+2. Use the data_preparation/cut_by_vad.py script of the Libri-Light repo to do the vad. For example,
+    "python cut_by_vad.py --input_dir path_to_Libri-Light/small --output_dir Libri-Light_vad/small_vad --target_len_sec 20"
+
+Author
+------
+ * Mirco Ravanelli, 2020
+ * Ju-Chieh Chou, 2020
+ * Loren Lugosch, 2020
+ * Pierre Champion, 2023
+ * Adel Moumen, 2024
```
Adel-Moumen, line 16: Not sure about hwo relevants those people are with respect to this file. If they didnt contributed to this file, then you can remove them

recipes/Libri-Light/self-supervised-learning/librilight_prepare.py
```diff
@@ -0,0 +1,270 @@
...
+    if skip_prep:
+        return
+    data_folder = data_folder
+    splits = vad_splits
+    save_folder = save_folder
+
+    # Other variables
+    # Saving folder
+    if not os.path.exists(save_folder):
+        os.makedirs(save_folder)
```
Adel-Moumen, line 90: -> `os.makedirs(save_folder, exist_ok=True)`

Adel-Moumen (commented):
Hi @shucongzhang, I left you some comments. Thanks again for the PR :) I will try to train an SSL model so that we have something to release along this PR :)

shucongzhang:
> Hi @shucongzhang, I left you some comments. Thanks again for the PR :) I will try to train an SSL model so that we have something to release along this PR :)

Hi @Adel-Moumen , thank you for the comments again! Sorry for the delay, I was ill last week. I have modified the PR based on your comments. I'll take holidays now and be back next year, and I wish a good holiday season to you! :)

recipes/Libri-Light/self-supervised-learning/hparams/BEST-RQ.yaml
```diff
@@ -0,0 +1,184 @@
...
+avoid_if_longer_than: 60.0
+avoid_if_shorter_than: 2.0
+log_interval: 500 # Logging every N optimizer steps
+max_grad_norm: 10
+precision: fp16 # bf16, fp16 or fp32
+
+# The training will either stops at number_of_epochs or optimizer_step_limit
+# I.e. the first that is reached.
+number_of_epochs: 3000
+optimizer_step_limit: 300000
```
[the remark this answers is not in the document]
shucongzhang: Yes, I just follow @TParcollet for this :)

recipes/Libri-Light/self-supervised-learning/librilight_prepare.py
```diff
@@ -0,0 +1,270 @@
...
+@dataclass
+class LLRow:
+    snt_id: str
+    duration: float
+    file_path: str
+
+
+def process_line(wav_file) -> LLRow:
+    # snt_id = wav_file.split("/")[-1].replace(".flac", "")
+    snt_id = wav_file
```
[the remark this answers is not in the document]
shucongzhang: Under different folders, there are duplicated names for the .flac files. I've double checked and under different folders the .flac file with the same names are different utterances. Thus, in the beginning, I just use the full audio path as the ID. Now, the ID is changed to the subfolder name under each split + the name of the .flac.

TParcollet (approved):
CI keeps crashing due to storage issue. But once it's green LGTM!

PR #2767, opened 2024-11-22
Add People's Speech (30,000 hours) Conformer ASR (Code from Samsung AI Center Cambridge)

TParcollet:
This PR introduces an ASR training and an optional data preparation for performing ASR with the People's Speech Dataset.

# To-do:
- [x] Finish training the model
- [x] Answer the question of to download or not to download?
- [x] Add recipe testing. **WE CAN T RECIPE TEST RECIPES BASED ON HF DATASET BECAUSE CSV ARE NOT USED**

**Bellow question has been partially answered here (see bellow)**
This PR raises an important discussion that we must have with core maintainers and people wishing to participate (@pplantinga, @mravanelli, @Adel-Moumen, @asumagic, @poonehmousavi, @ycemsubakan, @Gastron): What should we do with HuggingFace datasets download? HuggingFace datasets are going to be more and more around, and luckily, the brilliant @Gastron built us a DynamicItemDataset and a set of functions that just work out of the box for it... which means that it is ultra simple to integrate in SB (as seen in this recipe). **HOWEVER**, HuggingFace datasets must be downloaded before being loaded. SpeechBrain politic has always been to let the user download the dataset, and I think this totally make sense, and i'd like to keep it that way. However, for Gigaspeech for instance, we (I) already broke this rule (sigh). My idea (Adel's actually), would be to dissociate the data_prep and download scripts. We should provide the users another .py that can be run before the recipe to download the dataset. OR we can also not provide any script to avoid maintenance and just give a few instructions to our users. The problem with the latter is that downloading an huggingface dataset is slightly more complex that wget-ing a link -- and users could get it wrong i.e. struggle to match it with our further data prep (csv creation).

**The proposal in this PR**
 It is to the user to download the dataset via HuggingFace beforehand, much like we do for pretty much all the other recipes. I voluntarily deactivate the ability of HF to go look on the internet the dataset. This will force users to DL it where they want beforehand. Being consistent is the most important thing, and it will ease maintainance.


recipes/PeoplesSpeech/ASR/transformer/extra_requirements.txt
```diff
@@ -0,0 +1 @@
+datasets
```
Adel-Moumen, line 1: what about fixing this dep to a specific datasets version?

recipes/PeoplesSpeech/ASR/transformer/hparams/conformer_large.yaml
```diff
@@ -0,0 +1,287 @@
+# ############################################################################
+# Model: E2E ASR with Transformer
+# Encoder: Conformer Encoder
+# Decoder: Transformer Decoder + (CTC/ATT joint) beamsearch + TransformerLM
+# Tokens: BPE
+# losses: CTC + KLdiv (Label Smoothing loss)
+# Training: People's Speech 28,000 hours
+# Authors:  Titouan Parcollet
+# ############################################################################
+# Seed needs to be set at top of yaml, before objects with parameters are made
```
Adel-Moumen, line 10: duplicated line

recipes/PeoplesSpeech/ASR/transformer/hparams/conformer_large.yaml
```diff
@@ -0,0 +1,287 @@
...
+seed: 3407
+__set_seed: !apply:speechbrain.utils.seed_everything [!ref <seed>]
+output_folder: !ref results/conformer_transducer_large/<seed>
+output_wer_folder: !ref <output_folder>/
+save_folder: !ref <output_folder>/save
+train_csv: !ref <save_folder>/train.csv
+train_log: !ref <output_folder>/train_log.txt
+data_folder: /ephemeral/datasets/people_speech
+test_2: /fss/shared_saic-speech/t.parcollet/mnt/results/conformer_en/3407/test.csv
+
```
Adel-Moumen, line 22: nice hardcoded paths my dude

recipes/PeoplesSpeech/ASR/transformer/hparams/conformer_large.yaml
```diff
@@ -0,0 +1,287 @@
...
+win_length: 32
+
+# This setup works well for A100 80GB GPU, adapts it to your needs.
+# Or turn it off (but training speed will decrease)
+max_batch_length_train: 500
+max_batch_length_val: 100 # we reduce it as the beam is much wider (VRAM)
+num_bucket: 200
+shuffle: True # if true re-creates batches at each epoch shuffling examples.
+batch_ordering: random
+max_batch_ex: 256
```
Adel-Moumen, line 58: I am wondering if we shouldn't talk about htat in the README?  I have always found people struggling with the concept of minimal audio hours per optimisation step. Would be maybe nice to say that you need to play with these parameters to fit them in your GPU while trying to reach N hours of audio per adam step

recipes/PeoplesSpeech/ASR/transformer/train.py
```diff
@@ -0,0 +1,458 @@
...
+
+    # 2. Define audio pipeline:
+    @sb.utils.data_pipeline.takes("audio")
+    @sb.utils.data_pipeline.provides("sig")
+    def audio_pipeline(audio):
+        if audio["sampling_rate"] != 16000:
+            sig = torchaudio.transforms.Resample(
+                audio["sampling_rate"],
+                16000,
+            )(audio["array"])
```
Adel-Moumen, line 269: shouldn't we have sampling_rate arg in the yaml? Also, dont we have our own interface for resmapling ?

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,398 @@
...
+        f"Preparing CSV of the Peoples Speech dataset in {save_folder}..."
+    )
+
+    os.makedirs(save_folder, exist_ok=True)
+
+    i = 0
+    for split, output in save_output.items():
+        logger.info(f"Starting creating {output} using {split} split.")
+        HF_create_csv(output, hf_dataset[i], split)
+        i += 1
```
Adel-Moumen, line 141: can't you use enumerate instead of having an int variable that is incremented ?

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,398 @@
...
+
+    logger.info("Data preparation completed!")
+
+
+def load_and_concatenate_datasets(subsets, hf_download_folder):
+    """Load/download and concatenate all the specified subsets from People's
+    speech
+    """
```
Adel-Moumen, line 149: missing docstring for args

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,398 @@
...
+        msg += " hf_download_folder. Make sure to set these variables properly."
+        raise Exception(msg)
+
+    logger.info("Loading dataset from: " + str(hf_caching_dir))
+
+    import multiprocessing
+
+    nproc = (
+        multiprocessing.cpu_count() // 2 + 1
+    )  # we don't want to use all cores
```
Adel-Moumen, line 176: shouldn't it be something that the user is supposed to set?

recipes/PeoplesSpeech/peoples_speech_prepare.py
Adel-Moumen: i think function is lacking a lot of details of what is happening here e.g. caching etc

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,398 @@
...
+    Preprocess English text from the People's Speech dataset into space-separated
+    words.
+    This removes various punctuation and treats it as word boundaries.
+    It normalises and retains various apostrophes (’‘´) between letters, but not
+    other ones, which are probably quotation marks.
+    It capitalises all text.
+    This function may error out if new characters show up in the training, dev,
+    or test sets.
+    """
+
```
Adel-Moumen, line 345: the docstring is not SB compliant :p

recipes/PeoplesSpeech/ASR/transformer/train.py
```diff
@@ -0,0 +1,458 @@
...
+    train_batch_sampler = None
+    valid_batch_sampler = None
+
+    dynamic_hparams = hparams["dynamic_batch_sampler"]
+    dynamic_hparams_valid = hparams["dynamic_batch_sampler_valid"]
+    num_buckets = dynamic_hparams["num_buckets"]
+
+    train_batch_sampler = DynamicBatchSampler(
+        train_data,
+        dynamic_hparams["max_batch_length"] * 1000,
```
Adel-Moumen, line 312: are you converting the units to seconds or something like that? maybe you should create a proper variable to better understand what are you doing here

recipes/PeoplesSpeech/ASR/transformer/extra_requirements.txt
TParcollet: blblbl but which one.

TParcollet: No, I don't want to in the end.

recipes/PeoplesSpeech/ASR/transformer/hparams/conformer_large.yaml
TParcollet: No, we have a tutorial about that.

recipes/PeoplesSpeech/ASR/transformer/extra_requirements.txt
Adel-Moumen: then face the consequences of having this recipe not working in the next release of HF datasets because they decided to modify the way caching works. enjoy. : p

recipes/PeoplesSpeech/peoples_speech_prepare.py
TParcollet: No, I found a few bugs with HuggingFace multiprocessing. It shouldn't be too high or it starts to go blblblblblbl. Fixing it like that is good imho .

recipes/PeoplesSpeech/ASR/transformer/train.py
TParcollet: Nah and Nah. It should be 16000 anyway. And why use an interface when you can use the native function?

recipes/PeoplesSpeech/ASR/transformer/train.py
TParcollet: Yeah, I had put a comment at some point ... but it disapeared.

Adel-Moumen (approved):
lgtm

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,421 @@
...
+    Returns
+    -------
+    List of HuggingFace dataset.
+    """
+
+    try:
+        import datasets
+        from datasets import concatenate_datasets, load_dataset
+    except ImportError:
+        raise ImportError("HuggingFace datasets must be installed.")
```
Adel-Moumen, line 170: I think you shouldn't raise the exception like that. It can silently skip some errors (e.g. one missing package while you have datasets).

recipes/PeoplesSpeech/ASR/transformer/hparams/conformer_large.yaml
```diff
@@ -0,0 +1,284 @@
...
+__set_seed: !apply:speechbrain.utils.seed_everything [!ref <seed>]
+output_folder: !ref results/conformer_transducer_large/<seed>
+output_wer_folder: !ref <output_folder>/
+save_folder: !ref <output_folder>/save
+train_csv: !ref <save_folder>/train.csv
+train_log: !ref <output_folder>/train_log.txt
+
+# Data files
+hf_download_folder: !PLACEHOLDER
+subsets: ["clean", "dirty"] # _sa sets are removed.
```
Adel-Moumen, line 22: wait. this is for training right? 

What about evals? I think one should document somewhere that you must DL the evals splits as well :p

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,421 @@
...
+
+    hf_dataset = load_and_concatenate_datasets(subsets, hf_download_folder)
+
+    logger.info(
+        f"Preparing CSV of the Peoples Speech dataset in {save_folder}..."
+    )
+
+    os.makedirs(save_folder, exist_ok=True)
+
+    for i, split, output in enumerate(save_output.items()):
```
Adel-Moumen, line 137: this line is not working :P you need to replace it by `for i, (split, output) in ...`

recipes/PeoplesSpeech/ASR/transformer/train.py
```diff
@@ -0,0 +1,460 @@
...
+        prepare_peoples_speech,
+    )
+
+    # multi-gpu (ddp) save data preparation
+    run_on_main(
+        prepare_peoples_speech,
+        kwargs={
+            "hf_download_folder": hparams["hf_download_folder"],
+            "subsets": hparams["subsets"],
+            "save_folder": hparams["save_folder"],
```
Adel-Moumen, line 368: Question: I found many recipes saving `csv` directly inside of the `output_folder`. In this case, you made the choice of saving manifests in the `save_folder` (i.e. `output_folder/save`). I was wondering, but, which one should we stick to in the future?

Adel-Moumen:
Just left some new comments. I run the code and I could extract and train for a few steps the transformer model. Lets merge after the new batch of comments is fixed :)

recipes/PeoplesSpeech/ASR/transformer/train.py
TParcollet: Yes, it's a long standing issue, I don't have an answer for that to be honest..

recipes/PeoplesSpeech/peoples_speech_prepare.py
```diff
@@ -0,0 +1,421 @@
...
+
+    # Managing the download dir as HF can be capricious with this.
+    if "HF_HUB_CACHE" in os.environ:
+        hf_caching_dir = os.environ["HF_HUB_CACHE"]
+    elif "HF_HOME" in os.environ:
+        hf_caching_dir = os.environ["HF_HOME"]
+    else:
+        hf_caching_dir = os.environ["XDG_CACHE_HOME"]
+
+    if hf_caching_dir != hf_download_folder:
```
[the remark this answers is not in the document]
TParcollet: That is where the arrow files are extracted ... and we don't want our users to be confused by HF hiding super heavy stuff in random places (what an absolutely horrendous software design).

recipes/PeoplesSpeech/peoples_speech_prepare.py
Adel-Moumen: In my case, I had a missing package in datasets while having datasets. So the ImportError was not helpful at all to debug this (i didn't had pyarrow for a Compute Canada reason :p)

recipes/PeoplesSpeech/peoples_speech_prepare.py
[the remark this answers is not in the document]
Adel-Moumen: i need to think more about this but ok

pplantinga:
I agree with Titouan that we need some consistent way for dataset download. Here's one more place where this already exists in the repo:

https://github.com/speechbrain/speechbrain/blob/develop/recipes/Voicebank/voicebank_prepare.py#L394

In this case this is just a function that users can call if they want to download, but its not explicitly called in the recipe (iirc). I would be fine with either something like this or a short additional file that calls a function like this. that people can choose to run themselves.

Adel-Moumen:
Thanks @TParcollet :)

Issue #2764, opened 2024-11-21
SpeechBrain Quantization refactoring


Adel-Moumen:
### Describe the bug

I'd like to raise a concern about how quantization is currently handled in SpeechBrain. While training my own k-means quantizer on the last layer of an ASR model, I noticed that the interface was not modular or suitable enough for this purpose.

Indeed, the current interface hardcodes many assumptions about the models (e.g., expecting an SSL model—why?). Additionally, several functions that, in my opinion, don't belong in the core `speechbrain` library are included there; they seem more appropriate for `recipes`. For example, consider the following function:

```python
def accumulate_and_extract_features(
    batch, features_list, ssl_model, ssl_layer_num, device
):
    """Extract features (output of SSL model) and accumulate them on the CPU for clustering.

    Arguments
    ---------
    batch : tensor
        A single batch of data.
    features_list : list
        List to accumulate features.
    ssl_model : torch.nn.Module
        The SSL model used to extract features for clustering.
    ssl_layer_num : int
        Specifies which layer's output of the SSL model to use.
    device : str
        CPU or GPU.
    """
    batch = batch.to(device)
    wavs, wav_lens = batch.sig
    wavs, wav_lens = wavs.to(device), wav_lens.to(device)
    feats = ssl_model(wavs, wav_lens)[ssl_layer_num].flatten(end_dim=-2)
    features_list.extend(feats.to("cpu").detach().numpy())
```

This function is part of `speechbrain/speechbrain/utils/kmeans.py`. I find its presence here surprising because, so far, we've always loaded audio directly in our `recipes` and not in the main package (e.g., via our audio pipeline functions). By rewriting these bits of code, we could allow users to modify recipes easily without needing to touch `speechbrain/speechbrain`. I think parts of this code should be moved from `kmeans.py` to, say, `train.py` in the quantization folder.

Furthermore, by bypassing the `Brain` class, we lose several important features. For instance, we can no longer use bf16 for embedding extraction, which seems like a waste of computational efficiency. I believe there are ways to use the `Brain` class directly for training k-means. If the class isn't modular enough, we could adapt it to support such functionality.

I propose that we discuss whether we want to keep the current structure or refactor it. Personally, I favor relying more on the `Brain` class and having quantization implementations that resemble traditional SB recipes. Quantization is a hot topic nowadays, especially with speech LMs, and having a clean and user-friendly implementation could make our toolkit more appealing for this purpose.

Feel free to share your thoughts. Thanks! :)

[...]

PR #2775, opened 2024-12-03
Quantization refactor

lucadellalib:
- Add K-means model
- Add example recipe

See https://github.com/speechbrain/speechbrain/issues/2764.


TParcollet:
I will do a more precise review later when it's ready. The idea looks good to me. I'm wondering about having a kmeans into lobes ... maybe this is a place where the lobes/integrations folder will make sense? Or utils maybe? kmeans is sufficiently widely used to make it logical to have a utils/kmeans.py or smth like that, maybe.

mravanelli:
Thank you @lucadellalib for this PR! This significantly improves the quantization process and makes it more "SpeechBrain-like". Below are my comments:

- I tested the recipe, and everything runs as expected

- The quantization recipe in LJSpeech must also be updated. 
- We need to ensure that the new quantization recipe produces outputs compatible with HiFi-GAN discrete training. Have you tested this? See `recipes/LJSpeech/TTS/vocoder`
- This recipe uses `scikit-learn` as an extra dependency, which isn’t included in SpeechBrain by default. I suggest creating an extra-requirements.txt file and adding scikit-learn there. Please also document this new dependency in the README.md file.

- In the future, I will consider moving `speechbrain/lobes/models/kmeans.py` to the integration folder that @pplantinga is preparing. This aligns well with the idea of placing techniques relying on external dependencies in the integration folder.
- Please update the README.md with clear instructions for training. Include links to logs, the pre-trained model, etc.

recipes/LibriSpeech/quantization/train_kmeans.py
```diff
@@ -0,0 +1,258 @@
...
+            # Partial fit on current batch
+            self.hparams.quantizer.partial_fit(feats)
+
+            # For K-means the training loss is the drift between current centroids and old centroids
+            # If close to 0, it means that the training has converged
+            curr_cluster_centers = self.hparams.quantizer.cluster_centers
+            loss = (curr_cluster_centers - old_cluster_centers).norm()
+
+            # Workaround to keep consistency with standard supervised training
+            loss.requires_grad_()
```
Adel-Moumen, line 69: why is it the case? Maybe we could directly fix core.py

Adel-Moumen:
Hey @lucadellalib, I like what you proposed. I think its much more in the spirit of what we are doing in SpeechBrain. I went through the code and saw some lines were fixes to make Brain work with kmeans (e.g. the returned loss). Can't we directly modify core.py to make the code more compatible ? I think there's some flexibility that could be improved on Brain side

recipes/LibriSpeech/quantization/train_kmeans.py
```diff
@@ -0,0 +1,258 @@
...
+        stage_stats = {"loss": stage_loss}
+
+        if stage == sb.Stage.TRAIN:
+            self.avg_train_loss = 0.0
+            self.train_stats = stage_stats
+            self.stats_meta = {"epoch": epoch, "steps": self.optimizer_step}
+            if if_main_process():
+                self.checkpointer.save_and_keep_only(
+                    meta={"loss": stage_stats["loss"], "epoch": epoch},
+                    max_keys=["epoch"],
```
pplantinga, line 104: JSYK the checkpointer will always save the most recent checkpoints, the max_keys is not needed here, though its fine to have it to make things explicit.

speechbrain/lobes/models/kmeans.py
```diff
@@ -0,0 +1,168 @@
+"""K-means implementation.
+
+Authors
+* Luca Della Libera 2024
+"""
+
+import joblib
+import torch
+from sklearn.cluster import MiniBatchKMeans
```
pplantinga, line 9: It would be cool to have this also in PyTorch, like this https://github.com/jokofa/torch_kmeans

But for practical reasons I guess it makes sense to keep it this way, though we will want to add it to `integrations` most likely.

speechbrain/lobes/models/kmeans.py
pplantinga, line 1: Should we deprecate the `speechbrain/utils/kmeans.py` in favor of this one?

recipes/LibriSpeech/quantization/train_kmeans.py
```diff
@@ -0,0 +1,258 @@
...
+                loss = torch.tensor(self.avg_train_loss)
+                # Workaround to keep consistency with standard supervised training
+                loss.requires_grad_()
+                return loss
+            self.feats = torch.cat(self.feats)
+            feats = self.feats[: self.hparams.kmeans_batch_size]
+            self.feats = [self.feats[self.hparams.kmeans_batch_size :]]
+            self.curr_batch_size = len(self.feats[0])
+
+        if stage == sb.Stage.TRAIN:
```
pplantinga, line 56: Minor comment: there are two `if stage == train` in a row
```suggestion
```

recipes/LibriSpeech/quantization/train_kmeans.py
```diff
@@ -0,0 +1,258 @@
...
+            self.optimizer_step += 1
+            assert self.optimizer_step == self.modules.quantizer.n_steps, (
+                self.optimizer_step,
+                self.modules.quantizer.n_steps,
+            )
+        else:
+            # For K-means the validation/test loss is the inertia
+            # The lower the inertia, the better should be the clustering
+            # It is useful to monitor progress across epochs
+            # However, when saving checkpoints we always keep the last one (i.e. max_keys=["epoch"])
```
pplantinga, line 79: I don't suppose anyone has tested if clusters with lower inertia provide better quantization, i.e. better performance on downstream tasks...

pplantinga (commented):
Everything looks good, only minor comments.

recipes/LibriSpeech/quantization/train_kmeans.py
lucadellalib: SpeechBrain expects a loss with a gradient. This is a very special case, I don't think it's worth addressing it in `core.py`, as doing so might lead to silent bugs in other situations (typically the loss gradient should come from the model's parameters). I've added additional comments to clarify this.

lucadellalib:
@TParcollet @mravanelli @Adel-Moumen @pplantinga I've addressed the comments. I agree that moving this to `integration` makes sense. I compared my implementation to the original code, and when run with the same parameters and implementation details, it produces identical results. However, I made two modifications to the original code:
1) **Accumulated features**:  It looks like for some reason, in the original code, the feature list is emptied instead of keeping the remaining accumulated features for the next iteration (https://github.com/speechbrain/speechbrain/blob/v1.0.2/speechbrain/utils/kmeans.py#L197) 
2) **K-means batch size**: I believe the batch size is set incorrectly (`batch_size=hparams["batch_size"]` instead of `batch_size=hparams["kmeans_batch_size"]`), not sure what's the impact (https://github.com/speechbrain/speechbrain/blob/v1.0.2/recipes/LJSpeech/quantization/train.py#L119)

@poonehmousavi any insights?

recipes/LibriSpeech/quantization/train_kmeans.py
lucadellalib: Not that I'm aware of

speechbrain/lobes/models/kmeans.py
lucadellalib: Public PyTorch implementations typically implement K-means instead of mini-batch K-means, this requires several modifications. I had a PR for that (https://github.com/speechbrain/speechbrain/pull/2411), but it was decided to move on with the scikit-learn implementation. It would be hard now to switch while maintaining backward compatibility, since we have a lot of models trained using the scikit-learn recipe.

mravanelli (approved):

mravanelli:
I ran the recipes again and everything works as expected. I think we can merge it, but I would like to give @TParcollet and @Adel-Moumen the opportunity for the last comments (if any).

recipes/LJSpeech/quantization/hparams/train_discrete_ssl.yaml
```diff
@@ -1,78 +1,102 @@
...
+output_folder: !ref results/<experiment_name>/<seed>
+save_folder: !ref <output_folder>/save
+cache_folder: !name:huggingface_hub.constants.HUGGINGFACE_HUB_CACHE
 
-# Dataloader options
-train_dataloader_opts:
-    batch_size: !ref <batch_size>
-    drop_last: True
+# Preprocessing parameters
+# For some reason the duration is not available in the manifest file...
```
Adel-Moumen, line 27: abt this, can't we simply modify the prepare recipe to have an extra duration column ?

recipes/LJSpeech/quantization/train.py
```diff
@@ -1,142 +1,268 @@
...
-    for dataset in hparams["splits"]:
-        datasets[dataset] = sb.dataio.dataset.DynamicItemDataset.from_json(
-            json_path=data_info[dataset],
-            replacements={"data_root": hparams["data_folder"]},
-            dynamic_items=[audio_pipeline],
-            output_keys=["id", "sig"],
+        sig = torchaudio.functional.resample(
+            sig, original_sample_rate, hparams["sample_rate"]
         )
+        yield sig
```
Adel-Moumen, line 188: why are you using this function with yield instead of the typical return?

recipes/LJSpeech/quantization/hparams/train_discrete_ssl.yaml
lucadellalib: Done

recipes/LJSpeech/quantization/train.py
lucadellalib: I typically use `yield` in pipelines so they can be readily extended to return multiple values if needed

recipes/LJSpeech/quantization/README.md
```diff
@@ -1,74 +1,102 @@
...
+  url     = {http://jmlr.org/papers/v25/24-0991.html}
 }
-@misc{speechbrain,
-  title={{SpeechBrain}: A General-Purpose Speech Toolkit},
-  author={Mirco Ravanelli and Titouan Parcollet and Peter Plantinga and Aku Rouhe and Samuele Cornell and Loren Lugosch and Cem Subakan and Nauman Dawalatabad and Abdelwahab Heba and Jianyuan Zhong and Ju-Chieh Chou and Sung-Lin Yeh and Szu-Wei Fu and Chien-Feng Liao and Elena Rastorgueva and François Grondin and William Aris and Hwidong Na and Yan Gao and Renato De Mori and Yoshua Bengio},
-  year={2021},
-  eprint={2106.04624},
-  archivePrefix={arXiv},
-  primaryClass={eess.AS},
-  note={arXiv:2106.04624}
```
Adel-Moumen, line 72: why did you replace this one?

recipes/LJSpeech/quantization/README.md
```diff
@@ -1,74 +1,102 @@
...
+@article{ravanelli2024open,
+  author  = {Mirco Ravanelli and Titouan Parcollet and Adel Moumen and Sylvain de Langen and Cem Subakan and Peter Plantinga and Yingzhi Wang and Pooneh Mousavi and Luca {Della Libera} and Artem Ploujnikov and Francesco Paissan and Davide Borra and Salah Zaiem and Zeyu Zhao and Shucong Zhang and Georgios Karakasidis and Sung-Lin Yeh and Pierre Champion and Aku Rouhe and Rudolf Braun and Florian Mai and Juan Zuluaga-Gomez and Seyed Mahed Mousavi and Andreas Nautsch and Ha Nguyen and Xuechen Liu and Sangeet Sagar and Jarod Duret and Salima Mdhaffar and Ga{{\"e}}lle Laperri{{\`e}}re and Mickael Rouvier and Renato De Mori and Yannick Est{{\`e}}ve},
+  title   = {Open-Source Conversational {AI} with {SpeechBrain} 1.0},
+  journal = {Journal of Machine Learning Research},
+  year    = {2024},
+  volume  = {25},
+  number  = {333},
+  pages   = {1--11},
+  url     = {http://jmlr.org/papers/v25/24-0991.html}
 }
```
Adel-Moumen, line 91: I understnad that this is the official new citation, but you could please replace `ravanelli2024open` by `speechbrainV1` to remain consistent with our past formatting. I will do the modifications for the other ones.

recipes/LJSpeech/quantization/train.py
```diff
@@ -1,142 +1,267 @@
...
+        loss = (curr_cluster_centers - old_cluster_centers).norm()
+
+        # Keep compatibility with standard supervised training
+        # (SpeechBrain expects a tensor with gradient)
+        loss.requires_grad_()
+        self.optimizer_step += 1
+        assert self.optimizer_step == self.modules.quantizer.n_steps, (
+            f"optimizer_step: {self.optimizer_step}",
+            f"quantizer.n_steps: {self.modules.quantizer.n_steps}",
+        )
```
Adel-Moumen, line 91: remark: I actually do think that SpeechBrain Brain class will need at some point a small refactor to allow the possibility to return a Loss object that is not from PyTorch and where you won't do BP on it.

Adel-Moumen (approved):
LGTM. Thanks @lucadellalib for the refactoring.

PR #2780, opened 2024-12-06
Fix multi-head-attention with adapters

### LINKED ISSUE #2779: Using Speechbrain MultiheadAttention with PEFT attribute error
### Describe the bug

Hi all,
Particularly @TParcollet @pplantinga and @poonehmousavi 

### Description
When attempting to apply the LoRA and Adapter from #2563  to Speechbrain [Conformer](https://huggingface.co/speechbrain/asr-conformer-transformerlm-librispeech). In the `MultiheadAttention` module in SpeechBrain, an `AttributeError` is raised. The error traceback is as follows:

```
  File "/venv/lib/python3.12/site-packages/torch/nn/modules/activation.py", line 1380, in forward
    self.out_proj.weight,
    ^^^^^^^^^^^^^^^^^^^^
File "/venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1931, in __getattr__
    raise AttributeError(
AttributeError: 'LoRA' object has no attribute 'weight'
```
### Cause
The issue arises from the implementation of the MultiheadAttention class in the SpeechBrain library, specifically in the file `speechbrain/speechbrain/nnet/attention.py` which is a wrapper from Pytorch MultiheadAttention. The relevant code snippet is located on line 864:

```
        self.att = nn.MultiheadAttention(
            embed_dim=d_model,
            num_heads=nhead,
            dropout=dropout,
            bias=bias,
            add_bias_kv=add_bias_kv,
            add_zero_attn=add_zero_attn,
            kdim=kdim,
            vdim=vdim,
        )
[...]
        output, attention_weights = self.att(
            query,
            key,
            value,
            attn_mask=attn_mask,
            key_padding_mask=key_padding_mask,
            need_weights=return_attn_weights,
        )
```

In the Pytorch MultiheadAttention class   implementation, it directly accesses self.out_proj.weight in its forward method, as shown in the following excerpt from the PyTorch documentation:
```
 1369             attn_output, attn_output_weights = F.multi_head_attention_forward(
 1370                 query,
 1371                  key,
 1372                 value,
 1373                 self.embed_dim,
 1374                 self.num_heads,
 1375                 self.in_proj_weight,
 1376                 self.in_proj_bias,
 1377                 self.bias_k,
 1378                 self.bias_v,
 1379                 self.add_zero_attn,
 1380                 self.dropout,
 1381                 **self.out_proj.weight,**
 1382                 **self.out_proj.bias,**
 1383                 training=self.training,
 1384                 key_padding_mask=key_padding_mask,
 1385                 need_weights=need_weights,
 1386                 attn_mask=attn_mask,
 1387                 average_attn_weights=average_attn_weights,
 1388                 is_causal=is_causal,
 1389             )
```
In the current implementation of Speechbrain and Adapter/LoRA, the `out_proj` attribute of the `MultiheadAttention` class is an instance of the `LoRA` class, which does not have a `weight` attribute. The structure is as follows:

```
MultiheadAttention(
  (out_proj): LoRA(
    (pretrained_module): NonDynamicallyQuantizableLinear(in_features=512, out_features=512, bias=True)
    (adapter_down_proj): Linear(in_features=512, out_features=8, bias=False)
    (adapter_up_proj): Linear(in_features=8, out_features=512, bias=False)
  )
)

```
Since the PyTorch MultiheadAttention implementation does not invoke the forward method of the LoRA class, but directly attempts to access self.out_proj.weight, it leads to the AttributeError.

Did I miss anything while applying the Adapter/LoRA?
Do you have any suggestions on how to resolve or prevent this issue?

### Expected behaviour

The MultiheadAttention should use the out-projection linear layer in its forward method to use LoRA/Adapter forward function.

### To Reproduce

**conformer_large_adapter.yaml**

```

seed: 3407
__set_seed: !apply:speechbrain.utils.seed_everything [!ref <seed>]
output_folder: !ref results/conformer_large_test_LORA/<seed>
output_wer_folder: !ref <output_folder>/
save_folder: !ref <output_folder>/save
train_log: !ref <output_folder>/train_log.txt

pretrained_lm_tokenizer_path: speechbrain/asr-transformer-transformerlm-librispeech
pretrained_path: speechbrain/asr-conformer-transformerlm-librispeech

data_folder: foo  # e.g., /path/to/LibriSpeech

train_splits: ["train-clean-100", "train-clean-360", "train-other-500"]
dev_splits: ["dev-clean"]
test_splits: ["dev-clean", "test-clean", "test-other"]
skip_prep: False
train_csv: !ref <output_folder>/train.csv
valid_csv: !ref <output_folder>/dev-clean.csv
test_csv:
    - !ref <output_folder>/dev-clean.csv
    - !ref <output_folder>/test-clean.csv
    - !ref <output_folder>/test-other.csv

number_of_epochs: 1
batch_size: 16 # This works for 2x GPUs with 32GB
ctc_weight: 0.3
grad_accumulation_factor: 1
max_grad_norm: 5.0
loss_reduction: 'batchmean'
sorting: random
num_workers: 4
precision: fp32 # bf16, fp16 or fp32
avg_checkpoints: 10 # Number of checkpoints to average for evaluation

lr_adam: 0.0008

sample_rate: 16000
n_fft: 512
n_mels: 80
win_length: 32

dynamic_batching: True
max_batch_length_train: 500
max_batch_length_val: 100 # we reduce it as the beam is much wider (VRAM)
num_bucket: 200
shuffle: True # if true re-creates batches at each epoch shuffling examples.
batch_ordering: random
max_batch_ex: 256

dynamic_batch_sampler_train:
    max_batch_length: !ref <max_batch_length_train>
    num_buckets: !ref <num_bucket>
    shuffle: !ref <shuffle>
    batch_ordering: !ref <batch_ordering>
    max_batch_ex: !ref <max_batch_ex>

dynamic_batch_sampler_valid:
    max_batch_length: !ref <max_batch_length_val>
    num_buckets: !ref <num_bucket>
    shuffle: !ref <shuffle>
    batch_ordering: !ref <batch_ordering>
    max_batch_ex: !ref <max_batch_ex>

train_dataloader_opts:
    batch_size: !ref <batch_size>
    shuffle: True
    num_workers: !ref <num_workers>

valid_dataloader_opts:
    batch_size: 1

test_dataloader_opts:
    batch_size: 1


d_model: 512
nhead: 8
num_encoder_layers: 12
num_decoder_layers: 6
d_ffn: 2048
transformer_dropout: 0.1
activation: !name:torch.nn.GELU
output_neurons: 5000

blank_index: 0
label_smoothing: 0.1
pad_index: 0
bos_index: 1
eos_index: 2

min_decode_ratio: 0.0
max_decode_ratio: 1.0
valid_search_interval: 10
valid_beam_size: 10
test_beam_size: 66
lm_weight: 0.60
ctc_weight_decode: 0.40


CNN: !new:speechbrain.lobes.models.convolution.ConvolutionFrontEnd
    input_shape: (8, 10, 80)
    num_blocks: 2
    num_layers_per_block: 1
    out_channels: (64, 32)
    kernel_sizes: (3, 3)
    strides: (2, 2)
    residuals: (False, False)

Transformer: !new:speechbrain.lobes.models.transformer.TransformerASR.TransformerASR # yamllint disable-line rule:line-length
    input_size: 640
    tgt_vocab: !ref <output_neurons>
    d_model: !ref <d_model>
    nhead: !ref <nhead>
    num_encoder_layers: !ref <num_encoder_layers>
    num_decoder_layers: !ref <num_decoder_layers>
    d_ffn: !ref <d_ffn>
    dropout: !ref <transformer_dropout>
    activation: !ref <activation>
    encoder_module: conformer
    attention_type: RelPosMHAXL
    normalize_before: True
    causal: False


**new_Transformer: !new:speechbrain.nnet.adapters.AdaptedModel
    model_to_adapt: !ref <Transformer>
    adapter_class: !name:speechbrain.nnet.adapters.LoRA
    all_linear: True
    manual_adapter_insertion: True
    adapter_kwargs:
        rank: 8**

lm_model: !new:speechbrain.lobes.models.transformer.TransformerLM.TransformerLM # yamllint disable-line rule:line-length
    vocab: !ref <output_neurons>
    d_model: 768
    nhead: 12
    num_encoder_layers: 12
    num_decoder_layers: 0
    d_ffn: 3072
    dropout: 0.0
    activation: !name:torch.nn.GELU
    normalize_before: False

tokenizer: !new:sentencepiece.SentencePieceProcessor

ctc_lin: !new:speechbrain.nnet.linear.Linear
    input_size: !ref <d_model>
    n_neurons: !ref <output_neurons>

seq_lin: !new:speechbrain.nnet.linear.Linear
    input_size: !ref <d_model>
    n_neurons: !ref <output_neurons>

normalize: !new:speechbrain.processing.features.InputNormalization
    norm_type: global
    update_until_epoch: 4

**modules:
    CNN: !ref <CNN>
    Transformer: !ref <new_Transformer>
    seq_lin: !ref <seq_lin>
    ctc_lin: !ref <ctc_lin>
    normalize: !ref <normalize>**

Adam: !name:torch.optim.AdamW
    lr: !ref <lr_adam>
    betas: (0.9, 0.98)
    eps: 0.000000001

**model_pretrained: !new:torch.nn.ModuleList
    - [!ref <CNN>, !ref <Transformer>, !ref <seq_lin>, !ref <ctc_lin>]

model: !new:torch.nn.ModuleList
    - [!ref <CNN>, !ref <new_Transformer>, !ref <seq_lin>, !ref <ctc_lin>]**


ctc_scorer: !new:speechbrain.decoders.scorer.CTCScorer
    eos_index: !ref <eos_index>
    ctc_fc: !ref <ctc_lin>


transformerlm_scorer: !new:speechbrain.decoders.scorer.TransformerLMScorer
    language_model: !ref <lm_model>
    temperature: 1.15

scorer_test_search: !new:speechbrain.decoders.scorer.ScorerBuilder
    full_scorers: [!ref <transformerlm_scorer>, !ref <ctc_scorer>]
    weights:
        ctc: !ref <ctc_weight_decode>
        transformerlm: !ref <lm_weight>

scorer_valid_search: !new:speechbrain.decoders.scorer.ScorerBuilder
    full_scorers: [!ref <ctc_scorer>]
    weights:
        ctc: !ref <ctc_weight_decode>

valid_search: !new:speechbrain.decoders.S2STransformerBeamSearcher
    modules: [!ref <new_Transformer>, !ref <seq_lin>]
    bos_index: !ref <bos_index>
    eos_index: !ref <eos_index>
    min_decode_ratio: !ref <min_decode_ratio>
    max_decode_ratio: !ref <max_decode_ratio>
    beam_size: !ref <valid_beam_size>
    using_eos_threshold: False
    length_normalization: True
    scorer: !ref <scorer_valid_search>

test_search: !new:speechbrain.decoders.S2STransformerBeamSearcher
    modules: [!ref <new_Transformer>, !ref <seq_lin>]
    bos_index: !ref <bos_index>
    eos_index: !ref <eos_index>
    min_decode_ratio: !ref <min_decode_ratio>
    max_decode_ratio: !ref <max_decode_ratio>
    beam_size: !ref <test_beam_size>
    temperature: 1.15
    using_eos_threshold: False
    length_normalization: True
    scorer: !ref <scorer_test_search>

log_softmax: !new:torch.nn.LogSoftmax
    dim: -1

ctc_cost: !name:speechbrain.nnet.losses.ctc_loss
    reduction: !ref <loss_reduction>

seq_cost: !name:speechbrain.nnet.losses.kldiv_loss
    label_smoothing: !ref <label_smoothing>
    reduction: !ref <loss_reduction>

noam_annealing: !new:speechbrain.nnet.schedulers.NoamScheduler
    lr_initial: !ref <lr_adam>
    n_warmup_steps: 30000

checkpointer: !new:speechbrain.utils.checkpoints.Checkpointer
    checkpoints_dir: !ref <save_folder>
    recoverables:
        model: !ref <model>
        noam_scheduler: !ref <noam_annealing>
        normalizer: !ref <normalize>
        counter: !ref <epoch_counter>

epoch_counter: !new:speechbrain.utils.epoch_loop.EpochCounter
    limit: !ref <number_of_epochs>

speed_perturb: !new:speechbrain.augment.time_domain.SpeedPerturb
    orig_freq: !ref <sample_rate>
    speeds: [95, 100, 105]

time_drop: !new:speechbrain.augment.freq_domain.SpectrogramDrop
    drop_length_low: 15
    drop_length_high: 25
    drop_count_low: 4
    drop_count_high: 4
    replace: "mean"

freq_drop: !new:speechbrain.augment.freq_domain.SpectrogramDrop
    drop_length_low: 10
    drop_length_high: 20
    drop_count_low: 4
    drop_count_high: 4
    replace: "mean"
    dim: 2

time_warp: !new:speechbrain.augment.freq_domain.Warping

fea_augment: !new:speechbrain.augment.augmenter.Augmenter
    min_augmentations: 3
    max_augmentations: 3
    augment_prob: 1.0
    augmentations: [
        !ref <time_drop>,
        !ref <freq_drop>,
        !ref <time_warp>]

compute_features: !new:speechbrain.lobes.features.Fbank
    sample_rate: !ref <sample_rate>
    n_fft: !ref <n_fft>
    n_mels: !ref <n_mels>
    win_length: !ref <win_length>


train_logger: !new:speechbrain.utils.train_logger.FileTrainLogger
    save_file: !ref <train_log>

error_rate_computer: !name:speechbrain.utils.metric_stats.ErrorRateStats
acc_computer: !name:speechbrain.utils.Accuracy.AccuracyStats

**pretrainer: !new:speechbrain.utils.parameter_transfer.Pretrainer
    collect_in: !ref <save_folder>
    loadables:
        lm: !ref <lm_model>
        tokenizer: !ref <tokenizer>
        model: !ref <model_pretrained>
    paths:
        lm: !ref <pretrained_lm_tokenizer_path>/lm.ckpt
        tokenizer: !ref <pretrained_lm_tokenizer_path>/tokenizer.ckpt
        model: !ref <pretrained_path>/asr.ckpt**

```

**train.py**
```
[...]

     hparams["pretrainer"].collect_files()

     # Add Adapters
     hparams["new_Transformer"].insert_adapters()


     # Trainer initialization
     asr_brain = ASR(
         modules=hparams["modules"],
         opt_class=hparams["Adam"],
         hparams=hparams,
         run_opts=run_opts,
     )

[....]
```

pplantinga:
Closes #2779 

PyTorch's multi-head attention apparently directly accesses the weight of linear layers. `peft` handles this by adding a shortcut. We'll just do the same, although I'm not confident that MHA is getting adapted appropriately if its short-circuiting the adapter.

Relevant peft line:
https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/model.py#L254


pplantinga:
I tested this with our TransformerASR and it seems to fix the issue.

TParcollet:
Is this only due to PEFT or our adapters as well?

pplantinga:
> Is this only due to PEFT or our adapters as well? 

This problem exists in our adapters when applied to torch's MHA -- I only looked to peft to see their solution.

> (Also WTF is this design from torch?!)

Yeah, seems like bad design from torch folks. I'm wondering if it was done for speed? On this line they call the C interface for `linear` which takes the weight and bias directly:

https://github.com/pytorch/pytorch/blob/274223d719cf5895d641dabac08f03cad9a323bf/torch/nn/functional.py#L6378

I'd guess this means the output projection cannot have adapters added, so the most proper solution might be to detect when we are applying adapters to MHA and emit a warning that these cannot be adapted.

TParcollet:
@pplantinga Should we add a small unittest for that? Checking if it gets adapted properly?

pplantinga:
```python
import torch
from peft import LoraConfig, get_peft_model
from speechbrain.lobes.models.transformer.TransformerASR import TransformerASR
from speechbrain.nnet.adapters import AdaptedModel, LoRA

src = torch.rand([4, 120, 512], requires_grad=True)
tgt = torch.randint(0, 720, [4, 120])

model = TransformerASR(720, 512, causal=False)
enc_out, dec_out = model(src, tgt)

adapted_layer = "encoder.layers.0.self_att.att.out_proj"
unfrozen_layer = "encoder.layers.1.self_att.att.out_proj"

adapted = AdaptedModel(model, LoRA, target_layers=[adapted_layer], unfrozen_layers=[unfrozen_layer])
layer = adapted.get_submodule(adapted_layer + ".adapter_down_proj").weight
layer.register_hook(lambda x: print("grads updated in SB LoRA"))
unfroz = adapted.get_submodule(unfrozen_layer).weight
unfroz.register_hook(lambda x: print("grads updated in SB (no lora)"))

enc_out, dec_out = adapted(src, tgt)

model2 = TransformerASR(720, 512, causal=False)
lora_config = LoraConfig(r=16, target_modules=[adapted_layer], modules_to_save=[unfrozen_layer])
lora_model = get_peft_model(model2, lora_config)
lora_model.print_trainable_parameters()
layer2 = lora_model.get_submodule(adapted_layer + ".lora_A.default").weight
layer2.register_hook(lambda x: print("grads updated in peft LoRA"))
unfroz2 = lora_model.get_submodule(unfrozen_layer).weight
unfroz2.register_hook(lambda x: print("grads updated in peft (no lora)"))

enc_out2, dec_out2 = lora_model(src, tgt)

loss = torch.nn.functional.mse_loss(enc_out, src)
loss.backward()

loss2 = torch.nn.functional.mse_loss(enc_out2, src)
loss2.backward()
```

So I ran this script and the output shows that the adapters on the output projections are not getting updated in either speechbrain nor peft. Perhaps the path forward is to merge this change now, at least its not worse than peft, and  either create an issue at pytorch to see if we can get it fixed upstream or create an issue at peft to see if there's a workaround.

pplantinga:
I'd guess the reason this hasn't come up for `peft` before is that the majority of HF `transformers` do not use the native PyTorch implementation of MHA, for example Whisper uses:

https://github.com/huggingface/transformers/blob/11ba1d472c61eaacdc58a12e31156d4436b132ce/src/transformers/models/whisper/modeling_whisper.py#L473

But this interesting note in `peft` shows that the native one uses a fused kernel -- probably a lot faster, but impossible to modify. Perhaps we should give an option in our own transformer implementation to use the optimized one or a less-optimized but configurable one.

https://github.com/huggingface/transformers/blob/11ba1d472c61eaacdc58a12e31156d4436b132ce/src/transformers/modeling_utils.py#L4961-L4970

pplantinga:
I changed my mind. Accessing the weight directly (as PyTorch MHA does) should fail, as it bypasses the adapters. Instead, I skip adding MHA layers to the list of adapted layers and emit a warning. This should be fine for now, it solves the immediate issue.

The potential longer term issue is that the speechbrain MHA layer projections are not adaptable. I'm not sure how important the parameters in the MHA projection layers are in the overall performance of adapting transformers, but if it ends up being important I suppose we should consider incorporating an alternative MHA  in the transformer architecture for adaptation. Its possible it could even load weights from the pre-trained PyTorch MHA layer and seamlessly continue training. But I think this is a separate issue that deserves another PR to solve.

pplantinga:
Any comments from @TParcollet or @Adel-Moumen regarding the most recent suggested approach? I could add an issue if we want to measure the performance difference between adapting our current transformer implementation and a fully adaptable one.

Adel-Moumen (approved):
LGTM! 

As discussed privately, I think it would be great for us to have our own implementation of MHA relying on flash attention/pytorch one.

PR #2782, opened 2024-12-10
Move files with optional dependencies to integrations folder

pplantinga:
In internal discussions, we decided to move as much of our external-dependent code as possible into a single folder for a couple of reasons:

- so we could add disclaimers that this stuff could stop working at any time
- additional testing requirements for code placed in this part of the repo, so we know if it stops working
- testing is only run before each release, alongside recipe tests, rather than for every CI build
- A readme and requirements file for using each integration

The basic concept is that we will add a python module `speechbrain.integrations` that contains any speechbrain blocks that depend on external libraries. So far we have:

- [x] speechbrain.lobes.models.huggingface_transformers => speechbrain.integrations.huggingface
- [x] speechbrain.wordemb => speechbrain.integrations.huggingface.wordemb
- [x] speechbrain.k2_integration => speechbrain.integrations.k2_fsa
- [x] speechbrain.lobes.models.discrete.speechtokenizer => speechbrain.integrations.audio_tokenizers.speechtokenizer_interface
- [x] speechbrain.lobes.models.discrete.wavtokenizer => speechbrain.integrations.audio_tokenizers.wavtokenizer_interface
- [x] speechbrain.lobes.models.flair => speechbrain.integrations.nlp
- [x] speechbrain.lobes.models.spacy => speechbrain.integrations.nlp
- [x] speechbrain.decoders.language_model => speechbrain.integrations.decoders.kenlm_scorer
- [x] speechbrain.alignment.ctc_segmentation => speechbrain.integrations.alignment.ctc_segmentation
- [x] speechbrain.lobes.models.kmeans => speechbrain.integrations.audio_tokenizers.kmeans
- [x] speechbrain.utils.bleu => speechbrain.integrations.nlp.bleu
- [x] speechbrain.processing.diarization => speechbrain.integrations.alignment.diarization
- [ ] ~~speechbrain.nnet.losses.transducer => speechbrain.integrations.transducer~~

Deprecations:
- [x] speechbrain.lobes.models.fairseq_wav2vec
- [x] speechbrain.utils.kmeans

After this approach is verified, I will:

- [x] Find and replace all deprecated references to the corresponding new paths
- [x] Update path for any integrations that I've missed, e.g. kmeans, audio tokenizers (now merged)
- [x] Add documentation to explain how to add to this folder, testing requirements, etc.

We need to ensure all integrated third-party folders have at least one doctest. Still needed:

- [x] spacyNLP
- [x] flairNLP


poonehmousavi:
Thanks @pplantinga … this change would be really helpful in adding new tokenizers.  Following are also needed to be transferred to integration folder:
- https://github.com/speechbrain/speechbrain/tree/develop/speechbrain/lobes/models/discrete
- https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/beats.py

pplantinga:
> Thanks @pplantinga … this change would be really helpful in adding new tokenizers. Following are also needed to be transferred to integration folder:
> 
> * https://github.com/speechbrain/speechbrain/tree/develop/speechbrain/lobes/models/discrete
> * https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/beats.py

I've added `discrete` to the list above, but I don't see any external imports in `beats.py`, does this code depend on an external library? IMO if the code is now internal (even if the original code was not) we don't need to put it in `integrations`

pplantinga:
> I've added `discrete` to the list above, but I don't see any external imports in `beats.py`, does this code depend on an external library? IMO if the code is now internal (even if the original code was not) we don't need to put it in `integrations`

In addition, are we keeping a folder for discrete in `lobes.models`? This makes the redirect a bit more complex.

poonehmousavi:
> > Thanks @pplantinga … this change would be really helpful in adding new tokenizers. Following are also needed to be transferred to integration folder:
> > 
> > * https://github.com/speechbrain/speechbrain/tree/develop/speechbrain/lobes/models/discrete
> > * https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/lobes/models/beats.py
> 
> I've added `discrete` to the list above, but I don't see any external imports in `beats.py`, does this code depend on an external library? IMO if the code is now internal (even if the original code was not) we don't need to put it in `integrations`

Actually, I transferred all their code to SB so it not dependant on external library, we just initialized the weights with their checkpoints

Adel-Moumen (commented):
Hi @pplantinga, the changes looks good to me. I think you also need to move Flair and spacy folder since they also depend on an external library. I would also advocate for having a README.md in each subfolder within integrations to explain how to install and make the dep and speechbrain working together (e.g. specifying the compatible versions). Finally, I would say that we should also discuss about the folders lobes and the naming conventions. I tend to think that we could increase the readability by having better names and maybe even subfolders? Idk. :)

pplantinga:
One more question for the crew, up to this point, we've been ignoring the CI tests for the integrations because it adds to the CI time and increases the likelihood that the CI breaks because we have to install a bunch of extra packages (and I'm wondering if some of them might not work on the CI server anyway). Do we want to add these tests back in, or is this something we want to continue skipping and just say that integrations are at your own risk. @TParcollet etc.

pplantinga:
There is  a long-standing issue with fairseq in newer Python versions (2.11+) that doesn't seem likely to get addressed anytime soon. Should we just go ahead and deprecate our integration?

https://github.com/facebookresearch/fairseq/issues/5012

pplantinga:
This is ready for review. I ran the coverage test listed in the `README.txt` and this is the result:

```
---------- coverage: platform linux, python 3.12.7-final-0 -----------
Name                                                          Stmts   Miss  Cover
---------------------------------------------------------------------------------
speechbrain/integrations/alignment/ctc_seg.py                   191     13    93%
speechbrain/integrations/discrete/kmeans.py                      51     10    80%
speechbrain/integrations/discrete/speechtok.py                   27      3    89%
speechbrain/integrations/discrete/wavtok.py                      32      5    84%
speechbrain/integrations/flairnlp/__init__.py                     6      2    67%
speechbrain/integrations/flairnlp/embeddings.py                  24      1    96%
speechbrain/integrations/flairnlp/sequencetagger.py              18      9    50%
speechbrain/integrations/huggingface/__init__.py                 17      5    71%
speechbrain/integrations/huggingface/discrete_ssl.py            100     12    88%
speechbrain/integrations/huggingface/encodec.py                 108      8    93%
speechbrain/integrations/huggingface/gpt.py                      30      9    70%
speechbrain/integrations/huggingface/hubert.py                    6      0   100%
speechbrain/integrations/huggingface/huggingface.py             119     41    66%
speechbrain/integrations/huggingface/labse.py                    30      7    77%
speechbrain/integrations/huggingface/llama2.py                  113     83    27%
speechbrain/integrations/huggingface/mbart.py                    49     11    78%
speechbrain/integrations/huggingface/mert.py                      6      0   100%
speechbrain/integrations/huggingface/mimi.py                     38      3    92%
speechbrain/integrations/huggingface/nllb.py                      6      0   100%
speechbrain/integrations/huggingface/textencoder.py              22      5    77%
speechbrain/integrations/huggingface/vocos.py                    46      4    91%
speechbrain/integrations/huggingface/wav2vec2.py                 69     15    78%
speechbrain/integrations/huggingface/wavlm.py                     6      0   100%
speechbrain/integrations/huggingface/weighted_ssl.py             29      3    90%
speechbrain/integrations/huggingface/whisper.py                 196     78    60%
speechbrain/integrations/huggingface/wordemb/__init__.py          0      0   100%
speechbrain/integrations/huggingface/wordemb/transformer.py      90     67    26%
speechbrain/integrations/huggingface/wordemb/util.py             11      0   100%
speechbrain/integrations/k2_fsa/__init__.py                       8      4    50%
speechbrain/integrations/k2_fsa/graph_compiler.py               117     50    57%
speechbrain/integrations/k2_fsa/lattice_decoder.py              108     68    37%
speechbrain/integrations/k2_fsa/lexicon.py                      158     33    79%
speechbrain/integrations/k2_fsa/losses.py                        11      0   100%
speechbrain/integrations/k2_fsa/prepare_lang.py                 194      8    96%
speechbrain/integrations/k2_fsa/utils.py                         51     28    45%
speechbrain/integrations/lm/ken.py                               96     74    23%
speechbrain/integrations/metrics/bleu.py                         51      9    82%
speechbrain/integrations/processing/diarization.py              317    157    50%
speechbrain/integrations/spacynlp/__init__.py                     5      2    60%
speechbrain/integrations/spacynlp/nlp.py                         19      1    95%
speechbrain/integrations/tests/test_ctc_segmentation.py          32      0   100%
speechbrain/integrations/tests/test_k2.py                       167      0   100%
---------------------------------------------------------------------------------
TOTAL                                                          2774    828    70%
```

So we  should decide which (if any) of these need to be increased before merge.

TParcollet:
"speechbrain.nnet.losses.transducer => speechbrain.integrations.transducer ??? Should we deprecate in favor of e.g. torchaudio?" --> No, torchaudio loss is crazily slower than our Numba impl. For the tests, i'd be in favor of having them, but it sounds like a nightmare to maintain. At the same time, we provide this folder as a main part of the library... so our users may expect it to work properly. Could we have a specific CI for this folder? Like one that is only triggered if the code in it is touched?

TParcollet:
Also that is a big PR, reviewing / testing will be slow.

pplantinga:
> Also that is a big PR, reviewing / testing will be slow.

You are right, its big but perhaps not as big as it looks since most of the changes are to module path names. For reviewing I would recommend just taking a look around the new integrations folder and seeing if the organization / readmes make sense. Some testing should be done too but again the functional changes are mostly just to module path names which should be unlikely to break anything. Biggest risk I guess is that I somehow put the wrong path for something, which recipe tests should catch.

speechbrain/__init__.py
```diff
@@ -34,6 +46,9 @@ def make_deprecated_redirections():
         also_lazy_export=True,
     )
 
+    for old_path, new_path in deprecations.items():
+        deprecated_redirect(old_path, new_path, also_lazy_export=True)
```
Adel-Moumen, line 50: should we also add an `extra_reason` ?

speechbrain/integrations/discrete/README.md
Adel-Moumen, line 1: rather than `discrete`, maybe we could have been more specific i.e. `audio_tokenizers` or something like that.

speechbrain/integrations/discrete/speechtok.py
Adel-Moumen, line 1: i think keeping the full name is better no? `speech_tokenizer.py` or `speechtokenizer.py`

speechbrain/integrations/discrete/wavtok.py
Adel-Moumen, line 1: same for this one `wavtokenizer` or `wav_tokenizer.py`

speechbrain/integrations/huggingface/discrete_ssl.py
Adel-Moumen, line 1: shouldn't this file go to `discrete` folder instead? The model doesn't seems to use the `HFTransformersInterface` abstract class that we have in `huggingface.py`. It seems to be more an orchestrator of different pre-trained blocks.

speechbrain/lobes/models/discrete/dac.py
Adel-Moumen, line 1: `dac` or `DAC` ?

speechbrain/lobes/models/fairseq_wav2vec.py
Adel-Moumen, line 1: Is there a specific reason why we are dropping FairSeq support? I tend to agree but maybe we should be more specific about the reason why?

speechbrain/lobes/models/kmeans.py
```diff
@@ -1,178 +1,11 @@
-"""K-means implementation.
+"""This file ensures old links to kmeans continue to work while providing a Deprecation warning"""
```
Adel-Moumen, line 1: this is something I am not fully sure to understand, isn't the `deprecation` module allows to remove files and still have the compatibility of doing let's say `speechbrain.lobes.models.kmeans` ?

speechbrain/integrations/discrete/speechtok.py
pplantinga: We can't keep the same name as the imported library as this leads to some sort of import error. But perhaps there's a better option than `speechtok.py` like `speechtokenizer_interface.py` or something.

speechbrain/integrations/discrete/wavtok.py
pplantinga: `wavtokenizer_interface.py` ?

speechbrain/integrations/huggingface/discrete_ssl.py
pplantinga: Ah, good idea!

speechbrain/lobes/models/kmeans.py
pplantinga: I just looked again, and I don't think there's any automatic way to do this.

speechbrain/lobes/models/discrete/dac.py
pplantinga: I think the Python standard for file names is all lower case -- due to different operating systems treating upper/lower case differently.

speechbrain/lobes/models/fairseq_wav2vec.py
pplantinga: FB has stopped active development of FairSeq (last released version was 2022) and there is some dependency issue with a new release of some package though I can't remember which one (cuda? pytorch?)

speechbrain/__init__.py
pplantinga: Sure, this can be added.

speechbrain/integrations/discrete/speechtok.py
Adel-Moumen: Yeah I understand now. I know that ESPNet is using the convention name `tokenizer`. E.g. from `espnet.espnet2.speechlm.tokenizer import X` The only issue with this is that `tokenizer` is a broad term and can be applied to anything. I am honestly not against having a folder that can have audio/speech tokenizers, as well as text (if one day this happen) or vision tokenizers.

speechbrain/integrations/discrete/wavtok.py
Adel-Moumen: what about `speechbrain.integrations.tokenizer.wavtokenizer` ?

speechbrain/lobes/models/fairseq_wav2vec.py
Adel-Moumen: FB stopping FairSeq, now Torchaudio :(

speechbrain/integrations/discrete/speechtok.py
Adel-Moumen: Btw, shouldn't `sentencepiece` be moved in the `.integrations.` folder ?

pplantinga:
Unfortunately the MERT huggingface model is not compatible with the latest version of transformers (see https://github.com/huggingface/transformers/issues/36134). I'll set up the test with an earlier version but we may have to deprecate at some point.

speechbrain/integrations/discrete/speechtok.py
pplantinga: Well currently `sentencepiece` is a non-optional dependency, but `integrations` is only for optional dependencies. I think we use it enough throughout the toolkit that perhaps its worth keeping it that way.

Adel-Moumen: I understand!

pplantinga:
Okay, this should be pretty much ready to go @TParcollet @Adel-Moumen 

Any last comments?

speechbrain/__init__.py
pplantinga: This is the message, do you think there's something else we should add here? Like version or something?

```
>>> from speechbrain.wordemb import transformer
<stdin>:1: UserWarning: Module 'speechbrain.wordemb' was deprecated, redirecting to 'speechbrain.integrations.huggingface.wordemb'. Please update your script.
```

PR #2799, opened 2025-01-13
Rotary Position Embedding (RoPE) for ASR (code from Samsung Cambridge)

shucongzhang:
## What does this PR do?
This PR implements  Rotary Position Embedding (RoPE) https://arxiv.org/pdf/2104.09864.

It improves the training speech of LibriSpeech by 13%, while giving slightly better ASR results. 

RoPE is also useful for other datasets. We have tested it with Lirbiheavy, CommonVoice, and Voxpopuli. The full results are shown in this short paper https://arxiv.org/pdf/2501.06051. In this PR, we only submit the recipe of RoPE on LibriSpeech for simplicity. If this can be merged, we would like to also submit recipes for other datasets.


TParcollet:
@pplantinga @Adel-Moumen @mravanelli this is an important PR in the sense that we should, from now on, use RoPE for most models we develop. The reason is that it's definitely faster, and better. We may want to retrain some models with it ... it's also a good distinction with ESPnet and NeMO I believe.

recipes/LibriSpeech/ASR/transformer/hparams/conformer_large_rope.yaml
```diff
@@ -0,0 +1,338 @@
...
+# Please, set your parameters accordingly.
+number_of_epochs: 120
+batch_size: 16 # This works for 2x GPUs with 32GB
+ctc_weight: 0.3
+grad_accumulation_factor: 1
+max_grad_norm: 5.0
+loss_reduction: 'batchmean'
+sorting: random
+num_workers: 4
+precision: fp32 # bf16, fp16 or fp32
```
TParcollet, line 58: change to fp16 by default please.

recipes/LibriSpeech/ASR/transformer/hparams/conformer_large_rope.yaml
```diff
@@ -0,0 +1,338 @@
+# ############################################################################
```
TParcollet, line 1: This guy looks VERY similar to our conformer_large.yaml, right? Why not changing the main one with RoPE. The model is just better, so I don't see any problem with that.

speechbrain/lobes/models/transformer/Conformer.py
```diff
@@ -730,6 +738,11 @@ def forward(
                 raise ValueError(
                     "The chosen attention type for the Conformer is RelPosMHAXL. For this attention type, the positional embeddings are mandatory"
                 )
+        elif self.attention_type == "RoPEMHAXL":
```
TParcollet, line 741: Why the XL? Why not just RoPEMHA?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -228,7 +242,7 @@ def __init__(
                 activation=activation,
                 normalize_before=normalize_before,
                 causal=True,
-                attention_type="regularMHA",  # always use regular attention in decoder
+                attention_type="regularMHA",
```
TParcollet, line 245: Do not remove this comment please.

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -787,16 +806,17 @@ def forward(
...
 
         # multi-head attention over the target sequence and encoder states
-
         tgt2, multihead_attention = self.multihead_attn(
             query=tgt1,
             key=memory,
             value=memory,
-            attn_mask=memory_mask,
-            key_padding_mask=memory_key_padding_mask,
+            attn_mask=memory_mask,  # none
```
TParcollet, line 813: remove comments?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -787,16 +806,17 @@ def forward(
...
+            key_padding_mask=memory_key_padding_mask,  # a mask for padding
             pos_embs=pos_embs_src,
         )
 
+        # breakpoint()
```
TParcollet, line 818: remove.

speechbrain/lobes/models/transformer/TransformerASR.py
```diff
@@ -262,7 +263,7 @@ def __init__(
         conformer_activation: Optional[nn.Module] = Swish,
         branchformer_activation: Optional[nn.Module] = nn.GELU,
         attention_type: Optional[str] = "regularMHA",
-        max_length: Optional[int] = 2500,
+        max_length: Optional[int] = 10000,
```
TParcollet, line 266: This breaks backward compatibility. I 100% agree that 2500 is too little, but it will break backward compatibility when loading old checkpoints... can we keep it to 2500?

speechbrain/nnet/attention.py
```diff
@@ -938,3 +939,346 @@ def forward(self, x):
...
+
+
+class RotationMatrix(nn.Module):
+    """
+    The rotation matrice for the :class:`~RoPEMHA`.
+    The shape of these matrices is supposedly to be max_lengh * d_attnhead * d_attnhead.
+    However, due to the sparsity of the rotation matrices, we only need to store max_lengh * d_attnhead elements
+    of the matrix.
+    see eq(15) of https://arxiv.org/pdf/2104.09864
+
```
TParcollet, line 951: Add a docstring example please.  We may want to unittest that as well, see the unitary tests that we already have!

speechbrain/nnet/attention.py
```diff
@@ -938,3 +939,346 @@ def forward(self, x):
...
+            0,
+            max_len,
+            dtype=torch.float32,
+        )
+
+        # equation (15) without zeros in the matrix
+        positions_inv_freq = torch.outer(positions, inv_freq)
+
+        cosines = torch.cos(positions_inv_freq)
+        # (cos(m*theta_0), cos(m*theta_0), cos(m*theta_1), cos(m*theta_1) ,... ) for equantion (34)
```
TParcollet, line 980: equation (typo)

speechbrain/nnet/attention.py
```diff
@@ -938,3 +939,346 @@ def forward(self, x):
...
+        cosines = torch.stack([cosines, cosines], dim=-1).reshape(
+            max_len, emb_dim
+        )
+
+        sines = torch.sin(positions_inv_freq)
+        # (sin(m*theta_0), sin(m*theta_0), sin(m*theta_1), sin(m*theta_1) ,... ) for equantion (34)
```
TParcollet, line 986: equation (typo)

TParcollet (changes requested):
Thanks @shucongzhang, just a few minor comments. Also as discussed together, let see how the discussion for Torch attention vs homemade attention goes with the others.

speechbrain/nnet/attention.py
```diff
@@ -1432,15 +1432,17 @@ def forward(
             key_padding_mask = key_padding_mask.view(bsz, 1, 1, klen).expand(
                 bsz, self.num_heads, klen, qlen
             )
+            torch.logical_not(key_padding_mask)
```
pplantinga, line 1435: Does this line do anything? Shouldn't it be stored back into a variable?

TParcollet:
I think that there is a software/hardware curiosity that could be of interest to @asumagic @pplantinga and @Adel-Moumen here. I just added a unit test comparing our homemade attention to torch attention (see the unit test folder). You will quickly see that this test passes easily when using CPU, but it fails miserably when using GPU. I am wondering if this is an actual CUDNN issue or something else?

asumagic:
I don't really know, can it just be expected precision loss? Or are the values significantly wrong?

TParcollet:
@asumagic it seems to be a precision issue, but that remains very surprising to me. It's a single layer. Also, when training models, those with torch attention always obtain slightly worst performance than the one with our homemade attention. If it was a CUDNN kernel precision issue, I would expect to see the models being sometimes better, sometimes worst, no? Otherwise it means that the kernels are just flawed and always lead to imprecise calculation? This would be insane.

asumagic:
Have you tried `SB_DISABLE_QUIRKS=allow_tf32`?

TParcollet:
@asumagic wait it actually works, wtf????

asumagic:
Yes, possibly, TF32 isn't supported everywhere. Some kernels (mostly matmul in this case, I guess) are going to compute at a reduced precision out-of-the-box when TF32 is supported (e.g. A100), but this is accuracy loss is probably very similar to enabling fp16/bf16 autocast to start with.

For inference I don't think it matters all that much but I haven't tried on a lot of models.

TParcollet:
@asumagic but then, it also means that for fp16, we're just doomed entirely? Because most models we have trained are trained with fp16, and this is where we see the difference in performance at the end...

asumagic:
I don't really know how significant the issue is TBH, especially when it comes to the difference between the custom SB attention vs torch attention.  
It's not so surprising to me that TF32 SB attention vs torch attention would have different accuracy characteristics because of the amount of fusing(?) that might happen, but I am really not sure, and I have no idea how much of a relative WER% loss I would expect from that.

recipes/LibriSpeech/ASR/transformer/train.py
```diff
@@ -135,7 +142,7 @@ def compute_objectives(self, predictions, batch, stage):
         loss_ctc = self.hparams.ctc_cost(
             p_ctc, tokens, wav_lens, tokens_lens
         ).sum()
-
+
```
rogiervd, line 145: This is unintentional, right? (Above as well.)

speechbrain/nnet/attention.py
```diff
@@ -938,3 +939,518 @@ def forward(self, x):
...
+        # (sin(m*theta_0), sin(m*theta_0), sin(m*theta_1), sin(m*theta_1) ,... ) for equantion (34)
+        sines = torch.stack([sines, sines], dim=-1).reshape(max_len, emb_dim)
+
+        self.register_buffer("cosines", cosines)
+        self.register_buffer("sines", sines)
+
+    @torch.no_grad()
+    def forward(self, x: torch.Tensor):
+        """
+        Builds the rotation matrice.
```
rogiervd, line 995: Matrix. But also, it doesn't.

rogiervd (commented):
On the basis of the performance, this is worthwhile implementing. I'm assuming this works well. However, I am unable to ascertain that the rotation is correct. My main recommendations are:

1. Rename `RotationMatrix` to `SinusoidCache` or something else that is actually correct, and make it an implementation detail.
2. I may be wrong, but it looks like there's a huge amount of code duplication going on.
3. I see no actual test for the RoPE bit.

For the test, I think the rotation speeds need to be customisable, so you can actually test anything. If all the if-statements for masks and so on remain in the new code, I would prefer to see these checked too. I would recommend:
* A test for the actual rotation, checked against a dumb implementation. The actual rotation would be in a separate function, which would probably be good form anyway.
* One or both of these:
  * A test like the following. Set the input, K, and V to identity matrices and Q to a permutation matrix (maybe in pairs of dimensions). Dimension 8 or so. That should make it easy to check the correct result manually.
  * A test where you input a few random tensors and test against a dumb (Numpy?) implementation of RoPE.

TParcollet:
@Adel-Moumen ready for the review. All the above comments can be only very briefly checked as we have worked intenally (all three of us) to make it better. You can give your input now.

recipes/LibriSpeech/ASR/transformer/README.md
```diff
@@ -37,7 +37,9 @@ Following table contains whisper-finetuning results for 1 epoch using Whisper mo
 
 | Release | hyperparams file | Dev Clean WER (No LM, small beam) | Test Clean WER (Transformer LM) | Test Other WER (Transformer LM) | HuggingFace link | Model link | GPUs |
 |:-------------:|:-------------:|:-------------:|:---------------------------:| :-----:| :-----:| :-----:| :--------:|
+| 30-09-24 | conformer_large_rope.yaml |1.85 with LM | 1.96 | 4.50 | Not Avail. | Not Avail. | 4xA40 46GB |
```
TParcollet, line 40: @shucongzhang I think we can remove this line tbh, results are not significatively different with RoPE so that is fine.

speechbrain/nnet/attention.py
```diff
@@ -938,3 +939,484 @@ def forward(self, x):
...
+    Memoises a function which has as its first argument a value that indicates a
+    minimum value to call the underlying function with.
+
+    Args:
+    function
+        The function to call.
+    round_up
+        A function that rounds up.
+        The fewer values this rounds up to, the less likely it is that the
+        function will be called repeatedly.
```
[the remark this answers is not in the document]
TParcollet: We cannot add a docstring for this class function. However it is properly unit tested, and there is an example of the decorator bellow in the code.

pplantinga (commented):
Very minor last comments cleanup

pplantinga (approved):

TParcollet (approved):
I am happy with the code. But would need a last check

Adel-Moumen:
I just ran the recipe tests and they worked for this recipe.

PR #2800, opened 2025-01-13
Readthedocs now requires a `sphinx.configuration` key

pplantinga:
Docs builds are failing due to a new requirement from Read the Docs -- a certain `sphinx.configuration` key, as outlined in this blog post:

https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

This change should add the required key.


Adel-Moumen (approved):
LGTM

PR #2801, opened 2025-01-14
Convert data download methods to run on main process only

pplantinga:
Currently recipes with noise/rir augmentations have to add lines to the `train.py` and cannot rely solely on the yaml due to the multiprocessing requirement. This fix allows us to download the noise/rir directly in the `yaml` using something like:

```yaml
# Download and prepare the dataset of noisy sequences for augmentation
prepare_noise_data: !apply:speechbrain.augment.preparation.prepare_dataset_from_URL
    URL: !ref <NOISE_DATASET_URL>
    dest_folder: !ref <data_folder_noise>
    ext: wav
    csv_file: !ref <noise_annotation>
```

### DIFF
--- a/speechbrain/augment/preparation.py
+++ b/speechbrain/augment/preparation.py
@@ -12,12 +12,14 @@ import os
 import torchaudio
 
 from speechbrain.utils.data_utils import download_file, get_all_files
+from speechbrain.utils.distributed import main_process_only
 from speechbrain.utils.logger import get_logger
 
 # Logger init
 logger = get_logger(__name__)
 
 
+@main_process_only
 def prepare_dataset_from_URL(URL, dest_folder, ext, csv_file, max_length=None):
     """Downloads a dataset containing recordings (e.g., noise sequences)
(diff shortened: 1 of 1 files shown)


TParcollet (approved):
LGTM (feel free to merge!) (won't this be backward incompatible with recipe changing the train.py? Like downloading twice?)

speechbrain/augment/preparation.py
```diff
@@ -51,6 +53,7 @@ def prepare_dataset_from_URL(URL, dest_folder, ext, csv_file, max_length=None):
         prepare_csv(filelist, csv_file, max_length)
 
 
+@main_process_only
 def prepare_csv(filelist, csv_file, max_length=None):
```
Adel-Moumen, line 57: is this function often called outside of `prepare_dataset_from_URL` ?

pplantinga: I doubt it, but it shouldn't hurt anything to add the decorator here just in case anyone does.

PR #2805, opened 2025-01-21
Fix Mimi Bug

poonehmousavi:
## What does this PR do?

- Mimi does not function correctly on CUDA because the embeddings are always loaded onto the CPU (in the __init__ function) and are never transferred to CUDA. To address this, I followed the approach in the original code by calling _compute_embedding within the forward or encode function on the first call to ensure the embeddings are loaded onto the correct device.

- Added a sampling_rate attribute for both WavTokenizer and SpeechTokenizer.

-->

Fixes #<issue_number>

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/lobes/models/discrete/speechtokenizer.py
+++ b/speechbrain/lobes/models/discrete/speechtokenizer.py
@@ -37,6 +37,8 @@ class SpeechTokenizer(nn.Module):
         HuggingFace hub name: e.g "fnlp/SpeechTokenizer"
     save_path : str
         Path (dir) of the downloaded model.
+    sample_rate : int (default: 16000)
+        The audio sampling rate
 
     Example
     -------
@@ -57,6 +59,7 @@ class SpeechTokenizer(nn.Module):
         self,
         source,
         save_path,
+        sample_rate=16000,
     ):
 
         # Lazy import to avoid circular dependency issues
@@ -84,6 +87,7 @@ class SpeechTokenizer(nn.Module):
             config_path, ckpt_path
         )
         self.model.eval()
+        self.sample_rate = sample_rate
 
     def forward(self, wav, wav_lens=None):
         """Takes an input waveform and return its corresponding wav2vec encoding.
--- a/speechbrain/lobes/models/huggingface_transformers/mimi.py
+++ b/speechbrain/lobes/models/huggingface_transformers/mimi.py
@@ -78,8 +78,8 @@ class Mimi(HFTransformersInterface):
 
         super().__init__(source=source, save_path=save_path, freeze=freeze)
         self.num_codebooks = num_codebooks
-        self.sampling_rate = sample_rate
-        self.embeddings = self._compute_embedding()
+        self.sample_rate = sample_rate
+        self.embeddings = None
 
     @torch.no_grad()
     def _compute_embedding(self):
@@ -115,6 +115,9 @@ class Mimi(HFTransformersInterface):
         audio : torch.Tensor
             the reconstructed audio
         """
+        if self.embeddings is None:
+            self.embeddings = self._compute_embedding()
+
         tokens, embedding = self.encode(inputs, length)
         audio = self.decode(tokens, length)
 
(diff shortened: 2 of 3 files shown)


speechbrain/lobes/models/huggingface_transformers/mimi.py
```diff
@@ -115,6 +115,9 @@ def forward(self, inputs, length):
         audio : torch.Tensor
             the reconstructed audio
         """
+        if self.embeddings is None:
+            self.embeddings = self._compute_embedding()
+
```
pplantinga, line 120: Not sure this is needed since `forward` directly calls `encode`

speechbrain/lobes/models/huggingface_transformers/mimi.py
```diff
@@ -178,6 +184,9 @@ def decode(self, tokens, length=None):
...
```
pplantinga, line 189: Not sure if this is needed, since the embeddings don't seem to be used later in `decode`

pplantinga (commented):
Looks pretty good, just a few questions to answer before mreging

speechbrain/lobes/models/huggingface_transformers/mimi.py
```diff
@@ -78,8 +78,8 @@ def __init__(
 
         super().__init__(source=source, save_path=save_path, freeze=freeze)
         self.num_codebooks = num_codebooks
-        self.sampling_rate = sample_rate
-        self.embeddings = self._compute_embedding()
+        self.sample_rate = sample_rate
+        self.embeddings = None
```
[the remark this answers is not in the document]
poonehmousavi: Apologies if my explanation wasn't clear earlier. I added self.embedding, which caused this [function](https://github.com/huggingface/transformers/blob/6bc0fbcfa7acb6ac4937e7456a76c2f7975fefec/src/transformers/models/mimi/modeling_mimi.py#L1261) to be called during initialization. As a result, self._embed becomes non-None.

In the original Mimi code, this function is first invoked when quantize is called, ensuring that self._embed is on the correct device. However, because I called the [_compute_embedding](https://github.com/poonehmousavi/speechbrain/blob/e5796d2d6a2a2be725520b83b7ac2130b32ae14d/speechbrain/lobes/models/huggingface_transformers/mimi.py#L85) function in the __init__ method, it leads to a "not on the same device" error.

To resolve this, I followed a similar approach and called _compute_embedding in the encode function instead. I hope this provides more clarity. Based on this explanation, what would you suggest?

speechbrain/lobes/models/huggingface_transformers/mimi.py
poonehmousavi: You are right, i could remove it.

speechbrain/lobes/models/huggingface_transformers/mimi.py
poonehmousavi: iIn the original code, the decode function calls the [self.emb](https://github.com/huggingface/transformers/blob/6bc0fbcfa7acb6ac4937e7456a76c2f7975fefec/src/transformers/models/mimi/modeling_mimi.py#L1285) function. This could potentially lead to the same "not on the same device" issue if decode is called before encode.

Such a scenario is possible when only the decoder module is used to convert tokens to signals, bypassing the encode function entirely

speechbrain/lobes/models/discrete/speechtokenizer.py
```diff
@@ -37,6 +37,8 @@ class SpeechTokenizer(nn.Module):
         HuggingFace hub name: e.g "fnlp/SpeechTokenizer"
     save_path : str
         Path (dir) of the downloaded model.
+    sample_rate : int (default: 16000)
+        The audio sampling rate
```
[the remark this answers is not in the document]
poonehmousavi: Most of these codecs are used in [DASB benchmark](https://github.com/speechbrain/benchmarks/tree/DASB/benchmarks/DASB), we always check sampling rate of the tokenizers and up/dowsample the audio according to this sampling rate. Example : [here](https://github.com/speechbrain/benchmarks/blob/9f64966ca6f0467dc140b7047ac2d908038fefc1/benchmarks/DASB/utils/tokens.py#L194)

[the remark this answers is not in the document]
pplantinga: Aha, this makes sense!

speechbrain/lobes/models/huggingface_transformers/mimi.py
pplantinga: Aha, makes sense!

pplantinga (approved):
I tried this, it works, so LGTM!

PR #2806, opened 2025-01-23
Conformer recipe for LargeScaleASR (code from Samsung AI Center Cambridge)

TParcollet:
This is the training recipe for our ASR conformer on [LargeScaleASR](https://github.com/speechbrain/speechbrain/pull/2802). 

This is by far our best speech recognizer in SpeechBrain, and we should put an entry in the HF leaderboard. I'll try to figure out how to release the ckpt from Samsung if possible!


TParcollet:
This is ready for review @Adel-Moumen and @mravanelli . We cannot add a recipe test because the recipe is only based on huggingface datasets and we have not uploaded a small test dataset there yet .... The recipe testing would need to be adapted for this use case, but that's the job for an engineer that we don't have anymore.

Adel-Moumen (approved):
LGTM!

PR #2814, opened 2025-02-03
[URGENT] Fix DDP broadcasting

TParcollet:
This is a serious issue. @peter842323 @Adel-Moumen  and @mravanelli - anyone technically sound should have an eye on it and review. **Our DDP, under some circumstances, is broken.** Basically, any module that does not contain trainable parameters but contains non trainable ones, will not be broadcasted properly. For BEST-RQ, for instance, it means that if you train with 4 GPUs, each GPU will get a different Quantizer. This is also true for input normalization! Up until now, if we use global as an input normalisation, the running mean and var will be calculated per GPU ... 

This PR addresses this by removing the offset to the seed of the processes. This makes sure that all processes initialise with the same seed i.e. same parameters. 

The module solution did not work as it's breaking most of the code base.

TODO:
- [x] Adel Check if it breaks stuff to redirect self.modules.module to self.modules (not good)
- [x] Adapt the input normalisation to use all_reduce operation to average the statistics.

### DIFF
--- a/recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
+++ b/recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
@@ -69,7 +69,7 @@ d_ffn: 2048
 transformer_dropout: 0.1
 activation: !name:torch.nn.GELU
 output_neurons: 5000
-encoder_layerdrop: 0.05
+encoder_layerdrop: 0.0
 
 # Masking parameters
 mask_length: 4
--- a/recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
+++ b/recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
@@ -35,7 +35,7 @@ number_of_epochs: 3000
 optimizer_step_limit: 300000
 
 # This setup is for 8 V100.
-seconds_per_batch: 400
+max_batch_len: 400 # In second and per gpu
 train_num_buckets: 150
 grad_accumulation_factor: 2
 
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -1715,22 +1715,20 @@ class Brain:
         if not self.distributed_launch and not self.data_parallel_backend:
             return
         elif self.distributed_launch:
-            for name, module in self.modules.items():
-                if any(p.requires_grad for p in module.parameters()):
-                    module = SyncBatchNorm.convert_sync_batchnorm(module)
-                    if self.distributed_backend == "gloo":
-                        module = DDP(
-                            module,
-                            device_ids=None,
-                            find_unused_parameters=self.find_unused_parameters,
-                        )
-                    else:
-                        module = DDP(
-                            module,
-                            device_ids=[self.device],
-                            find_unused_parameters=self.find_unused_parameters,
-                        )
-                    self.modules[name] = module
+            self.modules = SyncBatchNorm.convert_sync_batchnorm(self.modules)
+            if self.distributed_backend == "gloo":
+                self.modules = DDP(
+                    self.modules,
+                    device_ids=None,
+                    find_unused_parameters=self.find_unused_parameters,
+                )
+            else:
+                self.modules = DDP(
+                    self.modules,
+                    device_ids=[self.device],
+                    find_unused_parameters=self.find_unused_parameters,
+                )
+            self.modules = self.modules.module
         else:
             # data_parallel_backend
             for name, module in self.modules.items():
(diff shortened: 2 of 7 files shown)


speechbrain/core.py
```diff
@@ -1715,22 +1715,20 @@ def _wrap_distributed(self):
...
+                    device_ids=None,
+                    find_unused_parameters=self.find_unused_parameters,
+                )
+            else:
+                self.modules = DDP(
+                    self.modules,
+                    device_ids=[self.device],
+                    find_unused_parameters=self.find_unused_parameters,
+                )
+            self.modules = self.modules.module
```
pplantinga, line 1731: What is the purpose of this change? from `self.modules.module` to `self.modules`

Is it related to the fix for DDP? Because I thought removing the line above:

```
```

was the main fix here.

TParcollet: If you don't do this, then you can't do self.modules.Transformer. But you will need to do self.modules.module.Transformer... DDP wrapping ass an extra .module I.e refactor 100% of the recipes

TParcollet: We have changed of strategy (see above and commits)

Adel-Moumen (approved):
lgtm

recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
```diff
@@ -35,7 +35,7 @@ number_of_epochs: 3000
 optimizer_step_limit: 300000
 
 # This setup is for 8 V100.
-seconds_per_batch: 400
+max_batch_len: 400 # In second and per gpu
 train_num_buckets: 150
```
pplantinga, line 39: Are the BEST-RQ recipe changes related to the DDP fixes?

speechbrain/processing/features.py
```diff
@@ -1153,6 +1154,15 @@ def forward(self, x, lengths, spk_ids=torch.tensor([]), epoch=0):
...
 
+            # If multi GPU, we need to sync the statistics across processes.
+            if is_distributed_initialized():
+                torch.distributed.all_reduce(
+                    current_mean, op=torch.distributed.ReduceOp.AVG
+                )
+                torch.distributed.all_reduce(
+                    current_std, op=torch.distributed.ReduceOp.AVG
+                )
+
```
pplantinga, line 1165: I'm wondering if we should add something to the distributed tools to ensure this won't hang if its in a `if_main_process` block, similar to this broadcast code:

https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/distributed.py#L205

speechbrain/utils/seed.py
```diff
@@ -21,7 +20,12 @@
 def seed_everything(
     seed: int = 0, verbose: bool = True, deterministic: bool = False
 ) -> int:
-    r"""Function that sets the seed for pseudo-random number generators in: torch, numpy, and Python's random module.
+    r"""Function that sets the seed for pseudo-random number generators in: torch, numpy, and Python's random module. Important note on DDP: all DDP
+    process have the same seed. This is important to ensure that parameters
+    without a require_grad set to True are the same across processes. This
+    must be taken into account if one wants to build a custom data sampler as
+    the processes would pick the same samples... SpeechBrain takes care of that
+    internally.
```
pplantinga, line 28: Have we verified that the sampler actually does work correctly when they're all the same seed? What were the original reasons for setting separate seeds? Also, do we need to provide a mechanism to access a process-specific seed?

pplantinga (commented):
From the updates to the PR it looks like it is fine to only wrap modules with DDP if they have trainable parameters. I'm not sure how buffers work in pytorch, if a module consists only of e.g. batchnorm or other buffer-related blocks, would that fail here? I'm also not sure how SyncBatchNorm works, if this could somehow be leveraged for InputNorm or something like this.

Overall, I think using the same seed on the different processes is probably the way to go, I just want to confirm once more that everything is working with this

speechbrain/processing/features.py
TParcollet: Oh right, good catch.

speechbrain/utils/seed.py
TParcollet: The reason was that we thought it was not the case, but @Gastron  did the distributed sampler well :p

recipes/LibriSpeech/self-supervised-learning/BEST-RQ/hparams/BEST-RQ.yaml
TParcollet: Not really... but they are here on purpose -- these are minor "cosmetic" changes that I don't want to do a PR for..

speechbrain/processing/features.py
TParcollet: Done!

TParcollet:
@pplantinga i've followed your advice and added a ddp_all_reduce utility in our DDP code making sure that it does not hang if called within a main_process_only context.

pplantinga (approved):
LGTM!

PR #2822, opened 2025-02-11
Fix BEST RQ norm error and efficiency changes

whettenr:
## What does this PR do?

This PR overall fixes and improve the BEST-RQ recipe. More specifically it does the following:
- fixes issue with normalizing on the wrong dimension in the ```RandomProjectionQuantizer```
- fixes inconsistency with normalizer in pre-training and fine-tuning scripts
- makes ```compute_forward``` function more efficient by only computing pseudo-targets on masked area


whettenr:
Hi @TParcollet,
Do you think you could look at this? 
I think you would be a good reviewer for this as you are familiar with this recipe .

TParcollet (approved):
LGTM

PR #2824, opened 2025-02-13
Changing the default parameters of --compile to make it work more often

TParcollet:
This PR changes the default value of torch compile to make it work most of the time. The dynamic graph computation should be True by default --- SB only has variable sequences.. 
For instance, with this, our conformer Transforer / seq_lin and ctc_lin can be compiled and get a 10-15% training speed boost. There is still a problem with the Convolution Frontend that I will try to address.


Adel-Moumen (approved):
LGTM. Curious to know but isnt torch.compile going to re-compile for each new shape a newly optimised kernel?

TParcollet:
@Adel-Moumen not really, at least not from what I can see from the trace. I suspect that you can compile kernel where some dimensions are unspecified (Tensorflow did it)... same story with Flex Attention. But I mean, it's not crazy faster anyway.

PR #2827, opened 2025-02-14
Small improvements to core.py for AMP/gradscaler/tqdm

Adel-Moumen:
## What does this PR do?

This PR adds the support of a new context manager for AMP `TorchAutocast` which automatically handles the `precision` being used for training and evaluation. In `precision: fp32` it acts as a `nullcontext()` without any effects. This addition simplify quite a bit the `fit_batch` function. Additionally, I made a fix to avoid the ckpt adding the `scaler` object to save when it is an empty object (i.e. fp32/bf16). 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ pygtrie>=2.1,<3.0
 scipy>=1.4.1,<1.13.0
 sentencepiece>=0.1.91
 SoundFile; sys_platform == 'win32'
-torch>=2.1.0
-torchaudio>=2.1.0
+torch>=2.1.0,<=2.6.0
+torchaudio>=2.1.0,<=2.6.0
 tqdm>=4.42.0
 transformers>=4.30.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ pygtrie>=2.1,<3.0
 scipy>=1.4.1,<1.13.0
 sentencepiece>=0.1.91
 SoundFile; sys_platform == 'win32'
-torch>=2.1.0,<=2.6.0
-torchaudio>=2.1.0,<=2.6.0
+torch>=2.1.0
+torchaudio>=2.1.0
 tqdm>=4.42.0
 transformers>=4.30.0
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -756,12 +756,10 @@ class Brain:
             f"Gradscaler enabled: {gradscaler_enabled}. Using precision: {self.precision}."
         )
         self.scaler = torch.GradScaler(self.device, enabled=gradscaler_enabled)
-
         dtype = AMPConfig.from_name(self.precision).dtype
         self.training_ctx = TorchAutocast(
-            enabled=gradscaler_enabled, device=self.device, dtype=dtype
+            enabled=self.precision == "fp16" or self.precision == "bf16", device_type=self.device, dtype=dtype
         )
-        print(self.training_ctx)
         if (
             gradscaler_enabled
             and self.checkpointer is not None
@@ -1191,7 +1189,6 @@ class Brain:
                 loss = self.compute_objectives(
                     outputs, batch, sb.Stage.TRAIN
                 )
-    
             scaled_loss = self.scaler.scale(
                 loss / self.grad_accumulation_factor
             )
--- a/speechbrain/core.py
+++ b/speechbrain/core.py
@@ -753,12 +753,19 @@ class Brain:
             )
 
         logger.info(
-            f"Gradscaler enabled: {gradscaler_enabled}. Using precision: {self.precision}."
+            f"Gradscaler enabled: {gradscaler_enabled}. Using training precision: {self.precision}."
+        )
+        logger.info(
+            f"Evaluation precision: {self.eval_precision}."
         )
         self.scaler = torch.GradScaler(self.device, enabled=gradscaler_enabled)
-        dtype = AMPConfig.from_name(self.precision).dtype
+        train_dtype = AMPConfig.from_name(self.precision).dtype
         self.training_ctx = TorchAutocast(
-            enabled=self.precision == "fp16" or self.precision == "bf16", device_type=self.device, dtype=dtype
+            enabled=self.precision == "fp16" or self.precision == "bf16", device_type=self.device, dtype=train_dtype
+        )
+        eval_dtype = AMPConfig.from_name(self.eval_precision).dtype
+        self.evaluation_ctx = TorchAutocast(
+            enabled=self.eval_precision == "fp16" or self.eval_precision == "bf16", device_type=self.device, dtype=eval_dtype
         )
         if (
             gradscaler_enabled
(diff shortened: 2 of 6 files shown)


Adel-Moumen:
If you are ok with those changes, I will broadcast the new Autocast obj to all the recipes cc: @TParcollet and @pplantinga

speechbrain/core.py
```diff
@@ -793,17 +753,24 @@ def __init__(  # noqa: C901
...
-        self.use_amp = False
-        if self.device == "cpu" and self.precision == "bf16":
-            self.use_amp = True
-        elif "cuda" in self.device and self.precision in ["fp16", "bf16"]:
-            self.use_amp = True
-
-        if self.use_amp and self.checkpointer is not None:
+        logger.info(
+            f"Evaluation precision: {self.eval_precision}."
+        )
```
Adel-Moumen, line 760: I will most likely merge both messages together.

TParcollet:
Is this compatible with our lowest torch version required?

Adel-Moumen:
> Is this compatible with our lowest torch version required?

I just tried and yes it does!

speechbrain/utils/autocast.py
```diff
@@ -3,13 +3,131 @@
...
+    --------
+    >>> with TorchAutocast(enabled=True, device="cpu", dtype=torch.bfloat16):
+    ...     # Run model inference with autocasting enabled.
+    """
+
+    def __init__(self, enabled: bool, *args, **kwargs):
+        if enabled:
+            self.context = torch.autocast(*args, **kwargs)
+        else:
+            self.context = nullcontext()  # no-op context manager
```
pplantinga, line 81: Just to make sure I understand what is happening here, the change here is to add the `enabled` flag, otherwise `torch.autocast` would work fine as-is, is that right?

speechbrain/utils/autocast.py
```diff
@@ -3,13 +3,131 @@
...
+        """
+        try:
+            return self.context.__enter__()
+        except RuntimeError as e:
+            if hasattr(self.context, "device") and hasattr(self.context, "fast_dtype"):
+                device = self.context.device
+                dtype = self.context.fast_dtype
+                raise RuntimeError(
+                    f"Error during autocasting with dtype={dtype} on device={device}.\n"
+                    "If you are on the FAIR Cluster, you might need to use autocast_dtype=float16."
```
pplantinga, line 107: Do we really want a cluster-specific message here? Perhaps we could make it a bit more general, like "Ensure your hardware supports the dtype you are attempting to use."

speechbrain/core.py
```diff
@@ -793,17 +753,24 @@ def __init__(  # noqa: C901
...
+        self.scaler = torch.GradScaler(self.device, enabled=gradscaler_enabled)
+        train_dtype = AMPConfig.from_name(self.precision).dtype
+        self.training_ctx = TorchAutocast(
+            enabled=self.precision == "fp16" or self.precision == "bf16", device_type=self.device, dtype=train_dtype
+        )
```
pplantinga, line 765: Could we just make `TorchAutocast` a no-op if dtype == "fp32"?

speechbrain/core.py
```diff
@@ -1430,7 +1382,7 @@ def _fit_train(self, train_set, epoch, enable):
         steps_since_ckpt = 0
         with tqdm(
             train_set,
-            initial=self.step,
+            initial=self.optimizer_step,
```
pplantinga, line 1385: This is a pretty significant change to behavior IMO... I'm wondering if we want a separate issue to discuss this. For example, we might want it to be configurable or alternatively to explain somehow that the TQDM bar now iterates optimizer updates rather than batches. Basically I think this point needs more discussion because it seems a major change to me.

pplantinga (commented):
Overall, I approve of moving complexity out of `core.py` where possible. My only concern with this PR is that the change from `tqdm` iterating batches to iterating updates is pretty major and perhaps there's something we can do to mitigate confusion when someone comes across the behavior change.

speechbrain/utils/autocast.py
Adel-Moumen: oops this is a reliquat from some code i have stolen / adapted

speechbrain/core.py
pplantinga: Here are yet more possible approaches to consider:

- If grad_accumulation_factor is specified, add a second progress bar with the optimizer update progress
- If grad_accumulation_factor is specified, add an entry to the progress bar with `opt_ups=4/7` or whatever

Adel-Moumen: Thanks Peter! But I think I can just remove this for now. We could about it a bit better later on, but I don't see this as something critical for the next release?

speechbrain/utils/autocast.py
Adel-Moumen: Yeah. If `enabled` is `False` then it is just a no-op context manager, meaning nothing is applied while if you have `enabled` to be `True` then it acts as a regular autocast ctx.

speechbrain/core.py
Adel-Moumen: Done.

requirements.txt
```diff
@@ -10,7 +10,7 @@ pygtrie>=2.1,<3.0
 scipy>=1.4.1,<1.13.0
 sentencepiece>=0.1.91
 SoundFile; sys_platform == 'win32'
-torch>=2.1.0
-torchaudio>=2.1.0
+torch>=2.1.0,<=2.6.0
+torchaudio>=2.1.0,<=2.6.0
```
pplantinga, line 14: I suppose the version cap is to prevent errors due to deprecated GradScaler, but I think I would prefer an approach that more smoothly transitions to the new GradScaler. Maybe we can warn that if you want to use GradScaler you have to use a more recent version of torch (I don't know when they introduced it).

Mainly, it is annoying for someone who wants to use the most recent version of torch to be forced to use an older version with SpeechBrain.

pplantinga: Also, the new version allows someone to use a device other than cuda.

pplantinga (commented):
Looks pretty good. The one last thing I think we should consider is how to handle the deprecation of `torch.cuda.amp.GradScaler` etc. I think we should try to move to the new version if possible, see my comment.

requirements.txt
Adel-Moumen: > Also, the new version allows someone to use a device other than cuda.

I think you could still use gradscaler with a non-cuda device, cpu allows fp16 (which is a recent addition) but in practice people are using bf16 which doesn't requires gradscaler. Overall, I think I totally agree on what you suggested. What do you think of directly fetching the right gradscaler object based on the torch version? 

Something similar to checkpoint.py in which we would check the version of torch, and based on that, fetch the appropriate ,ethod.

pplantinga: Sure, that makes sense to me!

pplantinga (approved):
I ran it and everything looks good. Only a minor comment about how it is structured that can easily be ignored.

PR #2828, opened 2025-02-14
Create `FetchConfig` for standardizing use of `fetch`

pplantinga:
Code for calling `fetch` uses a variety of subsets of the actual available options for `fetch`. This simplifies and standardizes the use of `fetch` by putting the configuration options in a single place, a `FetchConfig` dataclass. This should probably sit in `develop` for awhile before releasing a new version to ensure nothing is broken.

This PR also makes `fetch` multiprocessing-safe (and therefore save to use with DDP).


Adel-Moumen:
Wondering about this PR. Are you aware of any other place external to the speechbrain codebase that are using the `fetch` function? I am thinking of potentially custom Hf interfaces or google colab tutorials that might need to accommodate the new changes.

speechbrain/utils/parameter_transfer.py
```diff
@@ -181,12 +186,7 @@ def split(src):
         else:
             return split(path)
 
-    def collect_files(
-        self,
-        default_source=None,
-        use_auth_token=False,
-        local_strategy: LocalStrategy = LocalStrategy.SYMLINK,
-    ):
+    def collect_files(self, default_source=None, fetch_config=FetchConfig()):
```
Adel-Moumen, line 189: Wondering about this. `fetch_config=FetchConfig()` is evaluated once at function definition time, so every call to this function that doesn't pass a `fetch_config` will reuse the same instance. So, if one call for whatever reason is modifying one attribute of `FetchConfig`, then it will backpropagate to all `collect_files` functions. Wouldn't it be better to have this parameter set to `None`, and then in the function do something like:
```
if fetch_config is None:
   fetch_config = FetchConfig()
```

pplantinga: Good catch.

*turns toward void and inhales deeply* CURSE YOU PYTHON AND YOUR FRIGGEN UNINTUITIVE DEFAULT PARAMTERS

speechbrain/inference/interfaces.py
```diff
@@ -39,28 +39,89 @@
 
 def foreign_class(
     source,
-    hparams_file="hyperparams.yaml",
+    hparams_file,
```
Adel-Moumen: what's the motivation behind remove the default `hyperparams.yaml` value? Hopefully, no one is using the default naming.

speechbrain/inference/interfaces.py
```diff
@@ -39,28 +39,89 @@
...
     classname="CustomInterface",
-    overrides={},
-    overrides_must_match=True,
     savedir=None,
-    use_auth_token=False,
-    download_only=False,
-    huggingface_cache_dir=None,
-    local_strategy: LocalStrategy = LocalStrategy.SYMLINK,
+    local_strategy=LocalStrategy.SYMLINK,
+    fetch_config=FetchConfig(),
```
Adel-Moumen: we are again using `FetchConfig()` as a default value. Can't we just set it to `None` and if it is `None`, then init `FetchConfig()` ?

speechbrain/inference/interfaces.py
```diff
@@ -39,28 +39,89 @@
...
 
+    Arguments
+    ---------
+    source : str or Path or FetchSource
+        The location to use for finding the model. See
+        ``speechbrain.utils.fetching.fetch`` for details.
+    hparams_file : str
+        The name of the hyperparameters file to use for constructing
+        the modules necessary for inference. Must contain two keys:
+        "modules" and "pretrainer", as described.
```
Adel-Moumen: As described where? Not sure to see where you referring to (not the docstring of this function?)

speechbrain/inference/interfaces.py
```diff
@@ -39,28 +39,89 @@
...
+        savedir=savedir,
+        save_filename=None,
+        local_strategy=local_strategy,
+        fetch_config=fetch_config,
+    )
+    sys.path.append(str(pymodule_local_path.parent))
+
+    # Import class and create instance
+    module = import_from_path(pymodule_local_path)
+    cls = getattr(module, classname)
```
Adel-Moumen: this seems a bit too magical to me haha. Would it be possible to have some comments around this pls?

speechbrain/inference/interfaces.py
```diff
@@ -39,28 +39,89 @@
...
+def pretrained_from_hparams(
+    cls,
+    source,
+    hparams_file="hyperparams.yaml",
+    overrides={},
+    overrides_must_match=True,
+    savedir=None,
+    download_only=False,
+    local_strategy=LocalStrategy.SYMLINK,
+    fetch_config=FetchConfig(),
```
Adel-Moumen: same as before?

speechbrain/inference/interfaces.py
```diff
@@ -69,66 +130,43 @@ def foreign_class(
...
     ---------
+    cls : Type[Pretrained]
+        The class to construct an instance of, usually a sub-type of Pretrained
     source : str or Path or FetchSource
         The location to use for finding the model. See
         ``speechbrain.utils.fetching.fetch`` for details.
     hparams_file : str
         The name of the hyperparameters file to use for constructing
         the modules necessary for inference. Must contain two keys:
         "modules" and "pretrainer", as described.
```
Adel-Moumen, line 155: I guess same here?

speechbrain/inference/interfaces.py
```diff
@@ -69,66 +130,43 @@ def foreign_class(
...
+        Configuration options for caching and other fetch behavior.
     **kwargs : dict
         Arguments to forward to class constructor.
 
     Returns
     -------
-    object
-        An instance of a class with the given classname from the given pymodule file.
+    object : Pretrained
+        An instance of a Pretrained class, constructed from the hparams.
```
Adel-Moumen: And can be None as well

speechbrain/inference/interfaces.py
pplantinga: I wonder if I should use a NamedTuple here? This is immutable so we don't have to worry about default values, but it might not play as nicely with type hints if we're going to add those.

pplantinga: Or, perhaps even better is to use `frozen=True`

https://docs.python.org/3/library/dataclasses.html#frozen-instances

speechbrain/inference/interfaces.py
pplantinga: Is the new version better?

speechbrain/inference/interfaces.py
pplantinga: What did you mean by this exactly? That this argument should be explained better?

speechbrain/inference/interfaces.py
Adel-Moumen: Yes!

speechbrain/inference/interfaces.py
Adel-Moumen: Yes, with `frozen=True` it makes it immutable. I guess that if we don't plan to have mutable arguments in `FetchConfig`, we can just set it as a default value for `fetch_config`. Otherwise, if we really want to be safe (but I don't think this is really necessary), we can just do `fetch_config: Optional[FetchConfig] = None` and: 
```
if fetch_config is None:
        fetch_config = FetchConfig()
```
Both are fine, but `fetch_config=FetchConfig()` is a bit more idiomatic so we can leave this like that:)

Adel-Moumen (approved):
LGTM!

Quick question: is there a way to tests these functions and make sure they do what they are intended to do?

pplantinga:
The functions are called as part of the tests, several unit tests and doctests download things via fetch. I suppose we could test it more systematically with some unit tests, especially loading from local

PR #2835, opened 2025-02-25
Fix input normalization and global normalization variance calculation

### LINKED ISSUE #2815: InputNormalization with "global" is incorrect
### Describe the bug

This issue reports an incorrect behavior w.r.t InputNormalization when used with the "global" argument:
1. The [aggregated std deviation](https://github.com/speechbrain/speechbrain/blob/79c4fd16978bec74329c04bf682b10bbe5672d59/speechbrain/processing/features.py#L1155) is mathematically wrong. The average of multiple std is not equivalent to the std of all the values. 
2. The function does not do what we think it does. If I understand correctly the intent, this function should compute statistics across the dataset (with running mean and std) and normalise each sample in the batch with this running statistics. Hopefully, the running statistics should converge to the true values of the mean and std over the dataset. Unfortunately, this is not what the code does for the std (the mean is correct). In our code, the running std is computed on a per-batch basis, which is very different from getting the statistics of a whole set. For instance two vectors (let's say 2 utterances of 3 frames with one value) [4, 4, 4] and [-4, -4, -4] gives you a "dataset" variance of 16, while with our implementation (mean over the sentences in a batch), you'll get 0. Fixing this requires an entire revamp of the InputNormalisation, and since it seems to be working like that... i'd be more inclined in documenting this behavior properly AND fixing 1. 

@pplantinga / @mravanelli / @Adel-Moumen what do you think?

### Expected behaviour

Discuss about the issue. 

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

pplantinga:
Fixes #2815 

`InputNormalization` (and `GlobalNorm`) has several bugs, including the one listed above.

* Mainly, it is mathematically incorrect to average the standard deviation across inputs, which more closely mimics the behavior of `BatchNorm`. Instead, the variance should be computed against the running mean, not the tensor mean.
* mean and variance normalization was applied to the padding (it should be ignored)
* if `mean_norm` is false and `std_norm` is true, the mean is not subtracted and added back, leading to incorrect variance normalization.
* `update_until_epoch` was set to 3 by default, which in most cases is twice the number of updates needed.
* Updates are applied directly to `.data` which is not recommended because it bypasses the gradient system. Probably doesn't matter for normal cases where this is used at the input, but it could matter if its used anywhere else in the network.

Maybe some of these settings were not getting used (a cursory `git grep` shows mostly default settings), but we should fix the bugs anyway. This PR adds extensive unittests (thanks Claude) and hopefully catches most of the edge cases.


pplantinga:
The failing test here is one of the ASR CTC integration tests, looking across all three examples in the folder, the train loss is always higher than before. I'm not quite sure what to make of this, but it seems to be a real effect of correcting the standard deviation calculation. I went back to the develop version and changed one line:

```diff
- current_std = torch.std(x, dim=0).detach().data
+ current_std = (x - self.glob_mean.detach()).square().mean().sqrt().data
```
And got a very similar increase in the training loss.

I suppose a next step could be to try training a full recipe of some kind that includes input norm and see if the result is comparable to what we had before.

pplantinga:
I ran the ASR template example using current develop and this branch, which seems to me to alleviate any concern that the train loss is higher.

Current `develop`:
```
epoch: 1, lr: 1.00e+00 - train loss: 1.54 - valid loss: 1.33, valid CER: 6.42, valid WER: 13.37
epoch: 2, lr: 1.00e+00 - train loss: 1.25 - valid loss: 1.36, valid CER: 12.60, valid WER: 22.30
epoch: 3, lr: 8.00e-01 - train loss: 1.07 - valid loss: 1.34, valid CER: 9.63, valid WER: 15.76
epoch: 4, lr: 8.00e-01 - train loss: 9.96e-01 - valid loss: 1.35, valid CER: 8.09, valid WER: 14.64
epoch: 5, lr: 8.00e-01 - train loss: 9.72e-01 - valid loss: 1.34, valid CER: 6.59, valid WER: 12.12
epoch: 6, lr: 8.00e-01 - train loss: 1.39 - valid loss: 1.36, valid CER: 7.63, valid WER: 14.27
epoch: 7, lr: 6.40e-01 - train loss: 1.32 - valid loss: 1.33, valid CER: 6.06, valid WER: 12.22
epoch: 8, lr: 6.40e-01 - train loss: 1.33 - valid loss: 1.33, valid CER: 4.68, valid WER: 10.35
epoch: 9, lr: 6.40e-01 - train loss: 1.28 - valid loss: 1.33, valid CER: 7.39, valid WER: 13.90
epoch: 10, lr: 5.12e-01 - train loss: 1.27 - valid loss: 1.32, valid CER: 10.23, valid WER: 19.99
epoch: 11, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.31, valid CER: 8.17, valid WER: 14.97
epoch: 12, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.31, valid CER: 7.17, valid WER: 13.73
epoch: 13, lr: 4.10e-01 - train loss: 1.24 - valid loss: 1.30, valid CER: 4.96, valid WER: 11.54
epoch: 14, lr: 4.10e-01 - train loss: 1.22 - valid loss: 1.30, valid CER: 5.46, valid WER: 10.87
epoch: 15, lr: 4.10e-01 - train loss: 1.22 - valid loss: 1.31, valid CER: 6.17, valid WER: 12.36
Epoch loaded: 8 - test loss: 1.32, test CER: 4.25, test WER: 7.36
```

This PR:
```
epoch: 1, lr: 1.00e+00 - train loss: 1.57 - valid loss: 1.34, valid CER: 7.63, valid WER: 14.22
epoch: 2, lr: 1.00e+00 - train loss: 1.25 - valid loss: 1.36, valid CER: 6.43, valid WER: 12.91
epoch: 3, lr: 1.00e+00 - train loss: 1.12 - valid loss: 1.38, valid CER: 7.67, valid WER: 13.50
epoch: 4, lr: 8.00e-01 - train loss: 9.99e-01 - valid loss: 1.35, valid CER: 5.82, valid WER: 12.32
epoch: 5, lr: 8.00e-01 - train loss: 9.74e-01 - valid loss: 1.35, valid CER: 9.89, valid WER: 18.54
epoch: 6, lr: 6.40e-01 - train loss: 1.37 - valid loss: 1.33, valid CER: 9.37, valid WER: 16.54
epoch: 7, lr: 6.40e-01 - train loss: 1.34 - valid loss: 1.33, valid CER: 9.50, valid WER: 16.81
epoch: 8, lr: 5.12e-01 - train loss: 1.28 - valid loss: 1.32, valid CER: 6.62, valid WER: 12.88
epoch: 9, lr: 5.12e-01 - train loss: 1.26 - valid loss: 1.31, valid CER: 8.46, valid WER: 16.59
epoch: 10, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.30, valid CER: 6.54, valid WER: 13.33
epoch: 11, lr: 4.10e-01 - train loss: 1.26 - valid loss: 1.30, valid CER: 5.38, valid WER: 10.58
epoch: 12, lr: 4.10e-01 - train loss: 1.24 - valid loss: 1.31, valid CER: 5.15, valid WER: 11.32
epoch: 13, lr: 3.28e-01 - train loss: 1.22 - valid loss: 1.30, valid CER: 7.03, valid WER: 14.19
epoch: 14, lr: 2.62e-01 - train loss: 1.21 - valid loss: 1.29, valid CER: 5.59, valid WER: 12.02
epoch: 15, lr: 2.62e-01 - train loss: 1.21 - valid loss: 1.30, valid CER: 5.33, valid WER: 11.12
Epoch loaded: 11 - test loss: 1.30, test CER: 3.64, test WER: 6.89
```

pplantinga:
Okay, I found the issue: the mean/stdev should be computed *only* over the time dimension, not all dimensions. I will fix this and update.

TParcollet:
@pplantinga @mravanelli the speaker normalisation is making this class so complicated for no reasons. By no reasons I mean -- we litteraly don't have a recipe using it. What about just removing it?

speechbrain/processing/features.py
```diff
@@ -1321,6 +1369,93 @@ def _load(self, path, end_of_epoch=False):
...
+    # Update mean for use in variance
+    delta = new_tensor - mean
+    mean += delta.mean() * new_weight / (weight + new_weight)
+
+    # Update variance, including sync
+    if var is not None:
+        delta2 = new_tensor - mean
+        new_var = (delta * delta2).sum() / (delta.numel() - 1)
+        var = (var * weight + new_var * new_weight) / (weight + new_weight)
+        var = ddp_all_reduce(var, torch.distributed.ReduceOp.AVG)
```
rogiervd, line 1413: Assuming that `var` means "variance" and not "(non-central) second moment", this does not at first glance look correct.

rogiervd (commented):
I'm new to this code, so maybe I've got possibly naive questions (@TParcollet suggested I look at this so feel free to blame him). So can I check that my assumptions are correct?

* Under `norm_type=global`, the existing implementation was not correct, so this changes the behaviour only of that, and not of the other three `norm_type`s.
* Under `norm_type=global, avg_factor=None`, the expected behaviour is that the statistics represent those of all the data seen so far.
* Normalisation is meant to be per-dimension, i.e. the global mean and variance have the shape of a single feature vector. (This assumption I'm getting not from the code but because e.g. HTK used to do this.)

What is the desired behaviour under `norm_type=global, avg_factor=0.5` (or another value)? I don't think I understand what `avg_factor` does.

Shouldn't the change to the algorithm, by using Welford's algorithm, be made to the `norm_type=speaker` too, if `norm_type=speaker` is retained?

pplantinga:
> @pplantinga @mravanelli the speaker normalisation is making this class so complicated for no reasons. By no reasons I mean -- we litteraly don't have a recipe using it. What about just removing it?

I'm okay with removing it, but we would still have to handle the old checkpoints which all have this.

> Under norm_type=global, the existing implementation was not correct, so this changes the behaviour only of that, and not of the other three norm_types.

Well, speaker norm was not correct either, but looking at my implementation here I can see that I haven't fixed it. I'll go ahead and fix it if we end up keeping it.

> What is the desired behaviour under norm_type=global, avg_factor=0.5 (or another value)? I don't think I understand what avg_factor does.

The `avg_factor` updates the stats according to `(1 - avg_factor) * new_stat + avg_factor * old_stat`, e.g. if `avg_factor` is 0.1, we would include 10% of the new stat and 90% of the old stat. This might be desirable if you expect some distribution shift over training and want to weight newer samples stronger.

> Shouldn't the change to the algorithm, by using Welford's algorithm, be made to the norm_type=speaker too, if norm_type=speaker is retained?

Perhaps. This change should be straightforward if we decide to keep speaker normalization.

speechbrain/processing/features.py
pplantinga: Or did you mean that the reduction is not done correctly? This could easily be the case, let me look at the reduction part once more.

pplantinga:
One more annoying issue: the original code averages over time first, then batch, but here I've averaged over time and batch at the same time. I wouldn't expect this to make a huge difference on new runs but it can affect the performance of loading old checkpoitns. I guess I'll edit this to average time first and then batch, even though this is probably slower and adds complexity (especially for standard deviation).

TParcollet:
We honestly don't want to make the code slower for the sake of retro compatibility imho ... so if it's actually slower ... well ... maybe we should also check with the profiler how impactful this new normalisation is

pplantinga:
Okay, should have rechecked this part as the normalization dimension didn't matter for global norm, although it still affects `norm_type=batch` -- the real cause of the performance diff when loading old checkpoints was due to the *padding* not getting normalized (I explicitly excluded it). I can certainly update the PR to normalize the padding again, but do we want to have an option to exclude the padding? My assumption is that most would not want the padding normalized, but it seems our checkpoints might depend on this point.

For the `norm_type=batch`, is it more important to keep backwards compatibility or to update to the correct algorithm? I don't see any instances of `norm_type=batch` in the repo. I guess this norm type is equivalent to `torch.nn.BatchNorm1d(keep_running_stats=False)`. We also could deprecate this one in addition to the speaker norm.

speechbrain/processing/features.py
rogiervd: What I mean is that if you want to get the variance of all data (which I still assume is what you want for `norm_type=global, avg_factor=None`), then this is not correct.

Imagine one node has data `[4,6]` and another node `[-4,-6]`. Each node has a local variance of 1, so the average you compute here is also 1, but the overall variance is 26.

pplantinga:
> It would still be really helpful for me to understand whether these are true:
> 
> > * Under `norm_type=global, avg_factor=None`, the expected behaviour is that the statistics represent those of all the data seen so far.

Yes, this is the expected behavior.

> > * Normalisation is meant to be per-dimension, i.e. the global mean and variance have the shape of a single feature vector. (This assumption I'm getting not from the code but because e.g. HTK used to do this.)

I think the answer is yes here, assuming you meant that the reduction is done over *time and batch* dimensions only. The first version of the PR did not have this, but I have updated it to match.

> Welford's algorithm is an incremental method for computing the overall mean and variance of all the data you've put into it. Is there a similar statement you can make when you use `avg_factor`?
> 
> For clarity, my feeling is that the answer may be something like: "It seems like with what we're trying to achieve in this new version, there is no meaning any more `avg_factor` and we should take it out." This is all based on the idea that the standard behaviour is to compute the mean and variance of the whole data. I might be wrong though.

Aha, I understand your question now. I think the `avg_factor` is used to achieve behavior similar to BatchNorm on the input, which the PyTorch documentation calls "momentum". But if that use case is covered by BatchNorm, perhaps it is better to just deprecate it here.

pplantinga:
Okay, this should be ready to review once more @TParcollet @rogiervd 

I have verified the performance is the same when loading old checkpoints. The only remaining step I will complete before merge is to train a new model to see if performance is roughly the same.

TParcollet:
@pplantinga I think that what @rogiervd meant is that Welford's algorithm renders the momentum meaningless?

pplantinga:
> @pplantinga I think that what @rogiervd meant is that Welford's algorithm renders the momentum meaningless?

I don't think Welford's algorithm or the variance computation fix changes the meaning -- it is now simply a momentum-based update instead of an equal-weight update. However, this seems to follow the same update algorithm as BatchNorm -- the mean and variance are momentum-weighted so at test time it should perform quite similarly -- so I have deprecated it here.

TParcollet:
@pplantinga thanks! Just to make sure: when using "norm_type=global", and "avg_factor=None" we now try to compute the exact per-feature dimension mean and variance over the whole dataset iteratively such that the outcome is a vector of dimension equal to F in a [B,T,F] input Tensor (i.e. computed over Batch and Time). Is that correct?

speechbrain/processing/features.py
```diff
@@ -995,250 +996,240 @@ def forward(self, x):
 
 @register_checkpoint_hooks
 class InputNormalization(torch.nn.Module):
-    """Performs mean and variance normalization of the input tensor.
+    """Performs mean and variance normalization over the time dimension of the input.
```
TParcollet, line 999: Batch and/or Time?

speechbrain/processing/features.py
```diff
@@ -995,250 +996,240 @@ def forward(self, x):
...
+
+    Running mean and running variance calculation is done using Welford's Algorithm.
+
+    WARNING: at first, the running statistics do not represent the "true" mean
+    and variance, but are estimates based on the data seen so far. Once enough
+    data has been seen, the stats should closely approximate the "true" values.
+
+    WARNING: Using global normalization, the first call of `forward()` will
+    throw an error if no updates have been performed (including the current batch),
+    i.e. on first call the epoch >= update_until_epoch or in eval() mode.
```
TParcollet, line 1009: the end of this sentence is not clear to me.

speechbrain/processing/features.py
```diff
@@ -995,250 +996,240 @@ def forward(self, x):
...
+         * 'batch' computes norms per input tensor (no running stats)
+         * 'global' computes norms over all inputs (single mean, variance)
+         * 'speaker' - DEPRECATED
+    avg_factor : float, optional
+        Passing avg_factor is DEPRECATED as this exactly matches the
+        behavior of BatchNorm. To maintain this behavior, use
+        `speechbrain.nnet.normalization.BatchNorm1d(momentum=avg_factor)`.
+    length_dim : int, default 1
+        The dimension for which to mask out the padding positions.
+    update_until_epoch : int, default 2
```
TParcollet, line 1029: default is 2 not 1 as stated in the docstring.

speechbrain/processing/features.py
```diff
@@ -995,250 +996,240 @@ def forward(self, x):
...
+        The epoch for which updates to the norm stats should stop.
+        By default, stops after one epoch of updates.
+    avoid_padding_norm : bool, default False
```
TParcollet, line 1032: Do we want to keep this at False by default?

speechbrain/processing/features.py
```diff
@@ -1316,11 +1286,92 @@ def _load(self, path, end_of_epoch=False):
...
+    # Add dimensions other than (batch, length) back into the mask
+    for dim in range(1, x.ndim):
+        if dim != length_dim:
+            mask = mask.unsqueeze(dim)
+
+    # Repeat values (without copying) into singleton dimensions
+    return mask.expand_as(x)
+
+
 class GlobalNorm(torch.nn.Module):
```
TParcollet, line 1375: Was this function wrong as well? I see it being used in AudioMnist.

TParcollet (commented):
Not an in-depth functional review yet. I think Rogier will do this better :p See my comments / questions however.

speechbrain/processing/features.py
```diff
@@ -1316,11 +1286,92 @@ def _load(self, path, end_of_epoch=False):
...
         """
         del end_of_epoch  # Unused here.
-        device = "cpu"
-        stats = torch.load(path, map_location=device)
+        stats = torch.load(path, map_location=self.device)
         self._load_statistics_dict(stats)
 
 
+def mean_std_update(x, mask, dim, run_count, run_mean, run_std=None):
+    """Welford's algorithm for running mean, variance (from Wikipedia)
```
rogiervd, line 1294: I would remove mention of Welford's algorithm (the caller shouldn't care about the implementation) and say something like "Update the mean and variance statistics `run_mean` and `run_std` that have been computed on `run_count` samples to integrate the new samples `x`.

speechbrain/processing/features.py
```diff
@@ -1316,11 +1286,92 @@ def _load(self, path, end_of_epoch=False):
...
+    run_count : float or torch.Tensor
+        The running number of samples seen so far.
+    run_mean : float or torch.Tensor
+        The running mean of samples seen so far.
+    run_std : float or torch.Tensor, optional
+        The running standard deviations from the mean. If `None`, the
+        variance is not computed and this returns 1.0 for the `run_std`.
+
+    Returns
+    -------
```
rogiervd, line 1315: Maybe say "Statistics that represent the union [concatenation?] of the previously-seen data and the new data"?

speechbrain/processing/features.py
```diff
@@ -1316,11 +1286,92 @@ def _load(self, path, end_of_epoch=False):
...
+    run_count : torch.Tensor
```
rogiervd, line 1316: I'd call these `new_*` instead of `run_*`, since they have different values from `run_*`.

tests/unittests/test_input_norm.py
```diff
@@ -0,0 +1,318 @@
...
+    tensor = torch.tensor([[1.0, 2.0, 3.0]])
+    mask = torch.tensor([[True, True, True]])
+    dims = (1,)
+    run_count = 2
+    run_mean = torch.tensor([2.0])
+    run_std = torch.tensor([1.0])
+
+    # Mock the distributed functions
+    def mock_ddp_all_reduce(tensor, op):
+        return tensor  # Just return the tensor, simulating no distribution
```
rogiervd, line 278: How convinced are you that the code actually works correctly? It seems worth testing ddp_all_reduce for real (with more extensive mocking, I guess).

tests/unittests/test_input_norm.py
```diff
@@ -0,0 +1,318 @@
...
+        return real_import(name, *args, **kwargs)
+
+    builtins.__import__ = mock_import
+
+    # Call the function
+    try:
+        from speechbrain.processing.features import mean_std_update
+
+        updated_count, updated_mean, updated_std = mean_std_update(
+            tensor,
```
rogiervd, line 299: Are you putting the same tensor in every time? Does that even test that the change to Welford's algorithm has changed anything?

rogiervd (commented):
A few comments for now on Welford's algorithm.

speechbrain/processing/features.py
pplantinga: The code will stop when epoch <= update_until_epoch, so "2" means one epoch of updates.

speechbrain/processing/features.py
pplantinga: I agree, `True` makes more sense here, but this breaks backwards compatibility

speechbrain/processing/features.py
pplantinga: Yes, it used the same wrong std update algorithm.

speechbrain/processing/features.py
pplantinga: Sure, this is fair, as I'm not sure whether the changes I've made to accomodate DDP and tensors may still count as the algorithm. However, perhaps there should be a mention somewhere in case someone is confused about the two `x - mean` computations done for numerical stability.

speechbrain/processing/features.py
pplantinga: Sure, good point. I'll update the name to indicate they've changed.

tests/unittests/test_input_norm.py
pplantinga: This is a good idea. I'll try to add more extensive tests here

speechbrain/processing/features.py
pplantinga: If we want to eventually make this `True` by default we could add a deprecation notice here, and then in v1.1.0 or whatever we could update the critical HF checkpoints with the parameter set to `False` and add a readme note to use SB v1.0.3 or greater.

pplantinga:
> @pplantinga thanks! Just to make sure: when using "norm_type=global", and "avg_factor=None" we now try to compute the exact per-feature dimension mean and variance over the whole dataset iteratively such that the outcomes are two vectors of dimension equal to F in a [B,T,F] input Tensor (i.e. computed over Batch and Time). Is that correct?

Yes, this computes the global per-feature mean and standard deviation over the whole dataset. Once you've seen the full set, the glob_mean and glob_std will each contain (out of [B, T, F, ...] ) dimensions equal to the [F, ...] part, representing the mean and standard deviation of each feature. Also, I have deprecated `avg_factor` so this won't have an effect anymore.

tests/unittests/test_input_norm.py
rogiervd: Having thought of this a bit more, I think it is not obvious how to mock ddp_all_reduce. But I've got an idea for a tool to do this which I could put together in a separate PR if helpful.

tests/unittests/test_input_norm.py
TParcollet: @pplantinga I agree with the question here.

pplantinga:
I've verified that it doesn't hurt results (if anything, it seems to help a little, train&valid loss are consistently lower)

Current `develop`:

```
epoch: 1, lr: 1.00e+00 - train loss: 1.54 - valid loss: 1.33, valid CER: 6.42, valid WER: 13.37
epoch: 2, lr: 1.00e+00 - train loss: 1.25 - valid loss: 1.36, valid CER: 12.60, valid WER: 22.30
epoch: 3, lr: 8.00e-01 - train loss: 1.07 - valid loss: 1.34, valid CER: 9.63, valid WER: 15.76
epoch: 4, lr: 8.00e-01 - train loss: 9.96e-01 - valid loss: 1.35, valid CER: 8.09, valid WER: 14.64
epoch: 5, lr: 8.00e-01 - train loss: 9.72e-01 - valid loss: 1.34, valid CER: 6.59, valid WER: 12.12
epoch: 6, lr: 8.00e-01 - train loss: 1.39 - valid loss: 1.36, valid CER: 7.63, valid WER: 14.27
epoch: 7, lr: 6.40e-01 - train loss: 1.32 - valid loss: 1.33, valid CER: 6.06, valid WER: 12.22
epoch: 8, lr: 6.40e-01 - train loss: 1.33 - valid loss: 1.33, valid CER: 4.68, valid WER: 10.35
epoch: 9, lr: 6.40e-01 - train loss: 1.28 - valid loss: 1.33, valid CER: 7.39, valid WER: 13.90
epoch: 10, lr: 5.12e-01 - train loss: 1.27 - valid loss: 1.32, valid CER: 10.23, valid WER: 19.99
epoch: 11, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.31, valid CER: 8.17, valid WER: 14.97
epoch: 12, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.31, valid CER: 7.17, valid WER: 13.73
epoch: 13, lr: 4.10e-01 - train loss: 1.24 - valid loss: 1.30, valid CER: 4.96, valid WER: 11.54
epoch: 14, lr: 4.10e-01 - train loss: 1.22 - valid loss: 1.30, valid CER: 5.46, valid WER: 10.87
epoch: 15, lr: 4.10e-01 - train loss: 1.22 - valid loss: 1.31, valid CER: 6.17, valid WER: 12.36
Epoch loaded: 8 - test loss: 1.32, test CER: 5.65, test WER: 11.36
```

This PR:

```
epoch: 1, lr: 1.00e+00 - train loss: 1.55 - valid loss: 1.35, valid CER: 9.90, valid WER: 15.85
epoch: 2, lr: 1.00e+00 - train loss: 1.25 - valid loss: 1.37, valid CER: 6.99, valid WER: 12.49
epoch: 3, lr: 1.00e+00 - train loss: 1.13 - valid loss: 1.38, valid CER: 6.98, valid WER: 13.35
epoch: 4, lr: 8.00e-01 - train loss: 1.00 - valid loss: 1.35, valid CER: 7.33, valid WER: 14.05
epoch: 5, lr: 6.40e-01 - train loss: 9.48e-01 - valid loss: 1.34, valid CER: 8.62, valid WER: 16.57
epoch: 6, lr: 5.12e-01 - train loss: 1.34 - valid loss: 1.32, valid CER: 7.43, valid WER: 13.77
epoch: 7, lr: 5.12e-01 - train loss: 1.31 - valid loss: 1.32, valid CER: 7.42, valid WER: 13.28
epoch: 8, lr: 5.12e-01 - train loss: 1.28 - valid loss: 1.32, valid CER: 7.84, valid WER: 15.62
epoch: 9, lr: 4.10e-01 - train loss: 1.25 - valid loss: 1.30, valid CER: 4.93, valid WER: 10.87
epoch: 10, lr: 4.10e-01 - train loss: 1.24 - valid loss: 1.30, valid CER: 8.36, valid WER: 15.83
epoch: 11, lr: 3.28e-01 - train loss: 1.26 - valid loss: 1.29, valid CER: 6.30, valid WER: 12.24
epoch: 12, lr: 3.28e-01 - train loss: 1.23 - valid loss: 1.30, valid CER: 7.92, valid WER: 14.43
epoch: 13, lr: 2.62e-01 - train loss: 1.21 - valid loss: 1.29, valid CER: 7.31, valid WER: 12.80
epoch: 14, lr: 2.62e-01 - train loss: 1.20 - valid loss: 1.29, valid CER: 9.50, valid WER: 17.03
epoch: 15, lr: 2.10e-01 - train loss: 1.21 - valid loss: 1.29, valid CER: 4.08, valid WER: 9.15
Epoch loaded: 15 - test loss: 1.28, test CER: 5.12, test WER: 11.31
```

rogiervd:
Great! Actually using `torch.multiprocessing` is clearly the right way of doing this.

I must admit that I am not sure that I fully understand what the implementation of `mean_std_update` is doing in terms of the variance computation. And I also am not sure that the test isn't too simple.

So I wrote a test that uses random vectors and compares against the statistics on the flattened vectors, which fails on your implementation. I have a ~60% confidence it's a legitimate failure. What shall I do? Shall I put my implementation and test for Welford's algorithm in a separate PR and then we can take it from there?

pplantinga:
I think the extra test should just be added to this PR if possible. Or do you have to be a maintainer at SperchBrain to be able to edit the PR? Honestly wherever is easiest for you to put the code, if it is possible to make a PR against the branch you can do that too.

pplantinga:
I added one more test and I'm like 93% sure I fixed the last remaining bugs.

Can you run your test once more to verify?

rogiervd:
I have to admit I still can't wrap my head around `x - run_mean` with the old mean. I also don't know that it's worth getting unbiased variance estimates (dividing by `n-1` instead of `n`), and how this is undone when combining statistics. In #2857 I offer an alternative implementation for this which in my mind is simpler. At least, I can understand it.

pplantinga:
> I have to admit I still can't wrap my head around `x - run_mean` with the old mean. 

Perhaps we should ditch Welford's algorithm for the sake of readability then.

> I also don't know that it's worth getting unbiased variance estimates (dividing by `n-1` instead of `n`), and how this is undone when combining statistics. 

I'm fine either way honestly. The main reason to keep it imo is so that the examples/tests have the variance that people expect.

> In #2857 I offer an alternative implementation for this which in my mind is simpler. At least, I can understand it.

I'll do my best to integrate your PR into this one so this one is more understandable, while maintaining your commit / authorship credit.

rogiervd:
> > I have to admit I still can't wrap my head around `x - run_mean` with the old mean.
> 
> Perhaps we should ditch Welford's algorithm for the sake of readability then.

Ah, maybe now I understand what's going on. Welford's algorithm per Wikipedia works for adding a single data point. That's not what we want to do, right? Presumably what we need to implement is https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm.

> > I also don't know that it's worth getting unbiased variance estimates (dividing by `n-1` instead of `n`), and how this is undone when combining statistics.
> 
> I'm fine either way honestly. The main reason to keep it imo is so that the examples/tests have the variance that people expect.

Do people expect an unbiased estimate of the variance? I could be mistaken but I don't think the square root of the unbiased estimate of the variance is an unbiased estimate of the standard deviation.

My main concern about this is that I'm not sure whether, in terms of the Wikipedia page, the code correctly deals with the distinction between s^2_n and sigma^2_n.

> > In #2857 I offer an alternative implementation for this which in my mind is simpler. At least, I can understand it.
> 
> I'll do my best to integrate your PR into this one so this one is more understandable, while maintaining your commit / authorship credit.

In that case, wouldn't you want to review and merge my PR? Even it is pretty long.

pplantinga:
Okay, I have merged @rogiervd 's branch into this one, to hopefully clarify the computation of the gaussian statistics. This no longer includes unbiased variance calculation nor Welford's algorithm. I added the mask and std_norm=False back in. So this should be ready to review @TParcollet.

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
         return cw_x
 
 
+def gaussian_statistics(
```
TParcollet, line 999: Do we want to doctest this function maybe?

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
...
+    mean
+        The mean.
+    variance
+        The variance.
+    """
+
+    if isinstance(dim, int):
+        dim = (dim,)
+
+    # Use output size to compute N from the mask. Assumes N will
```
TParcollet, line 1037: Assuming N is the count? Maybe we should clarify this.

TParcollet: I think I don't understand what this 5 lines do :/

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
...
+):
+    """
+    Compute first- and second-order moments of data, and return them as the
+    count, mean, and variance of a vector over one or more dimensions.
+
+    Arguments
+    ---------
+    x: torch.Tensor
+        The tensor to compute the statistics over
+    mask: torch.Tensor
```
TParcollet, line 1013: Do we want to specify what 0 means and what 1 means?

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
...
+
+    return (global_count, global_mean, global_variance)
+
+
+def mean_std_update(x, mask, dim, run_count, run_mean, run_std=None):
+    """Update the mean and variance statistics run_mean and run_std that
+    have been computed on run_count samples to integrate the new samples x.
+
+    For more information about running variance statistic calculation, see
+    https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
```
TParcollet, line 1177: Are we or not still using this algorithm?

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
...
+        The dimension or dimensions to reduce (e.g. 1 for length).
+    run_count : float or torch.Tensor
+        The running number of samples seen so far.
+    run_mean : float or torch.Tensor
+        The running mean of samples seen so far.
+    run_std : float or torch.Tensor, optional
+        The running standard deviations from the mean. If None, returns
+        1.0 for the standard deviation without computation.
+
+    Returns
```
TParcollet, line 1197: We should doct test this function.

speechbrain/processing/features.py
```diff
@@ -1316,11 +1515,45 @@ def _load(self, path, end_of_epoch=False):
             Here for compatibility, but not used.
         """
         del end_of_epoch  # Unused here.
-        device = "cpu"
-        stats = torch.load(path, map_location=device)
+        stats = torch.load(path, map_location=self.device)
         self._load_statistics_dict(stats)
 
 
+def get_mask(x, lengths=None, length_dim=1):
```
TParcollet, line 1522: We have 1 billion function doing this in SB, I wonder if we should do smth about it.

speechbrain/processing/features.py
```diff
@@ -993,252 +996,469 @@ def forward(self, x):
...
+    new_run_count : torch.Tensor
+        Updated count all samples, now including x.
+    new_run_mean : torch.Tensor
+        Updated running mean of all samples, now including x.
+    new_run_std : torch.Tensor
+        Updated running standard deviations of all samples, now including x.
+    """
+
+    compute_variance = run_std is not None
+    new_statistics = combine_gaussian_statistics_distributed(
```
TParcollet, line 1208: why distributed here without checking if DDP is up? How is DDP handled?

TParcollet (commented):
Only minor comment! I trust @rogiervd and @pplantinga for the correctness! This is an important change, so we may want a last review from @rogiervd to make sure we are all happy with this.

TParcollet:
@pplantinga any chance that you could run a simple time.time comparison to check the impact on this new code compared to the previous one on a big random batch?

pplantinga:
Hm, yes this is probably worth checking. I was a bit concerned about this given the complexity/length of the code and checks.

speechbrain/processing/features.py
```diff
@@ -1110,12 +1121,17 @@ def combine_gaussian_statistics(
         The combined mean.
     variance
         The combined variance, relative to the new mean.
+        Returns None if either statistics set has variance of None
```
rogiervd, line 1124: Instead of "Returns", "Is"?

speechbrain/processing/features.py
```diff
@@ -1162,292 +1182,335 @@ def combine_gaussian_statistics_distributed(
...
-    mask: Optional[torch.Tensor],
-    dim: Union[int, tuple, None],
-    run_count: int,
-    run_mean: torch.Tensor,
-    run_std: torch.Tensor,
-):
-    """
-    Update the running count, running mean, and running standard deviation
-    by integrating new data x from multiple processes.
+def mean_std_update(x, mask, dim, run_count, run_mean, run_std=None):
```
rogiervd, line 1207: Why remove types?

speechbrain/processing/features.py
```diff
@@ -1162,292 +1182,335 @@ def combine_gaussian_statistics_distributed(
...
+        Ones / trues are valid positions, and zeros / falses are padding positions.
+    dim : tuple or int
+        The dimension or dimensions to reduce (e.g. 1 for length).
+    run_count : float or torch.Tensor
+        The running number of samples seen so far.
+    run_mean : float or torch.Tensor
+        The running mean of samples seen so far.
+    run_std : float or torch.Tensor, optional
+        The running standard deviations from the mean. If None, returns
+        1.0 for the standard deviation without computation.
```
rogiervd, line 1229: run_std doesn't return anything, but also, the returned standard deviation is `None`, right?

speechbrain/processing/features.py
```diff
@@ -1162,292 +1182,335 @@ def combine_gaussian_statistics_distributed(
...
+            x,
+            mask=mask,
+            dim=(0, self.length_dim),
+            run_count=self.count,
+            run_mean=self.glob_mean,
+            run_std=self.glob_std if self.std_norm else None,
         )
 
-        return current_mean, current_std
+    def _compute_current_stats(self, x, mask, dim):
```
rogiervd, line 1498: Why not just call `gaussian_statistics`, which has been unit-tested?

speechbrain/processing/features.py
```diff
@@ -1525,11 +1567,46 @@ def _load(self, path, end_of_epoch=False):
...
+    mask : torch.Tensor
+        A boolean tensor with `True` for valid positions and `False`
+        for padding positions.
+    """
+    if lengths is None:
+        lengths = torch.ones(x.size(0), device=x.device)
+
+    # Convert relative lengths to absolute lengths, then compute boolean mask
+    max_len = x.size(length_dim)
+    abs_lengths = lengths.unsqueeze(1) * max_len
```
rogiervd, line 1598: `(lengths * max_len).unsqueeze(1)` should be faster and use less memory, right?

rogiervd: And shouldn't it be `(lengths * max_len + some_small_constants).unsqueeze(1)` to stop rounding errors?

Note e.g. on my computer (presumably IEEE 754)
```
>>> (15 / 22) * 22
14.999999999999998
```

tests/unittests/test_input_norm.py
```diff
@@ -9,12 +9,272 @@
...
 )
 
 
+class TestInputNormalization:
+    """Test suite for the InputNormalization class."""
+
+    @pytest.fixture
+    def sample_data(self):
+        """Create sample data for testing."""
+        # Create a batch of 3 sequences with 4 features and variable lengths
```
rogiervd, line 26: Is a two-dimensional tensor good enough?

tests/unittests/test_input_norm.py
```diff
@@ -9,12 +9,272 @@
...
+
+# Utility tests for the helper functions
+def test_get_mask():
+    """Test the get_mask function."""
+
+    # Create a batch of sequences
+    x = torch.ones(3, 4, 2)  # Batch size 3, seq len 4, feature dim 2
+
+    # Relative lengths: 100%, 75%, 50%
+    lengths = torch.tensor([1.0, 0.75, 0.5])
```
rogiervd, line 257: Don't you want to test with values like 15/22?

Found by running
```
for numerator in range(1, 50):
  for denominator in range(numerator, 100):
    if (numerator/denominator) * denominator < numerator:
      print(numerator, denominator)
```

speechbrain/processing/features.py
```diff
@@ -998,8 +998,9 @@ def forward(self, x):
 
 def gaussian_statistics(
     x: torch.Tensor,
-    dim: Union[int, tuple, None] = None,
     mask: Optional[torch.Tensor] = None,
+    dim: Union[int, tuple, None] = None,
+    compute_var: bool = True,
```
rogiervd, line 1003: Do you test this? I didn't see it.

speechbrain/processing/features.py
rogiervd: Why should you want to use `True`? Much too dangerous in case there is a one-off difference in the padding somewhere.

rogiervd (changes requested):
Thanks a lot! This looks generally good.

I do have a number of small remaining questions and comments.

TParcollet:
@pplantinga I've looked both at a profiler trace and a simple time.time() check, and this is not making the normalisation slower. I think that once you have fixed the small changes requested by Rogier, we can simply merge! (after the release of 1.0.2, just to have a bit of safety net :p)

tests/unittests/test_input_norm.py
pplantinga: Added a test for this.

rogiervd (approved):
Thanks a lot for humouring my questions and suggestions. I think this is good!

TParcollet:
@pplantinga feel free to merge this PR!

PR #2850, opened 2025-03-04
Reforging LLaMA lobe 

TParcollet:
This PR refactor our LLaMA lobe to become generic enough and version agnostic. I cannot share the associated recipe for now, but it works ok. I still may have a few question open w.r.t checkpointing, but I think that this should be ok. I think that the current state of our LLaMA lobe is a real problem as it may hide the capacity of SB to adapt and to enable people to develop their custom code. 

@poonehmousavi Multiwoz will need to be slightly refactor around this code I think. As a rule of thumb here are the potential changes:
1. BNB and PeFT are not expected to be on the lobe. These are external things that must be managed at the yaml level.
2. SB has its own adapter scheme, we should rely on this as it's more transparent than PeFT. This does not prevent function like preparation of kbit training to be applied in the training recipe, but not in the lobe. We can also use PeFT adapter class with our adapter method.


speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(
+        self,
+        source: str,
+        save_path: str,
+        bnb_config: BitsAndBytesConfig = None,
+        freeze: bool = False,
+        pad_token: str = "[PAD]",
+        padding_side: Literal["left", "right"] = "right",
+        weight_precision_load: torch.dtype = torch.bfloat16,
```
Adel-Moumen, line 68: That's why our HF interface is bad. The `from_pretrained` method gives you the possibility to override the precision of the backend through `torch_dtype`, and also the `attn_implementation`. E.g.:
```python
AutoModelForCausalLM.from_pretrained(
          params.backbone,
          torch_dtype=torch.bfloat16,
          attn_implementation="flash_attention_2"
)
```

I would therefore suggest to modify `huggingface.py` directly to allow both kwargs, and just define through the yaml those params (this class should also have a `*args` to accommodate this).

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,133 @@
...
+    ) -> None:
+        self.pad_token = pad_token
+        self.padding_side = padding_side
```
Adel-Moumen, line 71: where is it being used?

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,133 @@
...
+        if self.bnb_config is None:
+            self.model = self.model.to(weight_precision_load)
+            logger.info(
+                f"Converted {self.__class__.__name__} to {weight_precision_load}"
+            )
+
+        self.load_tokenizer(source=source, pad_token=self.pad_token)
+        self.model.resize_token_embeddings(
+            len(self.tokenizer), pad_to_multiple_of=8
+        )  # TODO: optimal value of pad_to_multiple_of varies? see https://github.com/huggingface/tokenizers/issues/991.
```
Adel-Moumen, line 98: I believe it should in `huggingface.py`

TParcollet: Not all models have embeddings.

speechbrain/lobes/models/huggingface_transformers/llama.py
TParcollet: I found these attributes to potentially be useful to access in recipes, but maybe they can be removed yeah.

speechbrain/lobes/models/huggingface_transformers/llama.py
TParcollet: makes sense

TParcollet: @Adel-Moumen but wait, our interface already passes kwargs, there is nothing to change exept that I don't need to manage this on the llama lobe but just pass the dtype to the interface. right?

TParcollet:
@pplantinga do you have any idea why the doctest keep running on the llama file while I excluded it? This is frustrating ahah

Adel-Moumen:
> @pplantinga do you have any idea why the doctest keep running on the llama file while I excluded it? This is frustrating ahah

You need to add ` # doctest: +SKIP` (see: https://github.com/speechbrain/speechbrain/pull/2850#issuecomment-2729190180)

TParcollet:
@Adel-Moumen But like, isn't my change to the pytest file supposed to avoid testing llama.py? this sounds weird to me.

speechbrain/lobes/models/huggingface_transformers/llama.py
Adel-Moumen: you just need to add a kwargs to this interface, and then in the yaml to pass the `torch_dtype`. Also, you should maybe better document this behavior and especially the fact that you can dynamically modify the attn implemtantion through `attn_implementation`.

speechbrain/lobes/models/huggingface_transformers/llama.py
Adel-Moumen: There's specifics methods in HF to get the input/output embedding of a model. Look at `get_input_embeddings` and `get_output_embeddings`. I think it would be better to move this code in the `huggingface.py` as I believe this is a feature that could be used by other decoder models.

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,128 @@
...
+            Please refer to HuggingFace documentation and map it to your Llama model of interest.
+
+        Returns
+        -------
+        hyp : torch.Tensor
+            Contains tokenized (indices) outputs.
+        """
+
+        with torch.no_grad():
+            return self.model.generate(**kwargs)
```
Adel-Moumen, line 128: I am surprised that the default behavior of our `huggingface.py` class is to not call `self.model` methods

speechbrain/lobes/models/huggingface_transformers/llama.py
TParcollet: @Adel-Moumen yes, this is what I ended up doing (see the CoVoST ast recipe). For the documentation, I guess you are right, i'll add a few words in the docstring.

speechbrain/lobes/models/huggingface_transformers/llama.py
TParcollet: @Adel-Moumen I'll check. But I see the fact of putting it in huggingface.py as a problem. The reason is that we are dynamically changing the graph of the model, and this should not be hidden to the user. I feel like, actually, it should stay here but also be much better documented in the docstring. No?

speechbrain/lobes/models/huggingface_transformers/llama.py
TParcollet: You just don't like our HF interface :p

speechbrain/lobes/models/huggingface_transformers/llama.py
Adel-Moumen: wait what happen to this?

speechbrain/lobes/models/huggingface_transformers/llama.py
Adel-Moumen: If its not being used then it would be better to remove it :)

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,135 @@
...
+            **kwargs,
+        )
+
+        self.load_tokenizer(source=source, pad_token=self.pad_token)
+
+        # We resize the token embeddings size to a factor of 8 to maximise
+        # the use of tensorcores.
+        self.model.resize_token_embeddings(
+            len(self.tokenizer), pad_to_multiple_of=8
+        )
```
Adel-Moumen, line 100: I still tend to think that this should be part of `huggingface.py` but if you really think its better here then ok but i dont see valid reasons to not move it to the abstract class

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def forward(self, **kwargs):
+        """This function wraps the HuggingFace forward function. See the HuggingFace documentation of your Llama model of interest to know which
+        parameters to pass, typically the input tokens or embeddings and attention masks.
+
+        Arguments
+        ---------
+        **kwargs : list
+            Please refer to HuggingFace documentation and map it to your Llama model of interest.
```
Adel-Moumen, line 109: isn't it a dict?

speechbrain/lobes/models/huggingface_transformers/llama.py
```diff
@@ -0,0 +1,135 @@
...
+        return self.model(**kwargs)
+
+    def generate(self, **kwargs):
+        """This function wraps the HuggingFace generate function. See the HuggingFace documentation of your Llama model of interest to know which
+        parameters to pass, typically the input tokens or embeddings, attention masks and a transformers.GenerationConfig.
+
+        Arguments
+        ---------
+        **kwargs : list
+            Please refer to HuggingFace documentation and map it to your Llama model of interest.
```
Adel-Moumen, line 126: same comment

Adel-Moumen (approved):
LGTM

PR #2851, opened 2025-03-04
Add mock for ddp_all_reduce

rogiervd:
This allows testing of functions that uses `ddp_all_reduce`.

## What does this PR do?

This adds a utility function to test functions that use `ddp_all_reduce`, without using threads or multiple processors.

It does this by laboriously calling the functions that would be running on separate threads one by one, cancelling them at calls to `ddp_all_reduce` where not all results have been seen yet, and then re-running them for a bit longer every time.

The test runs two simple examples where this can be used.

This PR was prompted by #2835.

<details>

</details>

## PR review

<details>

</details>


pplantinga:
Looks cool! I may have a hard time verifying that it works in general, as I'm not very familiar with how to do mocks and so on.  Let me just try to apply it to my problem and verify that it works for that.

Before I do that though, I'm a bit confused why the machinery is needed for checking if the function is complete? Couldn't you do something a bit simpler like:

```
fake_process_tensors = [torch.tensor([1.]), torch.tensor([2.])]
function make_mock_ddp_reduce(fake_process_tensors):
    function mock_ddp_reduce(tensor, op):
         if op == SUM:
             return fake_process_tensors.sum()
    return mock_ddp_reduce
```

pplantinga:
I realize that the above is not as beautifully general as your solution, and tweaks would need to be made for multiple calls etc. However, the benefit IMO is more easily understanding how it works.

One more option to consider is just using threads or processes -- we already do that for testing the checkpointer:

https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_checkpoints.py#L431

TParcollet:
Wait @pplantinga I thought that the Free tier of CI that we are using did not allow multithreading? Aren't CI job supposed to run with only one thread?

rogiervd:
> I realize that the above is not as beautifully general as your solution, and tweaks would need to be made for multiple calls etc. However, the benefit IMO is more easily understanding how it works.
> 
> One more option to consider is just using threads or processes -- we already do that for testing the checkpointer:
> 
> https://github.com/speechbrain/speechbrain/blob/develop/tests/unittests/test_checkpoints.py#L431

From Titouan I had understood that it was impossible to use the real `all_reduce` in test. If this is possible, then that seems the right way of testing this.

If we need to use threads but still mock `ddp_all_reduce`, then the question becomes how to safely implement inter-thread communication of tensors, so then this PR may still be valuable.

pplantinga:
Thanks for explaining, I think I understand now why its necessary, and the revisions do make it clearer. Let me just try this on the PR to ensure it works there, and we can go ahead with this.

rogiervd:
It turns out a better idea to test `ddp_all_reduce` in multiple threads, see #2835.

PR #2857, opened 2025-03-11
Add functions to compute and combine Gaussian statistics (contribution from Samsung, AI Center, Cambridge)

rogiervd:
## What does this PR do?

This adds functions to compute and combines Gaussian statistics (mean and variance) on data. This comes out of me reviewing #2835, which also implements this.

However, I'm not clever enough to be confident that I fully understand this aspect of #2835, so I hereby offer an implementation which is simple enough that even I can understand it. Since the code is split up into different functions that each do one thing and can easily be tested and each require a docstring, the line count is a bit higher.

`combine_gaussian_statistics` and `combine_gaussian_statistics_distributed` compute the overall mean and variance from multiple means and variances exactly as if they were computed on all the data concatenated. After formulating the desired behaviour this way, the unit tests can test exactly this. For the DDP tests, I've copy-pasted from #2835.

@pplantinga if this looks interesting, I will still add support for `std=None` (assuming this is really necessary), and for masks.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -1185,16 +1185,14 @@ def mean_std_update(
     x: torch.Tensor,
     mask: Optional[torch.Tensor],
     dim: Union[int, tuple, None],
-    run_count,
-    run_mean,
-    run_std=None,
+    run_count: int,
+    run_mean: torch.Tensor,
+    run_std: torch.Tensor,
 ):
     """
     Update the running count, running mean, and running standard deviation
     by integrating new data x from multiple processes.
     """
-
-    # TODO implement run_std is None
     current_statistics = (run_count, run_mean, torch.square(run_std))
     new_statistics = combine_gaussian_statistics_distributed(
         gaussian_statistics(x, dim=dim, mask=mask)
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -1012,7 +1012,10 @@ def gaussian_statistics(
     dim: int | tuple | None
         The dimension or dimensions that the statistics should be computed over.
         The other dimensions are retained in the output.
-        If None, then scalar-valued statistics will be returned.
+        If None, then statistics will be computed over all dimensions and
+        scalar-valued statistics will be returned.
+        () has the same effect as None, which is nonsensical but it consistent
+        with torch.sum and friends.
     mask: torch.Tensor | None
         A boolean tensor with True for elements that should be considered, and
         False for elements that should not be considered (e.g. that are after
@@ -1038,7 +1041,9 @@ def gaussian_statistics(
     ) -> Tuple[tuple, tuple]:
         """Normalise "dim" and return (reduce_dimensions, keep_dimensions)."""
         all_dimensions = range(len(x.shape))
-        if dim is None:
+        if dim is None or dim == ():
+            # dim == () is an exceptional case and replicates the strangeness
+            # of torch.sum(.., dim=()) and friends.
             return (tuple(d for d in all_dimensions), ())
         elif isinstance(dim, int):
             return ((dim,), tuple(d for d in all_dimensions if d != dim))
@@ -1059,13 +1064,6 @@ def gaussian_statistics(
         for d in keep_dimensions:
             assert mask.size(d) == 1
 
-    if reduce_dimensions == ():
-        if mask is None:
-            return 1, x, torch.zeros_like(x)
-        else:
-            # mask.numel == 1
-            return int(torch.sum(mask)), mask * x, torch.zeros_like(x)
-
     if mask is None:
         number = math.prod(x.size(d) for d in reduce_dimensions)
     else:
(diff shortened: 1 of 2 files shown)


pplantinga:
This code looks good, its clean and fits the speechbrain design philosophy of providing clear reference algorithms. I admit that its a bit wordy for my taste and handles edge cases that I feel are unnecessary, but perhaps that's for the best.

I'm not sure we really need to support `std=None` but the padding masks are a crucial part of this, so go ahead and add them.

rogiervd:
Great. I've started to add support for masks, which I should be able to get in by Tuesday.

What edge cases could I get rid of you think? I do agree it has ended up wordy.

speechbrain/processing/features.py
```diff
@@ -993,6 +996,169 @@ def forward(self, x):
...
+    """
+
+    if dim is None:
+        number = math.prod(x.shape)
+    elif isinstance(dim, int):
+        number = x.shape[dim]
+    else:
+        assert isinstance(dim, tuple)
+        if dim == ():
+            return 1, x, torch.zeros_like(x)
```
pplantinga, line 1031: Mainly I was thinking of this edge case, I'm not too concerned about someone passing an empty tuple here, and if they did I'm guessing it would break in predictable ways. But honestly this is only two lines, so, who cares?

pplantinga: Meaning we can probably just leave it as-is.

speechbrain/processing/features.py
```diff
@@ -993,6 +996,215 @@ def forward(self, x):
...
+        () has the same effect as None, which is nonsensical but it consistent
+        with torch.sum and friends.
+    mask: torch.Tensor | None
+        A boolean tensor with True for elements that should be considered, and
+        False for elements that should not be considered (e.g. that are after
+        the end of utterances).
+        This tensor should have the same number of dimensions as "x".
+        The dimensions indicated by "dim" should have the same size as the
+        matching dimensions in "x".
+        The other dimensions should have size 1.
```
rogiervd, line 1026: @pplantinga would you mind checking that this is indeed the correct specification for the mask?

speechbrain/processing/features.py
```diff
@@ -993,6 +996,215 @@ def forward(self, x):
...
+    ---------
+    x: torch.Tensor
+        The tensor to compute the statistics over
+    dim: int | tuple | None
+        The dimension or dimensions that the statistics should be computed over.
+        The other dimensions are retained in the output.
+        If None, then statistics will be computed over all dimensions and
+        scalar-valued statistics will be returned.
+        () has the same effect as None, which is nonsensical but it consistent
+        with torch.sum and friends.
```
rogiervd, line 1018: I realised that `dim=()` worked strangely since this is what `torch.sum` implements. Now I just inherit this behaviour.

rogiervd (commented):
@pplantinga sorry about the delay. This should now be ready. Adding support for masks was a bit more of a hassle than I thought.

pplantinga (commented):
I am of the opinion that we should re-think the way the masks are done. This also still doesn't support skipping the variance calculation.

TParcollet:
@pplantinga and @rogiervd we should start working on a single PR, because I think we are approaching the end of this refactoring. #2835 is merging the approach from this PR with the original "philosophy", I reviewed it and left a few comments. I think that we would need a final round from @rogiervd and then we can proceed (@pplantinga I don't remember what we said w.r.t retro compatibility of existing checkpoints, is it supported? otherwise we should document it somewhere somehow.)

tests/unittests/test_input_norm.py
```diff
@@ -0,0 +1,348 @@
...
+        (2,),
+        (3,),
+        (0, 1),
+        (0, 2),
+        (0, 1, 2),
+        (0, 1, 3),
+    ],
+)
+@pytest.mark.parametrize("use_mask", [False, True])
+@pytest.mark.parametrize("random_seed", [20250304, 20250326, 20250327])
```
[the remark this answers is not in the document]
rogiervd: It did make a difference in the tests. Would you prefer a for-loop and an extra level of indentation? (I don't, hence the current format.)

speechbrain/processing/features.py
rogiervd: Well the function is general enough! Why restrict use to input normalisation?

pplantinga:
> I think the problem is `dim`. I feel like writing down and checking the shape of the `lengths` tensor is going to be just as complicated.

Perhaps you're right that the problem is `dim`. For the purpose of input norm (which started all of this) we have constant `dim`, it will always be `(0, 1)`. The reason for me starting to make it more general was because the "sentence" norm only normed over `dim=1` but that can already not be supported here due to the `number` being different across different features. So one option is to delete the `dim` parameter and always norm those two dimensions. Or we can just keep it as-is for potential future uses. This way is more complex, but its already done and we could just move ahead with it.

tests/unittests/test_input_norm.py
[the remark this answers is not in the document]
pplantinga: Okay, this is fair if it makes a difference.

pplantinga (approved):
I think we should just go ahead with the mask version because it is more flexible -- handles more cases than just using the length, even if it is a little more complex.

TParcollet:
Feel free to merge @pplantinga

PR #2865, opened 2025-03-17
SpeechLLM (with LLaMA) and Conformer recipe for speech translation on CoVoST (Code from Samsung AI Center Cambridge) 

TParcollet:
This PR introduces two speech translation recipes: one training a conformer encoder decoder with MT training (ASR+ST) from scratch. The other finetuning an XLS-R into an adapter before a LLama 3 7B model. I'd need some help to test these ones. These two recipes are basic things that we can find on the literature, nothing new, but should serve a good entry points for baselines of papers. 

This is based on the new LLaMA interface of #2850 

I only tried for En - De.


TParcollet:
I could use a first quick review on looking briefly at the code (forget about llama.py as it's from another PR). Maybe @Adel-Moumen / @pplantinga ?

@poonehmousavi this is an example as how to refactor multiwoz.

recipes/CoVoST/AST/train.py
```diff
@@ -0,0 +1,498 @@
...
+logger = get_logger(__name__)
+
+
+# Define training procedure
+class AST(sb.core.Brain):
+    def compute_forward(self, batch, stage):
+        """Forward computations from the waveform batches to the output probabilities."""
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        wavs, wav_lens = wavs.to(self.device), wav_lens.to(self.device)
```
Adel-Moumen, line 33: unnecessary line as you are already moving tensors on `device` on line 31

recipes/CoVoST/AST/train.py
```diff
@@ -0,0 +1,498 @@
...
+        if stage != sb.Stage.TRAIN:
+            current_epoch = self.hparams.epoch_counter.current
+            valid_search_interval = self.hparams.valid_search_interval
+            if current_epoch % valid_search_interval == 0 or (
+                stage == sb.Stage.TEST
+            ):
+
+                tgt_detokenizer = MosesDetokenizer(
+                    lang=self.hparams.tgt_language
+                )
```
Adel-Moumen, line 119: hmmm I think this might require some documentation somewhere. Is there a point of having a wrapper on top of that in speechbrain? Also, I think it would make sense to init this tokenizer outside of this block of code, as every test step you are going to re-init this tokenizer for "nothing". It would add more visibility by defining elsewhere (and only once)

recipes/CoVoST/AST/train.py
```diff
@@ -0,0 +1,498 @@
...
+    # 2. Define audio pipeline:
+    @sb.utils.data_pipeline.takes("wav")
+    @sb.utils.data_pipeline.provides("sig")
+    def audio_pipeline(wav):
+        info = torchaudio.info(wav)
+        sig = sb.dataio.dataio.read_audio(wav)
+        resampled = torchaudio.transforms.Resample(
+            info.sample_rate,
+            hparams["sample_rate"],
+        )(sig)
```
Adel-Moumen, line 277: QUestioN: don't we have already a `Resampler` in SpeechBrain? Why not using it ?

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+logger = get_logger(__name__)
+
+
+# Define training procedure
+class AST(sb.core.Brain):
+    def compute_forward(self, batch, stage):
+        """Forward computations from the waveform batches to the output probabilities."""
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        wavs, wav_lens = wavs.to(self.device), wav_lens.to(self.device)
```
Adel-Moumen, line 34: same as before, can be deleted thanks to line 32

recipes/CoVoST/AST/hparams/w2v2_llama3.yaml
```diff
@@ -0,0 +1,256 @@
...
+skip_prep: False # Skip data preparation
+convert_to_wav: True # Switch this to True to convert all mp3 files to wav.
+
+# URL for the HuggingFace model we want to load (BASE here)
+wav2vec2_hub: microsoft/wavlm-large
+wav2vec2_folder: !ref <save_folder>/wav2vec2_checkpoint
+wav2vec2_frozen: False
+
+# LLM options
+llm_path: !PLACEHOLDER # /scratch-01/llama-2/llama-2-7b-hf, or /scratch-01/llama-2/llama-2-7b-chat-hf
```
Adel-Moumen, line 37: Can't we use HF path directly ? E.g `Qwen/Qwen2-0.5B` ?

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+        text_len = tokens_prompt_translation.shape[1]
+        audio_prompt_len = (audio_len + prompt_len)[0]
+
+        # Then for the full embedding prompt sequence
+        if hasattr(self.modules.lora_llm, "module"):
+            embeddings = (
+                self.modules.lora_llm.module.model.get_input_embeddings()
+            )
+        else:
+            embeddings = self.modules.lora_llm.model.get_input_embeddings()
```
Adel-Moumen, line 73: this makes me believe we really need to have a pointer to the raw parameters (i.e. before wrapping the module to DDP). This way, we can just call specific methods that don't need forward passes.

Adel-Moumen: this is cost-free to do. We just need to add a single line in core.py that does `self.unwrapped_modules = self.module` before wrapping it in DDP. And this way we can then do `self.unwrapped_modules.lora_llm.get_input_embeddings()` without having to check for `.module.`

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+            inputs_embeds=inputs_embeds, attention_mask=attn_mask
+        ).logits
+
+        # output layer for seq2seq log-probabilities
+        p_seq = self.hparams.log_softmax(llm_logits)
+
+        if hasattr(self.modules.lora_llm, "module"):
+            gen_func = self.modules.lora_llm.module.model.generate
+        else:
+            gen_func = self.modules.lora_llm.model.generate
```
Adel-Moumen, line 98: same as before this is so ugly xD

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+
+        elif stage == sb.Stage.VALID:
+
+            # Define generation config depending on runtime values
+            config = transformers.GenerationConfig(
+                num_beams=self.hparams.valid_beam_size,
+                pad_token_id=tokenizer.pad_token_id,
+                eos_token_id=tokenizer.eos_token_id,
+                max_new_tokens=int(2 * audio_len),
+            )
```
Adel-Moumen, line 112: can't we define this in the yaml or outside this function?

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+
+            hyps = gen_func(
+                inputs_embeds=inputs_embeds[
+                    :, :audio_prompt_len
+                ],  # give model audio features and prompt for inference
+                attention_mask=attn_mask[:, :audio_prompt_len],
+                generation_config=config,
+            )
```
Adel-Moumen, line 120: hm. Question. From my understanding, the point of doing this is to avoid giving to the model padded tokens. Don't you think that this can be fixed by just making the `attention_mask` corresponding to that? i.e. setting to `False` entries that are padded so that the causal mask transformer doesn't attend to those tokens?

recipes/CoVoST/AST/train_w2v2_llama.py
Adel-Moumen: Yeah. I think something should be done on that side. Especially given the fact that this `generate` method has other attributes like `beam_size` etc. It would be nice to have a way to define it elsewere + custom

Adel-Moumen: Wait. Why are you doing `max_new_tokens=int(2 * audio_len)` ?

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+        # We are only interested in computing the loss over the logits after
+        # the audio + prompt embeddings. Tokens_translation does not start with bos, so we just need to make sure to shift the logits to the last token of the prompt (to ensure next word prediction)
+
+        p_seq_translation_only = p_seq[:, audio_prompt_len - 1 :]
+
+        loss = self.hparams.nll_loss(
+            p_seq_translation_only,
+            tokens_translation,
+            length=tokens_translation_len,
+        )
```
Adel-Moumen, line 164: I don't know what is more beautiful. Setting the target labels for `audio embds + prompt embeddings` to be `-100` or just cropping the input + target?

recipes/CoVoST/AST/train_w2v2_llama.py
```diff
@@ -0,0 +1,550 @@
...
+    def audio_pipeline(wav):
+        info = torchaudio.info(wav)
+        sig = sb.dataio.dataio.read_audio(wav)
+        resampled = torchaudio.transforms.Resample(
+            info.sample_rate,
+            hparams["sample_rate"],
+        )(sig)
+        resampled = torch.nn.functional.layer_norm(
+            resampled, normalized_shape=resampled.shape
+        )
```
Adel-Moumen, line 358: whats that? :o

Adel-Moumen (changes requested):
First round (more details will come later when I will have the chance to run the recipe)

How is the LLM processing Speech embeddings ? I see in one doctstring that the LLM takes `[speech embds] + [prompt] + [translation]`. I was wondering but in the vLLM literature, its more common to do `<image_start> <image embeddings> <image_end> + [prompt] + [translation]` and I belive this might be easier for the LLM to reason about what is an image thanks to the `<image>` token i.e. you can ask him question and relate to the `<image>` token. My question is more, how easy it is for us to modify the way the model is processing the input? Can we modify it easily to accommodate this?


Also, I tend to think that some things could be improved. I understnd the point of exposing the forward pass to make it easier to understand, but maybe we want here to hide some ugly tricks. For instance, I tend to think we could just have a custom generate function in our LLM interface + a custom GenerationCofing. We also need to have a pointer to the un-DDP wrapped modules to avoid having to check if `module` key is present etc.

recipes/CoVoST/AST/hparams/w2v2_llama3.yaml
```diff
@@ -0,0 +1,256 @@
...
+llm_prompt: " Translate English speech to German text: "
```
Adel-Moumen, line 38: Now that I'm thinking about it, but does it make sense to have a prompt if the LoRA adapters are only trained to perform English -> German?

recipes/CoVoST/AST/hparams/w2v2_llama3.yaml
```diff
@@ -0,0 +1,256 @@
...
+# llm: !new:speechbrain.nnet.adapters.AdaptedModel
+#    model_to_adapt: !ref <llm>
+#    adapter_class: !name:speechbrain.nnet.adapters.LoRA
+#    all_linear: True
+#    adapter_kwargs:
+#        rank: !ref <lora_rank>
+
+feat_downsampler: !new:speechbrain.lobes.downsampling.ConcatDownsampler
+    downsampling_factor: !ref <downsampling_factor>
+
```
Adel-Moumen, line 180: there's no such class as `ConcatDownsampler` :p are you an LLM? 👀

recipes/CoVoST/AST/hparams/w2v2_llama3.yaml
```diff
@@ -0,0 +1,256 @@
...
+# URL for the HuggingFace model we want to load (BASE here)
+wav2vec2_hub: microsoft/wavlm-large
+wav2vec2_folder: !ref <save_folder>/wav2vec2_checkpoint
+wav2vec2_frozen: False
+
+# LLM options
+llm_path: !PLACEHOLDER # /scratch-01/llama-2/llama-2-7b-hf, or /scratch-01/llama-2/llama-2-7b-chat-hf
+llm_prompt: " Translate English speech to German text: "
+llm_emb_size: 3072
+lowercase: True # Transcript are given to the LLM in uppercase or lowercase
```
Adel-Moumen, line 40: is it actually being used?

recipes/CoVoST/AST/train_w2v2_llama.py
Adel-Moumen: wondering if maybe instead of calling this at each step, you could just save the embedding in the brain class e.g. `brain.txt_embeddings = get_input_embeddings()` so that in the code you can just call it like this `brain.txt_embeddings(...)` (similar to `brain.tokenizer`).

Adel-Moumen: wait. does this code works if we are not using LORA? Here you are making the assumption that the lora block of code is not commented

recipes/CoVoST/AST/train_w2v2_llama.py
Adel-Moumen: AFAIK, we could just plug the LLaMA decoder in our seq2seq and it would/should work with minimal tuning. Usually, we do expect to see in the decoder audio frames, but we could make this optional.

recipes/CoVoST/AST/train.py
TParcollet: We shouldn't have one tbh.

recipes/CoVoST/AST/train_w2v2_llama.py
TParcollet: Yes, this is absurdly ugly. I have strictly no idea how we can fix it despite spending a significant amount of time on the problem. We would have to rethink entirely our DDP structure and even then, not sure that we can do something about that.

recipes/CoVoST/AST/train_w2v2_llama.py
TParcollet: I agree on all comments. I don't have time to do it however, and this is non blocking.

Adel-Moumen:
FYI, currently working on reviewing / making some edits. Then, will run tests and merge.

Adel-Moumen:
Added a recipe test for Conformer in the AST recipe. Couldn't do it for LLaMA since it's relying on a large pretrained LLM.  

Gonna fix the pre-commit (couldn't run them on CCA).

Adel-Moumen (approved):
LGTM!

We still have minor edits to do such as the generation part, how we construct the prompt etc but I think we need to move on on that topic. Will try to get back to this one and refactor a bit.

Thanks Sensei.

PR #2493, opened 2024-04-04
Temporarily pin max scipy version due to removed function breaking a dep

asumagic:
## What does this PR do?

gensim, an indirect dependency of SB (through flair, which is optional but installed in CI), is currently broken with SciPy `1.13.0`, pending the merge for https://github.com/piskvorky/gensim/pull/3524. In the mean time, fix CI by pinning scipy < `1.13.0`.

This is not otherwise a major problem for users outside of CI because flair is only imported if you use functionality from `lobes/models/flair`.

<details>
  <summary><b>Before submitting</b></summary>

- [x] Did you read the [contributor guideline](https://speechbrain.readthedocs.io/en/latest/contributing.html)?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing [tests](https://github.com/speechbrain/speechbrain/tree/develop/tests) pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Does your code adhere to project-specific code style and conventions?

</details>

## PR review

<details>
  <summary>Reviewer checklist</summary>

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified
- [ ] Confirm that the changes adhere to compatibility requirements (e.g., Python version, platform)
- [ ] Review the self-review checklist to ensure the code is ready for review

</details>

PR #2867, opened 2025-03-23
Update requirements.txt

mravanelli:
It looks like there is an incompatibility in the binary of numpy and scipy (if we restrict scipy to be <1.13.0). I'm not sure why that was introduced, but it is now causing issues.

To replicate:
https://colab.research.google.com/drive/11-6UBSSCjMOKJNJuCY2AwOtj8cnZ9A8F?usp=sharing

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -13,6 +13,8 @@ def pytest_generate_tests(metafunc):
 collect_ignore = [
     "setup.py",
     "speechbrain/lobes/models/huggingface_transformers/mert.py",
+    "speechbrain/lobes/models/flair",
+    "speechbrain/lobes/models/spacy",
 ]
 try:
     import numba  # noqa: F401
--- a/conftest.py
+++ b/conftest.py
@@ -15,6 +15,7 @@ collect_ignore = [
     "speechbrain/lobes/models/huggingface_transformers/mert.py",
     "speechbrain/lobes/models/flair",
     "speechbrain/lobes/models/spacy",
+    "speechbrain/alignment/ctc_segmentation.py",
 ]
 try:
     import numba  # noqa: F401
--- a/tests/.run-doctests.sh
+++ b/tests/.run-doctests.sh
@@ -5,5 +5,5 @@ set -e -u -o pipefail
 # > pytest --doctest-modules speechbrain/
 # However, we take this more complex approach to avoid testing files not
 # tracked by git. We filter out tests that require optional dependencies.
-avoid="transducer_loss.py\|fairseq_wav2vec.py\|huggingface_wav2vec.py\|bleu.py\|ctc_segmentation.py\|check_url.py\|huggingface_whisper.py\|language_model.py\|vocos.py\|discrete_ssl.py\|lobes/flair|lobes/spacy"
+avoid="transducer_loss.py\|fairseq_wav2vec.py\|huggingface_wav2vec.py\|bleu.py\|ctc_segmentation.py\|check_url.py\|huggingface_whisper.py\|language_model.py\|vocos.py\|discrete_ssl.py\|lobes/models/flair\|lobes/models/spacy"
 git ls-files speechbrain | grep -e "\.py$" | grep -v $avoid | xargs pytest --doctest-modules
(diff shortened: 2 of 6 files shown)


mravanelli:
It looks like the problem is a bit broader as it looks like we are not currently fully compliant with numpy 2

pplantinga:
Looking at the git blame for that line, the change was introduced with the following PR for supporting a dependency of `flair` until a pending PR was merged:

https://github.com/speechbrain/speechbrain/pull/2493

It looks like the PR is merged now, and `flair` will move to the `integrations` folder soon anyway, so this looks good to me. I'll see if I can find why the tests are breaking.

pplantinga:
It looks to me like the `transformers` library we depend on also has scipy pinned to <1.13:

https://github.com/huggingface/transformers/blob/main/src/transformers/dependency_versions_table.py#L78

I think `transformers` could also be removed from our main repo dependencies with the integrations PR, but that doesn't solve the issue for this particular PR. Perhaps the temporary solution is to ignore all the `transformers` tests in our CI.

pplantinga:
> It looks to me like the `transformers` library we depend on also has scipy pinned to <1.13:

I was wrong, the pytorch-based version of `transformers` does not depend on scipy. I ran the test locally (with numpy == 2.2.4) and it worked fine, so this problem has to be specific to the CI setup somehow.

mravanelli:
I'm fine upgrading the torch version.

TParcollet:
@pplantinga hum... can we use it for the CI and still leave it to people to be able to use 2.2? We already bumped it recently.

pplantinga:
> @pplantinga hum... can we use it for the CI and still leave it to people to be able to use 2.2? We already bumped it recently.

I suppose we could do this, but it would lead to broken installs if anyone installed both numpy 2.0 and torch 2.2... But I guess we can just add a note somewhere and leave it to people to manage themselves?

TParcollet:
"but it would lead to broken installs if anyone installed both numpy 2.0 and torch 2.2" -- this would only happen if people install without following our recommendations, I guess?

conftest.py
```diff
@@ -13,6 +13,9 @@ def pytest_generate_tests(metafunc):
 collect_ignore = [
     "setup.py",
     "speechbrain/lobes/models/huggingface_transformers/mert.py",
+    "speechbrain/lobes/models/flair",
```
TParcollet, line 16: Why are we adding these now?

tests/.run-doctests.sh
```diff
@@ -5,5 +5,5 @@ set -e -u -o pipefail
 # > pytest --doctest-modules speechbrain/
 # However, we take this more complex approach to avoid testing files not
 # tracked by git. We filter out tests that require optional dependencies.
-avoid="transducer_loss.py\|fairseq_wav2vec.py\|huggingface_wav2vec.py\|bleu.py\|ctc_segmentation.py\|check_url.py\|huggingface_whisper.py\|language_model.py\|vocos.py\|discrete_ssl.py\|lobes/flair|lobes/spacy"
+avoid="transducer_loss.py\|fairseq_wav2vec.py\|huggingface_wav2vec.py\|bleu.py\|ctc_segmentation.py\|check_url.py\|huggingface_whisper.py\|language_model.py\|vocos.py\|discrete_ssl.py\|lobes/models/flair\|lobes/models/spacy\|llama2.py"
```
TParcollet, line 8: I've tried modifying this line recently, and it was not working (See the PR on reforging Llama).

TParcollet (commented):
See comments

pplantinga (approved):
I'm a fan of unpinning the scipy version which will happen soon anyway in the "integrations" PR

conftest.py
pplantinga: Flair is not compatible with numpy 2.0 and can't be run on the CI server anymore. The others started failing too for some reason.

tests/.run-doctests.sh
pplantinga: There's two places that tests need to be excluded, here and `conftest.py`

Issue #2869, opened 2025-03-31
Bugs found in Recipe Tests


pplantinga:
### Describe the bug

Aishell1Mix and Binaural WSJ0 Mix and DNS:
```
speechbrain.utils.checkpoints - Would load a checkpoint here, but none found yet.
speechbrain.utils.epoch_loop - Going into epoch 1
speechbrain.core - Exception:
Traceback (most recent call last):
  File "/home/competerscience/Documents/Repositories/speechbrain/recipes/Aishell1Mix/separation/train.py", line 680, in <module>
    separator.fit(
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/core.py", line 1575, in fit 
    self._fit_train(train_set=train_set, epoch=epoch, enable=enable)
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/core.py", line 1400, in _fit_train
    loss = self.fit_batch(batch)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/competerscience/Documents/Repositories/speechbrain/recipes/Aishell1Mix/separation/train.py", line 130, in fit_batch
    if self.use_amp:
       ^^^^^^^^^^^^
AttributeError: 'Separation' object has no attribute 'use_amp'
```

GigaSpeech:
```
speechbrain.core - Exception:
Traceback (most recent call last):
  File "/home/competerscience/Documents/Repositories/speechbrain/recipes/GigaSpeech/ASR/transducer/train.py", line 465, in <module>
    tokenizer = SentencePiece(
                ^^^^^^^^^^^^^^
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/tokenizers/SentencePiece.py", line 177, in __init__
    run_on_main(self._train_BPE)
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/utils/distributed.py", line 105, in run_on_main
    main_process_only(func)(*args, **kwargs)
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/utils/distributed.py", line 168, in main_proc_wrapped_func
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/competerscience/Documents/Repositories/speechbrain/speechbrain/tokenizers/SentencePiece.py", line 333, in _train_BPE
    spm.SentencePieceTrainer.train(query)
[...]

PR #2872, opened 2025-04-01
Fix "use_amp" errors in recipes

pplantinga:
Found via the recipe tests, see #2869 

"Brain" class no longer has the "use_amp" attribute due to changes to the mixed precision handling (from #2827 ? )

But this should make recipes a little cleaner as we don't need to repeat the optimizer update code twice.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,4 +13,4 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.12'
-      - uses: pre-commit/action@v4.2.0
+      - uses: pre-commit/action@v3.0.1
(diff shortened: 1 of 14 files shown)


Adel-Moumen (approved):
LGTM! 

I am just wondering about this `kaldilm` changes. Could you please detail a bit why are we removing the dep? I saw you added the `arpa` file in the ignored files but was wondering what was the reason ? 

Thanks and good job as always :)

pplantinga:
Ah, good question. The CI was failing to build/install `kaldilm` and this integration will be removed from the CI soon anyway with the `integrations` PR. Unfortunately I can't go back and see what the error was, but iirc it was a build failure. Not sure what change caused it to happen.

Issue #2856, opened 2025-03-10
Weights only load failed. In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`.


lixl4433:
### Describe the bug

(speechbrain) [root@zz231 speechbrain]# pip  list
Package                       Version
----------------------------- -----------
alabaster                     0.7.16
asttokens                     3.0.0
attrs                         25.1.0
babel                         2.17.0
better-apidoc                 0.3.2
certifi                       2025.1.31
charset-normalizer            3.4.1
click                         8.1.8
comm                          0.2.2
commonmark                    0.9.1
debugpy                       1.8.13
decorator                     5.2.1
docutils                      0.21.2
exceptiongroup                1.2.2
executing                     2.2.0
fastjsonschema                2.21.1
filelock                      3.17.0
fsspec                        2025.2.0
greenlet                      3.1.1
huggingface-hub               0.29.2
HyperPyYAML                   1.2.2
hypothesis                    6.127.7
idna                          3.10
imagesize                     1.4.1
importlib_metadata            8.6.1
ipykernel                     6.29.5
ipython                       8.18.1
jedi                          0.19.2
Jinja2                        3.1.6
joblib                        1.4.2
jsonschema                    4.23.0
jsonschema-specifications     2024.10.1
jupyter-cache                 1.0.1
jupyter_client                8.6.3
jupyter_core                  5.7.2
kenlm                         0.2.0
llvmlite                      0.43.0
Markdown                      3.7
markdown-it-py                3.0.0
MarkupSafe                    3.0.2
matplotlib-inline             0.1.7
mdit-py-plugins               0.4.2
mdurl                         0.1.2
mpmath                        1.3.0
myst-nb                       1.2.0
myst-parser                   3.0.1
nbclient                      0.10.2
nbformat                      5.10.4
nest-asyncio                  1.6.0
networkx                      3.2.1
numba                         0.60.0
numpy                         1.26.4
nvidia-cublas-cu12            12.4.5.8
nvidia-cuda-cupti-cu12        12.4.127
nvidia-cuda-nvrtc-cu12        12.4.127
nvidia-cuda-runtime-cu12      12.4.127
nvidia-cudnn-cu12             9.1.0.70
nvidia-cufft-cu12             11.2.1.3
nvidia-curand-cu12            10.3.5.147
nvidia-cusolver-cu12          11.6.1.9
nvidia-cusparse-cu12          12.3.1.170
[...]

PR #2875, opened 2025-04-03
Fix issue with weight loading for torch 2.6

TParcollet:
This is fixing #2856 .

This error may happen in other places, but i don't think we should fix them all as they should be considered individually to avoid introducing security issues.

### DIFF
--- a/speechbrain/processing/features.py
+++ b/speechbrain/processing/features.py
@@ -1526,7 +1526,7 @@ class InputNormalization(torch.nn.Module):
         """
         del end_of_epoch  # Unused here.
         device = "cpu"
-        stats = torch.load(path, map_location=device)
+        stats = torch.load(path, map_location=device, weights_only=False)
         self._load_statistics_dict(stats)
 
 


pplantinga (commented):
I don't think this should be needed here right? What part of the checkpoint is not a weight?

Adel-Moumen:
> I don't think this should be needed here right? What part of the checkpoint is not a weight?

I think here `stats` is not a `tensor` and might be a `dict` or something like that which would explain why we need to add `weights_only=False`.

pplantinga:
Dictionaries should be fine right?

From https://pytorch.org/docs/stable/generated/torch.load.html:
"weights_only – Indicates whether unpickler should be restricted to loading only tensors, primitive types, dictionaries and any types added via [torch.serialization.add_safe_globals()](https://pytorch.org/docs/stable/notes/serialization.html#torch.serialization.add_safe_globals). See [torch.load with weights_only=True](https://pytorch.org/docs/stable/notes/serialization.html#weights-only) for more details."

I haven't had any trouble loading the input norm checkpoints even though I'm using 2.6 already.

TParcollet:
Tbh, I couldn't reproduce the issue either, but it happened to our user. Isn't it ok to just merge this? It shouldn't break things for the people not affected, right?

PR #2876, opened 2025-04-03
Generic feature extraction POC

flexthink:
## What does this PR do?
(Work in progress) A universal feature extractor to extract arbitrary features from dataset (discrete tokens, continuous representations, etc) and save them using arbitrary formats.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/recipes/LibriTTS/extraction/extract.py
+++ b/recipes/LibriTTS/extraction/extract.py
@@ -64,6 +64,7 @@ if __name__ == "__main__":
         audio_features = modules.ssl_model(sig.data, sig.lengths)
         if selected_layers:
             audio_features = audio_features[selected_layers]
+        audio_features = audio_features.permute(1, 2, 0, 3)
         return PaddedData(audio_features, sig.lengths)
 
     @sb.utils.data_pipeline.takes("sig")


recipes/LJSpeech/extraction/extract.py
Adel-Moumen, line 1: I don't think that this script should be here. I think it should be dataset dependent similar to what we are doing for let's say `librispeech_preparation.py`

recipes/LJSpeech/extraction/sanity_check.py
Adel-Moumen, line 1: maybe you should move this a unit-test. I think the extraction will requires extensive tests to make sure they are correct in the loading/saving process

Adel-Moumen (commented):
Do you have an example of a `train.py` integration of your new tokens loader?

recipes/LJSpeech/extraction/extract.py
flexthink: This was borrowed from DASB - my older approach integrated that with preparation.

recipes/LJSpeech/extraction/sanity_check.py
flexthink: I will create unit tests

flexthink:
> Do you have an example of a `train.py` integration of your new tokens loader?

I have some private examples - but they are on new work in progress not ready to be merged yet, as well as older incarnations of Tokotron.

I would suggest choosing one existing recipe and integrating it.

Adel-Moumen:
Also, quick question to @pplantinga, don't you think we should maybe aim for a single backend? Given that we are trying to minimise the number of dependencies, I would find it better to just stick to the best and more general purpose solution (instead of having something too general). I believe that most of them share similar pos and cons. In our context, I am not sure if we really need something very sota, I would prefer having something easy to use, where we only need low efforts to maintain the integration. So maybe, something like numpy or h5 is enough.

flexthink:
See #2938 for a simplified H5-only version.

recipes/LJSpeech/extraction/sanity_check.py
flexthink: @Adel-Moumen: Unit tests created in #2938

Adel-Moumen:
Closed in favour of #2985.

PR #2885, opened 2025-04-11
SpeechLLM LibriSpeech recipe

Adel-Moumen:
## What does this PR do?

Not ready to be merged yet :) 

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/recipes/LibriSpeech/ASR/transformer/train_speechllm.py
+++ b/recipes/LibriSpeech/ASR/transformer/train_speechllm.py
@@ -1,0 +1,342 @@
+#!/usr/bin/env python3
+"""
+
+Speech Encoder -> Adapter -> LLM
+
+
+Authors
+ * Adel Moumen 2025
+"""
+
+import os
+import sys
+from pathlib import Path
+
+import torch
+from hyperpyyaml import load_hyperpyyaml
+
+import speechbrain as sb
+from speechbrain.utils.distributed import if_main_process, run_on_main
+from speechbrain.utils.logger import get_logger
+
+logger = get_logger(__name__)
+
+
+# Define training procedure
+class ASR(sb.core.Brain):
+    def compute_forward(self, batch, stage):
+        """Forward computations from the waveform batches to the output probabilities."""
+        batch = batch.to(self.device)
+        wavs, wav_lens = batch.sig
+        tokens_bos, _ = batch.tokens_bos
+
+        wavs = self.hparams.normalize(wavs, wav_lens)
+        feats = self.modules.ssl(wavs, wav_lens)
+        feats = self.modules.proj(feats)
+        # self.raw_modules.llm
+        txt_embds = self.txt_embedding(tokens_bos)
+        print(txt_embds.shape)
+        exit()
+        return None
+
+    def compute_objectives(self, predictions, batch, stage):
+        """Computes the loss (CTC+NLL) given predictions and targets."""
+
+        return None
+
+    def on_stage_start(self, stage, epoch):
+        """Gets called at the beginning of each epoch"""
+        if stage != sb.Stage.TRAIN:
+            self.acc_metric = self.hparams.acc_computer()
+            self.wer_metric = self.hparams.error_rate_computer()
+
+    def on_stage_end(self, stage, stage_loss, epoch):
+        """Gets called at the end of a epoch."""
+        # Compute/store important stats
+        stage_stats = {"loss": stage_loss}
+        if stage == sb.Stage.TRAIN:
+            self.train_stats = stage_stats
+        else:
+            stage_stats["ACC"] = self.acc_metric.summarize()
+            current_epoch = self.hparams.epoch_counter.current
+            valid_search_interval = self.hparams.valid_search_interval
+            if (
+                current_epoch % valid_search_interval == 0
+                or stage == sb.Stage.TEST
+            ):
+                stage_stats["WER"] = self.wer_metric.summarize("error_rate")
+
+        # log stats and save checkpoint at end-of-epoch
+        if stage == sb.Stage.VALID:
+            lr = self.hparams.noam_annealing.current_lr
+            steps = self.optimizer_step
+            optimizer = self.optimizer.__class__.__name__
+
+            epoch_stats = {
+                "epoch": epoch,
+                "lr": lr,
+                "steps": steps,
+                "optimizer": optimizer,
+            }
+            self.hparams.train_logger.log_stats(
+                stats_meta=epoch_stats,
+                train_stats=self.train_stats,
+                valid_stats=stage_stats,
+            )
+            self.checkpointer.save_and_keep_only(
+                meta={"ACC": stage_stats["ACC"], "epoch": epoch},
+                max_keys=["ACC"],
+                num_to_keep=self.hparams.avg_checkpoints,
+            )
+
+        elif stage == sb.Stage.TEST:
+            self.hparams.train_logger.log_stats(
+                stats_meta={"Epoch loaded": self.hparams.epoch_counter.current},
+                test_stats=stage_stats,
+            )
+            if if_main_process():
+                with open(
+                    self.hparams.test_wer_file, "w", encoding="utf-8"
+                ) as w:
+                    self.wer_metric.write_stats(w)
+
+            # save the averaged checkpoint at the end of the evaluation stage
+            # delete the rest of the intermediate checkpoints
+            # ACC is set to 1.1 so checkpointer only keeps the averaged checkpoint
+            self.checkpointer.save_and_keep_only(
+                meta={"ACC": 1.1, "epoch": epoch},
+                max_keys=["ACC"],
+                num_to_keep=1,
+            )
+
+    def on_fit_batch_end(self, batch, outputs, loss, should_step):
+        """At the end of the optimizer step, apply noam annealing."""
+        if should_step:
+            self.hparams.noam_annealing(self.optimizer)
+
+
+def dataio_prepare(hparams, tokenizer):
+    """This function prepares the datasets to be used in the brain class.
+    It also defines the data processing pipeline through user-defined functions.
+    """
+    data_folder = hparams["data_folder"]
+
+    train_data = sb.dataio.dataset.DynamicItemDataset.from_csv(
+        csv_path=hparams["train_csv"],
+        replacements={"data_root": data_folder},
+    )
+
+    if hparams["sorting"] == "ascending":
+        # we sort training data to speed up training and get better results.
+        train_data = train_data.filtered_sorted(sort_key="duration")
+        # when sorting do not shuffle in dataloader ! otherwise is pointless
+        hparams["train_dataloader_opts"]["shuffle"] = False
+
+    elif hparams["sorting"] == "descending":
+        train_data = train_data.filtered_sorted(
+            sort_key="duration", reverse=True
+        )
+        # when sorting do not shuffle in dataloader ! otherwise is pointless
+        hparams["train_dataloader_opts"]["shuffle"] = False
+
+    elif hparams["sorting"] == "random":
+        pass
+
+    else:
+        raise NotImplementedError(
+            "sorting must be random, ascending or descending"
+        )
+    valid_data = sb.dataio.dataset.DynamicItemDataset.from_csv(
+        csv_path=hparams["valid_csv"],
+        replacements={"data_root": data_folder},
+    )
+    valid_data = valid_data.filtered_sorted(sort_key="duration")
+
+    # test is separate
+    test_datasets = {}
+    for csv_file in hparams["test_csv"]:
+        name = Path(csv_file).stem
+        test_datasets[name] = sb.dataio.dataset.DynamicItemDataset.from_csv(
+            csv_path=csv_file, replacements={"data_root": data_folder}
+        )
+        test_datasets[name] = test_datasets[name].filtered_sorted(
+            sort_key="duration"
+        )
+
+    datasets = [train_data, valid_data] + [i for k, i in test_datasets.items()]
+
+    # 2. Define audio pipeline:
+    @sb.utils.data_pipeline.takes("wav")
+    @sb.utils.data_pipeline.provides("sig")
+    def audio_pipeline(wav):
+        sig = sb.dataio.dataio.read_audio(wav)
+        return sig
+
+    sb.dataio.dataset.add_dynamic_item(datasets, audio_pipeline)
+
+    bos_index = tokenizer.bos_token_id
+    eos_index = tokenizer.eos_token_id
+    print(bos_index, eos_index)
+
+    # 3. Define text pipeline:
+    @sb.utils.data_pipeline.takes("wrd")
+    @sb.utils.data_pipeline.provides(
+        "wrd", "tokens_list", "tokens_bos", "tokens_eos", "tokens"
+    )
+    def text_pipeline(wrd):
+        wrd = wrd[0] + wrd[1:].lower()
+        yield wrd
+        tokens_list = tokenizer(wrd, add_special_tokens=False).input_ids
+        yield tokens_list
+        tokens_bos = torch.LongTensor([bos_index] + tokens_list)
+        yield tokens_bos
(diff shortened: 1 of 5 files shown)


speechbrain/integrations/huggingface/llama.py
```diff
@@ -63,13 +66,22 @@ def __init__(
...
         self.source = source
         self.save_path = save_path
         self.bnb_config = bnb_config
 
+        # Capture config-only overrides to avoid passing them to from_pretrained
+        self._config_overrides = {}
+        if "output_hidden_states" in kwargs:
+            self._config_overrides["output_hidden_states"] = kwargs.pop(
+                "output_hidden_states"
+            )
```
Adel-Moumen, line 87: TBH, I don't remember in what scenario you want to set `output_hidden_states=True`

recipes/LibriSpeech/ASR/transformer/train_speechllm.py
```diff
@@ -0,0 +1,678 @@
...
+        """
+        yield wrd
+        tokens_list = tokenizer(wrd, add_special_tokens=False).input_ids
+        yield tokens_list
+        tokens_bos = torch.LongTensor(
+            [start_of_audio_index]
+            + [end_of_audio_index]
+            + prompt_ids
+            + [bos_index]
+            + tokens_list
```
Adel-Moumen, line 475: one thing: it's really hard to know if an LM requires bos / eos (e.g. https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior) ; so, I think, ideally, we need to have a proper prompt function that will create a prompt depending on the input tokens (e.g. is eos is None, then don't append etc).

recipes/LibriSpeech/ASR/transformer/hparams/speechllm_ssl_feats.yaml
pplantinga, line 1: Do we need a second yaml file for this? I'm legitimately curious -- its fine if the answer is yes!

speechbrain/integrations/huggingface/llama.py
```diff
@@ -63,13 +66,22 @@ def __init__(
         freeze: bool = False,
         pad_token: str = "[PAD]",
         torch_dtype: torch.dtype = torch.float16,
+        additional_special_tokens: List[str] = None,
+        pad_to_multiple_of: int = 8,
```
pplantinga, line 74: These new parameters are not documented.

speechbrain/nnet/activations.py
pplantinga, line 1: What's the benefit of adding this over `torch.nn.GELU`? It doesn't seem to behave any differently

recipes/LibriSpeech/ASR/transformer/hparams/speechllm_e2e.yaml
```diff
@@ -0,0 +1,252 @@
...
+                value: !ref <ignore_index>
+
+####################### Model Architecture Parameters #######################
+lora_rank: 16
+
+# Output & Decoding Parameters
+bos_index: !PLACEHOLDER # 0
+eos_index: !PLACEHOLDER # 0
+pad_token: !PLACEHOLDER # 128256
+prompt: "Transcribe speech to text."
```
pplantinga, line 140: This recipe doesn't support tasks other than transcription? Couldn't we at least do keyword spotting? "Is the word {word} present in the audio?" For SpeechLLMs my understanding is that we ultimately want multi-task machines, so it would be nice if we at least had a basic concept of how multi-task would be handled.

pplantinga (commented):
Overall, looks like a good addition -- it will be nice to have a starting framework for doing SpeechLLMs in SpeechBrain!

From my first read, I guess I'm wondering if we want to try to support more than one task, even a simple second task such as keyword spotting, just to show how it can be done, as it seems like the main benefit of SpeechLLMs over traditional ASR is the fact that they can support multiple tasks.

recipes/LibriSpeech/ASR/transformer/README.md
```diff
@@ -23,6 +22,20 @@ installed in your environment (see extra-requirements.txt)**
...
+- SpeechLLM with SSL features
+- SpeechLLM with E2E features
+
+In the first mode, the speech features are extracted from the audio waveforms using a pre-trained SSL model, and then projected to the LLM embedding space using a linear layer projection, where everything is trained jointly.
+
+In the second mode, the speech features are already being extracted offline (see: `extract_ssl_feats.py` script). The LLM is then trained on the frozen SSL representations. This mode is more efficient and faster to train, but at the cost of flexibility on the frozen SSL model.
+
+| Release | Model | hyperparams file | Dev Clean WER | Test Clean WER | Test Other WER | HuggingFace link | Model link | GPUs |
+|:-------------:|:-------------:|:-------------:|:---------------------------:| :-----:| :-----:| :-----:| :--------:|
+| 29-11-25 | WavLM Large + SmolLM2 1.7B + LoRA | speechllm_ssl_feats.yaml | N/A | 3.17 | 6.83 | [HuggingFace](https://huggingface.co/speechbrain/asr-speechllm-librispeech) | - | 1xH100 40GB |
```
TParcollet: These are not really good numbers given the architecture imho, but ok.

Adel-Moumen: I had much better results (2.X%), but I need to train the models a bit longer.

recipes/LibriSpeech/ASR/transformer/extract_ssl_feats.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+"""Script to extract SSL features from the audio waveforms.
+
+The script uses the `speechbrain.integrations.hdf5.cached_item` module to cache the features.
```
TParcollet, line 4: Tutorial

recipes/LibriSpeech/ASR/transformer/hparams/speechllm_e2e.yaml
Adel-Moumen: I think for now it would best to stay on a single task, since we already have a speechllm translation task (although, will need some udpates in the future as the current code is not optimal). One could just train LORA adapters for ASR, then again a new set of adapters for a different task while loading the same backbone (instead of training with multiple tasks which I think what is currently being done in industry)

recipes/LibriSpeech/ASR/transformer/hparams/speechllm_ssl_feats.yaml
Adel-Moumen: I think it's better to have two yamls so that we don't end up loading on the device the full architecture while we should only load the encoder for extracting features.

recipes/LibriSpeech/ASR/transformer/extract_ssl_feats.py
Adel-Moumen: Done! Please have a look at data-loading-pipeline which now features multiple sections about saving features, and also update the speaker example with loading saved features.

recipes/LibriSpeech/ASR/transformer/README.md
```diff
@@ -23,6 +22,20 @@ installed in your environment (see extra-requirements.txt)**
...
+- SpeechLLM with SSL features
+- SpeechLLM with E2E features
+
+In the first mode, the speech features are extracted from the audio waveforms using a pre-trained SSL model, and then projected to the LLM embedding space using a linear layer projection, where everything is trained jointly.
+
+In the second mode, the speech features are already being extracted offline (see: `extract_ssl_feats.py` script). The LLM is then trained on the frozen SSL representations. This mode is more efficient and faster to train, but at the cost of flexibility on the frozen SSL model.
+
+| Release | Model | hyperparams file | Dev Clean WER | Dev Other WER | Test Clean WER | Test Other WER | HuggingFace link | Model link | GPUs |
+|:-------------:|:-------------:|:-------------:|:---------------------------:| :-----:| :-----:| :-----:| :-----:| :--------:|
+| 29-01-26 | WavLM Large + LLama 3.2 1B + LoRA | speechllm_e2e.yaml | 2.79 | 5.03 | 2.72 | 5.34 | [HuggingFace](https://huggingface.co/speechbrain/asr-wavlm-large-llama3.2-1b-lora-librispeech) | - | 1xA100 80GB |
```
Adel-Moumen, line 37: results have been updated and now are much better! Will upload and turn the repo public ASAP.

speechbrain/nnet/activations.py
Adel-Moumen: The main reason is standardisation with the yamls. I found in some examples that we had to define a class in order to use it in speechbrain e.g. if this an apply function (i.e. return f(x)) or a standalone class that you need first to create (e.g. f = nn.ReLU(); f(x))

Adel-Moumen: I don't think this is necessary a bad idea, especially when you think about the fact that some classes can have dynamic activation instantiations (e.g. if i recall correctly RNNs can have specific input activations functions)

Adel-Moumen:
Hi guys @pplantinga @TParcollet @mravanelli , I think the PR is now ready. I went through the comments, and added a tutorial for the caching feature. Also, I improved the ASR SpeechLLM baseline and now gets more competitive results (2.72% on LS test-clean and 5.34% on test-other). 

I also created the model card and will upload the required files so that we can display this example for the community. This is based on Llama 3.2 1B with LORA adapters and WavLM-Large. 

Happy to consider other remarks/requests. 

:)

speechbrain/nnet/activations.py
pplantinga: I get that sometimes a class may be needed and sometimes a function, but torch has both a class `torch.nn.GELU` and a function `torch.nn.functional.gelu` so this shouldn't be needed here right?

pplantinga: If you need the default argument changed you can always use "name" to change it, e.g.

```yaml
activation: !name:torch.nn.GELU {approximate: tanh}
```

Which returns a constructor that will use tanh by default.

speechbrain/nnet/activations.py
```diff
@@ -169,3 +169,76 @@ def forward(self, x):
...
+            Input tensor.
+
+        Returns
+        -------
+        The GeLU output.
+        """
+        return F.gelu(x, approximate=self.approximate)
+
+
+class ReLU(torch.nn.Module):
```
pplantinga: Same here, I don't think this is needed.

speechbrain/nnet/schedulers.py
```diff
@@ -300,7 +300,6 @@ class LinearWarmupScheduler:
     from the initial lr set in the optimizer to 0, after
     a warmup period during which it increases linearly
     from 0 to the initial lr set in the optimizer.
-    * Ge Li 2022
```
pplantinga, line 303: Did Ge Li not work on this? Why would we remove it?

speechbrain/core.py
```diff
@@ -457,6 +457,12 @@ def __init__(  # noqa: C901
                 self.hparams.output_folder,
             )
 
+        self.raw_modules = (
+            self.modules.module
+            if hasattr(self.modules, "module")
+            else self.modules
+        )
+
```
pplantinga, line 465: I'm not sure about this... is this handling the extra layer from DDP? I'm wondering if this should just be handled by the recipe or if we do need a more general solution whether we need something more robust here somehow, like a function you can call to get the modules appropriately.

docs/tutorials/basics/data-loading-pipeline.ipynb
```diff
@@ -1,2167 +1,2691 @@
...
...Dr+Rf8q3aCxzFCrvcDegRrLKaNPZ9/mq281F9K5Hdo1pJfVUTAMgxLRWSvSrATcbNkglCjPJp7pYyz7O7IT7VBwYwF0fTPUljXD7R/mh4+KmVmVUfUhrw2nO6uUHl9ylLsU8ve2TcHhuNb4fcE/Cs+lRD+Te+wfl9E75iJIlOtlmin+D0fOKu85qy26j0tP3qkcOvsldqHKth2+/ggSLedQXLv3ECdsxrC2cq7KyyWoInyG+dYrsO92xuTNLzf26lXXoXsf0OTeUNkeSkFvtFveb8qy5rZw8/esy2aX6pboayPhwDiDzlfmygvVq/RyOnz3LQKF+uWtYZPtB/3mnhHDUXBjEbjp+kxI5jbHhThRu+XCaXH6z39RtUS2xEXluVrYECbO6J3ZbqO4CqxS/Bm5x9lvRo+clX8jLPxZ9TLc6LtH8WrTcJ3bNkm8AU/eaPdzf7Bffe4UXdkkZAWz8cczEv75J+s3YfzB9mmoIswKFIWMaeVj0UebPxZn7P1lAvN3s7+jJKfpnmuqvxX7e7jZvDmjddmHO/7UXIgFjIb/rLJp6nN8Ee5QcGMBuHovLJL2THMhA3xK0G6eBBc9U8pW4nAhOVX+NWbPEcMwDJLTsj6U812dpXW9To+3oZ+V8kRw7umR6LzyCm6+UT6/wzr2pdLjUikjtzLmNzG786GqCrKW6QalGDaMNE4sP3E66dMbpftxmZO9NhMRfvI4PsWpX/HGhhsaEj5m/4vbfdY4fysaSbWfVzVGzH3izcjYH/hvv/r8QgUE3ExkJppRcGMBuBqWKij4zEm9mU4e2KRlSeNOynvG4oZ3TJrqob2F+y6iVny47HFzNUvic8qZFTU7J8EPbPkYit0rMyZW5lweruynKZUy6DtvE35ZuBUMw+BFDLcrZITI6A1gNGy/oK16orvYfcB8V1FlKvz1IkKNEHg3fKV+q4z4+Kz5YvXfstuDp6/YRMOC1/mH2N3wc/+eLZgrVhx6JqaBghvCm8zhG02U5YrQdt6IPl5eVOyxSNWjax4AGEb1RMbGtwaoXZH15dllrDilf2K+zEncGzetkX9Cyc/z8tO3WJ04GKviBiAuIQkLly3S+7rKxP6Qv+uW/H9C9PcEw3ONZCr2KGOZ9ZH7kbj0gtvAXB/aDO+8iDHOTtI+GhIyZl99VEmo3V5sXIu+aVjw2vtac1a2c8lU6Vu45
```
pplantinga: I'm confused why this states "Compression: None" when "gzip" was specified earlier. Maybe this just needs to be run once more?

pplantinga (commented):
Looks like its getting close, thanks Adel! Just a few small comments before merge.

speechbrain/core.py
Adel-Moumen: The point is to exactly remove the DDP extra layering (similar to here: https://github.com/karpathy/nanoGPT/blob/3adf61e154c3fe3fca428ad6bc3818b27a3b8291/train.py#L253). That way, we can systematically access to methods e.g. `get_input_embedding` which wouldn't be accessible easily (e.g. here: `self.raw_modules.get_input_embedding`) while before we would have to do something like `x = self.modules.module.get_input_embedding if 'module' in self.modules else ...` so you would still need to unwrap the DDP container to get target module for calling the function. Here, we just have a pointer that does that instead of doing it manually. I have seen some recipes that had to do a ugly 'if' and I think this would solve the problem. 

having a function that unwrap would work but I think this is simpler and more modular. Happy to consider something else if you think this is not the way to go.

Adel-Moumen:
> Looks like its getting close, thanks Adel! Just a few small comments before merge.

I think I addressed all the comments! Fixed the notebook and HDF5 (we were never passing the `compression` arg to the constructor) + removed the activations + added back authors.

speechbrain/core.py
pplantinga: Well, I'm happy to go with this for now if needed, but maybe the longer term solution is to actually store the DDP module in a separate container. I mean for this part (simplified):

```
for name, module in self.modules.items():
        module = SyncBatchNorm.convert_sync_batchnorm(module)
        ddp_module = DDP( 
            module,
            device_ids=[self.device],
            find_unused_parameters=self.find_unused_parameters,
        )
        self.modules[name] = ddp_module
```

change the last line to `self.ddp_modules[name] = ddp_module` so we don't overwrite the old one.

PR #2911, opened 2025-05-14
Fix type annotations (from Samsung, AI Center, Cambridge)

rogiervd:
## What does this PR do?

Fix existing type annotations. These are meant to be uncontroversial.

There are a few types of categories:
* Ungrammatical or mis-spelled type annotations.
* Arguments `argument: int = None` where the type should be `Optional[int]`
* The opposite: arguments `argument: Optional[int] = 3072` where the type should be `int`, because `Optional` means "may be `None`".
* Various other cases.

These were flagged by Pyright (see https://github.com/speechbrain/speechbrain/pull/2901).

<details>

</details>

## PR review

<details>

</details>


speechbrain/utils/logger.py
```diff
@@ -128,7 +128,7 @@ def log(self, level: int, msg: str, *args: tuple, **kwargs: dict):
                 self.logger.log(level, msg, *args, **kwargs)
 
     @functools.lru_cache(None)
-    def warning_once(self, *args: tuple, **kwargs: dict):
+    def warning_once(self, *args, **kwargs):
```
rogiervd, line 131: From https://typing.python.org/en/latest/spec/callables.html#annotating-args-and-kwargs:
> Therefore, the definition:
```
def func(*args: str, **kwargs: int): ...
```
> means that the function accepts an arbitrary number of positional arguments of type str and an arbitrary number of keyword arguments of type int.

speechbrain/dataio/sampler.py
```diff
@@ -13,7 +13,7 @@
 
 from collections import Counter
 from operator import itemgetter
-from typing import List
+from typing import List, Optional, Union
```
pplantinga, line 16: Should `List` be removed here?

speechbrain/lobes/models/transformer/Transformer.py
```diff
@@ -104,26 +104,26 @@ def __init__(
         num_decoder_layers=6,
         d_ffn=2048,
         dropout=0.1,
-        activation=nn.ReLU,
+        activation: type = nn.ReLU,
```
pplantinga, line 107: Should we use `Type[nn.Module]` here?

speechbrain/nnet/RNN.py
```diff
@@ -1914,8 +1916,9 @@ def __init__(
         if self.output_gate:
             self.o_gate = nn.Sigmoid()
 
-    def forgetMult(self, f, x, hidden):
-        # type: (Tensor, torch.Tensor, Optional[Tensor]) -> torch.Tensor # noqa F821
+    def forgetMult(
+        self, f: torch.Tensor, x: torch.Tensor, hidden: Optional[torch.Tensor]
```
pplantinga, line 1920: Should this be `hidden: Optional[torch.Tensor] = None` ?

pplantinga (commented):
Overall, looks good. Just a few questions here

speechbrain/dataio/sampler.py
rogiervd: There are existing uses of `List` which I have not changed, but we could/should do this in another PR, now that python>=3.9.

speechbrain/lobes/models/transformer/Transformer.py
rogiervd: Yes but I can't get it to work. I may be missing something.

```
speechbrain/lobes/models/transformer/Transformer.py:228:28 - error: Argument of type "type[Module]" cannot be assigned to parameter "activation" of type "type[ReLU]" in function "__init__"
    "type[Module]" is not assignable to "type[ReLU]"
    Type "type[Module]" is not assignable to type "type[ReLU]" (reportArgumentType)
```

speechbrain/nnet/RNN.py
rogiervd: No default value was given in the original, and I haven't changed the behaviour. This may have been the original intention, or not. Of course `Optional[int]` is `int | None`: `Optional` does not mean that the argument is optional.

speechbrain/lobes/models/transformer/Transformer.py
pplantinga: I messed around with this a bit and still couldn't get it to work, I'm fine with `type` for now until we find something better.

pplantinga (approved):
LGTM

PR #2925, opened 2025-05-22
Re-ran tutorial to remove errors

### LINKED ISSUE #2924: Errors exist in Speech Classification From Scratch Tutorial
### Describe the bug

If you check our documentation, https://speechbrain.readthedocs.io/en/latest/tutorials/tasks/speech-classification-from-scratch.html

There are several errors in the code execution:

```
RuntimeError: Found no NVIDIA driver on your system. Please check that you have an NVIDIA GPU and installed a driver from http://www.nvidia.com/Download/index.aspx
```

```
ls: cannot access 'results/speaker_id/1986/save/CKPT*': No such file or directory
```

```
RuntimeError: Symlink loop from '/content/best_model/embedding_model.ckpt'
```

### Expected behaviour

Tutorial should not contain errors like this. Should be re-run to fix and then comitted to repo

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

pplantinga:
Fixes #2924 

Basically just reran the tutorial on colab, everything still works I guess.

The one thing this revealed that we might want to fix later is that the template now spits out a bunch of:

```
speechbrain.dataio.encoder - CategoricalEncoder.expect_len was never called: assuming category count of 28 to be correct! Sanity check your encoder using `.expect_len`. Ensure that downstream code also uses the correct size. If you are sure this does not apply to you, use `.ignore_len`.
```


Adel-Moumen:
> The one thing this revealed that we might want to fix later is that the template now spits out a bunch of:
I think this is an issue that we might want to address here as well. I knew we had a bunch of this popping, but I didn't thought it was every epochs... I can have a deeper look.

pplantinga:
> > The one thing this revealed that we might want to fix later is that the template now spits out a bunch of:
> 
> I think this is an issue that we might want to address here as well. I knew we had a bunch of this popping, but I didn't thought it was every epochs... I can have a deeper look.

Perhaps just switching to `warnings.warn` could be enough to reduce warnings to one

PR #2928, opened 2025-05-27
Renaming largescale ASR to the Loquacious Set

TParcollet:
This renaming is necessary because the Loquacious Set is the real name of the dataset as accepted at Interspeech 2025. The other name was used to preserve anonymity ...

To dos:
- [x] rename in code
- [x] rename HF dataset
- [x] rename HF model


Adel-Moumen (approved):
Gud!

Issue #2046, opened 2023-06-22
Annoying Warning ( CTCTextEncoder)


mravanelli:
### Describe the bug

When running recipes that involve CTC, we always output the following warning:
```

/scratch/ravanelm/speechbrain/speechbrain/dataio/encoder.py:722: UserWarning: CTCTextEncoder.expect_len was never called: assuming category count of 40 to be correct! Sanity check your encoder using `.expect_len`. Ensure that downstream code also uses the correct size. If you are sure this does not apply to you, use `.ignore_len`.
  warnings.warn(
```

This can be useful, but it is very annoying if it comes out every time. Aku, do you see a fix for that?

### Expected behaviour

-

### To Reproduce

-

### Versions

-

### Relevant log output

```shell
-
```

### Additional context

-

Adel-Moumen:
@AsuMagic

mravanelli:
@Gastron, any insight on that? It is pretty annoying as it appears every time we run a recipe with CTC.

mravanelli:
Similarly, we often see this warning: "`UserWarning: CategoricalEncoder.expect_len was never called: assuming category count of 4 to be correct! Sanity check your encoder using `.expect_len`. Ensure that downstream code also uses the correct size. If you are sure this does not apply to you, use `.ignore_len`.`". How can we avoid that to happen when running a recipe?

mravanelli:
It looks like it is a change you made @AsuMagic . This causes many unnecessary warnings while running some recipes (e.g,, TIMIT)

AsuMagic:
Yes, the change was intentional to prevent accidental mismatches. At the time of the PR I didn't update recipes (partly because IIRC recipe tests weren't that ready back then). Will need to look into it.

AsuMagic:
I don't think I will have the time to update all recipes given the work time I have left (on top of existing PRs+streaming), so I spent some time thinking about how the warning should be improved instead. Right now, it looks like:

> CTCTextEncoder.expect_len was never called: assuming category count of 1000 to be correct! Sanity check your encoder using `.expect_len`. Ensure that downstream code also uses the correct size. If you are sure this does not apply to you, use `.ignore_len`.

I had some ideas for shorter messages but discussing it with an (uninvolved) developer, it seems to me that the verbosity remains useful/necessary and that my attempts at shorter warnings were not better.

So, would just revising the tone of the first sentence and adding a notice in the warning that states "This may be normal, users can ignore this." suffice?

[...]

PR #2932, opened 2025-05-31
Attempt to fix warning calls with Logger

Adel-Moumen:
## What does this PR do?

Attempt to fix warnings #2046 and #2925. 

Unfortunately, all my attempts to fix this problem failed. Each worker seems to call the `warning_once` method and trigger the message. Even when we try to cache the message. My understanding is that each worker is created a new logger, which doesn't share the same cached message. I decided to fallback to `logger.debug` to avoid having this messages. And now, thinking about it, it might make more sense. We should only inspect the different between output tokens and target vocab size if we are seeing NaNs. 

If we agree to merge this PR, we will need to re-run #2925 to remove the outputs.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/dataio/encoder.py
+++ b/speechbrain/dataio/encoder.py
@@ -8,6 +8,7 @@ Authors
 import ast
 import collections
 import itertools
+import warnings
 
 import torch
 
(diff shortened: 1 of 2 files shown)


Adel-Moumen:
> I tried this using the speaker-id template, and the dataloader still seems to spit out multiple messages, and I wasn't able to figure out why.
> 
> Maybe its not worth fixing this bug, as it seems a bit complicated and maybe not worth it for the couple of errors we have in the dataloader?

Yeah. I don'tthink its worth it. I can try to revisit this PR latter on as I am bit surprised by how difficult it is to solve this problem haha.

PR #2937, opened 2025-06-12
Saveable Generator: initial import

flexthink:
## What does this PR do?

This PR improves reproducibility for scenarios where training or inference can be highly non-deterministic, especially in cases where the selection of data, layers to train, etc is done randomly and it is unreasonable to expect training to complete in a single run.

It provides an opt-in wrapper called `SaveableGenerator` that will save the state of the available global generations (or custom generators if specified) together with the checkpoint.

Sample usage:

```yaml
generator: !new:model.custom_model.SaveableGenerator # <-- Include the wrapper

checkpointer: !new:speechbrain.utils.checkpoints.Checkpointer
    checkpoints_dir: !ref <save_folder>
    recoverables:
        model: !ref <model>
        lr_scheduler: !ref <lr_annealing>
        counter: !ref <epoch_counter>
        generator: !ref <generator> # <-- Add to the checkpoint
```

Usage is entirely opt-in. There are no changes to existing recipes or libraries.

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/utils/repro.py
+++ b/speechbrain/utils/repro.py
@@ -5,9 +5,10 @@ Author:
 """
 
 import re
-import speechbrain as sb
+
 import torch
 
+import speechbrain as sb
 from speechbrain.utils.logger import get_logger
 
 logger = get_logger(__name__)
@@ -48,9 +49,7 @@ class SaveableGenerator:
 
     def __init__(self, generators=None):
         if generators is None:
-            generators = {
-                "default": torch.default_generator
-            }
+            generators = {"default": torch.default_generator}
             if torch.cuda.is_available():
                 for idx, generator in torch.cuda.default_generators:
                     generators[f"cuda:{idx}"] = generator
@@ -72,10 +71,14 @@ class SaveableGenerator:
             match = re.match(r"cuda:(\d+)", key)
             if match:
                 if not torch.cuda.is_available():
-                    logger.warning("Unable to restore RNG for %s, CUDA unavailable", key)
+                    logger.warning(
+                        "Unable to restore RNG for %s, CUDA unavailable", key
+                    )
                     continue
                 idx = match.group(1)
                 if idx > torch.cuda.device_count() - 1:
-                    logger.warning("Unable to restore RNG for %s, device not found", key)
+                    logger.warning(
+                        "Unable to restore RNG for %s, device not found", key
+                    )
                     continue
             self.generators[key].set_state(state)
--- a/speechbrain/utils/repro.py
+++ b/speechbrain/utils/repro.py
@@ -36,7 +36,7 @@ class SaveableGenerator:
             model: !ref <model>
             lr_scheduler: !ref <lr_annealing>
             counter: !ref <epoch_counter>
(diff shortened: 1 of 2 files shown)


speechbrain/utils/repro.py
```diff
@@ -0,0 +1,130 @@
...
+    consider implementing a custom generator.
+
+    Running it on an unsupported device not using the Torch
+    generator interface will simply fail to restore the
+    state but will not cause an error.
+
+    Arguments
+    ---------
+    generators : list, optional
+        A list of generator objects. If not provided,
```
pplantinga: Sentence not complete, what happens if not provided?

pplantinga: Also, can you add an example to the docstring?

pplantinga: Also, I think the type should be `dict` or `Mapping` instead of `list`, and this should describe how the mapping should be defined, e.g. `Mapping[str, Generator]` or something like this. I think the torch generator is of the generic type:

https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator

tests/unittests/test_repro.py
```diff
@@ -0,0 +1,33 @@
+"""Unit tests for reproducibility utilities"""
+
+import torch
+
+
+def test_repro(tmpdir):
```
pplantinga, line 8: Some of our unit tests take a device parameter for testing on e.g. cuda. Perhaps we can do something similar here to ensure it works at least locally (I guess the CI is running on cpu).

pplantinga (commented):
I like the idea of saving the random generator state and this seems like a nice implementation to me. Am I understanding correctly that one could fix a recipe to   restart the random seed correctly just by adding a `SaveableGenerator` object to the checkpointer without any arguments? Maybe we could consider a separate PR to convert many of our recipes.

tests/unittests/test_repro.py
flexthink: @pplantinga : Done in a separate test. However, device support is currently limited given that:
- Torch has inconsistencies in how generators are handled
- Even for Cuda, default generators are no longer exposed *as generators*
- Some devices like MPS don't even expose RNG state

So for now, this feature will be only for the most common use cases. Other devices can be added later by writing wrappers similar to the one I had for Cuda - if they support the functionality at all.

speechbrain/utils/repro.py
flexthink: @pplantinga : Addressed... However, in reality, this takes a "generator-like object", only getting/setting the state is required - no need to implement the full interface.

flexthink:
> I like the idea of saving the random generator state and this seems like a nice implementation to me. Am I understanding correctly that one could fix a recipe to restart the random seed correctly just by adding a `SaveableGenerator` object to the checkpointer without any arguments? Maybe we could consider a separate PR to convert many of our recipes.

That is absolutely correct. Further customization may be needed if the recipes use custom generators, but for the base case that's all it takes.

pplantinga (approved):
This LGTM!

PR #2946, opened 2025-07-14
Adopt pyproject and move to Ruff

TParcollet:
This PR makes SpeechBrain move away from setup.py to pyproject. Then it moves all the old CI tools to ruff, which integrates most of them in a rust-based (i.e. MUCH faster) and unique tool. I'll need the help of @pplantinga and @Adel-Moumen to validate this one. Actually even our good ol' friend @asumagic could be intrigued if he passes bye GitHub. 

In particular @pplantinga I am unsure how moving away impacts releases and pypi? 

Now I also need help from both @pplantinga and @Adel-Moumen on the rules. If you look into the pyproject, you'll find a massive list of excluded errors. This is to make the precommit pass, but it also raises invalid things in our code. I think i'd like to hear from you about what should be fixed (and therefore kept as a rule) or not fixed and therefore kept as an excluded rule. Also, we may want to soften some rules to have less files impacted ... or not? xD

To test, just do:

uv sync --extra dev
uv run pre-commit run -a

To dos:
- [x] Move to pyproject
- [x] Move to ruff
- [ ] Debate on the rules.


asumagic:
I think that's a generally good thing, switching to Ruff won't make CI faster other than maybe fail faster (since that's the pre-commit action running on its own runner), but it's an upside for developer use in IDEs or just running the tool.

As for the pyproject switch I'm not really familiar with it but I imagine that it shouldn't need any changes outside the [release](https://github.com/speechbrain/speechbrain/blob/develop/.github/workflows/release.yml) action (if it even needs any). You could check if those commands generate a functional `.whl` that you can properly install and use in a clean venv.

pplantinga:
> In particular @pplantinga I am unsure how moving away impacts releases and pypi?

As @asumagic said, I doubt this requires changes to the `release` action (which probably isn't even working anyway) as `python -m build` should work with `pyproject.toml` anyway.

lint-requirements.txt
```diff
@@ -1,8 +1,4 @@
-black==24.3.0
-click==8.1.7
-flake8==7.0.0
-isort==5.13.2
-pycodestyle==2.11.0
-pydoclint==0.4.1
+codespell>=2.3.0
 pytest==7.4.0
+ruff>=0.3.0
```
pplantinga: Maybe as Mirco said we can fix this to a specific version (latest, 0.12.4?) to avoid errors when new versions are released.

.pre-commit-config.yaml
```diff
@@ -13,22 +13,14 @@ repos:
       - id: check-added-large-files
         args: [--maxkb=1024]
 
-  - repo: https://github.com/psf/black
-    rev: 24.3.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.3.0
```
pplantinga: Should we use the latest one here? v0.12.4

pyproject.toml
```diff
@@ -1,23 +1,133 @@
...
+    "huggingface_hub",
+]
+keywords = ["speech", "audio", "pytorch", "deep-learning"]
+
+[project.urls]
+Homepage = "https://speechbrain.github.io/"
+
+[project.optional-dependencies]
+dev = [
+    "ruff>=0.3.0",
```
pplantinga: As before, let's use a specific version.

pplantinga (commented):
LGTM, my only suggestion is to fix the ruff version

TParcollet:
I've changed a few things, and we will want to solve one after the other the different ignore placed in the pyproject. I've fixed the major ones, but other may exist. @Adel-Moumen please merge.

Adel-Moumen (approved):
LGTM.

Tested on my mac and it works neatly (quite fast compared to our previous pipeline!).

PR #2949, opened 2025-07-17
Move sentencepiece to extra_requirements.txt

### LINKED ISSUE #2944: `sentencepiece` is no longer maintained and doesnt support newer python versions
### Describe the bug

Please check this thread to see the issues sentencepiece is having and maintainers are ignoring: https://github.com/google/sentencepiece/pull/1120

pyannote is having issues since it depends on speechbrain which depends on sentencepiece: https://github.com/pyannote/pyannote-audio/issues/1890

Even mistral-common and transformers packages are having issues and thinking of removing sentencepiece from dependencies: https://github.com/mistralai/mistral-common/issues/75

### Expected behaviour

I expect speechbrain to work in python 3.13.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

pplantinga:
Fixes #2944 

We will work on new recipes using different a different tokenizer (e.g. HF `tokenizers` or `torchtext`) but for now we just remove the requirement from the main library and put it in `extra_requirements.txt` for every recipe that uses it.

### DIFF
--- a/docs/coverage.md
+++ b/docs/coverage.md
@@ -138,7 +138,6 @@ our tests cover one PyTorch version only, _the latest_.
        "numpy",
        "packaging",
        "scipy",
-       "sentencepiece",
        "torch>=1.9",
        "torchaudio",
        "tqdm",
--- a/recipes/AISHELL-1/ASR/seq2seq/extra_requirements.txt
+++ b/recipes/AISHELL-1/ASR/seq2seq/extra_requirements.txt
@@ -1,4 +1,1 @@
-librosa
-pesq
-pystoi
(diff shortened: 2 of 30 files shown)


pplantinga:
Its not as easy as I thought to move sentencepiece to extra requirements slash integrations because many of our inference models depend on it. One alternative is a community build of sentencepiece that supports 3.13:

https://pypi.org/project/dbowring-sentencepiece/

pplantinga:
Closing, as I think a better approach is to convert to `tokenizers` a la https://github.com/huggingface/transformers/issues/25316

PR #2952, opened 2025-07-18
Fix missing defaults in run options

### LINKED ISSUE #2950: RunOpts no longer working with DDP (single and multinode)
### Describe the bug

I had to checkout to this commit to bypass the issue: ` ba9364af76f95683348f83b211c90b4301788734 `

Since the RunOpts dataclass refactor, my scripts are no longer working on the HPC cluster Jean-Zay.
The `distributed_backend` attribute is always set to `None`. The expected behaviour is to for it to be automatically set to `nccl`.

I tried the following: 
1. Setting it manually to `nccl` via arg prams or in the code: The train starts until the train log prints, then a deadlock happens and the NCCL watchdog stops the processes.
2. Re-creating a `run_opts` object in the train script, and loading the hparams after:  it is then impossible to run this: 
```python
with open(hparams_file, encoding="utf-8") as fin:
        hparams = load_hyperpyyaml(fin, overrides)
    hparams.update(run_opts) # => Throws error: "Could not update hprams from run_opts: attribute name must be string, not 'int'"
```

I was using the latest commit as of the 17/08/2025.

I tested multiple scripts from different projects of mine. Furthermore, I tested both single-node and multi-node DDP.

Here's an example of the error I get when trying to fine-tune Whisper for an ASR task using 4 GPUs on the same node: 
```log
  File "<my_path>/model/train_whisper.py", line 275, in <module>
    sb.utils.distributed.ddp_init_group(run_opts)
  File "<hidden>/speechbrain/speechbrain/utils/distributed.py", line 323, in ddp_init_group
  File "<hidden>/model/train_whisper.py", line 275, in <module>
    run_opts["distributed_backend"]
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

Could it be that I need to change the way I run my scripts and not an error ?  I currently run my scripts like follows:
```bash
<SLURM config stuff>
<Loading relevant modules>
conda activate <env>
cd <train_script_path>
torchrun --nproc_per_node=4 \
        --nnodes=${SLURM_JOB_NUM_NODES} \
        --node_rank=${SLURM_NODEID} \
        --rdzv_id=12345 \
        --rdzv_backend=c10d \
        --rdzv_endpoint=$RDZV_ENDPOINT \
        --rdzv_conf="timeout=600" \
        train_whisper.py hparams/whisper.yaml
```

Thank you for your help and your work on Speechbrain ! :D

### Expected behaviour

For the training to.art as it previously did.

### To Reproduce

Use multiple GPUs:
```bash
torchrun --nproc_per_node=4 \
        --nnodes=${SLURM_JOB_NUM_NODES} \
        --node_rank=${SLURM_NODEID} \
        --rdzv_id=12345 \
        --rdzv_backend=c10d \
        --rdzv_endpoint=$RDZV_ENDPOINT \
        --rdzv_conf="timeout=600" \
        train_whisper.py hparams/whisper.yaml
```

### Environment Details

Speechbrain installed from source.
Python version: 3.11
Using the last commit as of 17/08/2025

### Relevant Log Output

```shell
  File "/path/to/your/project/model/train_whisper.py", line 275, in <module>
    sb.utils.distributed.ddp_init_group(run_opts)
  File "/path/to/speechbrain/speechbrain/utils/distributed.py", line 323, in ddp_init_group
  File "/path/to/your/project/model/train_whisper.py", line 275, in <module>
    run_opts["distributed_backend"]
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

### Additional Context

Used to work fine before updating my Speechbrain repo.
A checkout to this commit to "solves" the issue: ba9364af76f95683348f83b211c90b4301788734

pplantinga:
Fixes #2950 

String-based defaults were mistakenly missing from the run opts command line interface

### DIFF
--- a/speechbrain/utils/run_opts.py
+++ b/speechbrain/utils/run_opts.py
@@ -316,6 +316,7 @@ class RunOptions:
                 kwargs["action"] = "store_true"
             elif default is not None:
                 kwargs["type"] = type(default)
+                kwargs["default"] = default
 
             # Any options with "precision" in the name can only take these values
             if "precision" in option:


pplantinga:
Looks like this does solve at least one issue, maybe we can go ahead with it while we wait for more details on the next issue.

pplantinga:
Another person has run into the issue, so I'll go ahead and merge.

PR #2963, opened 2025-08-05
fix unnecessary pos_emb for RoPE

shucongzhang:
This PR deletes the `pos_emb` check for RoPE in `forward_streaming` of the Conformer. The rotation matrix of RoPE will not be passed as an argument so this line shouldn't be here.


TParcollet:
@adityadev11 could you please check if this is fixing your issue?

adityadev11:
Yes @TParcollet , training seems to work with the change^, thanks!

Issue #2753, opened 2024-11-08
GigaSpeech: 1% absolute WER discrepancy between training script eval and evaluation through StreamingASR


asumagic:
### Describe the bug

When evaluating the Conformer-Transducer model on GigaSpeech (related: #2752), I observed a major discrepancy (12.0% WER -> 10.9% WER) between the regular evaluation logic (inside the training script) and using [**a custom script that evaluates using `StreamingASR`**](https://gist.github.com/asumagic/66aaf43156c31866cf2ebcffda46ea01). See:

```
%WER 11.97 [ 46777 / 390744, 7308 ins, 12193 del, 27276 sub ]
%WER 10.94 [ 42749 / 390744, 7734 ins, 8980 del, 26035 sub ]
```

Notice that the `StreamingASR` path increases insertions, but greatly reduces deletion count.

For that test, I emulated full context by setting a huge chunk size.

This is a really significant difference. I wonder if it affects other datasets we are testing against with this model.

This might indicate:
- differences in padding? (streaming logic does some amount of left-padding -- would explain reduced deletions)
- a masking bug of some sort? (not sure where that would be)

The detailed WER outputs are listed below (inferencepath uses the custom script):
- [test_wer.txt](https://github.com/user-attachments/files/17679031/test_wer.txt)  
- [test-wer-inferencepath.txt](https://github.com/user-attachments/files/17679026/test-wer-inferencepath.txt)

### Expected behaviour

The discrepancy should be much smaller.

asumagic:
Also it largely doesn't seem to care about batch size for the StreamingASR script, as a batch size of 1 yields very close results to a batch size of 32 even with the bogus `lengths` tensor. That said the test set seems sorted in size so the impact is limited.

PR #2975, opened 2025-09-25
Adapting transducer greedy decoding

younessdkhissi:
Adding a decoding loop on each decoding timestep. It is more adapted to RNN-T loss and it improves significantly the results (especially for streaming recipes)

## What does this PR do?

Fixes the transducer greedy decoding (this probably fixes the issue https://github.com/speechbrain/speechbrain/issues/2753)

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/docs/tutorials/basics/data-loading-pipeline.ipynb
+++ b/docs/tutorials/basics/data-loading-pipeline.ipynb
@@ -41,7 +41,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {
     "id": "5spDEaL1yddy"
    },
@@ -55,7 +55,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {
     "id": "DPX-4BBbX5L9"
    },
@@ -76,16 +76,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {
     "id": "vtSWx0KAX5L9"
    },
    "outputs": [],
    "source": [
     "%%capture\n",
-    "# here we download the material needed for this tutorial: images and an example based on mini-librispeech\n",
-    "!wget https://www.dropbox.com/s/b61lo6gkpuplanq/MiniLibriSpeechTutorial.tar.gz?dl=0\n",
-    "!tar -xvzf MiniLibriSpeechTutorial.tar.gz?dl=0\n",
     "# downloading mini_librispeech dev data\n",
     "!wget https://www.openslr.org/resources/31/dev-clean-2.tar.gz\n",
     "!tar -xvzf dev-clean-2.tar.gz"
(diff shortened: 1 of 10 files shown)


Adel-Moumen:
Hi @younessdkhissi, 

Thanks for the PR. 

Could you please develop a bit more what are your suggested changes, and why do you think they are necessary? Furthermore, can you report the general improvements you've got in a streaming setting? Ideally, we would like to imrpove our transfucer decoding interfaces by being more aligned to the literature, so if you could provide some references it would be great as well. 

Thanks Youness!

younessdkhissi:
Thanks @Adel-Moumen for taking time to look at my PR.
The idea behind the changes made in the greedy search is to represent all the alignments we learn with the RNN-T loss during training. In the figure below, we see that many of the RNN-T alignment paths suppose that many tokens could be decoded in the same timestamp.
In the current speechbrain greedy search, we could decode only one token from each timestep (similar to CTC) but this is different from RNN-Transducer objective.
In Streaming ASR, transducer has more tendency to delay the predictions until it is very confident by visiting more future frames (http://arxiv.org/abs/2111.01690). This push the model to learn decoding many tokens at the end. With the current algorithm, a number of deletions is expected at the end of th transcriptions. That why I mentionned the issue https://github.com/speechbrain/speechbrain/issues/2753 where there is a difference between the evaluation inside the training script and the evaluation using StreamingASR class: in this class, there is a number of zero chunks injected at the end of the input stream which gives more chance to decode more tokens at the end.

I have tried to implement the following paper "DUAL-MODE ASR: UNIFY AND IMPROVE STREAMINGASR WITH FULL-CONTEXT MODELING" https://[openreview.net/pdf?id=Pz_dcqfcKW8](https://openreview.net/pdf?id=Pz_dcqfcKW8) (that could be a future PR) and these are the results I get on LibriSpeech using the streaming mode:
- Dual-mode ASR (streaming): 
 - Using the current greedy search:
   - test-clean: 4.32%
   - test-other: 11.04%
 - Using the greedy search proposed in the PR:
   - test-clean: **3.88%**
   - test-other: **9.88%**
- original paper results:
   - test-clean: 3.7%
   - test-other: 9.2%

rnn_t_alignment_paths

TParcollet:
Hi @younessdkhissi thanks for the work. I am really not sure what this PR does. I see that there is a new arbitrary for loop for each time steps. Do you have a paper describing what is being done here formally? This would be important to attach to such a change to greedy decoding.

younessdkhissi:
Hello @TParcollet
In this PR, I changed the greedy search so that for each timestep, We decode until we predict the blank token because the original Transducer paper (https://arxiv.org/abs/1211.3711?utm_source=openai) defines alignments over a T×U lattice (see the figure in the previous comment) with a blank symbol, allowing multiple label emissions inthe same time step.
To avoid infinite loop, I defined a for loop of 'max_iterations' which consiste that for each timetep we could decode at most "max_itarations" of non-blank token.
Here is a paper of NVIDIA where they describe their greedy search algorithm (https://www.isca-archive.org/interspeech_2024/galvez24_interspeech.pdf): the 9th line of Algorithm 1 is equilevent to the for loop that I have added in my PR.
The choice of "max_iterations" value was inpired from this paper too.

TParcollet:
Hi @younessdkhissi and thank you. I'll take a closer look into this asap. In the meantime, could you please provide a few measurements of how the decoding speed is impacted by this change? Many thanks!

younessdkhissi:
Hi @TParcollet
I measured the inference time using both decoding methods and take the mean of 5 different runs with different seeds.
I used the RTX 2080 Ti GPU card. The measurements have been made on 12-layer conformer transducer (similar architecture to the existing SpeechBrain recipe) using a chunk size of 320ms without taking any left context. The inference has been made on LibriSpeech test set using batch size of 4.
- With the current greedy decoding(test-clean/test-other):
  - WER : 3.85%/10.36%
  - Inference time : 2min 29s / 2mins 18s
- With the proposed greedy decoding: 
  - WER : 3.79%/10.30%
  - Inference time : 2min 47s / 2mins 35s
 
If you want more measurements let me know :)

TParcollet:
@younessdkhissi thanks. Can you fix the tests and then we'll merge.

TParcollet (approved):
@younessdkhissi one last thing, could you make the max_steps an aargument of the function isntead?

TParcollet:
@younessdkhissi sorry I should have been more precise. I think "max_steps" is a bit to generic and may be interpreted as "max number of decoding steps" which may be confusing. Thanks for adding it to the arguments, but could you give it a more adequate name?

Thanks!

younessdkhissi:
@TParcollet It's my fault for making such a generic name for this variable. I propose "max_symbols_per_step" to avoid any confusion. Let me know if there are more changes to do :)

TParcollet:
Thanks!

Issue #2983, opened 2025-10-17
Checkpoint save multi-gpu multi-nodes


mdhaffar:
### Describe the bug

Hi all!
Are you aware of some issues in the checkpointer management on multi-nodes within the release 1.0.3 ?  I'm working on porting the implementation of the SENSE (https://arxiv.org/pdf/2509.12093) models to SpeechBrain. It worked correctly with 1.0.2 but not with 1.0.3.
When I work on multi-gpu in a single node with 1.0.3, there is no issue, but when working in a multi-node environment, some files are missing:
brain.ckpt counter.ckpt  dataloader-TRAIN.ckpt (they are created when working on a single node with 1.0.3, but also when working on multi-node with 1.0.2).

The .yaml:
checkpointer: !new:speechbrain.utils.checkpoints.Checkpointer
    checkpoints_dir: !ref <save_folder>
    recoverables:
        model: !ref <model>
        wav2vec2: !ref <wav2vec2>
        lr_annealing_adam: !ref <lr_annealing_adam>
        lr_annealing_wav2vec: !ref <lr_annealing_wav2vec>
        counter: !ref <epoch_counter>

In the begining of epoch 2, I get this error:
Traceback (most recent call last):
  File “train_with_weighted_random_sampler_H1.py”, line 300, in <module>
    st_brain.fit(
  File “speechbrain/speechbrain/core.py”, line 1198, in fit
    self.on_fit_start()
  File “speechbrain/speechbrain/core.py”, line 738, in on_fit_start
    self.checkpointer.recover_if_possible()
  File “speechbrain/utils/checkpoints.py”, line 1018, in recover_if_possible
    self.load_checkpoint(chosen_ckpt)
  File “speechbrain/utils/checkpoints.py”, line 1031, in load_checkpoint
    self._call_load_hooks(checkpoint)
  File “speechbrain/utils/checkpoints.py”, line 1163, in _call_load_hooks
    if self.optional_recoverables[name]:
KeyError: ‘counter’
Any suggestions?
Thank you,

### Expected behaviour

I want to correctly save brain.ckpt counter.ckpt  dataloader-TRAIN.ckpt 

### Relevant Log Output

```shell

```

PR #2992, opened 2025-11-01
Add functions for running code once per node

pplantinga:
Adds the capability for users to choose whether to run certain code (like saving) only once (main proc) vs. once per node (local_rank = 0), solving issues like #2983.

In addition adds brief documentation and a convenience function for updating checkpointer to save on each node.

### DIFF
--- a/speechbrain/utils/distributed.py
+++ b/speechbrain/utils/distributed.py
@@ -59,12 +59,29 @@ def get_rank() -> Optional[int]:
     return None
 
 
+def get_local_rank() -> Optional[int]:
+    r"""Get the local rank of the current process on the current node.
+
+    Returns
+    -------
+    int or None
+        The local rank of the current process, or None if the local rank could not be determined.
+    """
+    rank_keys = ("LOCAL_RANK", "SLURM_LOCALID")
+    for key in rank_keys:
+        rank = os.environ.get(key)
+        if rank is not None:
+            return int(rank)
+    # None to differentiate whether an environment variable was set at all
+    return None
+
+
 def infer_device() -> str:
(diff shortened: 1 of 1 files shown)


speechbrain/utils/distributed.py
```diff
@@ -154,6 +243,14 @@ def if_main_process() -> bool:
         return True
 
 
+def is_local_rank_zero() -> bool:
+    r"Returns whether the current process has local rank of 0."
+    if is_distributed_initialized():
+        return int(os.environ["LOCAL_RANK"]) == 0
```
Adel-Moumen: do you think it makes sense to call the `get_rank` method?

pplantinga: `get_rank` looks at the `RANK` env var before `LOCAL_RANK`, we want the latter here

Adel-Moumen: yes I know, but I think i should have been clearer sorry: my point was mostly, looking at get_rank and how pt lighning tries to fetch the rank attribute, it seems to me that its possible to have more than one variable that describes local_rank depending on the context. So I think I have two questions: one is, are you aware of this specific problem? i.e. local_rank not being setup but used as a different name e.g. SLURM_LOCALID, and second, I think that given the fact we tend to check LOCAL_RANK quite frequently, it could make sense having a proper helping function fetching this value.

pplantinga: Okay, fair enough, we can check the SLURM_LOCALID as well. I haven't been able to test this on SLURM so far, so I have no idea if it works but I don't see any harm in just adding it here.

PR #2994, opened 2025-11-04
fix speed perturb inversion

### LINKED ISSUE #2993: slow/fast inverted in speed perturb
### Describe the bug

In the Speed Perturbation augmentation, speed changes are given as integer percentages. For example, passing 90 implies that the speed will be slowed to 90%. In this case, we would be multiplying the original sampling rate by 90 and then dividing by 100, effectively multiplying by 0.9. This is inverted and will actually speed up the signal to 110% speed. The original sampling rate should be divided by this factor, not multiplied by it.

In practice, this is not that bad, as by default, the speed changes are symmetric. A model will still *see* both speed changes during training. That being said, this should still be corrected for those who are using non-symmetric speed change lists so they don't slow down their signals when they are trying to speed them up. 

```
def __init__(self, orig_freq, speeds=[90, 100, 110], device="cpu"):
        ..
        ..
        ..
        for speed in self.speeds:
            config = {
                "orig_freq": self.orig_freq,
                "new_freq": self.orig_freq * speed // 100,
            }
        ...
```

### Expected behaviour

The expected behavior is that the signal slows down when 90 is passed, not sped up.

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

gfdb:
## What does this PR do?
This PR fixes the inverted speed perturb logic, updates the unit tests, and replaces the incorrect "expected" output for the integration tests.

Fixes #2993

<details>

</details>

## PR review

<details>

</details>

### DIFF
--- a/speechbrain/augment/time_domain.py
+++ b/speechbrain/augment/time_domain.py
@@ -8,6 +8,7 @@ All classes are implemented as `torch.nn.Module`, enabling end-to-end differenti
 Authors:
 - Peter Plantinga (2020)
 - Mirco Ravanelli (2023)
+- Gianfranco Dumoulin Bertucci (2025)
 """
 
 # Importing libraries
@@ -477,7 +478,7 @@ class SpeedPerturb(torch.nn.Module):
     >>> clean.shape
     torch.Size([1, 52173])
     >>> perturbed.shape
-    torch.Size([1, 46956])
+    torch.Size([1, 57971])
     """
 
     def __init__(self, orig_freq, speeds=[90, 100, 110], device="cpu"):
@@ -493,7 +494,7 @@ class SpeedPerturb(torch.nn.Module):
         for speed in self.speeds:
             config = {
                 "orig_freq": self.orig_freq,
-                "new_freq": self.orig_freq * speed // 100,
+                "new_freq": round(self.orig_freq * 100 / speed),
             }
             self.resamplers.append(Resample(**config))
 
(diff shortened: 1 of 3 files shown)


speechbrain/augment/time_domain.py
```diff
@@ -493,7 +494,7 @@ def __init__(self, orig_freq, speeds=[90, 100, 110], device="cpu"):
         for speed in self.speeds:
             config = {
                 "orig_freq": self.orig_freq,
-                "new_freq": self.orig_freq * speed // 100,
+                "new_freq": int(self.orig_freq / (speed / 100)),
```
pplantinga: Should this be the following to prevent rounding errors?

```suggestion
                "new_freq": self.orig_freq * 100 // speed
```

I tried with orig_freq=44100 and speed=112 and got this:
`orig_freq / (speed / 100) = 39374.99999999999`
The `int` conversion makes this `39374` which is wrong.

pplantinga: On the other hand, perhaps truncating is the wrong way to go here, we could round instead:
```suggestion
                "new_freq": round(self.orig_freq * 100 / speed)
```

gfdb: you're right, I think round is indeed the best option

gfdb:
@pplantinga i rebased and force pushed so we only get 1 merge commit, you will have to hard reset the branch if you are testing locally

pplantinga (approved):
LGTM, thanks @gfdb !

PR #3043, opened 2026-03-29
fix float columns being converted to string in from_csv

### LINKED ISSUE #3027: DynamicItemDataset changes fields to string
### Describe the bug

Below, the column types are being changed to string, by DynamicItemDataset:

```
import speechbrain as sb
import pandas as pd
df=pd.DataFrame(columns=['ID', 'duration', 'rand'])
df.loc[0]=[1,1.3, 1.3]
df.loc[1]=[2, 2.1, 3.2]
df.to_csv('test_file.csv', index=False)
train_data = sb.dataio.dataset.DynamicItemDataset.from_csv(
        csv_path="test_file.csv",
        replacements={"data_root": "./"},
    )
```

train_data - Output ->

The column 'rand' has been changed to type 'string'. The data type of column 'duration' has been preserved.

```
train_data.data
{'1.0': {'duration': 1.3, 'rand': '1.3'},
 '2.0': {'duration': 2.1, 'rand': '3.2'}}
```

### Expected behaviour

```
train_data.data
{'1.0': {'duration': 1.3, 'rand': 1.3},
 '2.0': {'duration': 2.1, 'rand': 3.2}}
```

### To Reproduce

_No response_

### Environment Details

_No response_

### Relevant Log Output

```shell

```

### Additional Context

_No response_

knQzx:
`load_data_csv` was only casting the `duration` field to float, leaving all other numeric columns as strings. replaced that with a general int/float cast for all fields so numeric values are preserved properly.

fixes #3027

### DIFF
--- a/speechbrain/dataio/dataio.py
+++ b/speechbrain/dataio/dataio.py
@@ -99,7 +99,7 @@ def _recursive_format(data, replacements):
             # If not dict, list or str, do nothing
 
 
-def load_data_csv(csv_path, replacements=None):
+def load_data_csv(csv_path, replacements=None, col_types=None):
     """Loads CSV and formats string values.
 
     Uses the SpeechBrain legacy CSV data format, where the CSV must have an
@@ -107,6 +107,11 @@ def load_data_csv(csv_path, replacements=None):
     If there is a field called duration, it is interpreted as a float.
     The rest of the fields are left as they are (legacy _format and _opts fields
     are not used to load the data in any special way).
+
+    When ``col_types`` is provided, the specified columns are explicitly cast
+    to the given types. This is useful when numeric columns (e.g. duration,
+    speaker id) need to be stored as a specific Python type rather than as
+    strings.
 
     Bash-like string replacements with $to_replace are supported.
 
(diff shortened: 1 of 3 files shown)


Adel-Moumen:
Hi @knQzx , thanks for your PR! 

I think your PR try to address an important problem, but I am not so convinced by the way you've done it. I think this way of automatically -- and silently -- converting columns to "compatible" types might lead to incorrect behaviours. For instance, a string "00123" will be converted into an int "123" which loses information, the first two digits that might be important. 

I think another solution could be to pass to `DynamicItemDataset.from_csv` a types argument (e.g. a dict) that explictly tells the type of each column e.g. "filename: str, duration: float, speak_id: int". This way we leave the conversion to the user rather than something automatic. 

cc @pplantinga

knQzx:
hey, thanks for the feedback - you're right, auto-casting is a bad idea since it silently breaks stuff like "00123" -> 123.

reworked it to use an explicit col_types dict like you suggested. now load_data_csv and from_csv accept col_types={"duration": float, "spk_id": int} etc - only casts what the user asks for, default behavior is unchanged.

added a test for it too. lmk if this looks better!

pplantinga (approved):
LGTM

PR #3047, opened 2026-04-02
Add hooks to ensure saving of kmeans model works correctly

pplantinga:
Saving of kmeans model doesn't work because methods are not registered. This fixes the issue.


Adel-Moumen (approved):
LGTM.

PR #3067, opened 2026-07-13
Use safeloader to load checkpoint meta files

pplantinga:
While loading checkpoints is known to be unsafe (see issue #1930), a related potential security issue is loading checkpoint meta files (usually `CKPT.yaml`) which is currently also unsafe, and has the potential to be slightly more surprising as all meta files in the save directory can be loaded for sorting etc., not just the target checkpoint. This adjusts the behavior so that meta files are loaded with SafeLoader, reducing the chance that a malicious meta file could be loaded without knowing.

Adopting this PR would mean that the old supported behavior of storing objects (e.g. Tensors) in the meta would be no longer supported, so we should run recipe tests and make sure no recipes depend on this. Also, this would merit a warning in the release notes and potentially a minor version marker rather than patch (i.e. 1.2 rather than 1.1.1).

### DIFF
--- a/recipes/ESC50/interpret/interpreter_brain.py
+++ b/recipes/ESC50/interpret/interpreter_brain.py
@@ -366,13 +366,13 @@ class InterpreterBrain(sb.core.Brain):
         comp_tensor = torch.Tensor(self.comp.scores)
         comp_tensor = comp_tensor[~torch.isnan(comp_tensor)]
         tmp = {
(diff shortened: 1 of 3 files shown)


pplantinga:
After having gemini flash 3.7 look through all the recipes, there seems to be only one recipe that saves tensors (ESC50/interpret/interpreter_brain.py). I have fixed this one and run the ESC50 recipe tests (and a few others) so we should be good to go.

Adel-Moumen (approved):

